PR #9, opened 2018-03-08
[ML] First pass implementation of support functionality for change detection and modelling 

tveasey:
**Description**
This implements 1) a naive Bayes classifier, using our distribution models, which will be used for modelling the probability of a change, and 2) a change detector framework, currently supporting detecting level shifts and time shifts, which works by comparing BIC of the various possible hypotheses against one another and a null hypothesis that there is no change. (Note that this work is going to initially be implemented on a feature branch to enable incremental review of the changes and so we can evaluate on all our QA data sets before merging to master.)

**Effects**
The functionality are not used in this change, so it has no effect on our results.


droberts195:

PR #11, opened 2018-03-12
[ML] Wire in change detection/modelling to our univariate time series model

tveasey:
**Description**
This is work in progress, but I'm submitting to avoid this PR growing larger. This wires in change detection and starts some unit testing of `CTimeSeriesModel` with change points. There is some more work to be done to avoid using level shifts to try and fit other types of change points, such as scaling.

**Effects**
This will affect all our results, but it is work in progress so it isn't anticipated that we will test results without further changes.


droberts195:

PR #21, opened 2018-03-21
[ML] Implements an absolute goodness-of-fit test to accept a change 

tveasey:
The key issue we had with change detection, prior to this PR, was that all the tests were relative, i.e. in terms of relative evidence to no change. This meant that if the time series changed in a way which was not well described by one of the possible changes we consider, it was quite possible to accept a change versus the hypothesis that the time series hadn't changed. This led to degraded adaption of the model: whose parameters should be rapidly relearnt as we do now in this case.

This PR implements an absolute "goodness-of-fit" test for each change, by additionally testing versus its expected BIC given the residual distribution. It means we will only accept changes which are a reasonably accurate description of the change currently occurring in the time series.


droberts195:

PR #27, opened 2018-03-27
[ML] Add include categorical filter type to detector rules

davidkyle:
Detector rules have a `E_Categorical` condition that applies to text fields, when used in conjunction with the action `E_FilterResults` the effect is to filter the results i.e. don't generate results for these field values 

This PR introduces the inverse functionality i.e. generate results only for those field values in the filter list. 

I renamed the enum `E_Categorical` to `E_Match` but left the JSON parsing code as is for backwards compatibility. I'm not entirely happy with the naming any suggestions are welcome or perhaps this is best left for the anticipated rewrite of rules


davidkyle:
Marking as a non-issue for release notes purposes as this code was reworked significantly in #119 and was never released in this form.

Issue #53, opened 2018-04-13
[ML] Analytics Runtime Improvements


tveasey:
This is a placeholder for a number of analytics runtime improvements we have identified based on profiling high cardinality population analysis (see [oprofile_report.400k.txt](https://github.com/elastic/ml-cpp/files/1906967/oprofile_report.400k.txt), [oprofile_report.1m.txt](https://github.com/elastic/ml-cpp/files/1906971/oprofile_report.1m.txt) and [oprofile_report.2m.txt](https://github.com/elastic/ml-cpp/files/1906979/oprofile_report.2m.txt) for more details):

- [x] _Fixed Size Weights Array and Direct Indexing_
Rework weight styles to use an array of size equal to number of possible weights and then directly address by the weight enum. The fact that that methods like maths_t::countForUpdate are cropping up near the top of the profiles means this is an easy win.
- [ ] _Cache Probability Calculation Results for Multivariate Features_
Caching probabilities was a big win for univariate features. The caching strategy is harder for multivariate features such as lat/long, but should still be feasible.
- [ ] _Investigate Approaches for Speeding Up Multimodal Probability Calculation_
Investigate ways of speeding up ml::maths::CTools::CMixtureProbabilityOfLessLikelySamples, possibly using a different strategy at the expense of reducing accuracy where we don't need it.
- [ ] _Reduce the Time Spent Computing Logarithms_
std::log continues to show up as high on the profiles. We have an approximate fast log implementation ml::maths::CTools::fastLog which we should be able to roll out to more places in the code. This can't be done indiscriminately because not all our uses can suffer the loss in precision, but there are definitely more places where this would be possible.

   There are also approaches for using SSE instructions to accelerate logs on collection of values and we could perhaps refactor parts of the code to exploit these. There is quite a lot of relevant material online: of particular interest are the references in [this](https://stackoverflow.com/questions/8902971/logarithm-with-sse-or-switch-to-fpu) post and [this](https://github.com/romeric/fastapprox) repository of fast approximate special functions.
- [ ] _Investigate Alternative Unordered Container Implementations_
We have had situations in the past where data gatherering has taken a significant proportion of the runtime. In these cases various specific hash map lookups were the bottleneck. We should investigate alternative hash map implementations.

[...]

PR #73, opened 2018-04-24
[ML] Speed up trend model component prediction

tveasey:
Whilst profiling end-of-bucket processing to evaluate #54 another hotspot was uncovered.

We have multiple trend models and use a weighted combination of these to make a prediction at some future time. However, depending on the time the prediction is required and the current decay rate, a number of these will have very low weight. We can exclude these altogether from the prediction and save the time to compute corresponding prediction with minimal impact on the result.

This saves another 7% on the end-of-bucket processing for the test scenario described in #53. It has a small impact on predictions for count and metric model and so a small impact on results.


droberts195:
When this is backported to 6.x please also backport 001291a25c2ef280aade119a4096a92c2b00fe96

Issue #57, opened 2018-04-16
[ML] Nginx remote_ip_url_count results regression


sophiec20:
Forst noticed in: `{
  "build" : {
    "hash" : "a47564f",
    "date" : "2018-04-09T21:45:51.347364Z"
  },
`

The ML Module Nginx jobs in 6.3 are producing significantly different results to those found in 6.0. Job config:
```
high_dc(nginx.access.url) over nginx.access.remote_ip
bucket_span=1h
incluencer=nginx.access.remote_ip
```

6.0.1 Overall score
![image](https://user-images.githubusercontent.com/4185750/38807171-ad8fe75a-4173-11e8-8295-7102dfe0cdb0.png)

6.3.0 Overall score
![image](https://user-images.githubusercontent.com/4185750/38807161-a01f01b4-4173-11e8-9927-b4e493a355bf.png)

The 6.0.1 top influencer is `72.57.0.53` on `2017-02-23`. This is not in the top 10 list for 6.3.0.

The top 10 influencer lists only share 2 remote_ips: `196.89.216.188` and `154.187.183.122`

A very notable difference is for `95.142.104.2`...
  
 `95.142.104.2` 6.0.1
![image](https://user-images.githubusercontent.com/4185750/38807635-2aae400a-4175-11e8-8e53-e91eed399f06.png)

 `95.142.104.2` 6.3.0
![image](https://user-images.githubusercontent.com/4185750/38807663-3d536e88-4175-11e8-905e-52afda362ad2.png)

Is this an intended or unintended consequence of recent improvements?

tveasey:
The latest behaviour is broadly similar to 6.0.1:
<img width="1665" alt="Screenshot 2020-12-08 at 20 31 55" src="https://user-images.githubusercontent.com/7591487/101538101-8688fe00-3994-11eb-9572-1748358653b5.png">
I think there is a case for always selecting a model using population analysis, but this is a significant change which we would need to assess across our full regression suite.

PR #74, opened 2018-04-25
[ML] Always combine duplicate samples when updating population models

tveasey:
Whilst investigating #57 it became clear that at least for this analysis we get many exact duplicates per bucket. It is reasonable to expect this may be relatively common, at least for integer valued data.

We already have a mechanism (for fuzzy de-duplication) to combine equivalent samples and re-weight the unique values. Therefore, it is relatively easy to always combine identical values.

De-duplicating has two advantages: 1) it makes the job of clustering somewhat easier since we only visit the values serially, 2) for cases such as the analysis described in #57 we get significant runtime improvements. In particular, with this change we get more stable model selection and the runtime to process all 14M records on my laptop drops from around 9m40s to around 6m40s for this job.

This will potentially affect count and metric population analysis results. The changes should be relatively small and will depend on the degree of duplication for the particular data set and function.

(I created a separate change for test changes to aid review.)

### DIFF
--- a/lib/model/CModelTools.cc
+++ b/lib/model/CModelTools.cc
@@ -124,8 +124,7 @@ void CModelTools::CFuzzyDeduplicate::add(TDouble2Vec value) {
 
 void CModelTools::CFuzzyDeduplicate::computeEpsilons(core_t::TTime bucketLength,
                                                      std::size_t desiredNumberSamples) {
-    m_Quantize = m_Count > 0;
-    if (m_Quantize) {
+    if (m_Count > 0) {
         m_QuantizedValues.reserve(std::min(m_Count, desiredNumberSamples));
         m_TimeEps = std::max(bucketLength / 60, core_t::TTime(1));
         m_ValueEps.assign(m_RandomSample[0].size(), 0.0);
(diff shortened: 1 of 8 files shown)


lib/model/CModelTools.cc
```diff
@@ -169,7 +164,7 @@ CModelTools::TDouble2Vec CModelTools::CFuzzyDeduplicate::quantize(TDouble2Vec va
 }
 
 core_t::TTime CModelTools::CFuzzyDeduplicate::quantize(core_t::TTime time) const {
-    return maths::CIntegerTools::floor(time, m_TimeEps);
+    return m_TimeEps > 0 ? maths::CIntegerTools::floor(time, m_TimeEps) : time;
```
davidkyle: Is there no need for a similar check in `quantize(TDouble2Vec ` where `m_ValueEps` has not be initialised with a call to `computeEpsilons`? That is `m_ValueEps` has length > 0.

tveasey: Yes good catch, the loop should have been changed to be over `m_ValueEps.size()` as part of the changes to make this de-duplicate identical values even when `computeEpsilons` hasn't been called.

tveasey:
In terms of the non-test code changes, these breakdown as: 1) always de-duplicate identical population samples, but only fuzzily de-duplicate when the number of people per bucket is `> s_MinimumToFuzzyDeduplicate`, 2) some changes to get `CFuzzyDeduplicate` to support 1) i.e. work as expected when the epsilons are "zero", 3) changes to move some otherwise unnecessary work to after the checks to decide whether or not to update the model, and 4) changes to variable naming to reflect the new behaviour.

I did also do some tidying up in `CEventRatePopulationModelTest` because I needed to fix one of these tests as fall out and `CModelToolsTest` because I was adding some testing here anyway.

lib/model/CModelTools.cc
davidkyle: Odd though as that [unit test](https://github.com/tveasey/ml-cpp-1/blob/081b7a4bcd96e07b4f5d062f8f9bd26eae896582/lib/model/unittest/CModelToolsTest.cc#L95) hits that code and I assume it passes

tveasey: I know, it is surprising. Even more surprisingly is it worked as expected on a run of population analysis on nginx from #57. Strictly the behaviour is "undefined" so it can/does work magically.

In fact, I think the reason it often works is that the type is a small vector, which uses static memory for the first two elements. I suspect that these can easily be <= 0. I tried running valgrind on the unit test and it did indeed show up `conditional jump or move depends on uninitialised value`, so the fact that the `m_ValueEps[0]` turns out to often be less than or equal to zero is chance, it is just seemingly highly likely to occur.

Issue #76, opened 2018-04-27
[ML] Probable SIGSEGV due to out-of-bounds access in updateRecycledModels


tveasey:
We've had a crash reported against 5.6.8 (possibly due to incorrect state upgrade from 5.5 -> 5.6.8) with stack trace:

```
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `./autodetect --jobid=webscan_rare_responses_v2 --licenseValidation=213'.
Program terminated with signal 11, Segmentation fault.
...
(gdb) bt
#0  0x00007f4b52d48eaf in ml::model::CModel::updateRecycledModels() () from /mnt/Downloads/libs/libMlModel.so
#1  0x00007f4b52d73466 in ml::model::CPopulationModel::updateRecycledModels() () from /mnt/Downloads/libs/libMlModel.so
#2  0x00007f4b52d6d6e9 in ml::model::CPopulationModel::createUpdateNewModels(long, ml::model::CResourceMonitor&) () from /mnt/Downloads/libs/libMlModel.so
#3  0x00007f4b52bb98ce in ml::model::CEventRatePopulationModel::sample(long, long, ml::model::CResourceMonitor&) () from /mnt/Downloads/libs/libMlModel.so
#4  0x00007f4b52aee6c9 in ml::model::CAnomalyDetector::sample(long, long, ml::model::CResourceMonitor&) () from /mnt/Downloads/libs/libMlModel.so
[...]

PR #79, opened 2018-05-01
[ML] Check range in bounds when clearing recycled models

tveasey:
Add failsafes to ensure that we never try and reset state for identifiers outside vector bounds. This shouldn't happen, but this is the likely cause of #76 and there are no side effects from checking and issuing an error in such cases. Investigating and fixing the root cause of #76 will be the subject of separate PR.

### DIFF
--- a/lib/model/CAnomalyDetectorModel.cc
+++ b/lib/model/CAnomalyDetectorModel.cc
@@ -451,8 +451,8 @@ void CAnomalyDetectorModel::updateRecycledModels() {
         if (pid < m_PersonBucketCounts.size()) {
             m_PersonBucketCounts[pid] = 0.0;
         } else {
-            LOG_ERROR("Recycled person identifier '"
-                      << pid << "' out-of-range [," << m_PersonBucketCounts.size() << ")");
+            LOG_ERROR(<< "Recycled person identifier '" << pid << "' out-of-range [,"
+                      << m_PersonBucketCounts.size() << ")");
         }
     }
     people.clear();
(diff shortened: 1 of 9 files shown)


lib/model/CPopulationModel.cc
```diff
@@ -445,16 +445,24 @@ void CPopulationModel::createNewModels(std::size_t n, std::size_t m) {
...
-            m_PersonAttributeBucketCounts[cid] = *m_NewPersonBucketCounts;
+        if (cid < m_AttributeFirstBucketTimes.size()) {
+            m_AttributeFirstBucketTimes[cid] = CAnomalyDetectorModel::TIME_UNSET;
+            m_AttributeLastBucketTimes[cid] = CAnomalyDetectorModel::TIME_UNSET;
+            m_DistinctPersonCounts[cid] = m_NewDistinctPersonCounts;
+            if (m_NewPersonBucketCounts) {
+                m_PersonAttributeBucketCounts[cid] = *m_NewPersonBucketCounts;
+            }
+        } else {
+            LOG_ERROR("Recycled attribute identifier '"
```
droberts195: There's a missing `<<` after the `LOG_ERROR(`.

Also, have you run `clang-format` on this file?  CI is set to fail now if the formatting isn't what `clang-format` would choose.

droberts195:
I think it would be good to make the same fix in 5.6.10 as well, since that was where the problem was reported, and some people will be using 5.6 until early next year.

lib/model/CAnomalyDetectorModel.cc
```diff
@@ -448,7 +448,12 @@ void CAnomalyDetectorModel::createNewModels(std::size_t n, std::size_t /*m*/) {
 void CAnomalyDetectorModel::updateRecycledModels() {
     TSizeVec& people{m_DataGatherer->recycledPersonIds()};
     for (auto pid : people) {
-        m_PersonBucketCounts[pid] = 0.0;
+        if (pid < m_PersonBucketCounts.size()) {
+            m_PersonBucketCounts[pid] = 0.0;
+        } else {
+            LOG_ERROR("Recycled person identifier '"
```
edsavage: Missing '<<' at the start of the error message

tveasey:
Agreed Dave, I was planning to make the necessary changes in that repo too.

PR #81, opened 2018-05-02
[ML] Correct function description for population lat_long

### LINKED ISSUE #80: [ML] function_description set to mean for lat_long population detector
`function_description` is being set to `mean` in records for a `lat_long` detector with an `over` field. This should be set to `lat_long`, in line with the behaviour for non-population analyses:

image

Found in 7.0.0-alpha1, but I believe this has been the long-standing behaviour.

Detector config used:

```
"analysis_config": {
    "bucket_span": "15m",
    "detectors": [
      {
        "detector_description": "lat-long(location)",
        "function": "lat_long",
        "field_name": "lat-lon",
        "over_field_name": "apache2.access.remote_ip",
        "detector_index": 0
      }
    ],
    "influencers": [
      "apache2.access.remote_ip"
    ]
  }
```

Note that `function_description` is used in the UI as a display hint field, and affects for example the columns in the anomalies table.

dimitris-athanasiou:
Closes #80

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -41,6 +41,8 @@
 
 === Bug Fixes
 
+Function description for population lat_long results should be lat_long instead of mean ({pull}81[#81])
+
 === Regressions
 
 === Known Issues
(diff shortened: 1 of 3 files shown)


droberts195:
I think the new process is to add a Changelog entry as part of the PR.  Eventually a bot is supposed to enforce this but it hasn't been set up yet.

dimitris-athanasiou:
@droberts195 I added an entry in the changelog.

PR #84, opened 2018-05-02
[ML] Improve and use periodic boundary condition in seasonal component

tveasey:
We were previously not using periodic boundary conditions when fitting seasonal components. This is because the value need not be the same at the start and end of the period because of variation in the gradient within the component. A better strategy is to use the predicted value in the next and previous period, since we expect the component to be continuous. As a result, we can enable periodic boundary conditions by default.

This gives a smoother fit to periodic signals. We get lower maximum prediction errors for smooth signals as a result which is reflected in some of the tighter unit test thresholds and also some manual testing I've done. This will affect results for event rate and metric analysis of periodic signals.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -15,6 +15,8 @@
 
 === Enhancements
 
+Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
+
 === Bug Fixes
 
 === Regressions
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -15,8 +15,6 @@
 
 === Enhancements
 
-Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
-
 === Bug Fixes
 
 === Regressions
@@ -28,6 +26,8 @@ Improve and use periodic boundary condition for seasonal component modeling ({pu
 === New Features
 
 === Enhancements
+
+Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
 
 === Bug Fixes
 
--- a/lib/maths/CAdaptiveBucketing.cc
+++ b/lib/maths/CAdaptiveBucketing.cc
@@ -345,6 +345,8 @@ bool CAdaptiveBucketing::knots(core_t::TTime time,
                                TDoubleVec& knots,
                                TDoubleVec& values,
                                TDoubleVec& variances) const {
+    using TSizeVec = std::vector<std::size_t>;
+
     knots.clear();
     values.clear();
     variances.clear();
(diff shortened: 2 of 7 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -15,6 +15,8 @@
 
 === Enhancements
 
+Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
```
dimitris-athanasiou: This is labelled to go in 6.4 so I think the release note should also be in the 6.4 section. Still getting my head around it though :-)

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
```diff
@@ -1872,11 +1873,11 @@ void CTimeSeriesDecompositionTest::testYearly() {
         }
     }
 
-    std::ofstream file;
-    file.open("results.m");
-    TDoubleVec f;
-    TTimeVec times;
-    TDoubleVec values;
+    //std::ofstream file;
```
dimitris-athanasiou: Should these be removed?

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
```diff
@@ -1890,9 +1891,9 @@ void CTimeSeriesDecompositionTest::testYearly() {
             maths::CBasicStatistics::mean(decomposition.baseline(time, 0.0));
         double error = std::fabs((prediction - trend) / trend);
         meanError.add(error);
-        times.push_back(time);
-        values.push_back(trend);
-        f.push_back(prediction);
+        //times.push_back(time);
```
dimitris-athanasiou: Should these be removed?

lib/maths/CAdaptiveBucketing.cc
```diff
@@ -400,16 +403,68 @@ bool CAdaptiveBucketing::knots(core_t::TTime time,
...
-                values[0] = (values[0] + values.back()) / 2.0;
-                variances[0] = (variances[0] + variances.back()) / 2.0;
-                knots.push_back(m_Endpoints[n]);
-                values.push_back(values[0]);
-                variances.push_back(variances[0]);
+                // We search for the value in the last and next period which
+                // are adjacent. Note that values need not be the same at the
+                // start and end of the period because the gradient can vary,
+                // but we expect them to be continuous.
+                for (std::size_t j = n - 1; j > 0; --j) {
```
dimitris-athanasiou: Have you compared performance after this change?

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
tveasey: These are useful to have in if there are changes in results. You can uncomment these and look at the modelling we do.

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
tveasey: As above.

lib/maths/CAdaptiveBucketing.cc
tveasey: This only happens very infrequently (when we interpolate which is once per period) so typically once per day, week, etc. As a result these functions are never bottlenecks. For this specific code, n is small and typically this loop exits after the first iteration (only carries on for sparse data). The unit tests didn't noticeably slowdown, so I think this is probably safe from that perspective.

Issue #87, opened 2018-05-03
[ML] Large outliers can impair our ability to correctly identify and model periodicity


tveasey:
Currently, we don't have explicit handling of outliers when 1) testing for periodic patterns in the data and 2) when initialising new seasonal components. This can sometimes slowdown or prevent us detecting the correct periodic components present in a signal. We had a support case raised where this was the underlying reason for us failing to detect an outage with sufficient severity.

tveasey:
I have a reasonable implementation for this. The strategy is relatively simple: we iteratively refine definition of outliers with respect to the period under test and reweight those samples when computing test statistics such as variance reduction and cyclic autocorrelation. I use the same iterative reweighting scheme to initialise the seasonal components. As a result, we get faster detection of the correct seasonal components in signals which are distorted by anomalous features as we learn and more accurate initial modelling.

For example,
Before:
![screen shot 2018-05-04 at 14 28 05](https://user-images.githubusercontent.com/7591487/39631070-20a39050-4fa9-11e8-9ce3-efbc3f5c77ed.png)
After:
![screen shot 2018-05-04 at 14 28 44](https://user-images.githubusercontent.com/7591487/39631084-29a8c350-4fa9-11e8-9b55-665854d6d56f.png)

Similarly,
Before:
![screen shot 2018-05-04 at 14 57 59](https://user-images.githubusercontent.com/7591487/39632262-7c275512-4fac-11e8-834b-24dd9b1b68de.png)
![screen shot 2018-05-04 at 14 59 23](https://user-images.githubusercontent.com/7591487/39632327-b4d76410-4fac-11e8-8334-34a1ddbf0f51.png)

After:
![screen shot 2018-05-04 at 15 01 12](https://user-images.githubusercontent.com/7591487/39632282-8a6bc054-4fac-11e8-8006-b56bdbd2e3ee.png)
![screen shot 2018-05-04 at 14 59 34](https://user-images.githubusercontent.com/7591487/39632373-d02b6748-4fac-11e8-9915-1b1754b58310.png)

tveasey:
Closing as I don't have any further plans than what was implemented in #90 at present.

PR #88, opened 2018-05-03
[ML] Correct ageing of windowed seasonal components

tveasey:
Investigating a forecasting regression associated with changes targeting #87 showed up an issue with the way we age periodically repeating partitions of the data (such as weekday/end splits). In particular, we were ageing them too fast and their decay rate should be prorated by the fraction of values with which they are updated. For subtle reasons, related to interaction between components in the decomposition, this could lead to biased predictions.

This can affect results for metric and count analyses when the data has distinct weekday/end repeating pattern.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -31,6 +31,8 @@ Improve and use periodic boundary condition for seasonal component modeling ({pu
 
 === Bug Fixes
 
+Age seasonal components in proportion to the fraction of values with which they're updated ({pull}88[#88])
+
 === Regressions
 
 === Known Issues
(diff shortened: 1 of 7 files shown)


hendrikmuhs:
LGTM, only misses the changelog I think.

tveasey:
Thanks @hendrikmuhs, I updated the change log and made your suggested change.

hendrikmuhs:
@tveasey 

Just to clarify, when you say "regression" in your description you mean regression w.r.t. an unreleased version?

tveasey:
That's correct. Whilst evaluating changes related to #88, I noticed we were getting significantly worse forecasts in one of the unit tests. This turned out to be because we identified a weekday/end cyclic partition, which we didn't before. The fact that the results were worse was down to the problem this PR addresses.

PR #90, opened 2018-05-04
[ML] Improve robustness w.r.t. outliers of detection and initialisation of seasonal components

tveasey:
This makes two principle changes:
1) Iteratively reweights outliers w.r.t. the seasonal component under test and for initialisation. These are defined as a fraction of values with highest residual w.r.t. the component's predictions.
2) Switches marginal test decisions for decomposition components to use logistic regression on top of the various factors, i.e. variance reduction, autocorrelation, number of periods of data observed, etc.

I've tested this on a variety of synthetic and real examples where initial periodic patterns are distorted by outliers and this approach has proved effective, see #87 for more details. Otherwise, it seems to have no detrimental effects.

This will affect results on count and metric analyses when there is seasonality and significant distortion due to outliers.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@
 === Enhancements
 
 Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
+Improve robustness w.r.t. outliers of detection and initialisation of seasonal components ({pull}90[#90])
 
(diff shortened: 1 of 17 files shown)


hendrikmuhs:
Please additionally do

```
git cherry-pick 18ebdd67f7ba2b7f861607e248d79e1cca8c7891
```

when backporting this to 6.x

PR #91, opened 2018-05-11
[ML] Extend unit testing and make some improvements to deal with changes in seasonal components

tveasey:
In the course of other testing related to #6, I found evidence of potential instability in the decomposition when the seasonal components in the time series abruptly change. In addition, our lifecycle management of seasonal components was not reliably removing components which were no longer improving predictions. This was related to the way we were trying to preserve components which carry information about periodic variance in the data.

This adds a unit test to exercise these problems and makes some related improvements. As part of these I've implemented a failsafe to remove all seasonal components and start again if instability occurs. The current behaviour on this unit test (red actuals, blue predictions) is as follows:
screen shot 2018-05-10 at 18 11 58
screen shot 2018-05-10 at 18 14 53
screen shot 2018-05-10 at 18 15 42
screen shot 2018-05-10 at 18 17 16
screen shot 2018-05-10 at 18 17 56

For reference, this is what was happening before:
screen shot 2018-05-15 at 09 56 29

This can affect results on metric and count analysis with seasonal signals where seasonality changes significantly from time to time.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@
 
 Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
 Improve robustness w.r.t. outliers of detection and initialisation of seasonal components ({pull}90[#90])
+Extend unit testing and make some improvements to deal with changes in seasonal components ({pull}91[#91])
 
 === Bug Fixes
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -29,7 +29,7 @@
 
 Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
 Improve robustness w.r.t. outliers of detection and initialisation of seasonal components ({pull}90[#90])
-Extend unit testing and make some improvements to deal with changes in seasonal components ({pull}91[#91])
+Improve behavior when there are abrupt changes in the seasonal components present in a time series ({pull}91[#91])
 Explicit change point detection and modelling ({pull}92[#92])
 
 === Bug Fixes
--- a/lib/maths/CSeasonalComponent.cc
+++ b/lib/maths/CSeasonalComponent.cc
@@ -23,7 +23,7 @@
 #include <boost/math/distributions/chi_squared.hpp>
 #include <boost/math/distributions/normal.hpp>
 
-#include <ios>
+#include <cmath>
 #include <vector>
 
 namespace ml {
@@ -148,7 +148,6 @@ void CSeasonalComponent::interpolate(core_t::TTime time, bool refine) {
     if (refine) {
         m_Bucketing.refine(time);
     }
-
     TDoubleVec knots;
     TDoubleVec values;
     TDoubleVec variances;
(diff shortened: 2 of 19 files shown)


lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -255,6 +258,7 @@ const TSizeVecVec SC_TRANSITION_FUNCTION{
     TSizeVec{SC_NORMAL, SC_NORMAL, SC_NORMAL, SC_NORMAL}};
 
 const std::string VERSION_6_3_TAG("6.3");
+const std::string VERSION_6_4_TAG("6.4");
```
droberts195: We probably want to prevent a job that has written 6.4 state from running on a 6.3 node in a mixed 6.3/6.4 cluster.  I guess if this isn't done then a couple of fields won't get initialised.  The way to do that is to change [line 78 of CAnomalyJob.cc](https://github.com/elastic/ml-cpp/blob/d5211a76f9227fb9cbbb44dd10c0126502bc5ef7/lib/api/CAnomalyJob.cc#L78).

docs/CHANGELOG.asciidoc
```diff
@@ -29,6 +29,7 @@
 
 Improve and use periodic boundary condition for seasonal component modeling ({pull}84[#84])
 Improve robustness w.r.t. outliers of detection and initialisation of seasonal components ({pull}90[#90])
+Extend unit testing and make some improvements to deal with changes in seasonal components ({pull}91[#91])
```
droberts195: I don't think the release notes need to mention unit test changes - someone trying to find out what's new in ML isn't going to care.

lib/maths/CSeasonalComponent.cc
```diff
@@ -210,29 +207,18 @@ double CSeasonalComponent::delta(core_t::TTime time,
...
-
-        // We have a smooth decision boundary for whether to apply
-        // a delta for the case that the difference from the mean
-        // is 1/3 of the range. We force the delta to zero for values
-        // significantly smaller than this.
-        double scale{CTools::logisticFunction(3.0 * min[0] / minmax.range(), 0.1, 1.0)};
-        scale = CTools::truncate(1.002 * scale - 0.001, 0.0, 1.0);
-
-        return -scale * min[0] * CTools::sign(shortPeriodValue);
+        return bias.signMargin() != 0.0 ? bias.signMargin() : cancellation.signMargin();
```
droberts195: Would you want to treat very small numbers as 0?  For example, if `bias.signMargin()` was 1e-308 would you still want to use it?  (This is only a question - if the answer is yes then fine.)

docs/CHANGELOG.asciidoc
tveasey: Fair enough, I'll update.

lib/maths/CSeasonalComponent.cc
tveasey: One could, for example, not bother if it is << amplitude. The chances of this being very small, but not identically zero, is actually small. If there is no bias then we expect the differences at the different times in the period to *not* all have the same sign with a high probability because noise in the underlying values should cause them to randomly land on different sides of the mean. Also, it doesn't have a significant impact to have a small, but non-zero, delta. So on balance I don't think this warrants the extra complication.

tveasey:
This turned out to be more involved: testing with variable decay rate I again hit the instability. I've therefore significantly extended the original PR:
1) I did some more work on the delta calculation. In particular, it minimises the sum component amplitude.
2) I added in active monitoring for signs of instability, by inspecting the sum component amplitude, and reduce the gain if the modelling displays any symptoms of instability.
3) I made sure we remove *all* components if the prediction accuracy drops below the reference.

I split the extra changes into two commits. The first is a pure refactor to better encapsulate seasonal and calendar components. The second makes the functional changes. Can you take another look @droberts195.

PR #92, opened 2018-05-11
[ML] Change point detection and prediction

tveasey:
This implements a mechanism to detect specific types of change points in a time series; specifically, 1) step change, 2) linear scaling and 3) time shift (e.g. for daylight saving). These are then applied to the time series model in one shot (rather than gradually relearning its parameters). In addition, it introduces a new probabilistic model to predict when step changes might occur in the future and uses this to roll out possible future paths for the time series including steps.

The fact that we can now detect specific types of change points has also meant that we can increase the model robustness w.r.t. outliers, since behaviour at significant change points doesn't as strongly rely on being able to adjust the model parameters with unusual values.

From an anomaly detection perspective this means we can adjust the model with significantly less growth in prediction uncertainty when the time series undergoes one of the modelled class of changes. This results in fewer blind spots in detection for such signals. From a forecasting perspective, signals with predictable discontinuities, such a garbage collection, are much more accurately predicted.

This PR commits the long running task on change point detection and modelling to master. Note that these changes have already been individually reviewed.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -210,8 +210,9 @@ double tailWeight(const CPrior& prior, double derate, double scale, double value
     } else if (upperBound >= MINUS_LOG_TOLERANCE) {
         f = std::exp(-(lowerBound + upperBound) / 2.0);
         f = std::min(f, 1.0 - f);
-   } else if (!prior.minusLogJointCdfComplement(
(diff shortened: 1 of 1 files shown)


droberts195:
When this is backported to 6.x there are now _two_ extra fixes that need cherry-picking at the same time: d89d6b59914166cc6907556988dab9fd65f0e2fc and 799d6b9e8bdff9126d378a173e11d325b9cbe082

PR #95, opened 2018-05-16
[ML] Improve change point detection for long bucket lengths

tveasey:
This PR does some hardening to change detection based on further testing, particularly for longer bucket lengths. By way of example, the following is the current behaviour on a problematic data set:
30m bucketing
screen shot 2018-05-16 at 12 59 48
1h bucketing
screen shot 2018-05-16 at 13 03 50
2h bucketing
screen shot 2018-05-16 at 13 01 40

After these changes we get:
30m bucketing
screen shot 2018-05-16 at 12 21 55
1h bucketing
screen shot 2018-05-16 at 12 20 46
2h bucketing
screen shot 2018-05-16 at 12 21 25

I have committed these changes incrementally in logical groups, so it may be easier to review them a commit at a time.

The following are made in the last commit:
1. Improve the methodology for deciding if we'll eventually accept a change. Specifically, we can observe all the values for the decision function over the change detection window and use these to predict whether we'll eventually accept the change. We use this to:
    - Better adjust the weights of values added to the trend during change period,
    - Better decide if it is worth continuing to test.
2. Improve the decision to clear the data structures we use for periodicity testing based on the historical data characteristics and the nature of the change.
3. Require a longer period of unusual values to trigger change detection. I've traded this for a slightly higher threshold on the individual p-values. There are downsides to testing for a change: i) it temporarily increases the model memory size, ii) we don't have concurrent change detection, so if the test is already running, it can interfere with detection if the test window overlaps a real change point. On balance, it is therefore better to wait slightly longer to see if the event is short-lived before committing to run the full test.
4. Decrease the time we'll take to detect a change for long bucket lengths. We can get acceptable test accuracy observing fewer values and with the current parameterisation we were waiting too long for a common case of scaling daily periodicity if the job used long bucket lengths.

Making these changes also uncovered an improvement to make to periodicity testing following on from #90. Specifically, we should have been weighting the samples when computing the level for the amplitude test to better deal with outliers. This is made in the first commit.

This can affect results for data with change points particularly if using longer bucket lengths. Marking as a non-issue since this is a refinement to an unreleased version.


lib/maths/CTimeSeriesChangeDetector.cc
```diff
@@ -178,40 +195,36 @@ double CUnivariateTimeSeriesChangeDetector::decisionFunction(std::size_t& change
...
+        evidences[0] / EXPECTED_EVIDENCE_THRESHOLD_MULTIPLIER / expectedEvidence,
+        std::max(static_cast<double>(m_TimeRange.range() - m_MinimumTimeToDetect), 0.0) /
+            static_cast<double>(m_MaximumTimeToDetect - m_MinimumTimeToDetect)};
     double p{CTools::logisticFunction(x[0], 0.05, 1.0) *
              CTools::logisticFunction(x[1], 0.1, 1.0) *
              (x[2] < 0.0 ? 1.0 : CTools::logisticFunction(x[2], 0.2, 1.0)) *
              CTools::logisticFunction(x[3], 0.2, 0.5)};
-    LOG_TRACE("p(" << (*candidates[0].second)->change()->print() << ") = " << p
-                   << " | x = " << core::CContainerPrinter::print(x));
+    LOG_TRACE(<< "df(" << (*candidates[0].second)->change()->print() << ") = " << p / 0.03125
```
droberts195: It would make it clearer to the reader to use `MAXIMUM_DECISION_FUNCTION * p` rather than `p / 0.03125` even though they evaluate to the same thing.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -599,10 +599,22 @@ void CTimeSeriesDecompositionDetail::CPeriodicityTest::test(const SAddValue& mes
     }
 }
 
-void CTimeSeriesDecompositionDetail::CPeriodicityTest::clear(ETest test, core_t::TTime time) {
-    if (m_Windows[test] != nullptr) {
-        m_Windows[test].reset(this->newWindow(test));
-        m_Windows[test]->initialize(time);
+void CTimeSeriesDecompositionDetail::CPeriodicityTest::maybeClear(core_t::TTime time,
+                                                                  double shift) {
+    for (auto& test : {E_Short, E_Long}) {
```
droberts195: Could it just be `auto` for iterating simple enum values?

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -599,10 +599,22 @@ void CTimeSeriesDecompositionDetail::CPeriodicityTest::test(const SAddValue& mes
...
+        if (m_Windows[test] != nullptr) {
+            TDoubleVec values;
+            values.reserve(m_Windows[test]->values().size());
+            for (const auto& value : m_Windows[test]->values()) {
+                if (CBasicStatistics::count(value) > 0.0) {
+                    values.push_back(CBasicStatistics::mean(value));
+                }
+            }
+            if (shift > 1.4826 * CBasicStatistics::mad(values)) {
```
droberts195: The magic `1.4826` could be a symbolic constant or at least commented.

lib/maths/CTimeSeriesChangeDetector.cc
tveasey: agreed

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: This was an oversight from originally iterating over the windows rather than the enums

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: This corrects for bias when estimating the standard deviation (for normal data). I've pulled out the constant as you suggest.

PR #98, opened 2018-05-21
[ML] Add system call restrictions to the ML processes 

davidkyle:
The `autodetect`, `autoconfig`, `categorize` and `normalize` programs all require the ability to create a named pipe (mkfifo/CreateNamedPipe) open a connection to a named pipe and read & write to those pipes. All other system calls which are not required such as fork/exec should be restricted using platform specific APIs.

The code here is similar to [SystemCallFilter.java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/bootstrap/SystemCallFilter.java).

## LibSeccomp
This PR adds a new _static_ library `libseccomp` containing a single class `CSystemCallFilter` the library must be linked in the 4 main programs listed above. 

Due to static linkage remember to `make relink` the unit test and binaries after making a change.

## macOS
macOs has the most configurable sandboxing functionality the approach here is to blacklist everything by default but allow `read*` and `write*` for open, mkfifo, read and write. This imposes harder limits than the [SystemCallFilter.java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/bootstrap/SystemCallFilter.java#L448) which only denies fork and exec.

## Linux
Seccomp BPF was added to `prctl()` in kernal 3.5, no filters will be installed in older kernals. The dedicated `seccomp()` function was added in kernal 3.17 and provides the same the functionality as `prctl()` but with the [SECCOMP_FILTER_FLAG_TSYNC](http://man7.org/linux/man-pages/man2/seccomp.2.html) option that is not required here if the filters are installed by the main thread before any other threads are spawned. Because of this we use the older more compatible `prctl()` function.

[SystemCallFilter.java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/bootstrap/SystemCallFilter.java#L391) tries to use `seccomp()` and falls back to `prctl()` 

## Windows 
The Windows implementation uses [Job Objects](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684161(v=vs.85).aspx) to prevent the process spawning another.


droberts195:
> execve, fork and vfork are disallowed should any other system calls be? fexecve? execveat?

Yes, `execveat` should definitely be disallowed (if it exists).  I guess the complication here is that it only exists on kernel 3.19 and above.  (And I think `fexecve` is the libc function that uses `execveat` rather than a system call itself?)

I think also `socket`, `socketcall` and `socketpair` should be disallowed, because we wanted to prevent network access.  If we stop people getting hold of a socket then hopefully that will avoid the need to list every single other network function.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+        // is installed by the main thread before any other threads are
+        // spawned.
+        if (prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &prog)) {
+            LOG_ERROR("Unable to install Seccomp BPF");
+        } else {
+            LOG_DEBUG("Seccomp BPF installed");
+        }
+
+    } else {
+        LOG_DEBUG(<< "Seccomp BPF not available");
```
hendrikmuhs: could imagine this being info as it might be useful for customers to know

droberts195: I'm not sure.  Upgrading this would mean customers running on CentOS 6 would be spammed every time an ML process starts up.  And `normalize` can run pretty frequently...

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+    (allow file-read*) \
+    (allow file-read-data) \
+    (allow file-write*) \
+    (allow file-write-data)");
+
+std::string getTempDir() {
+    // In production this needs to match the setting of java.io.tmpdir.  We rely
+    // on the JVM that spawns our controller daemon setting TMPDIR in the
+    // environment of the spawned process.
+    const char* tmpDir(::getenv("TMPDIR"));
```
hendrikmuhs: consider using boost for that, it handles some more cases afaik and can be easily replaced with std once we switch to C++17

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+    if (path[path.length() - 1] != '/') {
+        path += '/';
+    }
+    return path;
+}
+
+std::string getTempRulesFilename() {
+    std::string tempDir = getTempDir();
+    // randomise the sandbox rules filename
+    char * tempName = ::tempnam(tempDir.c_str(), "ml");
```
hendrikmuhs: consider `boost::filesystem::unique_path`

include/seccomp/CSystemCallFilter.h
```diff
@@ -0,0 +1,28 @@
...
+#define INCLUDED_ml_seccomp_CSystemCallFilter_h
+
+namespace ml {
+namespace seccomp {
+
+//! \brief
+//! Installs Sandbox/sys call filters
+//!
+//! DESCRIPTION:\n
+//!
```
droberts195: Could say the intention is to prevent starting new processes and, where possible, preventing network access.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+
+#include <linux/audit.h>
+#include <linux/filter.h>
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+
+
+namespace ml {
+namespace seccomp {
+
```
droberts195: Everything from here down to the end of `canUseSeccompBpf()` should be in an unnamed namespace to enforce internal linkage.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+// The x64 architecture should fail these calls
+unsigned int UPPER_NR_LIMIT = 0x3FFFFFFF;
+
+// Offset to the nr field in struct seccomp_data
+unsigned int SECCOMP_DATA_NR_OFFSET   = 0x00;
+// Offset to the arch field in struct seccomp_data
+unsigned int SECCOMP_DATA_ARCH_OFFSET = 0x04;
+
+// Copied from seccomp.h
+// seccomp.h cannot be included as it was added in Linux kernal 3.17
```
droberts195: typo: kernal -> kernel

lib/seccomp/CSystemCallFilter_Linux.cc
davidkyle: It started as info level but as Dave points out every time normalize starts the message is logged and this can spam the log files

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+
+
+namespace ml {
+namespace seccomp {
+
+// The old x32 ABI always has bit 30 set in the sys call numbers.
+// The x64 architecture should fail these calls
+unsigned int UPPER_NR_LIMIT = 0x3FFFFFFF;
```
droberts195: I think this and the two below should be `const std::uint32_t` (plus `#include <cstdint>`).  The struct member they're used with is explicitly 32 bits, and whilst `unsigned int` is 32 bits on all the compilers we're currently using it's best practice to specify the size exactly where the consuming code does.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+
+// Offset to the nr field in struct seccomp_data
+unsigned int SECCOMP_DATA_NR_OFFSET   = 0x00;
+// Offset to the arch field in struct seccomp_data
+unsigned int SECCOMP_DATA_ARCH_OFFSET = 0x04;
+
+// Copied from seccomp.h
+// seccomp.h cannot be included as it was added in Linux kernal 3.17
+// and this must build on older versions.
```
droberts195: Could add a TODO to remove these once our oldest supported build platform has a suitable `seccomp.h`.

lib/seccomp/CSystemCallFilter_MacOSX.cc
davidkyle: I'm not sure what you mean is there a boost version of getenv? I'm using the temp dir set by elasticsearch

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+#define SECCOMP_RET_ERRNO   0x00050000U
+#define SECCOMP_RET_ALLOW   0x7fff0000U
+#define SECCOMP_RET_DATA    0x0000ffffU
+
+// Added in Linux 3.5
+#ifndef PR_SET_NO_NEW_PRIVS
+#define PR_SET_NO_NEW_PRIVS 38
+#endif
+
+struct sock_filter filter[] = {
```
droberts195: I think this should be `const` and called `FILTER`.  That will probably mean using `const_cast` at the point of use, but our coding style says to prefer making things that are logically constant `const`.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+    // with CONFIG_SECCOMP_FILTER
+    // http://man7.org/linux/man-pages/man2/prctl.2.html
+    int result = prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, nullptr);
+    int configError = errno;
+    if (result != -1) {
+        LOG_ERROR(<< "prctl set seccomp should have failed");
+        return false;
+    }
+
+    // If the kernal is not configured with CONFIG_SECCOMP_FILTER
```
droberts195: typo: kernal -> kernel

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+CSystemCallFilter::CSystemCallFilter() {
+    if (canUseSeccompBpf()) {
+        LOG_DEBUG(<< "Seccomp BPF filters available");
+
+        // Ensure more permissive privileges cannot be set in future.
+        // This must be set before installing the filter.
+        // PR_SET_NO_NEW_PRIVS was aded in kernal 3.5
+        prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0);
```
droberts195: We should check the return value here and log if the call fails.

lib/seccomp/CSystemCallFilter_MacOSX.cc
hendrikmuhs: I mean ` boost::filesystem::temp_directory_path()`

lib/seccomp/CSystemCallFilter_Linux.cc
davidkyle: Actually we could log this message in autodetect main only. It is a useful message

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+
+        struct sock_fprog prog = {
+           .len = static_cast<unsigned short>(sizeof(filter) / sizeof(filter[0])),
+           .filter = filter,
```
droberts195: It's worth noting that this initialization syntax is not actually valid C++.  It works in this Linux specific file because we build on Linux with `-std=gnu++14` and the GNU extensions allow this C syntax.  Fine to leave it here I guess though.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+        // seccomp(SECCOMP_SET_MODE_FILTER, 0, filter which was added in
+        // kernal 3.17. We choose the older more compatible function.
+        // Note this precludes the use of calling seccomp() with the
+        // SECCOMP_FILTER_FLAG_TSYNC which is acceptable if the filter
+        // is installed by the main thread before any other threads are
+        // spawned.
+        if (prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &prog)) {
+            LOG_ERROR("Unable to install Seccomp BPF");
+        } else {
+            LOG_DEBUG("Seccomp BPF installed");
```
droberts195: Nit: missing `<<`.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -0,0 +1,113 @@
...
+        // prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, filter) was introduced
+        // in kernal 3.5. This is functionally equivalent to
+        // seccomp(SECCOMP_SET_MODE_FILTER, 0, filter which was added in
+        // kernal 3.17. We choose the older more compatible function.
+        // Note this precludes the use of calling seccomp() with the
+        // SECCOMP_FILTER_FLAG_TSYNC which is acceptable if the filter
+        // is installed by the main thread before any other threads are
+        // spawned.
+        if (prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, &prog)) {
+            LOG_ERROR("Unable to install Seccomp BPF");
```
droberts195: Nit: missing `<<`.

Also, it would be good to log the actual error, so:

```
    LOG_ERROR(<< "Unable to install Seccomp BPF: " << std::strerror(errno));
```

and `#include <cstring>` and `<cerrno>`.

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <sandbox.h>
+
+namespace ml {
+namespace seccomp {
+
+// The Sandbox rules deny all actions apart from creating fifos,
+// opening files and reading and writing.
+// (allow file-write*) is required for mkfifo and that permission
+// can not be set using the more granular controls.
+static const std::string SANDBOX_RULES("\
```
droberts195: The modern way is to use an unnamed namespace rather than static.

The functions below can also go in it so they get internal linkage.

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+    (version 1) \
+    (deny default) \
+    (debug all) \
```
droberts195: Is this appropriate for production?

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+    (allow file-read*) \
+    (allow file-read-data) \
+    (allow file-write*) \
+    (allow file-write-data)");
+
+std::string getTempDir() {
+    // In production this needs to match the setting of java.io.tmpdir.  We rely
+    // on the JVM that spawns our controller daemon setting TMPDIR in the
+    // environment of the spawned process.
```
droberts195: This comment isn't true for this use of a temp directory (it is for the other use).

I think the reason for preferring `$TMPDIR` is that it's private to the current user on macOS, so reduces the risk of other users tampering with the file somehow.

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#include "seccomp/CSystemCallFilter.h"
+
+#include <core/CLogger.h>
+
+#include <paths.h>
+#include <cstdio>
+#include <fstream>
+#include <sandbox.h>
```
droberts195: We generally separate the C and C++ headers with a blank line, C++ headers first.

lib/seccomp/CSystemCallFilter_MacOSX.cc
droberts195: I think it should be `std::tempnam()` because you `#include`d `<cstdio>` rather than `<stdio.h>`.  The C++ versions guarantee the names are in the `std` namespace and the C versions guarantee the names are in the global namespace.

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+        std::string msg("Error initializing macOS sandbox");
+        if (errorbuf != nullptr) {
+            msg += ": ";
+            msg += errorbuf;
+            sandbox_free_error(errorbuf);
+        }
+        LOG_ERROR(<< msg);
+    } else {
+        LOG_DEBUG(<< "macOS sandbox initialized");
+    }
```
droberts195: Can we delete the file here?  I think it's not necessary to keep around for the lifetime of the program as the ES Java code deletes it immediately after the sandbox is initialized.

lib/seccomp/CSystemCallFilter_Windows.cc
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#include "seccomp/CSystemCallFilter.h"
+
+#include <core/CLogger.h>
+#include <core/CWindowsError.h>
+
+#include <WinBase.h>
```
droberts195: Please use `core/WindowsSafe.h`.  If you look at the Microsoft documentation it says to include `Windows.h` rather than `WinBase.h`, so directly including `WinBase.h` may cause some transitive requirements to be missing in a different Windows SDK version.  `core/WindowsSafe.h` is a wrapper around `Windows.h` that stops it messing up some of our code.

lib/seccomp/CSystemCallFilter_Windows.cc
```diff
@@ -0,0 +1,47 @@
...
+    HANDLE job = CreateJobObject(nullptr, nullptr);
+    if (job == nullptr) {
+        LOG_ERROR(<< "Failed to create Job Object: " << ml::core::CWindowsError().errorString());
+    }
+
+    JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+
+    if (QueryInformationJobObject(job, JobObjectBasicLimitInformation,
+                                  &limits, sizeof(limits), nullptr) == 0) {
+        LOG_ERROR(<< "Error querying Job Object information: " << ml::core::CWindowsError().errorString());
```
droberts195: You can just use `<< ml::core::CWindowsError()`.  There's no need to call `errorString()`.

Same in the other places it's used.

lib/seccomp/CSystemCallFilter_Windows.cc
```diff
@@ -0,0 +1,47 @@
...
+#include <WinBase.h>
+
+
+namespace ml {
+namespace seccomp {
+
+CSystemCallFilter::CSystemCallFilter() {
+    HANDLE job = CreateJobObject(nullptr, nullptr);
+    if (job == nullptr) {
+        LOG_ERROR(<< "Failed to create Job Object: " << ml::core::CWindowsError().errorString());
```
droberts195: Should `return` here.

lib/seccomp/CSystemCallFilter_Windows.cc
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+
+    if (QueryInformationJobObject(job, JobObjectBasicLimitInformation,
+                                  &limits, sizeof(limits), nullptr) == 0) {
+        LOG_ERROR(<< "Error querying Job Object information: " << ml::core::CWindowsError().errorString());
+        return;
```
droberts195: If we return here then it would be best to call `CloseHandle()` on the handle to avoid leaking it.  Same for the two error returns below.

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    virtual void shutdown() {}
+
+private:
+    std::string m_FileName;
+    std::string m_Data;
+};
+}
```
droberts195: The end of the unnamed namespace could be moved below the `systemCall()` function below.

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+        &CSystemCallFilterTest::testSystemCallFilter));
+
+    return suiteOfTests;
+}
+
+void CSystemCallFilterTest::testSystemCallFilter() {
+#ifdef Linux
+    std::string release{ml::core::CUname::release()};
+    ml::core::CRegex semVersion;
+    semVersion.init("(\\d)\\.(\\d{1,2})\\.(\\d{1,2}).*");
```
droberts195: Could `CPPUNIT_ASSERT` that this call succeeds.

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+    ml::core::CRegex::TStrVec tokens;
+    CPPUNIT_ASSERT(semVersion.tokenise(release, tokens));
+    // Seccomp is available in kernals since 3.5
```
droberts195: typo: kernals -> kernels

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+    if (std::stoi(tokens[0]) < 3 || std::stoi(tokens[1]) < 5) {
```
droberts195: This means the test won't run on kernel version 4.4, as used on Ubuntu 16.04.  The second condition should only be checked if the first number equals 3.

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+        LOG_INFO(<< "Cannot test seccomp on linux kernals before 3.5");
```
droberts195: typo: kernals -> kernels

lib/seccomp/unittest/CSystemCallFilterTest.cc
```diff
@@ -0,0 +1,216 @@
...
+        return;
+    }
+#endif // Linux
+
+    // Ensure actions are not prohibited before the
+    // system call filters are applied
+    CPPUNIT_ASSERT(systemCall());
+
+    // // Install the filter
```
droberts195: nit: double `//`

lib/seccomp/unittest/Makefile
```diff
@@ -0,0 +1,22 @@
...
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+include $(CPP_SRC_HOME)/mk/defines.mk
+
+TARGET=ml_test$(EXE_EXT)
+
+USE_BOOST=1
+USE_BOOST_REGEX_LIBS=1
```
droberts195: You shouldn't need this as `lib/core` has it and only `lib/core` is using the Boost regex lib directly.  (If you do need it then it's a sign something isn't working as expected.)

lib/seccomp/unittest/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+LIBS:=$(LIB_ML_SECCOMP)
+
```
droberts195: For consistency the linker flags should be added here too, i.e. `LDFLAGS:=$(ML_SECCOMP_LDFLAGS)`

mk/stdcpptest.mk
```diff
@@ -15,7 +15,7 @@ REQUIRED_LIBS=$(LIB_ML_TEST) $(LIB_ML_CORE) $(LIB_ML_VER) $(CPPUNITLIBS)
 LIBS:=$(LOCALLIBS) $(filter-out $(REQUIRED_LIBS), $(LIBS)) $(REQUIRED_LIBS)
 
 CPPFLAGS+=$(INCLUDE_PATH)
-LDFLAGS:=$(UTLDFLAGS) $(LDFLAGS) $(LIB_PATH) $(ML_VER_LDFLAGS)
+LDFLAGS:=$(UTLDFLAGS) $(LDFLAGS) $(LIB_PATH) $(ML_VER_LDFLAGS) $(ML_SECCOMP_LDFLAGS)
```
droberts195: Since we're not linking the seccomp library to every unit test program I think the linker flags should also be added just in the one unit test program that tests it.

lib/seccomp/CSystemCallFilter_Linux.cc
droberts195: Remember ES itself will have logged this once at startup.  Logging in autodetect will still log 100 times if 100 jobs are started.

docs/CHANGELOG.asciidoc
```diff
@@ -35,6 +35,10 @@ Explicit change point detection and modelling ({pull}92[#92])
 Forecasting of Machine Learning job time series is now supported for large jobs by temporarily storing
 model state on disk ({pull}89[#89])
 
+Secure the ML processes by preventing system calls such as fork and exec. The Linux implemenation uses
+Seccomp BPF to intercept system calls and is available in kernels since 3.5. On Windows Job Objects prevent
+new proceses being created and macOs uses the sandbox functionality ({pull}98[#98])
```
tveasey: proceses -> processes

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,80 @@
...
+    return path;
+}
+
+std::string getTempRulesFilename() {
+    std::string tempDir = getTempDir();
+    // randomise the sandbox rules filename
+    char * tempName = ::tempnam(tempDir.c_str(), "ml");
+    if (tempName != nullptr) {
+        return std::string(tempName) + ".sb";
+        ::free(tempName);
```
tveasey: free after return?

lib/seccomp/CSystemCallFilter_MacOSX.cc
davidkyle: `mkstemps` is the safest way to do this as it avoids a race condition between creating and opening the file. I pushed a change to do this, unfortunately it uses `strlcpy` and `strlcat` and  file descriptors.

lib/seccomp/CSystemCallFilter_MacOSX.cc
```diff
@@ -0,0 +1,86 @@
...
+
+    std::unique_ptr<char[]> templateBuff(new char[tempDir.size() + sizeof(FILE_NAME_TEMPLATE)]);
+    ::strlcpy(templateBuff.get(), tempDir.c_str(), tempDir.size() + 1);
+    ::strlcat(templateBuff.get(), FILE_NAME_TEMPLATE,
+              tempDir.size() + sizeof(FILE_NAME_TEMPLATE));
+
+    // Create and open a temporary file with a random name
+    // templateBuff is updated with the new filename.
+    // 3 is the size of the suffix ".sb"
+    int fd = mkstemps(templateBuff.get(), 3);
```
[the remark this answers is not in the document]
davidkyle: Is that safe? Is `&profileFilename[0]` null terminated?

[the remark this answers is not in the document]
droberts195: It is in C++11 and above, but not previous standards.

https://stackoverflow.com/a/11752722 summarizes the situation.

PR #100, opened 2018-05-23
[ML] Store expanding window bucket values in a compressed format

tveasey:
For some time I've thought it would be good to investigate storing some of the model state in a compressed format where the access pattern lends itself to doing this without paying prohibitive CPU cost. This PR is the outcome of applying this strategy to the window of historical data used for periodicity testing. It turns out that one gets very significant memory gains by doing this, particularly for sparse data.

This PR breaks into two pieces:
1. Extending `core::CCompressUtils` to support inflation and adding support for vectors of trivially copyable types. This covers the requirements for this work. 
2. Updating `maths::CExpandingWindow` to optionally deflate the bucket value vector when not in use.

Profiling before and after this change showed up the need to make a refinement to the simplest possible implementation. In particular, `maths::CExpandingWindow::addValue` would have needed one call to inflate/deflate per data bucket. This proved prohibitively expensive: around a 25% increase in runtime. I therefore switched to maintaining a small buffer of incremental changes to the window bucket values and only flushing this when it gets full, which is, therefore, the only time addValue triggers an inflate/deflate. I can make this buffer very small, since the window bucket length is typically significantly longer than the data bucketing length, and still get significant speedups. A length of 5 constitutes a pretty good tradeoff: costing us almost nothing in terms of memory overhead and reducing the runtime cost of inflation/deflation to around 2%.

The memory gains from this change are fairly data dependent, but I got good improvements across the board (and up to just under 44% for very sparse data where the cost of this window is a big chunk of the model size and we get really good compression, up to 95%, of the mostly empty bucket values). The following is a summary of the gain relative to master (including and excluding #97).

"Partition foo"

| # partitions | Old model memory | New model memory | % improvement | with #97 | % improvement
| :---: | :---: | :---: | :---: | :---: | :---: |
| 10 | 1.4MB | 1.2MB | 14.3 | 0.98MB | 30
| 21 | 1MB | 0.61MB | 39 | 0.54MB | 46
| 78 | 4.1MB | 2.3MB | 43.9 | 2.2MB | 46.3
| 16925 | 559MB | 353MB | 36.8 | 317MB | 43.3

"By foo"

| # by | Old model memory | New model memory | % improvement |
| :---: | :---: | :---: | :---: |
| 10 | 0.97MB | 0.87MB | 10.0 |
| 21 | 0.49MB | 0.33MB | 32.7 |
| 78 | 2MB | 1.4MB | 30.0 |
| 16925 | 244.1MB | 161.2MB | 34.0 |

This should have no impact on results except where the hard memory limit is in play.


include/core/CCompressUtils.h
```diff
@@ -37,50 +37,135 @@ namespace core {
 //!
 class CORE_EXPORT CCompressUtils : private CNonCopyable {
 public:
-    //! The output type
     using TByteVec = std::vector<Bytef>;
+    enum EOperation { E_Deflate, E_Inflate };
```
hendrikmuhs: devils advocate: What is the reason to have operations/modes over 2 classes/methods ?

include/core/CCompressUtils.h
```diff
@@ -37,50 +37,135 @@ namespace core {
...
+private:
+    static const size_t CHUNK_SIZE{4096};
 
 private:
-    enum EState { E_Unused, E_Compressing, E_Finished };
+    //! Get an unsigned character pointer to the address of the start
+    //! of the vector data.
+    template<typename T>
+    static Bytef* bytes(const std::vector<T>& input) {
+        return const_cast<Bytef*>(reinterpret_cast<const Bytef*>(input.data()));
```
hendrikmuhs: consider: zlib becomes much nice w.r.t. `const` if you:

```
#define ZLIB_CONST
```

But this requires zlib >= 1.2.8, I am not sure what versions we use on the different platforms.

droberts195: To use this we'd have to build zlib from source on Linux and ship it with our distribution.

CentOS 6 is using version 1.2.3 (patched for security vulnerabilities thankfully) and even CentOS 7 is only on version 1.2.7.  So it will be many years before we can use this flag with the standard OS version of zlib.

hendrikmuhs: :+1: ok, probably not worth the hassle. 1.2.8 is already >5 years old, but it seems not old enough ;-)

droberts195: CentOS 6 is basically a hardened version of Fedora 12, and Fedora 12 was released in 2009!

I think `ZLIB_CONST` went into version 1.2.7.1 though, not 1.2.8, so it is available on CentOS 7.  So maybe we can use it in 8.x if CentOS 7 is the oldest supported platform by then.

lib/api/unittest/CAnomalyJobLimitTest.cc
```diff
@@ -433,7 +433,7 @@ void CAnomalyJobLimitTest::testModelledEntityCountForFixedMemoryLimit() {
         LOG_DEBUG(<< "# partition = " << used.s_PartitionFields);
         LOG_DEBUG(<< "Memory status = " << used.s_MemoryStatus);
         LOG_DEBUG(<< "Memory usage = " << used.s_Usage);
-        CPPUNIT_ASSERT(used.s_PartitionFields > 280 && used.s_PartitionFields < 360);
+        CPPUNIT_ASSERT(used.s_PartitionFields > 390 && used.s_PartitionFields < 470);
```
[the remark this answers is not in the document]
tveasey: Dimitris removed the asserts which tended to cause the failures in the past in elastic/elasticsearch#30716, which this test supersedes, but I'll double check nothing else does.

include/core/CCompressUtils.h
tveasey: It seemed easier to share code this way than have two classes; although I could probably extract common functionality out into a private base class. It also felt at the outset like it was going to be the smaller change to CCompressUtils; although in practice it ended up changing quite a lot. I'll have a think about this alternative.

tveasey: I think this turns out nicer: I added separate implementations with shared base class for inflation and deflation. These can really just wrap up only the small number of library calls which are task specific, so nearly all the implementation is shared.

include/core/CCompressUtils.h
```diff
@@ -35,57 +35,183 @@ namespace core {
 //! a multi-threaded application it would be best to create
 //! one object for each thread.
 //!
-class CORE_EXPORT CCompressUtils : private CNonCopyable {
+class CORE_EXPORT CCompressUtil : private CNonCopyable {
```
droberts195: If you rename the class I think you should also rename the source file to match (plus the corresponding implementation and unit test files).

tveasey: I thought about this a bit.

My rationale for renaming was "utils" suggests to me a collection of utility functionality, i.e. multiple disparate functions and objects, rather than a single (utility) class.

The reason I didn't rename the file was that we actually have two `CCompressUtil` classes, i.e. the inflator and deflator, and these are the types one would actually use from this header. So this is one of the cases where our naming convention for files breaks down, i.e. there isn't a single natural class after which to name the file. Conversely, breaking these classes out into their own header files seems wrong to me too.

What do you think?

droberts195: Other files that contain collections of classes don't have the initial `C` - `FrequencyPredicates.h` and `ProbabilityAggregators.h` for example.  So maybe this file could be just `CompressUtils.h`?  Having the initial `C` suggests it is named after a class.

tveasey:
Thanks @droberts195. Just running them now.

PR #101, opened 2018-05-23
[ML] Make logger test more robust to slow VMs

droberts195:
The non-ASCII logging test could hang indefinitely if the VM it
was running on stalled for 200ms before reconfiguring the logger
to use the test pipe.

This change makes it more robust.  A stall of up to 5 seconds is
now tolerated, but with checks every 50ms to avoid slowing the
test down.

### DIFF
--- a/lib/core/unittest/CLoggerTest.cc
+++ b/lib/core/unittest/CLoggerTest.cc
@@ -150,11 +150,18 @@ void CLoggerTest::testNonAsciiJsonLogging() {
 
     std::ostringstream loggedData;
     std::thread reader([&loggedData] {
-        // wait a bit so that pipe has been created
-        ml::core::CSleep::sleep(200);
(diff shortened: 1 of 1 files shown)


droberts195:
This is an example of a CI failure due to the problem:

```
2018-05-22 23:26:39,785 BST [2932] DEBUG CTestTimer.cc@19 +----------------------------------------+
2018-05-22 23:26:39,785 BST [2932] DEBUG CTestTimer.cc@20 |  CLoggerTest::testNonAsciiJsonLogging  |
2018-05-22 23:26:39,785 BST [2932] DEBUG CTestTimer.cc@21 +----------------------------------------+
Build timed out (after 120 minutes). Marking the build as failed.
```

It's particularly annoying because the Jenkins job times out after 2 hours, so the CI machine is blocked from running anything else during this time.

PR #102, opened 2018-05-23
[ML] Correct the way internal linkage of constants is achieved

droberts195:
1. There's no need for constants declared in unnamed namespaces
   to be declared `static` - they already have internal linkage.
2. Constants of type `char*` pointing to literals should be
   declared `const char* const` - declaring them `const char*` does
   not result in internal linkage because it allows them to be
   changed to point at something completely different!

### DIFF
--- a/include/core/CoreTypes.h
+++ b/include/core/CoreTypes.h
@@ -17,17 +17,12 @@ using TTime = time_t;
 
 //! The standard line ending for the platform - DON'T make this std::string as
 //! that would cause many strings to be constructed (since the variable is
-//! static const at the namespace level, so is internal to each file this
-//! header is included in)
+//! const at the namespace level, so is internal to each file this header is
+//! included in)
 #ifdef Windows
-static const char* LINE_ENDING = "\r\n";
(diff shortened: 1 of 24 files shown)


include/core/CoreTypes.h
```diff
@@ -17,17 +17,12 @@ using TTime = time_t;
...
-//! header is included in)
+//! const at the namespace level, so is internal to each file this header is
+//! included in)
 #ifdef Windows
-static const char* LINE_ENDING = "\r\n";
+const char* const LINE_ENDING = "\r\n";
 #else
-#ifdef __GNUC__
-// Tell g++ that it's reasonable that this variable isn't used
-__attribute__((unused)) static const char* LINE_ENDING = "\n";
```
davidkyle: Not needed?

droberts195: No, now the variable really does have internal linkage it's not needed.

In fact, the warning from g++ was the canary in the mine warning that something was not right.  Suppressing the warning was the wrong way to avoid it.

PR #108, opened 2018-05-29
[ML] Convert std::shared_ptrs to std::unique_ptrs where possible

tveasey:
Many of our current uses of shared pointer date from when not all our target platforms were C++11 compliant. Now that they all are, we can switch these to std::unique_ptr. This has a few advantages:
1. It saves us 16 bytes (8 for the extra pointer to the reference count and 8 for the reference count) per instance
2. It avoids atomic updates of the reference count
3. It forces one to have the correct copy semantics in such cases. For example, this showed up that we had an implicit shallow copy (not appropriate) on our naive Bayes implementation and fixing this showed up a bug in restore.

Whilst making this change I also spotted an error in memory accounting for shared pointers, which wasn't including the extra memory for the reference count. Note that I've split out the fix to memory accounting and a change to provide support for unique pointers in some of our utility functionality into separate commits.

This has no direct impact on results. Again there can be an impact from the memory reduction causing a change of behaviour in jobs which hit the hard or soft memory limits.


lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -481,7 +481,7 @@ CTimeSeriesDecompositionDetail::CPeriodicityTest::CPeriodicityTest(const CPeriod
     // Note that m_Windows is an array.
     for (std::size_t i = 0u; !isForForecast && i < other.m_Windows.size(); ++i) {
         if (other.m_Windows[i]) {
-            m_Windows[i] = std::make_shared<CExpandingWindow>(*other.m_Windows[i]);
+            m_Windows[i].reset(new CExpandingWindow{*other.m_Windows[i]});
```
hendrikmuhs: I know `make_unique` is C++14, but can we maybe use a backport, which we can throw away when we switched?

Herb Sutter has an implementation here: https://herbsutter.com/gotw/_102/

droberts195: Another possibility for using `std::make_unique` is to change the `-std` argument for the Mac/Linux compilers to `c++14`/`gnu++14` respectively on the 6.x branch.

Even though Visual Studio 2013 doesn't support very much of C++14, one bit it _does_ support is `std::make_unique`.  So switching the standard filtering used by the other compilers would mean all of them support `std::make_unique`.

People would just need to remember that only a tiny fraction of C++14 is available in the 6.x branch.  From https://msdn.microsoft.com/en-us/library/hh567368.aspx:

> Visual C++ in Visual Studio 2013 implemented some key C++14 library features:
> * "Transparent operator functors" `less<>`, `greater<>`, `plus<>`, `multiplies<>`, and so on.
> * `make_unique<T>(args...)` and `make_unique<T[]>(n)`
> * `cbegin()`/`cend()`, `rbegin()`/`rend()`, and `crbegin()`/`crend()` non-member functions.

include/maths/CClustererStateSerialiser.h
```diff
@@ -43,7 +43,7 @@ struct SDistributionRestoreParams;
 //! to potential competitors.
 class MATHS_EXPORT CClustererStateSerialiser {
 public:
-    using TClusterer1dPtr = std::shared_ptr<CClusterer1d>;
+    using TClusterer1dPtr = std::unique_ptr<CClusterer1d>;
```
hendrikmuhs: General problem:

In other parts of the code we use UPtr, so `TClusterer1dPtr` would be `TClusterer1dUPtr`.

include/maths/CNaiveBayes.h
```diff
@@ -222,30 +226,47 @@ class MATHS_EXPORT CNaiveBayes {
...
+
         //! Initialize by reading state from \p traverser.
         bool acceptRestoreTraverser(const SDistributionRestoreParams& params,
                                     core::CStateRestoreTraverser& traverser);
         //! Persist state by passing information to \p inserter.
         void acceptPersistInserter(core::CStatePersistInserter& inserter) const;
+
+        //! Get the number of examples in this class.
+        double count() const;
+        //! Set the number of examples in this class.
```
hendrikmuhs: nit: It's not really a setter but exposes count for writes, maybe just "access to number of ... "

include/maths/CClustererStateSerialiser.h
tveasey: I thought about this. The problem with renaming the type of pointer is that it makes changes like this much more difficult. Everywhere the typedef is referred to suddenly has to change as well. So this change with become a whole lot bigger. Also we don't use SPtr for shared_ptr for example which would be consistent with this pattern.

Whilst we probably don't need to change this type again, I don't like that an "implementation detail" (type of smart pointer) gets mixed up with the name of the type. In some cases we can't avoid this because we have a `shared_ptr<T>` and a `unique_ptr<T>` in the same scope but in most cases we can. 

Thoughts?

include/maths/CNaiveBayes.h
tveasey: True. I started off making this a setter, but then decided to make it return a writable reference.

lib/maths/CTimeSeriesDecompositionDetail.cc
droberts195: I wondered if `boost::make_unique` might return `boost::unique_ptr`s, but just checked and it really does create `std::unique_ptr`s.  So I guess it was designed for exactly this case of a compiler supporting `std::unique_ptr` but not `std::make_unique`.

I'm happy to use it until 6.x is for bug fixes only and then someone can do a search and replace of `boost::make_unique` with `std::make_unique` on the master branch.

tveasey: Ok cool, I'll leave as I have it.

include/maths/CClustererStateSerialiser.h
hendrikmuhs: I am fine with that, actually I agree more to what you wrote, for me one purpose of a typedef/using is to abstract away the implementation detail, putting it back in the name defeats this. 

I just wanted to point it out, if you grep the codebase for `UPtr` you find a small number of uses but I could not find a rule in our new and old styleguide.

Any other comments on this?

lib/maths/CModelStateSerialiser.cc
```diff
@@ -9,6 +9,7 @@
 #include <maths/CTimeSeriesModel.h>
 
 #include <boost/bind.hpp>
+#include <boost/make_unique.hpp>
```
hendrikmuhs: superflous? this class does not use make_unique

include/maths/CClustererStateSerialiser.h
tveasey: Yes I think our naming of pointer types is not completely consistent. I would propose something like the following rules:
1. If there is only one type of smart pointer in scope then it should be suffixed with Ptr
2. If there would be a name collision then they should be all qualified: shared_ptr as SPtr, unique_ptr as UPtr, auto_ptr as APtr and weak_ptr as WPtr.

Let's take this discussion offline though. If we get agreement on this we can revisit the codebase and tidy up the naming to conform to this in a separate PR.

tveasey:
@hendrikmuhs, doing some more testing I realised that I had subverted the caching mechanism for new time series models and correlation residual priors. This means we ended up with duplicates of these objects. Also, there was no need to cache the correlation model: it can't be shared. We were previously caching, but then cloning the value passed in anyway. I now pass this in by rvalue reference and move into place. I have corrected this behaviour in my last commit which is worth reviewing.

lib/maths/CModelStateSerialiser.cc
tveasey: Although it wasn't, this is now being used (see [this](https://github.com/elastic/ml-cpp/pull/108/commits/f3b30df1c66a54e1f2ca570f975e01e8abb4b760) commit).

PR #123, opened 2018-06-13
[ML] Add option to specify output directory to dump_state

davidkyle:
And create missing directories


lib/api/dump_state/Main.cc
```diff
@@ -41,16 +41,53 @@
...
+            "The directory to write state files to");
+
+        boost::program_options::variables_map vm;
+        boost::program_options::parsed_options parsed =
+            boost::program_options::command_line_parser(argc, argv)
+                .options(desc)
+                .run();
+        boost::program_options::store(parsed, vm);
+        if (vm.count("help") > 0) {
+            std::cout << desc << std::endl;
```
droberts195: nit: in all the other programs where we parse command line arguments we write the help to `cerr`.

lib/api/dump_state/Main.cc
```diff
@@ -41,16 +41,53 @@
...
+        if (vm.count("outputDir") > 0) {
+            outputDir = vm["outputDir"].as<std::string>();
+            if (outputDir.back() != '/') {
+                outputDir.push_back('/');
+            }
+        }
+
+        return true;
+    } catch (std::exception& e) {
+        std::cout << "Error processing command line: " << e.what() << std::endl;
```
droberts195: nit: in all the other programs where we parse command line arguments we write exceptions to `cerr`.

lib/api/dump_state/Main.cc
```diff
@@ -41,16 +41,53 @@
...
 
+static const std::string TEST_FILES_PATH{"../unittest/testfiles/"};
+
+bool parseOptions(int argc, const char* const* argv, std::string& outputDir) {
+    try {
+        boost::program_options::options_description desc(
+            "Utility for creating ML model state files for use in BWC tests");
+        desc.add_options()("help", "Display this information and exit")(
+            "outputDir", boost::program_options::value<std::string>(),
+            "The directory to write state files to");
```
droberts195: You could say this is optional

lib/api/dump_state/Main.cc
```diff
@@ -41,16 +41,53 @@
...
+                .options(desc)
+                .run();
+        boost::program_options::store(parsed, vm);
+        if (vm.count("help") > 0) {
+            std::cout << desc << std::endl;
+            return false;
+        }
+        if (vm.count("outputDir") > 0) {
+            outputDir = vm["outputDir"].as<std::string>();
+            if (outputDir.back() != '/') {
```
droberts195: If someone runs `dump_state --outputDir ""` or (more likely) `dump_state --outputDir=$SOMETHING` where the `$SOMETHING` environment variable is not set then `back()` has undefined behaviour and could cause a SEGV.

It's probably best to print an error and return `false` from this method if `--outputDir` is specified with a zero length value.

PR #132, opened 2018-06-15
[ML] Refactor code to use std::lgamma instead of boost::math::lgamma

### LINKED ISSUE #128: [ML] Migrate from boost::math to std where possible
Follow-up from #126 

This is a meta-issue listing all possible re-factorings made possible by C++11 based on scanning the code at the time of writing this issue:

- [ ] use CTools::lgamma (wrapper for std::lgamma)
- [ ] `isnan`, `isinf`, `isfinite`
- [ ] `erf`

hendrikmuhs:
Migrate remaining codebase to `std::lgamma`, std::{isfinite,isinf,isnan}` and `std::erf`.

Notes:

- uses CTools::lgamma when possible, in some cases it was better to stay with `std::lgamma` e.g. if error handling was already in place
- improves some error handling to correctly report overflowing

fixes #128

### DIFF
--- a/include/core/CRapidJsonWriterBase.h
+++ b/include/core/CRapidJsonWriterBase.h
@@ -20,10 +20,10 @@
 #include <rapidjson/writer.h>
 
 #include <boost/iterator/indirect_iterator.hpp>
-#include <boost/math/special_functions/fpclassify.hpp>
 #include <boost/unordered/unordered_map.hpp>
 #include <boost/unordered/unordered_set.hpp>
 
+#include <cmath>
 #include <memory>
 #include <stack>
 
@@ -163,7 +163,7 @@ public:
 
     bool Double(double d) {
         // rewrite NaN and Infinity to 0
-        if (!(boost::math::isfinite)(d)) {
+        if (std::isfinite(d) == false) {
             return TRapidJsonWriterBase::Int(0);
         }
 
@@ -352,7 +352,7 @@ public:
     //! Adds a double field with the name fieldname to an object.
     //! \p fieldName must outlive \p obj or memory corruption will occur.
     void addDoubleFieldToObj(const std::string& fieldName, double value, TValue& obj) const {
-        if (!(boost::math::isfinite)(value)) {
+        if (std::isfinite(value) == false) {
             LOG_ERROR(<< "Adding " << value << " to the \"" << fieldName
                       << "\" field of a JSON document");
             // Don't return - make a best effort to add the value
@@ -486,7 +486,7 @@ private:
     //! Log a message if we're trying to add nan/infinity to a JSON array
     template<typename NUMBER>
     void checkArrayNumberFinite(NUMBER val, const std::string& fieldName, bool& considerLogging) const {
-        if (considerLogging && !(boost::math::isfinite)(val)) {
+        if (considerLogging && (std::isfinite(val) == false)) {
             LOG_ERROR(<< "Adding " << val << " to the \"" << fieldName
                       << "\" array in a JSON document");
             // Don't return - make a best effort to add the value
--- a/include/maths/CMathsFuncs.h
+++ b/include/maths/CMathsFuncs.h
@@ -38,8 +38,7 @@ namespace maths {
 //!
 class MATHS_EXPORT CMathsFuncs : private core::CNonInstantiatable {
 public:
-    //! Wrapper around std::isnan() which avoids the need to add
-    //! cryptic brackets everywhere to deal with macros.
+    //! Check if value is NaN.
     static bool isNan(double val);
     //! Check if any of the components are NaN.
     template<std::size_t N>
@@ -54,8 +53,7 @@ public:
     //! Check if an element is NaN.
     static bool isNan(const core::CSmallVectorBase<double>& val);
 
-    //! Wrapper around std::isinf() which avoids the need to add
-    //! cryptic brackets everywhere to deal with macros.
+    //! Check if value is infinite.
     static bool isInf(double val);
     //! Check if any of the components are infinite.
     template<std::size_t N>
--- a/include/maths/CMathsFuncs.h
+++ b/include/maths/CMathsFuncs.h
@@ -38,7 +38,7 @@ namespace maths {
 //!
 class MATHS_EXPORT CMathsFuncs : private core::CNonInstantiatable {
 public:
-    //! Wrapper around boost::math::isnan() which avoids the need to add
+    //! Wrapper around std::isnan() which avoids the need to add
     //! cryptic brackets everywhere to deal with macros.
     static bool isNan(double val);
     //! Check if any of the components are NaN.
@@ -54,7 +54,7 @@ public:
     //! Check if an element is NaN.
     static bool isNan(const core::CSmallVectorBase<double>& val);
 
-    //! Wrapper around boost::math::isinf() which avoids the need to add
+    //! Wrapper around std::isinf() which avoids the need to add
     //! cryptic brackets everywhere to deal with macros.
     static bool isInf(double val);
     //! Check if any of the components are infinite.
--- a/include/maths/CMultivariateNormalConjugate.h
+++ b/include/maths/CMultivariateNormalConjugate.h
@@ -28,11 +28,11 @@
 #include <maths/CNormalMeanPrecConjugate.h>
 #include <maths/CRestoreParams.h>
 #include <maths/CSampling.h>
+#include <maths/CTools.h>
 #include <maths/ProbabilityAggregators.h>
 
 #include <boost/make_unique.hpp>
 #include <boost/math/special_functions/beta.hpp>
-#include <boost/math/special_functions/gamma.hpp>
 #include <boost/numeric/conversion/bounds.hpp>
 #include <boost/optional.hpp>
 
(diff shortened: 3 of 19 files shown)


include/maths/CMathsFuncs.h
```diff
@@ -54,7 +54,7 @@ class MATHS_EXPORT CMathsFuncs : private core::CNonInstantiatable {
     //! Check if an element is NaN.
     static bool isNan(const core::CSmallVectorBase<double>& val);
 
-    //! Wrapper around boost::math::isinf() which avoids the need to add
+    //! Wrapper around std::isinf() which avoids the need to add
     //! cryptic brackets everywhere to deal with macros.
```
droberts195: Based on your other changes in this PR we no longer need to add cryptic brackets everywhere to deal with macros, so this comment can be removed too.

include/core/CRapidJsonWriterBase.h
```diff
@@ -163,7 +163,7 @@ class CRapidJsonWriterBase
 
     bool Double(double d) {
         // rewrite NaN and Infinity to 0
-        if (!(boost::math::isfinite)(d)) {
+        if (std::isfinite(d) == false) {
```
droberts195: The reason `boost::math::isfinite` was in brackets was to prevent it being corrupted if a macro called `isfinite` was defined.  Are you sure that will never happen?  Even on old versions of Linux like Fedora 10?  If such a macro could be defined for certain sets of system headers then `std::isfinite` not in brackets will suffer the same corruption by the macro.

lib/maths/CMathsFuncs.cc
```diff
@@ -8,7 +8,7 @@
 
 #include <maths/CMathsFuncsForMatrixAndVectorTypes.h>
 
-#include <boost/math/special_functions/fpclassify.hpp>
+#include <cmath>
 
 #ifdef isnan
```
droberts195: If this macro is still needed then `std::isfinite` should probably be in brackets in other places where it's used.  Otherwise the check on this macro and `isinf` immediately below can be removed.

tveasey:
I had a quick first scan; looks good. I would like to spend a bit more time looking through on Monday.

I have a couple of suggestions:
1. Nearly all calls to CTools::lgamma supply true (to trap overflow) I think this suggests changing the default parameter value from false to true.
2. I suspect there are cases where try blocks can now be removed, since they were previously only guarding lgamma. I think it would be nice to remove these if possible. (I'll double check those I think are safe and comment on the detailed review.)

lib/maths/CMathsFuncs.cc
hendrikmuhs: I checked this. We are not including math.h - the C header file - anymore, which should avoid this problem. But even then, according to this blog post we should be fine:

https://developers.redhat.com/blog/2016/02/29/why-cstdlib-is-more-complicated-than-you-might-think/

FWIW: Prior to this PR I build this code on both `master` and `6.x`, so at least on these platforms (afaik that includes Fedora) we should be fine.

I will remove these guards and re-run builds.

droberts195: > We are not including math.h - the C header file - anymore, which should avoid this problem.

Even if we're not including math.h from our code, we could be including a 3rd party header that includes it.

But that blog you linked is a good find because it contains the reason why the possibility of clashing macros is completely removed since we upgraded to GCC 6:

> For GCC 6 this has been fixed, by making the C++ library install its own version of <math.h>, which includes <cmath>, which then includes the real <math.h> from the C library.

So even if a 3rd party header includes math.h it will pick up the C++ versions of all the maths functions.  And since the fix is in GCC rather than the OS headers we're safe regardless of which Linux distribution we're building on, which is also nice.

include/maths/CMultivariateNormalConjugate.h
```diff
@@ -1079,41 +1079,42 @@ class CMultivariateNormalConjugate : public CMultivariatePrior {
...
-            double logGammaPostMinusPrior = 0.0;
-            for (std::size_t i = 0u; i < N; ++i) {
-                logGammaPostMinusPrior +=
-                    boost::math::lgamma(
-                        0.5 * (wishartDegreesFreedomPost - static_cast<double>(i))) -
-                    boost::math::lgamma(0.5 * (wishartDegreesFreedomPrior -
-                                               static_cast<double>(i)));
+        for (std::size_t i = 0u; i < N; ++i) {
+            if (CTools::lgamma(0.5 * (wishartDegreesFreedomPost - static_cast<double>(i)),
+                               logGammaPost, true) &&
```
tveasey: Suggest making default for checking the range true rather than false.

lib/maths/CGammaRateConjugate.cc
```diff
@@ -658,11 +657,15 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
             LOG_TRACE(<< "numberSamples = " << m_NumberSamples);
 
-            m_Constant = m_PriorShape * std::log(m_PriorRate) -
-                         boost::math::lgamma(m_PriorShape) +
+            m_Constant = m_PriorShape * std::log(m_PriorRate) - std::lgamma(m_PriorShape) +
                          logVarianceScaleSum - logGammaScaledLikelihoodShape -
-                         nResidual * boost::math::lgamma(m_LikelihoodShape) +
-                         boost::math::lgamma(m_ImpliedShape);
+                         nResidual * std::lgamma(m_LikelihoodShape) +
+                         std::lgamma(m_ImpliedShape);
```
tveasey: This would have the potential to be, for example, infinity - infinity := NaN, i.e. `- std::lgamma(m_PriorShape)` + ... + nResidual * std::lgamma(m_LikelihoodShape)`. Therefore, I think the following should also check for NaN and fail the calculation in this case. Also, nothing is left in try catch block that can throw, so this can be removed.

lib/maths/CGammaRateConjugate.cc
```diff
@@ -658,11 +657,15 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
+
+            if (std::isfinite(m_ImpliedShape) && std::isfinite(m_Constant) == false) {
```
tveasey: Surely the binding is wrong for this: don't you want this condition to be `(std::isfinite(m_ImpliedShape) && std::isfinite(m_Constant)) == false`.

lib/maths/CLogNormalMeanPrecConjugate.cc
```diff
@@ -524,9 +523,15 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
 
             m_Constant = 0.5 * (std::log(m_Precision) - std::log(impliedPrecision)) -
-                         0.5 * m_NumberSamples * LOG_2_PI - 0.5 * logVarianceScaleSum +
-                         boost::math::lgamma(impliedShape) -
-                         boost::math::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+                         0.5 * m_NumberSamples * LOG_2_PI -
+                         0.5 * logVarianceScaleSum + std::lgamma(impliedShape) -
+                         std::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+
+            if (std::isfinite(m_Constant) == false) {
```
tveasey: As above: check for NaN and no need for try catch.

lib/maths/CMultinomialConjugate.cc
```diff
@@ -660,25 +659,24 @@ CMultinomialConjugate::jointLogMarginalLikelihood(const TDouble1Vec& samples,
         LOG_TRACE(<< "# samples = " << numberSamples
                   << ", total concentration = " << m_TotalConcentration);
 
-        result = boost::math::lgamma(numberSamples + 1.0) +
-                 boost::math::lgamma(m_TotalConcentration) -
-                 boost::math::lgamma(m_TotalConcentration + numberSamples);
+        result = std::lgamma(numberSamples + 1.0) + std::lgamma(m_TotalConcentration) -
+                 std::lgamma(m_TotalConcentration + numberSamples);
```
tveasey: We've lost the throw on failure here. Whilst this definitely shouldn't overflow, I think I'd still prefer to keep checks for NaN and finite values as per other distribution models and return accordingly. Also, try catch redundant here.

lib/maths/CNormalMeanPrecConjugate.cc
```diff
@@ -384,9 +384,14 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
 
             m_Constant = 0.5 * (std::log(m_Precision) - std::log(impliedPrecision)) -
-                         0.5 * m_NumberSamples * LOG_2_PI - 0.5 * logVarianceScaleSum +
-                         boost::math::lgamma(impliedShape) -
-                         boost::math::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+                         0.5 * m_NumberSamples * LOG_2_PI -
+                         0.5 * logVarianceScaleSum + std::lgamma(impliedShape) -
+                         std::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+
+            if (std::isfinite(m_Constant) == false) {
```
tveasey: As above, for NaN and redundant try catch.

lib/maths/CPoissonMeanConjugate.cc
```diff
@@ -548,17 +547,17 @@ CPoissonMeanConjugate::jointLogMarginalLikelihood(const TDouble1Vec& samples,
...
         // Get the implied shape parameter for the gamma distribution
         // including the samples.
         double impliedShape = m_Shape + sampleSum;
         double impliedRate = m_Rate + numberSamples;
 
-        result = boost::math::lgamma(impliedShape) + m_Shape * std::log(m_Rate) -
+        result = std::lgamma(impliedShape) + m_Shape * std::log(m_Rate) -
                  impliedShape * std::log(impliedRate) - sampleLogFactorialSum -
-                 boost::math::lgamma(m_Shape);
+                 std::lgamma(m_Shape);
```
tveasey: Check not NaN and is finite and remove redundant try catch.

lib/maths/CGammaRateConjugate.cc
hendrikmuhs: good catch

lib/maths/CGammaRateConjugate.cc
hendrikmuhs: > I think the following should also check for NaN and fail the calculation

`isfinite()` fails for inite **and** NaN

lib/maths/CLogNormalMeanPrecConjugate.cc
hendrikmuhs: as above, I use `isfinite()` because it combines the check for infinite and NaN

lib/maths/CMultinomialConjugate.cc
hendrikmuhs: this is a bit hidden: if you expand the diff you see at the end:

```
maths_t::EFloatingPointErrorStatus status = CMathsFuncs::fpStatus(result);

return status
```

Therefore there was no need to add another check.

lib/maths/CPoissonMeanConjugate.cc
hendrikmuhs: its checked already on line 566

lib/maths/CGammaRateConjugate.cc
tveasey: ... except, the return type is wrong. I think we should really be returning `maths_t::E_FpFailed` in this case.

lib/maths/CMultinomialConjugate.cc
tveasey: Are yes, missed that.

lib/maths/CPoissonMeanConjugate.cc
tveasey: Ok cool. But try catch can go here I think.

lib/maths/CGammaRateConjugate.cc
tveasey: I guess the cleanest thing might be to keep the check as is and use `CMathsFuncs::fpStatus` for this.

lib/maths/CLogNormalMeanPrecConjugate.cc
```diff
@@ -486,55 +486,51 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
 
-        } catch (const std::exception& e) {
-            LOG_ERROR(<< "Error calculating marginal likelihood: " << e.what());
+        m_Constant = 0.5 * (std::log(m_Precision) - std::log(impliedPrecision)) -
+                     0.5 * m_NumberSamples * LOG_2_PI -
+                     0.5 * logVarianceScaleSum + std::lgamma(impliedShape) -
+                     std::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+
+        if (std::isnan(m_Constant)) {
+            LOG_ERROR(<< "Error calculating marginal likelihood, flooting point nan");
```
tveasey: typo: flooting |-> floating

lib/maths/CLogNormalMeanPrecConjugate.cc
```diff
@@ -486,55 +486,51 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
             this->addErrorStatus(maths_t::E_FpFailed);
+        } else if (std::isinf(m_Constant)) {
+            LOG_ERROR(<< "Error calculating marginal likelihood, flooting point overflow");
```
tveasey: As above

lib/maths/CNormalMeanPrecConjugate.cc
```diff
@@ -356,45 +356,41 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
+
+        double impliedShape = m_Shape + 0.5 * m_NumberSamples;
+        double impliedPrecision = m_Precision + m_WeightedNumberSamples;
+
+        m_Constant = 0.5 * (std::log(m_Precision) - std::log(impliedPrecision)) -
+                     0.5 * m_NumberSamples * LOG_2_PI -
+                     0.5 * logVarianceScaleSum + std::lgamma(impliedShape) -
+                     std::lgamma(m_Shape) + m_Shape * std::log(m_Rate);
+        if (std::isnan(m_Constant)) {
+            LOG_ERROR(<< "Error calculating marginal likelihood, flooting point nan");
```
tveasey: and here

lib/maths/CNormalMeanPrecConjugate.cc
```diff
@@ -356,45 +356,41 @@ class CLogMarginalLikelihood : core::CNonCopyable {
...
             this->addErrorStatus(maths_t::E_FpFailed);
+        } else if (std::isinf(m_Constant)) {
+            LOG_ERROR(<< "Error calculating marginal likelihood, flooting point overflow");
```
tveasey: and here

PR #136, opened 2018-07-02
[ML] Fix issues upgrading state leading to possible aborts

### LINKED ISSUE #135: [ML] Autodetect process crashes when trying to forecast on job upgraded from 6.2.4 to 6.3.1
Found in `6.3.1`. Can be reproduced in `6.3.0`.

The issue can be reproduced by running a job using the `high_count` function on the `dns_tunneling` dataset with a `30m` bucket span.

While running a `6.3.1` ES instance that was upgraded from version `6.2.4`,  if the user tries to forecast on the `dns_tunneling` job that was created in version `6.2.4`, the autodetect process will crash as soon as the forecast request is sent.

ES logs show:
```
[2018-06-29T16:08:06,742][INFO ][o.e.x.m.j.p.a.AutodetectProcessManager] [ml-2] Opening job [dns_tunneling1_20180629-1542_624_ga]
[2018-06-29T16:08:06,764][INFO ][o.e.x.m.j.p.a.AutodetectProcessManager] [ml-2] [dns_tunneling1_20180629-1542_624_ga] Loading model snapshot [1530283391] with latest_record_timestamp [2016-02-11T23:52:14.000Z], job latest_record_timestamp [2016-02-11T23:52:14.000Z]
[2018-06-29T16:08:06,773][INFO ][o.e.x.m.j.p.a.NativeAutodetectProcessFactory] Restoring quantiles for job 'dns_tunneling1_20180629-1542_624_ga'
[2018-06-29T16:08:07,032][INFO ][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CResourceMonitor.cc@67] Setting model memory limit to 1024 MB
[2018-06-29T16:08:07,072][INFO ][o.e.x.m.j.p.a.AutodetectProcessManager] [ml-2] Successfully set job state to [opened] for job [dns_tunneling1_20180629-1542_624_ga]
[2018-06-29T16:08:07,114][INFO ][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CAnomalyJob.cc@849] Processing is already complete to time 1455233400
[2018-06-29T16:08:07,126][INFO ][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CForecastRunner.cc@113] Start forecasting from 2016-02-11T23:30:00+0000 to 2016-02-25T23:30:00+0000
[2018-06-29T16:08:07,127][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CTrendComponent.cc@347] Failed calculating confidence interval: Error in function boost::math::normal_distribution<double>::normal_distribution: Scale parameter is 0, but must be > 0 !, variance = 0, confidence = 95
[2018-06-29T16:08:07,190][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CTrendComponent.cc@347] Failed calculating confidence interval: Error in function boost::math::normal_distribution<double>::normal_distribution: Scale parameter is 0, but must be > 0 !, variance = 0, confidence = 95 | repeated [671]
[2018-06-29T16:08:07,190][FATAL][o.e.x.m.j.p.l.CppLogMessageHandler] [dns_tunneling1_20180629-1542_624_ga] [autodetect/28425] [CStateMachine.cc@211] Invalid index '1'
[2018-06-29T16:08:08,044][INFO ][o.e.x.m.j.p.a.NativeAutodetectProcess] [dns_tunneling1_20180629-1542_624_ga] State output finished
[2018-06-29T16:08:08,052][ERROR][o.e.x.m.j.p.a.NativeAutodetectProcess] [dns_tunneling1_20180629-1542_624_ga] autodetect process stopped unexpectedly: Failed calculating confidence interval: Error in function boost::math::normal_distribution<double>::normal_distribution: Scale parameter is 0, but must be > 0 !, variance = 0, confidence = 95
Invalid index '1'
Fatal error: 'terminate called after throwing an instance of 'std::runtime_error''
Fatal error: '  what():  Ml Fatal Exception'
Fatal error: 'si_signo 6, si_code: -6, si_errno: 0, address: 0x7ff1c2ab0428, library: /lib/x86_64-linux-gnu/libc.so.6, base: 0x7ff1c2a7b000, normalized address: 0x35428'

[2018-06-29T16:08:08,052][WARN ][o.e.x.m.j.p.a.o.AutoDetectResultProcessor] [dns_tunneling1_20180629-1542_624_ga] some results not processed due to the termination of autodetect
[2018-06-29T16:08:08,087][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [controller/28259] [CDetachedProcessSpawner.cc@184] Child process with PID 28425 was terminated by signal 6
[2018-06-29T16:08:08,094][INFO ][o.e.x.m.j.p.a.AutodetectProcessManager] [ml-2] Successfully set job state to [failed] for job [dns_tunneling1_20180629-1542_624_ga]
```

## Steps to reproduce

1. Start a version `6.2.4` ES instance and run  a job using the `high_count` function on the `dns_tunneling` dataset with a `30m` bucket span.
2. Upgrade the instance to version `6.3.1`
3. Try to run a forecast on the `dns_tunneling` job that was created in version `6.2.4`
4. Notice that the autodetect process has crashed, and the job state is now `failed`

tveasey:
We weren't properly upgrading aspects of the state from 6.2 which triggered us to abort in some cases. Specifically, we should *not* be restoring the machine identifier, which is initialised correctly and restored with the wrong value and we need to update states where the set has changed.

This needs back porting to 6.3 asap. Discussing internally we prefer not to request any delay to 6.3.1 which contains critical fixes and instead advise people to delay updating to 6.3 if they have realtime ML jobs until 6.3.2 becomes available.

Closes #135.


lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -355,6 +357,8 @@ bool upgradeTrendModelToVersion6p3(const core_t::TTime bucketLength,
     return true;
 }
 
+const TSizeSizeMap SC_STATES_UPGRADING_TO_VERSION_6_3{{0, 0}, {1, 1}, {2, 1}, {3, 2}, {4, 3}};
+
```
hendrikmuhs: am I right that this is just a helper to fix the BWC issue? Do we need this on `master`? 

It would be good to document. Maybe we can introduce some comment prefix that indicates possible cleanups at a later stage. If I am right, this can be cleaned up for 7.0.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -355,6 +357,8 @@ bool upgradeTrendModelToVersion6p3(const core_t::TTime bucketLength,
     return true;
 }
 
+const TSizeSizeMap SC_STATES_UPGRADING_TO_VERSION_6_3{{0, 0}, {1, 1}, {2, 1}, {3, 2}, {4, 3}};
```
droberts195: Might be worth adding a comment to say the intention is to map states from `{"INITIAL", "SMALL_TEST", "REGULAR_TEST", "NOT_TESTING", "ERROR"}` to the best equivalent in `{"INITIAL", "TEST", "NOT_TESTING", "ERROR"}`.

lib/core/CStateMachine.cc
```diff
@@ -201,14 +210,15 @@ void CStateMachine::CMachineDeque::capacity(std::size_t capacity) {
...
+const CStateMachine::SMachine& CStateMachine::CMachineDeque::operator[](std::size_t pos_) const {
+    std::size_t pos{pos_};
     for (const auto& machines : m_Machines) {
         if (pos < machines.size()) {
             return machines[pos];
         }
         pos -= machines.size();
     }
-    LOG_ABORT(<< "Invalid index '" << pos << "'");
+    LOG_ABORT(<< "Invalid index '" << pos_ << "'");
```
hendrikmuhs: something for a follow up? The abort is a drastic decision as in this issue we effectively kill the process although only a feature broke. We are more forgiving in other areas, e.g. restore returns a bool.

lib/maths/CTimeSeriesDecompositionDetail.cc
droberts195: This code can't be removed in 7.0.  It's permissible to upgrade from 6.0 to 7.last as long as you do a full cluster restart.

It could possibly be removed in 8.0, but even then we'd have to do extra work on the Java side to disallow reverting to a model snapshot that dated back to 6.x.

hendrikmuhs: ok, good point. I think it's still worth a follow up. I assume there will be more reasons to cut the minimum upgradeable version. So independent of the decision to be made in future, a comment can help to find it when it's time.

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: Yes, good point. I'll add a comment.

lib/core/CStateMachine.cc
tveasey: Yes. In hindsight, this is better dealt with by trapping and reinitialising the offending object (which would have worked around this problem). I'll make that change in a follow up PR.

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: Note that the whole block containing this is labelled as relating to updating state.

I agree that it would be nice to remove the extra code we need for upgrading state over time. So far we haven't been disciplined about how we manage this process in this repo. Recently, I've introduced the formalism that we use completely separate code paths for each state version (where we need to upgrade) in the restore functions, which I think is a step in that direction. 

I'd propose we do an audit of the code w.r.t. state update (there are other cases than this) and decide what constraints we need to impose to permit us to gradually remove this code. At that point, I think we should have a definite guidelines on how to implement state updates and how we track what can be removed when. I'd prefer to delay any further comments in the code until we've done that.

PR #143, opened 2018-07-04
[ML] Fix bug upgrading state from 5.6 direct to 6.3 causing SIGSEGV on open job

### LINKED ISSUE #141: [ML] SIGSEGV upgrading a job directly from 5.6.10 to 6.3
See [this](https://github.com/elastic/ml-cpp/issues/135#issuecomment-402147574) comment for the original report and [this](https://github.com/elastic/ml-cpp/issues/135#issuecomment-402162558) for diagnostics.

tveasey:
This also corrects upgrading the trend component model. The current behaviour could result in large prediction errors immediately after upgrading.

Closes #141.


lib/maths/CSeasonalComponentAdaptiveBucketing.cc
```diff
@@ -333,7 +333,9 @@ bool CSeasonalComponentAdaptiveBucketing::acceptRestoreTraverser(core::CStateRes
             RESTORE(LAST_UPDATES_OLD_TAG,
                     core::CPersistUtils::fromString(traverser.value(), lastUpdates))
         } while (traverser.next());
-
+        if (lastUpdates.empty()) {
```
droberts195: Could add a comment to say this is needed because `lastUpdates` wasn't present in 5.6.

PR #149, opened 2018-07-10
[ML] skip calculation of influencer probabilities if no influences specified

hendrikmuhs:
skip calculation of influencer probabilities if no influences specified

### DIFF
--- a/lib/model/CProbabilityAndInfluenceCalculator.cc
+++ b/lib/model/CProbabilityAndInfluenceCalculator.cc
@@ -761,7 +761,7 @@ bool CProbabilityAndInfluenceCalculator::calculate(
 
     LOG_TRACE(<< "probability = " << probability);
 
-    if (m_InfluencerProbabilities.empty() == true) {
+    if (m_InfluencerProbabilities.empty()) {
         LOG_TRACE(<< "no influencers");
         return true;
     }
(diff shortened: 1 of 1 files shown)


lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -761,6 +761,11 @@ bool CProbabilityAndInfluenceCalculator::calculate(
 
     LOG_TRACE(<< "probability = " << probability);
 
+    if (m_InfluencerProbabilities.empty() == true) {
```
tveasey: I thought our guidelines was to prefer `== false` to `!`, but I'm not sure about `== true` this looks a bit funny to me. Also, I think this can safely go before `if (!m_Probability.calculate(probability))` to skip more redundant work; I don't actually think it is appropriate to return false in this case either if that were to fail.

hendrikmuhs: fair enough, I was unsure myself if I should add the redundant `== true`, will remove that. 

I wonder about the suggestions to move it up, probability is an out parameter and according to my investigations this is important to be calculated independent of influencer.

tveasey: > I wonder about the suggestions to move it up, `probability` is an out parameter

Sorry, yes you are completely right I didn't read this carefully.

PR #150, opened 2018-07-10
[ML] Fix influencer count and influence calculation

### LINKED ISSUE #24: [ML] Bad values for the variance scale
A user data set has shown up two issues with the variance scale calculation in version 6.2.2 of the analytics:
 1) it is sometimes negative(!),
 2) it is sometimes infinite.

In particular, we are seeing the following error messages logged:
`Error calculating joint distribution: Bad variance scale -5.75`
`Error calculating joint distribution: Bad variance scale inf`

There is no prospect of getting hold of the data set; however the data characteristics sound benign. There were two detectors:  
- detector high_mean(x) over y influencers y,z. 
- detector high_median(x) over y influencers y,z

For x we have min: 0, max: 4.34571, avg: 2.0736 and cardinality of y is 430.

This issue is to investigate routes by which this problem could occur. The initial areas to investigate are `CTimeSeriesDecomposition::scale` and the calculation of the count variance scale, particularly for influencers.

cc @LucaWintergerst.

hendrikmuhs:
This fixes counting of influencer per bucket, prior this fix the count has always been set to 1.

Notes:

- fix == 1st commit, the 2nd and 3rd add test coverage and improve doc/robustness
- it turned out that this functionality had a test (`testVarp`) but test was forgotten to be added in the test suite
- change affects results if influencers are used as influencer score are now changed

Fixes #24

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -59,7 +59,7 @@ Age seasonal components in proportion to the fraction of values with which they'
 Persist and restore was missing some of the trend model state ({pull}#99[#99])
 Stop zero variance data generating a log error in the forecast confidence interval calculation ({pull}#107[#107])
 Fix corner case failing to calculate lgamma values and the correspoinding log errors ({pull}#126[#126])
-Influence count per bucket was wrong and lead to wrong influencer scoring ({pull}#150[#150])
+Influence count per bucket for metric population analyses was wrong and lead to wrong influencer scoring ({pull}#150[#150])
 
 === Regressions
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -59,6 +59,7 @@ Age seasonal components in proportion to the fraction of values with which they'
 Persist and restore was missing some of the trend model state ({pull}#99[#99])
 Stop zero variance data generating a log error in the forecast confidence interval calculation ({pull}#107[#107])
 Fix corner case failing to calculate lgamma values and the correspoinding log errors ({pull}#126[#126])
+Influence count per bucket was wrong and lead to wrong influencer scoring ({pull}#150[#150])
 
 === Regressions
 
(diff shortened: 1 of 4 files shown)


lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -370,8 +416,11 @@ void doComputeInfluences(model_t::EFeature feature,
     for (auto i = influencerValues.begin(); i != influencerValues.end(); ++i) {
         params.weights(weights).updateAnomalyModel(false);
 
-        computeInfluencedValue(value, count, i->second.first, i->second.second,
-                               params, influencedValue[0]);
+        if (computeInfluencedValue(value, count, i->second.first, i->second.second,
+                                   params, influencedValue[0]) == false) {
+            LOG_ERROR(<< "Failed to compute influencer value");
```
hendrikmuhs: @tveasey do you have a better idea for this error message?

tveasey: I think, at least for us it is very useful to see the function arguments, i.e. `value`, `count`, `i->second.first` and `i->second.second`.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -246,19 +280,24 @@ class CMeanDifference {
...
+
+        return true;
     }
 };
 
 //! \brief Computes the value of the variance statistic on a set difference.
 class CVarianceDifference {
 public:
     //! Features.
-    void operator()(const TDouble1Vec& v,
```
tveasey: Maybe document the parameters to this function too? i.e. `v` == overall variance and mean, `n` == overall count, `vi` == influencer variance and mean, etc.

docs/CHANGELOG.asciidoc
```diff
@@ -59,6 +59,7 @@ Age seasonal components in proportion to the fraction of values with which they'
 Persist and restore was missing some of the trend model state ({pull}#99[#99])
 Stop zero variance data generating a log error in the forecast confidence interval calculation ({pull}#107[#107])
 Fix corner case failing to calculate lgamma values and the correspoinding log errors ({pull}#126[#126])
+Influence count per bucket was wrong and lead to wrong influencer scoring ({pull}#150[#150])
```
tveasey: On second thoughts, I'd say the "Influence bucket count for metric population analysis was wrong..."

PR #159, opened 2018-07-19
[ML] do not clear periodicity test after detecting a changepoint

hendrikmuhs:
Internal QA test revealed a problem with deleting the periodicity test after detecting a changepoint, causing a regression in an internal dataset. This fix removes the clearance part of changepoint detection (only a small part of cp detection, see #95), effectively restoring the old behavior.

Non-issue because this fixes an issue of an unreleased feature.

### DIFF
--- a/lib/maths/unittest/CTimeSeriesModelTest.cc
+++ b/lib/maths/unittest/CTimeSeriesModelTest.cc
@@ -2190,13 +2190,14 @@ void CTimeSeriesModelTest::testLinearScaling() {
         debug.addValueAndPrediction(time, sample, model);
         auto x = model.confidenceInterval(
             time, 90.0, maths_t::CUnitWeights::unit<TDouble2Vec>(1));
-        CPPUNIT_ASSERT(::fabs(sample - x[1][0]) < 1.2 * std::sqrt(noiseVariance));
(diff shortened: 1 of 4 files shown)


lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -2190,13 +2190,14 @@ void CTimeSeriesModelTest::testLinearScaling() {
...
-        CPPUNIT_ASSERT(::fabs(sample - x[1][0]) < 1.2 * std::sqrt(noiseVariance));
+        CPPUNIT_ASSERT(::fabs(sample - x[1][0]) < 1.3 * std::sqrt(noiseVariance));
         CPPUNIT_ASSERT(::fabs(x[2][0] - x[0][0]) < 3.3 * std::sqrt(noiseVariance));
         time += bucketLength;
     }
 
     // Scale by 2 / 0.3
-
+    // Disabled, see https://github.com/elastic/ml-cpp/pull/159
+/*
```
tveasey: That's unfortunate. This shows the problem with detecting a seasonal component when the window includes one or more change points.

I wonder if we can determine that the seasonal component we detect has worse prediction error than the current model and discard it. Such a test should avoid the other problem we hit.

This is something we might be able to achieve for 6.4, but let's look at that in a follow up PR.

PR #170, opened 2018-07-25
[ML] Fixes potentials for SIGSEGV in multivariate probability calculation

tveasey:
There were a couple of issues in the multivariate probability calculation for discovered correlations:
1. When detecting a trend the only the correlation models should have been cleared, since the univariate time series model still exists.
2. When copying and swapping the multivariate distribution models we need to handle the case that no clusterer is present (marginal and conditional distributions are initialised without one).
3. The indexing of some quantities relating to the different correlated variables was incorrect.

This also splits up the probability calculation into a correlated and uncorrelated version since the existing function was cumbersome.

This can only affect results, but only for analysis with multivariate by fields enabled.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -1030,65 +1030,75 @@ bool CUnivariateTimeSeriesModel::probability(const CModelProbabilityParams& para
         double mostAnomalousSample{0.0};
         TPriorPtr mostAnomalousCorrelationModel;
 
-        for (std::size_t i = 0u; i < variables.size(); ++i) {
-            if (!value[i].empty() || (!params.mostAnomalousCorrelate() ||
-                                      i == *params.mostAnomalousCorrelate())) {
-                variable[0] = variables[i][0];
-                sample[0][variables[i][0]] = m_TrendModel->detrend(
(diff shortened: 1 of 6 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -973,137 +973,163 @@ bool CUnivariateTimeSeriesModel::probability(const CModelProbabilityParams& para
...
-    if (value[0].size() == 1) {
-        core_t::TTime time{time_[0][0]};
-        TDouble1Vec sample{m_TrendModel->detrend(
-            time, value[0][0], params.seasonalConfidenceInterval())};
-        maths_t::TDoubleWeightsAry1Vec weights{unpack(params.weights()[0])};
+bool CUnivariateTimeSeriesModel::uncorrelatedProbability(const CModelProbabilityParams& params,
+                                                         const TTime2Vec1Vec& time_,
+                                                         const TDouble2Vec1Vec& value,
+                                                         double& probability,
+                                                         TTail2Vec& tail) const {
```
tveasey: This function is the `if (value[0].size() == 1)` branch and is otherwise unchanged.

droberts195:
Note that the 6.4 branch has been created, so if this is to go into 6.4.0 it will need backporting to both 6.x and 6.4.

PR #171, opened 2018-07-25
[ML] Create 1.0 probability results when skipped due to rules

### LINKED ISSUE #169: Unexpected job result when using rules
**Found in version**
6.4.0 b1620

**Steps to reproduce**
In the ML UI:
- Create a single metric farequote job: mean(responsetime), 15m buckets
- View the job results in single metric viewer
  - Notice the single major anomaly with a probability of `0.0012282000587277902`
fq_without_rules
- Say you're only interested in this major anomaly. Create the following rule for the detector via the gear icon in a result row:
fq_rule
- Clone the job
  - Notice the `custom_rules` part of the cloned job's JSON:
detector_json
- Run the cloned job and view results in the single metric viewer

**Expected result**
* The major anomaly is the only reported anomaly - and it's still major.

**Actual result**
* The formerly major anomaly is the only reported anomaly, but it's not major any more (with a probability of `0.0017024271634351763`):
fq_with_rules

dimitris-athanasiou:
Previously, when a result is skipped we do not add a probability
for it in the probability calculator. This results to skipped
results to be similar to not seeing any data with regard to the
quantiles. For example, in a single metric job where a rule
suppresses all results but the most severe, it is possible that
record will have quite low score due to the lack of confidence.

This commit changes the behaviour to add 1.0 probabilities for
skipped results. Note this is not needed for population models
as we will be seeing multiple other results.

Closes #169

### DIFF
--- a/lib/model/CMetricModel.cc
+++ b/lib/model/CMetricModel.cc
@@ -347,6 +347,7 @@ bool CMetricModel::computeProbability(const std::size_t pid,
     pJoint.addAggregator(maths::CJointProbabilityOfLessLikelySamples());
     pJoint.addAggregator(maths::CProbabilityOfExtremeSample());
 
+    bool skippedResults{false};
     for (std::size_t i = 0u, n = gatherer.numberFeatures(); i < n; ++i) {
         model_t::EFeature feature = gatherer.feature(i);
         if (model_t::isCategorical(feature)) {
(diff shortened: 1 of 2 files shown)


lib/model/CMetricModel.cc
```diff
@@ -379,15 +381,22 @@ bool CMetricModel::computeProbability(const std::size_t pid,
         }
     }
 
-    if (pJoint.empty()) {
+    if (pJoint.empty() && skippedResults == false) {
```
tveasey: As per above, maybe keep this together with other checks below, which it relates to.

PR #173, opened 2018-07-30
[ML] Pad estimated memory usage in anticipation of multi-bucket analysis

tveasey:
This implements a 6.4 only pad of the estimated model memory in anticipation of the memory which will be consumed by multi-bucket analysis, which will be released in 6.5.

This serves to avoid unnecessary churn in estimated memory usage between versions since there were a number of memory reductions in 6.4, partly in anticipation of multi-bucket analysis. This quantity is used for among other things the job allocation decision.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -2434,7 +2434,7 @@ std::size_t CMultivariateTimeSeriesModel::memoryUsage() const {
     return core::CMemory::dynamicSize(m_Controllers) +
            core::CMemory::dynamicSize(m_TrendModel) +
            /*see above*/
-           core::CMemory::dynamicSize(m_ResidualModel) +
+           2 * core::CMemory::dynamicSize(m_ResidualModel) +
            core::CMemory::dynamicSize(m_AnomalyModel) +
            core::CMemory::dynamicSize(m_SlidingWindow);
 }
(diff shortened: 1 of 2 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -2410,14 +2418,22 @@ void CMultivariateTimeSeriesModel::debugMemoryUsage(core::CMemoryUsage::TMemoryU
...
+    core::CMemoryDebug::dynamicSize("m_ResidualModelPad", m_ResidualModel, mem);
     core::CMemoryDebug::dynamicSize("m_AnomalyModel", m_AnomalyModel, mem);
     core::CMemoryDebug::dynamicSize("m_SlidingWindow", m_SlidingWindow, mem);
 }
 
 std::size_t CMultivariateTimeSeriesModel::memoryUsage() const {
     return core::CMemory::dynamicSize(m_Controllers) +
            core::CMemory::dynamicSize(m_TrendModel) +
+           /*see above*/
            core::CMemory::dynamicSize(m_ResidualModel) +
```
droberts195: Should it be `2 *` here as well?

tveasey: Doh! It should indeed.

tveasey: All tests pass. This indicates we should be testing for agreement of `debugMemoryUsage` and `memoryUsage` in some extra cases and ideally also checking partition and by field memory usage in these cases as well. I think I'd prefer to introduce further testing in a follow up PR on master + 6.5.

PR #175, opened 2018-07-30
[ML] Anomaly detection for multiple bucket features

tveasey:
Currently, we perform anomaly detection only on a single bucket feature, typically derived from  some aggregation applied to all the documents which fall in that bucket's time interval. However, In many cases interesting events span multiple buckets.

The most important case is where any given bucket is not especially unusual, given the variation w.r.t. our predictions we've historically seen, but a collection of contiguous buckets are jointly unusual.

Such cases can be detected by modelling and detecting unusual events w.r.t. collective properties of a number of contiguous buckets, i.e. features derived from multiple buckets; this PR uses the cumulative prediction error in a sliding window ending at the current bucket. This has proved effective for picking up the sort of events we want to detect, i.e. extended periods where are model is systematically in error.

This constitutes a reasonably significant change to results: increasing the relative importance we'll report for longer lasting anomalies. However, it is giving consistently better area under the P-R curve against our internal test suit. In addition, we pay some extra memory costs (both to compute the feature and for the extra model). This has been more than recouped in other preparatory changes, i.e. #127, #146, etc.


bin/autodetect/CCmdLineParser.cc
```diff
@@ -118,7 +117,7 @@ bool CCmdLineParser::parse(int argc,
             ("multivariateByFields",
                         "Optional flag to enable multi-variate analysis of correlated by fields")
             ("multipleBucketspans",  boost::program_options::value<std::string>(),
-                        "Optional comma-separated list of additional bucketspans - must be direct multiples of the main bucketspan")
+                        "Deprecated - ignored")
```
hendrikmuhs: what's the reason to keep it? Are there clients using this?

droberts195: I'm happy to completely remove this, although it does then commit us to removing the corresponding options on the Java side.  @dimitris-athanasiou said he would do this.  The functionality to silently drop any settings related to this functionality (which was never documented and not fully tested) will then live in the Java code.

include/maths/CTimeSeriesBulkFeatures.h
```diff
@@ -0,0 +1,163 @@
...
+
+#include <maths/CBasicStatistics.h>
+#include <maths/CLinearAlgebra.h>
+#include <maths/CTypeConversions.h>
+#include <maths/MathsTypes.h>
+
+#include <boost/bind.hpp>
+#include <boost/circular_buffer.hpp>
+#include <boost/iterator/counting_iterator.hpp>
+
```
hendrikmuhs: does not seem to be used in this place

lib/maths/CTimeSeriesModel.cc
```diff
@@ -701,68 +737,69 @@ CUnivariateTimeSeriesModel::addSamples(const CModelAddSamplesParams& params,
...
-    }
+    double multiplier{this->updateDecayRates(params, averageTime, samples_)};
 
     if (m_Correlations != nullptr) {
         m_Correlations->addSamples(m_Id, params, samples, multiplier);
     }
 
     if (randomSample) {
-        m_SlidingWindow.push_back({randomSample->first, randomSample->second});
+        m_RecentSamples.push_back({randomSample->first, randomSample->second});
```
hendrikmuhs: could be an emplace_back?

hendrikmuhs:
I had a first scan over it, looks good.

One remark with respect to naming: I find `multibucket` more descriptive than `bulk`. I think it would be easier of the code would also use `multibucket`.

bin/autodetect/CCmdLineParser.cc
dimitris-athanasiou: I think it's fine to drop everything from the c++ side. I'll prepare the java side and we will merge them both in `6.5`.

dimitris-athanasiou: I raised https://github.com/elastic/elasticsearch/pull/32496. @tveasey, this means you can completely remove the `multipleBucketspans` param in this PR.

include/maths/CTimeSeriesBulkFeatures.h
tveasey: I think this was left over from earlier versions, I'll tidy.

bin/autodetect/CCmdLineParser.cc
tveasey: Ok great. I'll tidy.

include/maths/CTimeSeriesBulkFeatures.h
```diff
@@ -0,0 +1,163 @@
...
+        using TValueType = typename std::iterator_traits<ITR>::value_type;
+        using TSecondType = typename SSecondType<TValueType>::Type;
+        using TPrecise = typename SPromoted<typename TSecondType::TValue>::Type;
+        using TMeanAccumulator = typename CBasicStatistics::SSampleMean<TPrecise>::TAccumulator;
+    };
+
+    //! Univariate implementation returns zero.
+    template<typename T>
+    static double conformable(const T& /*x*/, double value) {
+        return value;
```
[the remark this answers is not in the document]
tveasey: Indeed, comment is out of date.

lib/maths/CTimeSeriesModel.cc
tveasey: This is a boost circular buffer which doesn't support emplace_back unfortunately.

tveasey:
@hendrikmuhs and @dimitris-athanasiou I addressed all your review comments. I also refined aggregation to take account of correlation between the multi-bucket and bucket features, in [this](https://github.com/elastic/ml-cpp/pull/175/commits/3937f40584a9c3ae670bb81b7f2edf80d5ea0a7a) commit. Can you take another look please?

include/maths/CTimeSeriesMultibucketFeatures.h
```diff
@@ -0,0 +1,173 @@
...
+//! \brief Defines features on collections of time series values.
+//!
+//! DESCRIPTION:\n
+//! The intention of these is to provide useful features for performing
+//! anomaly detection. Specifically, unusual values of certain properties
+//! of extended time ranges are often the most interesting events in a
+//! time series from a user's perspective.
+class CTimeSeriesMultibucketFeatures {
+public:
+    //! The geometric weight applied the window.
```
hendrikmuhs: nit: "... applied **to** the window" ?

include/model/CAnomalyDetectorModelConfig.h
```diff
@@ -162,6 +162,10 @@ class MODEL_EXPORT CAnomalyDetectorModelConfig {
     //! The default maximum time to test for a change point in a time series.
     static const core_t::TTime DEFAULT_MAXIMUM_TIME_TO_TEST_FOR_CHANGE;
 
+    //! The default number of time buckets used to generate multibucket features
+    //! for anomaly detection.
+    static const std::size_t MULTIBUCKET_FEATURE_WINDOW_LENGTH;
```
hendrikmuhs: nit: `MULTIBUCKET_FEATURE_WINDOW_LENGTH` in other places it is `MULTIBUCKET_FEATURES_WINDOW_LENGTH`, feature - > features

lib/maths/CTimeSeriesModel.cc
```diff
@@ -132,14 +168,16 @@ const std::string IS_NON_NEGATIVE_OLD_TAG{"g"};
 const std::string IS_FORECASTABLE_OLD_TAG{"h"};
 
 // Anomaly model
-const std::string MEAN_ERROR_TAG{"a"};
-const std::string ANOMALIES_TAG{"b"};
-const std::string ANOMALY_FEATURE_MODEL_TAG{"d"};
+// Version 6.4
```
hendrikmuhs: up to 6.5?

include/maths/CTimeSeriesMultibucketFeatures.h
```diff
@@ -0,0 +1,173 @@
...
+namespace maths {
+
+//! \brief Defines features on collections of time series values.
+//!
+//! DESCRIPTION:\n
+//! The intention of these is to provide useful features for performing
+//! anomaly detection. Specifically, unusual values of certain properties
+//! of extended time ranges are often the most interesting events in a
+//! time series from a user's perspective.
+class CTimeSeriesMultibucketFeatures {
```
hendrikmuhs: nit: final?

lib/maths/unittest/CTimeSeriesMultibucketFeaturesTest.cc
```diff
@@ -0,0 +1,131 @@
...
+using namespace ml;
+
+using TDouble1Vec = core::CSmallVector<double, 1>;
+using TDouble10Vec = core::CSmallVector<double, 10>;
+using TDouble10Vec1Vec = core::CSmallVector<TDouble10Vec, 1>;
+
+void CTimeSeriesMultibucketFeaturesTest::testMean() {
+    // Test we get the value and weight we expect.
+
+    LOG_DEBUG(<< "Univariate");
```
hendrikmuhs: nit: why not make 2 tests?

lib/maths/unittest/CTimeSeriesMultibucketFeaturesTest.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef INCLUDED_CTimeSeriesMultibucketFeaturesTest_h
+#define INCLUDED_CTimeSeriesMultibucketFeaturesTest_h
+
+#include <cppunit/extensions/HelperMacros.h>
+
+class CTimeSeriesMultibucketFeaturesTest : public CppUnit::TestFixture {
+public:
+    void testMean();
+    void testContrast();
```
hendrikmuhs: not used

hendrikmuhs:
I had another pass, only nitpicks except the version tags which I think should be bumped to 6.5.

PR #179, opened 2018-08-03
[ML] Move version qualifier to variable

droberts195:
Preivously version qualifier like "alpha1" or "rc2" have been
part of the version number stored in the repo.  This means that
to move from a release candidate to a GA build we have to change
a file and rebuild, and this runs the risk that other files have
been changed too.  There is a desire for the GA build to be
built from identical code to the final release candidate.

This change moves the version qualifier from gradle.properties
to a Gradle system property and, in the inner workings of the
build system, an environment variable.

For this to fully work as intended release manager and CI also
need to be updated.  In this initial commit the version qualifier
defaults to "alpha1" if not specified.  The default will be
changed to no version qualifier once CI and release manager are
setting the required variables.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,9 @@ String artifactClassifier = (isWindows ? "windows-x86_64" : ((isMacOsX || cppCro
 project.ext.bash = isWindows ? "C:\\Program Files\\Git\\bin\\bash" : "/bin/bash"
 project.ext.make = (isMacOsX || isWindows) ? "gnumake" : (isLinux ? "make" : "gmake")
 project.ext.numCpus = Runtime.runtime.availableProcessors()
+project.ext.makeEnvironment = [ 'CPP_CROSS_COMPILE': cppCrossCompile,
+                                'VERSION_QUALIFIER': versionQualifier,
+                                'SNAPSHOT': (isSnapshot ? 'yes' : 'no') ]
 
 configurations.all {
   // check for updates every build
@@ -51,9 +54,7 @@ configurations.all {
 }
 
 task clean(type: Exec) {
-  environment 'CPP_CROSS_COMPILE', cppCrossCompile
-  environment 'VERSION_QUALIFIER', versionQualifier
-  environment 'SNAPSHOT', isSnapshot ? 'yes' : 'no'
+  environment makeEnvironment
   commandLine bash
   args '-c', 'source ./set_env.sh && rm -rf build && ' + make + ' clean'
   workingDir "${projectDir}"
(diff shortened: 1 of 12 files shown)


mk/make_rc_defines.sh
```diff
@@ -2,8 +2,11 @@
 
 ML_USER=`id | awk -F')' '{ print $1 }' | awk -F'(' '{ print $2 }'`
 ML_VERSION_STR=`cat $CPP_SRC_HOME/gradle.properties | grep '^elasticsearchVersion' | awk -F= '{ print $2 }' | xargs echo`
+if [ -n "$VERSION_QUALIFIER" ] ; then
+    ML_VERSION_STR="$ML_VERSION_STR-$VERSION_QUALIFIER"
+fi
 if [ "$SNAPSHOT" != no ] ; then
-    ML_VERSION_STR="$ML_VERSION_STR"-SNAPSHOT
+    ML_VERSION_STR="$ML_VERSION_STR-SNAPSHOT"
```
davidkyle: Was this intended? I don't think you want to move the closing quote

build.gradle
```diff
@@ -47,7 +52,8 @@ configurations.all {
 
 task clean(type: Exec) {
   environment 'CPP_CROSS_COMPILE', cppCrossCompile
-  environment 'SNAPSHOT', snapshot ? 'yes' : 'no'
+  environment 'VERSION_QUALIFIER', versionQualifier
```
davidkyle: Is this need for clean, strip and test?

mk/make_rc_defines.sh
droberts195: I moved it to be more like the new line for the qualifier a little higher up.  It doesn't make any difference functionally.  Happy to put it back to how it was though if you think it looked better before.

lib/ver/Makefile
```diff
@@ -16,7 +16,10 @@ CLEAN_CMDS=$(RM) CBuildInfo.cc
 BUILD_YEAR=$(shell date '+%Y')
 USER_NAME_CMD=id | awk -F')' '{ print $$1 }' | awk -F'(' '{ print $$2 }'
 USER_NAME=$(shell $(USER_NAME_CMD))
-PRODUCT_VERSION=$(shell cat $(CPP_SRC_HOME)/gradle.properties | grep '^elasticsearchVersion' | awk -F= '{ print $$2 }' | xargs echo)
+PRODUCT_VERSION:=$(shell cat $(CPP_SRC_HOME)/gradle.properties | grep '^elasticsearchVersion' | awk -F= '{ print $$2 }' | xargs echo)
+ifneq ($(VERSION_QUALIFIER),)
+PRODUCT_VERSION:=$(PRODUCT_VERSION)-$(VERSION_QUALIFIER)
+endif
 ifeq ($(SNAPSHOT),yes)
```
davidkyle: Suggestion: Instead of SNAPSHOT and VERSION_QUALIFIER both could be passed as a single string VERSION_QUALIFIER_SNAPSHOT (or just VERSION_QUALIFIER) then both wouldn't need to be passed in the gradle tasks (make, clean, etc)

droberts195: I think it's better to keep the two variables separate so there's a one-to-one correspondence between the Gradle side and the `make` side.  For example, it will make things easier for the release-manager team when they have to set similar variables for the Docker scripts.

I think there is another way to avoid the duplication, by using `configure` and a closure.  I'll give it a go and push another commit if it works.

droberts195: I couldn't get `configure` and a closure to work, but environment variables can be set as a map as well as individually, so I created a single project-wide map first and then each `Exec` task just has one line to use it.

Issue #181, opened 2018-08-07
[ML] Partition-wise Scoring


edsavage:
True per partition normalisation functionality is viewed as undesirable since if the cardinality of the partition field is high it bypasses the rate limiting we do for alerting. Users could be subjected to a continuous stream of alerts if misused. 

However, there is one important use case which does come up occasionally: anomaly detection is performed as a service to a group of customers and each customer only views their own record level results. In this case, we absolutely don't want the normalisation to interact between customers and since they only see results relating to them the concern about creating too many alerts doesn't apply.

First steps should be:

1. Investigate moving to maintaining per time series minimum probability and reassess partition normalisation in this context,
2. Investigate results of partition analysis verses multiple jobs (one for each partition) where cardinality makes this possible

droberts195:
This was released in 6.5 - closing

PR #182, opened 2018-08-10
[ML] Partition-wise max-scores

edsavage:
Introduces a greater amount of independence between partitions as to how
scores are normalized.

Maximum scores are stored in maps within each normalizer. These values
are effectively keyed on the set
{partitionName, partitionValue, personName, personValue}.

Corresponding changes are required to the ml plugin in elasticsearch in
order that the personValue field be passed to the normalizer process.

The old-style single m_MaxScore per normalizer is maintained alongside
the new per-partition scores for backwards compatibility and for a
viable upgrade path for existing jobs

Relates to https://github.com/elastic/ml-cpp/issues/181


lib/model/unittest/CHierarchicalResultsTest.cc
```diff
@@ -1992,3 +2016,13 @@ CppUnit::Test* CHierarchicalResultsTest::suite() {
 
     return suiteOfTests;
 }
+
+void CHierarchicalResultsTest::setUp() {
+    // Enable trace level logging for these unit tests
+    ml::core::CLogger::instance().setLoggingLevel(ml::core::CLogger::E_Trace);
```
droberts195: I'm not convinced it's worth committing this and the associated teardown to the repo because:

1. Nobody is likely to read the trace logging of this test when run as part of a full end-to-end run
2. Apart from the nightly debug build our CI runs will compile out trace logging altogether

edsavage:
I've removed `m_MaxScore` entirely as @tveasey suggested. I think this is fine in master, however for backwards compatibility in 6.5 I think we will need to continue to persist/restore it.

lib/api/CResultNormalizer.cc
```diff
@@ -71,14 +72,26 @@ bool CResultNormalizer::handleRecord(const TStrStrUMap& dataRowFields) {
...
     double probability(0.0);
 
-    bool isValidRecord = parseDataFields(dataRowFields, level, partition, person,
+    // As of version 6.5 the 'personValue' field is required for (re)normalization to succeed.
+    // In the case of renormalization the 'personValue' field must be included in the set of 
+    // parameters sent from the Java side ML plugin to Elasticsearch.
+    // In production the version of the native code application and the java application it is communicating directly
+    // with will always match so supporting BWC with versions prior to 6.5 is not necessary.
+    bool isNormalizable = this->parseDataFields(dataRowFields, level, partition,
+                                         partitionValue, person, personValue,
```
tveasey: This doesn't look like it has been reformatted. Might be worth reformatting all files in this change again.

lib/model/CHierarchicalResultsNormalizer.cc
```diff
@@ -123,16 +123,30 @@ void CHierarchicalResultsNormalizer::visit(const CHierarchicalResults& /*results
...
+    const std::string partitionName = (node.s_Spec.s_PartitionFieldName)
+                                          ? *(node.s_Spec.s_PartitionFieldName)
+                                          : "";
+    const std::string partitionValue = (node.s_Spec.s_PartitionFieldValue)
+                                           ? *(node.s_Spec.s_PartitionFieldValue)
+                                           : "";
+    const std::string personName =
+        (node.s_Spec.s_PersonFieldName) ? *(node.s_Spec.s_PersonFieldName) : "";
+    const std::string personValue =
+        (node.s_Spec.s_PersonFieldValue) ? *(node.s_Spec.s_PersonFieldValue) : "";
```
tveasey: Can we use a static const empty string here to avoid copying strings (maybe you can factor out as something like a `const std::string& dereferenceOrEmpty(const std::string&))`.

droberts195:
> I've removed `m_MaxScore` entirely as @tveasey suggested. I think this is fine in master, however for backwards compatibility in 6.5 I think we will need to continue to persist/restore it.

It’s permissible to upgrade directly from 6.0 to 7.last via a full cluster restart. Would the removal of 7.x’s ability to read `m_MaxScore` result in inferior quantiles following such an upgrade? If so it might be worth keeping until 8.0.

edsavage:
Thanks for the info regarding upgrade paths Dave. 

In the case of m_MaxScore being missing in 7.x after such an upgrade the quantiles will not be adversely affected. The maximum score ceiling is just one factor in the process of normalisation - scores will continue to be normalised by rank in its absence. Note that this is only an issue with re-normalisation as subsequent updates of the quantiles will transition to the new per-partition scheme.

PR #189, opened 2018-08-28
[ML] Bugfix. Add value counts to MetricModel results

edsavage:
Vector of value counts was never populated causing assertion failures
and hence CMetricModelTest::testInterimCorrectionsWithCorrelations to
fail when run from a debug enabled build.


droberts195:
Adding `>non-issue` label as the bug was never in a released version so doesn't need a release note.

PR #194, opened 2018-09-03
[ML] Add notice stating how to obtain the Eigen source code

droberts195:


legalastic:
@droberts195 Dave--I'm not sure whether to edit the text directly or send a comment (like here).  I would revise the text to add a sentence at the very beginning to read:  "Eigen is licensed under Mozilla Public License Version 2.0 (MPL 2.0).  A MPL 2.0 and FAQ's for MPL 2.0 can be found at:  https://www.mozilla.org/en-US/MPL/2.0/ and https://www.mozilla.org/en-US/MPL/2.0/FAQ/."

Then keep the sentence that follows as you've written it:  "A copy of the Eigen source code may be obtained from http://bitbucket.org/eigen/eigen/get/3.2.5.zip"

droberts195:
Thanks @legalastic.  I updated the notice with your extra sentence.

PR #195, opened 2018-09-03
[ML] Improvements to address QA regressions associated with multi-bucket

tveasey:
This makes some further changes to multi-bucket analysis to address regressions in our QA suite. In particular:
1) I fixed a bug in the feature persist which meant the sliding window length could be incorrect after persist and restore.
2) I have reverted to something closer to what we had previously for features used to model anomalous intervals: duration and average absolute difference from the prediction.
3) I only update the anomalous time interval features if the *bucket* value is unusual: this is to mitigate the interaction between the anomaly model and multi-bucket features, which otherwise (positively) skew our anomalous interval feature values.
4) I avoid the systematic increase in sparse metric probabilities associated with multi-bucket by applying a correction to both probabilities (bucket and multi-bucket features).

Since we are also unable to restore the 6.4 anomaly model, I also take the opportunity to slightly simplify anomaly modelling by removing the tagging mechanism. This would only be needed for population models, for which modelling anomalous time intervals is disabled.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -170,14 +170,14 @@ const std::string IS_FORECASTABLE_OLD_TAG{"h"};
 
 // Anomaly model
 // Version >= 6.5
-const std::string ANOMALY_6_5_TAG{"b"};
-const std::string ANOMALY_FEATURE_MODEL_6_5_TAG{"d"};
+const std::string ANOMALY_6_5_TAG{"e"};
+const std::string ANOMALY_FEATURE_MODEL_6_5_TAG{"f"};
 // Version < 6.5
 // Discarded on state upgrade because features have changed.
-// Anomaly model nested
-const std::string OPEN_TIME_TAG{"a"};
-const std::string SUM_PREDICTION_ERROR_TAG{"b"};
-const std::string MEAN_ABS_PREDICTION_ERROR_TAG{"c"};
+// Anomaly only restored for 6.5 state.
+const std::string OPEN_TIME_6_5_TAG{"a"};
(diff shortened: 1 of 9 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -170,15 +170,14 @@ const std::string IS_FORECASTABLE_OLD_TAG{"h"};
...
 // Version < 6.5
 // Discarded on state upgrade because features have changed.
 // Anomaly model nested
-const std::string TAG_TAG{"a"};
-const std::string OPEN_TIME_TAG{"b"};
-const std::string SUM_PREDICTION_ERROR_TAG{"c"};
-const std::string MEAN_MINUS_LOG_PROBABILITY_TAG{"d"};
+const std::string OPEN_TIME_TAG{"a"};
+const std::string SUM_PREDICTION_ERROR_TAG{"b"};
+const std::string MEAN_ABS_PREDICTION_ERROR_TAG{"c"};
```
droberts195: I haven't looked that closely, but something seems wrong here, or at least confusing.  If these tags are being used to parse 6.4 state which is then discarded then surely the values need to be the same as before?  Or if these tags used to be used to parse a completely different section of the state in 6.4 to what they're now being used to parse then I think the comment on lines 175-177 should be made clearer.

tveasey: These tags are not used to parse 6.4 state. This object is never restored from 6.4, i.e. see [this](https://github.com/elastic/ml-cpp/blob/master/lib/maths/CTimeSeriesModel.cc#L575) line. I'll update the comment to make this clear.

PR #198, opened 2018-09-04
[ML] Improve time series decomposition in the presence of change points

tveasey:
The problems related to #159 showed up because of a fundamental issue with the way we decompose a time series in the presence of change points. In particular, we don't consider that the window of values we sketch might include common types of change points, that we've handled in the rest of our modelling since #92. There are two potential issues as a result:
1) We aren't able to find significant components of the time series because they are distorted by changes
2) If we do detect a periodic component we can get poorly initialised components as a result (the unit test disabled by #159 is an example of this problem, which this change re-enables)

This PR tackles this problem head on by allowing for the common types of change point we handle, linear scaling and piecewise linear trend, when testing for and initialising new components.

The bulk of this change is in code to perform and test top-down segmentation of a window of values, `CTimeSeriesSegmentation`, allowing for either piecewise linear trend or piecewise constant modulation of a seasonal component. We use our standard approach of doing model selection based statistical significance of the explained variance to choose the segmentation. (We can do this relatively easily because each candidate partition creates a nested hypothesis.) On this front, although bottom up partitioning is generally considered preferable, this is primarily concerned with dealing with large change events, which will be identified by a top-down approach. It is also much more compute efficient and we do this multiple times for different hypotheses we test to find the best decomposition.


lib/maths/unittest/CTimeSeriesDecompositionTest.cc
```diff
@@ -469,88 +469,89 @@ void CTimeSeriesDecompositionTest::testMinimizeLongComponents() {
...
-    TDoubleVec trend;
-    for (core_t::TTime time = 0; time < 100 * WEEK; time += HALF_HOUR) {
-        double weight = weights[(time / DAY) % 7];
-        double daily = 100.0 * std::sin(boost::math::double_constants::two_pi *
-                                        static_cast<double>(time) /
-                                        static_cast<double>(DAY));
-        times.push_back(time);
-        trend.push_back(weight * daily);
-    }
+    for (auto offset : {0 * DAY, 5 * DAY}) {
```
tveasey: This is just the same test performed for different offsets of the start time: the bulk of the changes are indentation.

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1317,47 +1490,85 @@ void CPeriodicityHypothesisTests::hypothesis(const TTime2Vec& periods,
...
+                    momentsAtOffset.add(CBasicStatistics::mean(values[j]),
+                                        CBasicStatistics::count(values[j]));
+                }
+                moments += momentsAtOffset;
+                b += (CBasicStatistics::count(momentsAtOffset) > 0.0 ? 0.0 : 1.0);
+            }
+
+            stats.s_V0 = CBasicStatistics::variance(moments);
+            stats.s_DF0 = b + static_cast<double>(stats.s_Segmentation.size() - 2);
+        } else {
```
tveasey: The old function was this branch.

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1317,47 +1490,85 @@ void CPeriodicityHypothesisTests::hypothesis(const TTime2Vec& periods,
...
-            std::size_t a{window.first};
-            std::size_t b{window.second};
-            for (std::size_t j = a; j < b; ++j) {
-                CBasicStatistics::moment<0>(buckets[j % n]) -=
-                    stats.s_T0[i][(j - a) % period];
+
+    if (stats.s_Segmentation.size() > 0) {
+        buckets = CTimeSeriesSegmentation::removePiecewiseLinearScaledPeriodic(
+            buckets, stats.s_Segmentation, stats.s_T0[0], stats.s_Scales);
+    } else {
```
tveasey: The old function was this branch.

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1376,167 +1587,224 @@ bool CPeriodicityHypothesisTests::testPeriod(const TTimeTimePr2Vec& windows,
...
     periodicTrend(values, window, m_BucketLength, trend);
+    double v1{residualVariance<double>(trend, scale)};
+
+    double R;
+    double meanRepeats;
+    double pVariance;
+    return this->testVariance(window, values, period_, df1, v1, stats, R,
+                              meanRepeats, pVariance) ||
+           this->testAmplitude(window, values, period_, b, v, R, meanRepeats,
+                               pVariance, stats);
```
tveasey: I factored out `testVariance` and `testAmplitude` from this function (for reuse in `CPeriodicityHypothesisTests::testPeriodWithScaling`). These are what is being compared to the following code.

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1376,167 +1587,224 @@ bool CPeriodicityHypothesisTests::testPeriod(const TTimeTimePr2Vec& windows,
...
         LOG_TRACE(<< "Projecting onto " << core::CContainerPrinter::print(windows));
         TFloatMeanAccumulatorVec projection;
         project(values, windows, m_BucketLength, projection);
         values = std::move(projection);
     }
 
-    double B{static_cast<double>(std::count_if(
+    // Compute the number of non-empty buckets. This needs to account for
+    // projecting onto windows.
+    stats.s_B = static_cast<double>(std::count_if(
```
edsavage: It might be useful to take the opportunity to rename the data members of STestStats to something more descriptive?

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
```diff
@@ -1540,85 +1541,78 @@ void CTimeSeriesDecompositionTest::testNonDiurnal() {
...
-                        totalMaxResidual += maxResidual;
-                        totalSumValue += sumValue;
-                        totalMaxValue += maxValue;
-
-                        CPPUNIT_ASSERT(sumResidual / sumValue < 0.35);
-                        CPPUNIT_ASSERT(maxResidual / maxValue < 0.33);
+        rng.generateNormalSamples(0.0, 1.0, trend.size(), noise);
+
+        core_t::TTime startTesting{pad + 28 * HOUR};
+        //TDoubleVec thresholds[]{TDoubleVec{0.08, 0.08}, TDoubleVec{0.18, 0.15}};
```
edsavage: nitpick: delete this line rather than comment out?

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -281,7 +286,7 @@ void project(const TFloatMeanAccumulatorVec& values,
              core_t::TTime bucketLength,
              TFloatMeanAccumulatorVec& result) {
     result.clear();
-    if (!values.empty()) {
+    if (values.size() > 0) {
```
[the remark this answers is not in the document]
tveasey: Our standards now say to not use `!` but instead `== false` to match Elastic Java style. I personally find size() > 0 easier to read than empty() == false at a glance. I changed to be more compliant with our style guidelines because I was making quite a few changes in this file.

include/maths/CPeriodicityHypothesisTests.h
```diff
@@ -267,13 +283,19 @@ class MATHS_EXPORT CPeriodicityHypothesisTests {
...
+        std::size_t trendSegments() const;
 
     private:
         using THypothesisVec = std::vector<CNestedHypotheses>;
 
     private:
         //! The test.
         TTestFunc m_Test;
+        //! True if the hypothesis used a piecewise linear trend.
+        std::size_t m_TrendSegments;
```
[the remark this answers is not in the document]
tveasey: Agreed. This used to be a bool but we actually need the number of segments in the calling code so this changed. I'll update.

lib/maths/CPeriodicityHypothesisTests.cc
edsavage: Lovely, thanks Tom

include/maths/CPeriodicityHypothesisTests.h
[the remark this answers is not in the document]
tveasey: See [this](https://github.com/elastic/ml-cpp/pull/198/commits/a8b78ddee61405f2194fb4e2058e434afbd31018) commit.

PR #204, opened 2018-09-13
[ML] Global max score to trigger renormalizations

edsavage:
Using a global maximum score (the maximum score see for the normalizer
as a whole) to trigger renormalization when a big change occurs.

Current behaviour is to set the 'big change' flag on the normalizer when
the maximum score of any partition has changed a significant amount.
This ultimately results in the normalizer state being persisted
triggering a renormalizationi.

This change heavily reduces the number of renormalization events produced and
results in a significant performance increase.

### DIFF
--- a/include/model/CAnomalyScore.h
+++ b/include/model/CAnomalyScore.h
@@ -254,6 +254,10 @@ public:
         //! The dictionary to use to associate partitions to their maximum scores
         TDictionary m_Dictionary;
 
+        //! The maximum score seen for the normalizer as a whole
+        //! Strictly used to trigger renormalizations on big change
+        TMaxValueAccumulator m_MaxScore;
+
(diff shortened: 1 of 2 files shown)


edsavage:
Thanks for the review @tveasey.

Yes, tests that analysed performance with and without this change were performed.

With the change the number of renormalization events that were triggered were reduced to the same number that occurred prior to the per-portion max score change was introduced. In one of the jobs used for testing this was a factor of 6.

Timing of job duration has proved a little troublesome to get consistent results but with this change, on average the jobs are sped up a noticeable amount.

tveasey:
Perhaps we should wait until all the other changes have been through the regression suite so we can see where we are in terms of any runtime degradation before merging. Also, it might be worthwhile discussing options here a bit further.

PR #213, opened 2018-09-24
[ML] Add mechanism to remove "stale" normaliser maximum scores

### LINKED ISSUE #210: [ML] Per partition normalisation tracking maximum scores for distinct over field and all influencer fields
There were a couple of inadvertent changes introduced by #182. These have caused large slow downs, and also many more high severity anomalies, in some of our regression suite data sets. The changes we want to "revert" are:
- [x] We now have one maximum score per distinct over field value for population analysis
- [ ] We now have one maximum score per distinct value of *all* influencer fields
- [ ] We don't remove maximum scores after removing the corresponding models

Items 1 and 2, in the case that the influencer is not one of the detector partitioning fields, was not intended in the original plan for this work. The current behaviour is not really appropriate since the presence of particular values of these fields in the data set are often transitory.

Item 3 means we gradually "leak" maximum scores over time and could eventually cause us problems.

This hasn't been released so marking as a non-issue.

tveasey:
Currently, "per partition" maximum scores are never deleted. This means that over time if there is churn in the values present for a partition, by field or influencer field we will gradually accumulate state for them.

This provides a simple mechanism to remove scores which haven't received any update for some period of time. Field values which appear frequently are thus never removed. Because the maximum score is aged over time, removing old scores should have little impact on normalisation, since old maximum scores are unlikely to constrain the normalised score of a newly received value.

Since this only applies to unreleased code I've marked this as a non-issue. Closes #210.


include/model/CAnomalyScore.h
```diff
@@ -175,40 +173,72 @@ class MODEL_EXPORT CAnomalyScore {
...
+            TMaxValueAccumulator m_Score;
+            double m_TimeSinceLastScore;
+        };
+        using TWordMaxScoreUMap = TDictionary::CWordUMap<CMaxScore>::Type;
 
     private:
         //! Used to convert raw scores in to integers so that we
         //! can use the q-digest.
-        static const double DISCRETIZATION_FACTOR;
+        constexpr static const double DISCRETIZATION_FACTOR = 1000.0;
```
droberts195: Do you need both `const` and `constexpr`?  I would have thought `constexpr` implies `const`.

Also, since we used to write `const` after `static`, should we also write `constexpr` after `static` since it's like a stronger form of `const`?

lib/api/unittest/testfiles/new_quantilesState.json
```diff
@@ -1,6 +1,6 @@
...
+,{"mlcue":"sysChange","mlkey":"api","mlquantilesdescription":"root","mlversion":"3","time":"1486855800","a":{"a":"0","b":"470","c":"","d":{"a":"201","b":"479","c":{"a":"0","b":"32767","c":"0"},"c":{"a":"0","b":"3","c":"1"},"c":{"a":"0","b":"0","c":"470"},"c":{"a":"1","b":"1","c":"3"},"c":{"a":"4","b":"7","c":"1"},"c":{"a":"8","b":"8","c":"1"},"c":{"a":"9","b":"9","c":"1"},"c":{"a":"16384","b":"20479","c":"1"},"c":{"a":"20480","b":"24575","c":"1"}},"e":{"a":"201","b":"9","c":{"a":"0","b":"32767","c":"0"},"c":{"a":"1","b":"1","c":"3"},"c":{"a":"3","b":"3","c":"1"},"c":{"a":"4","b":"4","c":"1"},"c":{"a":"8","b":"8","c":"1"},"c":{"a":"9","b":"9","c":"1"},"c":{"a":"19875","b":"19875","c":"1"},"c":{"a":"21954","b":"21954","c":"1"}},"f":"1.000000","h":"0","g":{"d":"1","a":{"a":"7382284430162962205","b":{"a":"19.64787","b":"0"}}}}}
```
droberts195: It looks like this has changed `"b":"19.64787"` to `"b":{"a":"19.64787","b":"0"}`.

I think from a BWC perspective it would be nicer if it could instead be changed to `"b":"19.64787","c":"0"`.  Otherwise it will fail to parse if a job runs on a 6.5 but then on 6.4.

tveasey: This is a bigger change because we currently use the `CPersistUtils` framework to persist this container. Do we think that it is a realistic scenario that we will need to restore 6.5 state into a 6.4 analytics process because I think there will potentially be other issues around trying to load newer state into an older model.

droberts195: Turns out my worry was misplaced, because the whole `"b"` section is new in 6.5, which is not released.  So no changes necessary.

lib/model/CAnomalyScore.cc
```diff
@@ -68,14 +68,15 @@ double scoreToProbability(double score) {
 // We use short field names to reduce the state size
 const std::string HIGH_PERCENTILE_SCORE_TAG("a");
 const std::string HIGH_PERCENTILE_COUNT_TAG("b");
-const std::string MAX_SCORE_TAG("c");
```
hendrikmuhs: would be good, to leave a comment that `c` was in use prior 6.4, otherwise someone might accidentally re-add `c`.

lib/model/CAnomalyScore.cc
```diff
@@ -68,14 +68,15 @@ double scoreToProbability(double score) {
...
 const std::string RAW_SCORE_QUANTILE_SUMMARY("d");
 const std::string RAW_SCORE_HIGH_QUANTILE_SUMMARY("e");
 const std::string TIME_TO_QUANTILE_DECAY_TAG("f");
 
 // Since version >= 6.5
 const std::string MAX_SCORES_PER_PARTITION_TAG("g");
 const std::string IS_FOR_MEMBERS_OF_POPULATION_TAG("h");
+const std::string MAX_SCORE_TAG("a");
```
hendrikmuhs: would be nice to add a comment that this is nested in key ...

lib/model/CAnomalyScore.cc
tveasey: Done

PR #214, opened 2018-09-25
[ML] Remove hard limit for maximum forecast interval

tveasey:
We've had requests to support longer time interval forecasts. With the work done on the underlying forecast methodology in 6.3 and 6.4, we are now able to provide better long range forecasts and the old hard limit is too restrictive.

This introduces a new dynamic limiting mechanism for forecasts based on the amount of historical data observed and the amount of data the model can use. In particular, we limit the time we'll forecast ahead to the data end time plus the data interval or (if smaller) the "amount of history" in a largest timescale trend component.

As before, forecasts are truncated if they exceed the limit.


sophiec20:
Can we please pause merging this until we have resource to validate results? I suggest pushing to 6.6.

tveasey:
I think we can merge, just not back port, if you want to delay. This has no impact on any results so it is safe to have on master

lib/model/CForecastDataSink.cc
```diff
@@ -56,33 +56,19 @@ const std::string CForecastDataSink::STATUS("forecast_status");
...
+    : s_Feature(feature), s_FirstDataTime(firstDataTime), s_LastDataTime(lastDataTime),
+      s_ForecastModel(std::move(forecastModel)), s_ByFieldValue(byFieldValue) {
 }
 
 CForecastDataSink::SForecastResultSeries::SForecastResultSeries(const SModelParams& modelParams)
     : s_ModelParams(modelParams), s_DetectorIndex(), s_ToForecastPersisted(),
       s_ByFieldName(), s_MinimumSeasonalVarianceScale(0.0) {
 }
 
-CForecastDataSink::SForecastResultSeries::SForecastResultSeries(SForecastResultSeries&& other)
```
hendrikmuhs: be careful, this looks superfluous but is there for a reason: VS/Windows builds (though maybe only on 6.x). Would be good to run this change through `mybranch` prior commit

lib/maths/CTimeSeriesModel.cc
```diff
@@ -964,17 +966,28 @@ CUnivariateTimeSeriesModel::confidenceInterval(core_t::TTime time,
...
                                           core_t::TTime endTime,
                                           double confidenceInterval,
                                           const TDouble2Vec& minimum_,
                                           const TDouble2Vec& maximum_,
                                           const TForecastPushDatapointFunc& forecastPushDataPointFunc,
                                           std::string& messageOut) {
-    if (m_ResidualModel->isNonInformative()) {
+
+    core_t::TTime horizon{std::min(lastDataTime + (lastDataTime - firstDataTime),
+                                   lastDataTime + m_TrendModel->maximumForecastInterval())};
```
hendrikmuhs: I think we should still have a sensible upper limit just to prevent someone requesting a `1000 years forecast` although he might need to tweak the data set to do that.

tveasey: This is ensured by `m_TrendModel->maximumForecastInterval()`. For our current model parameterisation the limit works out at just over 1600 days. I think I'd prefer to tie this to the model behaviour rather than use a fixed hard limit. I'll add a comment to this effect.

lib/model/CForecastDataSink.cc
tveasey: Thanks, I'll check. I think if this works on master I'll leave with the default implementation and then adjust in the back port if necessary.

sophiec20:
Re internal discussion, please do not merge to master until results are stabilised. There is a bit too much going on atm. As this is not required work this week, please leave on hold.

PR #218, opened 2018-09-27
[ML] Improve initialisation of the residual model after detecting new decomposition component

tveasey:
Currently we use a small random sample of historical values to initialise the prediction residual model after we've detected new components of the time series decomposition. This sample is not large enough to be reliably representative of true variation we should expect and can occasionally lead to spurious anomalies immediately after a component is detected. As a result of the increased sensitivity related to #181 this has become more important.

We actually have a better sample available in the window of values we use to perform decomposition (albeit aggregated at a different time scale than the bucketing interval). This change switches to use these values instead to reinitialise the residual model.

### DIFF
--- a/include/maths/CTimeSeriesDecomposition.h
+++ b/include/maths/CTimeSeriesDecomposition.h
@@ -166,6 +166,9 @@ public:
     virtual maths_t::TDoubleDoublePr
     scale(core_t::TTime time, double variance, double confidence, bool smooth = true) const;
 
+    //! Get the values in a recent time window if they are available.
+    virtual TTimeDoublePrVec windowValues(core_t::TTime time, bool forced = false) const;
+
     //! Roll time forwards by \p skipInterval.
     virtual void skipTime(core_t::TTime skipInterval);
 
@@ -182,7 +185,7 @@ public:
     virtual std::size_t staticSize() const;
 
     //! Get the time shift which is being applied.
-    virtual core_t::TTime timeShift(void) const;
+    virtual core_t::TTime timeShift() const;
 
     //! Get the seasonal components.
     virtual const maths_t::TSeasonalComponentVec& seasonalComponents() const;
--- a/include/maths/CTimeSeriesDecomposition.h
+++ b/include/maths/CTimeSeriesDecomposition.h
@@ -166,8 +166,11 @@ public:
     virtual maths_t::TDoubleDoublePr
     scale(core_t::TTime time, double variance, double confidence, bool smooth = true) const;
 
-    //! Get the values in a recent time window if they are available.
-    virtual TTimeDoublePrVec windowValues(core_t::TTime time, bool forced = false) const;
+    //! Check if this might add components between now and \p time.
+    virtual bool mightAddComponents(core_t::TTime time) const;
+
+    //! Get the values in a recent time window.
+    virtual TTimeDoublePrVec windowValues() const;
 
     //! Roll time forwards by \p skipInterval.
     virtual void skipTime(core_t::TTime skipInterval);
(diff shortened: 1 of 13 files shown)


include/maths/CTimeSeriesDecomposition.h
```diff
@@ -166,6 +166,9 @@ class MATHS_EXPORT CTimeSeriesDecomposition : public CTimeSeriesDecompositionInt
     virtual maths_t::TDoubleDoublePr
     scale(core_t::TTime time, double variance, double confidence, bool smooth = true) const;
 
+    //! Get the values in a recent time window if they are available.
+    virtual TTimeDoublePrVec windowValues(core_t::TTime time, bool forced = false) const;
```
dimitris-athanasiou: Could you document what forcing means here?

lib/maths/CTimeSeriesModel.cc
```diff
@@ -148,12 +127,12 @@ const std::string VERSION_6_5_TAG("6.5");
 const std::string ID_6_3_TAG{"a"};
 const std::string IS_NON_NEGATIVE_6_3_TAG{"b"};
 const std::string IS_FORECASTABLE_6_3_TAG{"c"};
-const std::string RNG_6_3_TAG{"d"};
+//const std::string RNG_6_3_TAG{"d"};
```
dimitris-athanasiou: Do we usually just comment out removed tags? Might be better to include a message that includes the version of removal.

lib/maths/CTimeSeriesModel.cc
```diff
@@ -1250,7 +1214,7 @@ uint64_t CUnivariateTimeSeriesModel::checksum(uint64_t seed) const {
     seed = CChecksum::calculate(seed, m_CandidateChangePoint);
     seed = CChecksum::calculate(seed, m_CurrentChangeInterval);
     seed = CChecksum::calculate(seed, m_ChangeDetector);
-    seed = CChecksum::calculate(seed, m_RecentSamples);
+    seed = CChecksum::calculate(seed, m_AnomalyModel);
```
dimitris-athanasiou: This looks like it's being calculated twice now in this function

include/maths/CTimeSeriesDecomposition.h
tveasey: On reflection I think a slight refactor makes this easier to understand. I factored out the logic to test to see if the components might be added so this now always returns the decomposition window values. See [this](https://github.com/elastic/ml-cpp/pull/218/commits/01ff26346f4464f8428e8515ac9da7515f18b7ca) commit.

tveasey:
@dimitris-athanasiou, I addressed your comments in my last commit. Can you take another look?

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -579,6 +579,27 @@ void CTimeSeriesDecompositionDetail::CPeriodicityTest::handle(const SNewComponen
     // components only when we test.
 }
 
+bool CTimeSeriesDecompositionDetail::CPeriodicityTest::shouldTest(ETest test,
+                                                                  core_t::TTime time) const {
+    // We need to test more frequently than we compress because it
+    // only happens each 336 buckets and would significantly delay
```
dimitris-athanasiou: Where is this `336` coming from? It would be nice to explain that too in the comment.

tveasey: Added an explanation in [this](https://github.com/elastic/ml-cpp/pull/218/commits/bd8854ed218330671b07160db35c161c76a312d4) commit.

PR #222, opened 2018-09-28
[ML] Skip model update in anomaly model

### LINKED ISSUE #217: [ML] Skipping model updates should also skip the anomaly model
Let us imagine a job which has a simple `high_count` detector. The detector also has a rule to `skip_model_update` when the count is over `100`. Then our data has a normal event rate which is below `100`. But at some point there are several buckets over `100`. We shall expect the model to remain unaffected and that all those buckets are detected as anomalies.

However, that is not the case. The screenshot below shows what currently happens.

image

This is most probably because the anomaly model is not skipped. Thus, we start learning that anomalies as normal and their score drops. We should ensure the anomaly model is also skipped when such a rule exists.

dimitris-athanasiou:
Skip model update was not being communicated
to the anomaly model resulting to the score/probability
of results triggering the skip-model-update rule
to reduce.

This commit skips updates to the anomaly model as well.

Closes #217

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -78,4 +78,8 @@ Change linker options on macOS to allow Homebrew installs ({ml-pull}225[225])
 
 //=== Bug Fixes
 
+Rules that trigger the `skip_model_update` action should also apply to the anomaly model.
+This fixes an issue where anomaly scores and probabilities on results that triggered
+the rule would reduce is they occurred frequently. (See {pull}222[#222].)
+
 //=== Regressions
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -79,7 +79,7 @@ Change linker options on macOS to allow Homebrew installs ({ml-pull}225[225])
 //=== Bug Fixes
 
 Rules that trigger the `skip_model_update` action should also apply to the anomaly model.
-This fixes an issue where anomaly scores and probabilities on results that triggered
-the rule would reduce is they occurred frequently. (See {pull}222[#222].)
+This fixes an issue where anomaly scores of results that triggered the rule would decrease
+if they occurred frequently. (See {ml-pull}222[#222].)
 
 //=== Regressions
--- a/lib/maths/unittest/CTimeSeriesModelTest.cc
+++ b/lib/maths/unittest/CTimeSeriesModelTest.cc
@@ -2361,7 +2361,7 @@ void CTimeSeriesModelTest::testSkipAnomalyModelUpdate() {
                 currentComputeProbabilityParams.skipAnomalyModelUpdate(true);
             } else {
                 model.addSamples(addSampleParams(weights),
-                                {core::make_triple(time, TDouble2Vec{sample}, TAG)});
+                                 {core::make_triple(time, TDouble2Vec{sample}, TAG)});
             }
             maths::SModelProbabilityResult result;
             model.probability(currentComputeProbabilityParams, {{time}}, {{sample}}, result);
@@ -2375,7 +2375,8 @@ void CTimeSeriesModelTest::testSkipAnomalyModelUpdate() {
 
         // Assert probs are decreasing
         double previousProbability = probabilities[0];
-        for (std::size_t currentProbability = 1; currentProbability < probabilities.size(); ++currentProbability) {
+        for (std::size_t currentProbability = 1;
+             currentProbability < probabilities.size(); ++currentProbability) {
             CPPUNIT_ASSERT(probabilities[currentProbability] <= previousProbability);
             previousProbability = probabilities[currentProbability];
         }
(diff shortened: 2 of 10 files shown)


dimitris-athanasiou:
Still owing tests for this but I thought @tveasey could have a look at the code already.

lib/model/CMetricModel.cc
```diff
@@ -577,6 +579,8 @@ void CMetricModel::fill(model_t::EFeature feature,
     const TSize2Vec1Vec& correlates{model->correlates()};
     const TTimeVec& firstBucketTimes{this->firstBucketTimes()};
     core_t::TTime bucketLength{gatherer.bucketLength()};
+    bool skipAnomalyModelUpdate = this->shouldIgnoreSample(feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID,
+                                                           model_t::sampleTime(feature, bucketTime, bucketLength));
```
dimitris-athanasiou: I wasn't quite sure as to what the right `time` parameter is. Could you confirm @tveasey ?

docs/CHANGELOG.asciidoc
```diff
@@ -51,4 +51,8 @@ series decomposition. (See {pull}218[218].)
 
 Fix cause of "Bad density value..." log errors whilst forecasting. ({pull}207[207])
 
+Rules that trigger the `skip_model_update` action should also apply to the anomaly model.
+This fixes an issue where anomaly scores and probabilities on results that triggered
+the rule would reduce is they occurred frequently. (See {pull}222[#222].)
```
droberts195: According to the comment on line 6, `{pull}` should be `{ml-pull}`.

lib/model/CMetricModel.cc
tveasey: Yes I think this is correct. The difference is small, but this means that for metric valued conditions we get the prediction at exactly the time of the the data point was received.

docs/CHANGELOG.asciidoc
dimitris-athanasiou: I fixed this. However, the rest of the `pull`s in the changelog should also be `ml-pull`. I'll raise a PR to fix those in a separate commit.

droberts195: Thanks.  At the same time please could you also add some missing `#` symbols.  Looking at the old release notes the PR numbers that get rendered in HTML start with hashes, e.g. `#123`, but in our changelog not all the numbers have hashes.

lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -2330,6 +2330,106 @@ void CTimeSeriesModelTest::testDaylightSaving() {
...
+            if (bucket >= 1700 && bucket < 1710) {
+                sample = 100.0;
+                currentComputeProbabilityParams.skipAnomalyModelUpdate(true);
+            } else {
+                model.addSamples(addSampleParams(weights),
+                                 {core::make_triple(time, TDouble2Vec{sample}, TAG)});
+            }
+            maths::SModelProbabilityResult result;
+            model.probability(currentComputeProbabilityParams, {{time}}, {{sample}}, result);
+            if (result.s_Probability < 1e-6) {
```
tveasey: Tying this to probability seems a bit fragile to me. Could you perhaps just move
```
maths::SModelProbabilityResult result;
model.probability(currentComputeProbabilityParams, {{time}}, {{sample}}, result);
probabilities.push_back(result.s_Probability);
```
inside the `if (bucket >= 1700 && bucket < 1710)` case?

lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -2330,6 +2330,106 @@ void CTimeSeriesModelTest::testDaylightSaving() {
...
+
+        LOG_DEBUG(<< "probabilities = " << core::CContainerPrinter::print(probabilities));
+
+        // Assert probs are decreasing
+        double previousProbability = probabilities[0];
+        for (std::size_t currentProbability = 1;
+             currentProbability < probabilities.size(); ++currentProbability) {
+            CPPUNIT_ASSERT(probabilities[currentProbability] <= previousProbability);
+            previousProbability = probabilities[currentProbability];
+        }
```
tveasey: As above.

lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -2330,6 +2330,106 @@ void CTimeSeriesModelTest::testDaylightSaving() {
...
```
tveasey: How about `std::is_sorted()`?

lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -2330,6 +2330,106 @@ void CTimeSeriesModelTest::testDaylightSaving() {
...
+            } else {
+                model.addSamples(addSampleParams(weights),
+                                 {core::make_triple(time, TDouble2Vec(sample), TAG)});
+            }
+
+            maths::SModelProbabilityResult result;
+            model.probability(currentComputeProbabilityParams, {{time}}, {(sample)}, result);
+            if (result.s_Probability < 1e-6) {
+                probabilities.push_back(result.s_Probability);
+            }
```
tveasey: As above.

dimitris-athanasiou:
jenkins please test this

PR #225, opened 2018-10-02
[ML] Add -headerpad_max_install_names to macOS linker args

droberts195:
This is to support Homebrew, which changes the @rpaths of libraries
it installs to (potentially very long) absolute paths.


dev-tools/docker/build_macosx_build_image.sh
```diff
@@ -16,7 +16,7 @@
 
 ACCOUNT=droberts195
 REPOSITORY=ml-macosx-build
-VERSION=3
+VERSION=5
```
droberts195: 4 is for 6.x

PR #229, opened 2018-10-03
[ML-Dataframe][FEATURE] A dense data frame

tveasey:
This is a first step towards supporting ML tasks on dense (i.e. mostly no missing elements) Elastic Dataframes.

It provides infrastructure for creating and reading (pieces of) data frames in a process to perform some ML task on them.

The interface is purposely minimal at this point: it supports threaded non-contending reads, sequential writes and the ability to reserve space to append new columns. This should cover many of our planned use cases. The read and write operations are via callbacks to maximally decouple from calling code and also to allow us to do certain operations efficiently. For example, a stream can be read/written directly to/from the data frame storage by wrapping in the appropriate writer/reader functions.

We will have to extend the interface somewhat (for example providing append column values as a minimum). I prefer to wait until the calling code is written to do this, so I've left a couple of TODOs since this is on a feature branch. There is also work to do to communicate errors back to the calling process and some further instrumentation to perform.

These objects come in two flavours: 1) in main memory and 2) on disk to allow us to efficiently process across a range of data scales.

The in main memory variety has very low overhead on top of the storage for the frame elements, typically a few hundred bytes **in total**. For example, using the float storage type, this gives us just under 13.5 million rows with 20 columns per GB. Together with memory mapped linear algebra types (coming soon) we should be able to process reasonably large frames in memory as a result.

The on disk variety has very little main memory usage (typically a few hundred bytes in total) and stores in a specified directory in binary format to maximise read and write speed. (Note these will be local to any machine (container) running the process, so we don't need to worry about the architecture changing between writing and reading.) The intention is that the client code will copy in as much as they need, subject to the memory constraint, and be responsible for partitioning strategies. These are problem specific in any case. Reads and writes from and to storage format are optionally asynchronous to support this variety.

### DIFF
--- a/include/core/CDataFrame.h
+++ b/include/core/CDataFrame.h
@@ -131,12 +131,11 @@ private:
 //! IMPLEMENTATION:\n
 //! This is a fairly lightweight container which is essentially responsible
 //! for managing the read and write process to some underlying store format.
-//! The store format is determined by the user implementing callbacks which are
-//! passed to the data frame constructor to read and write state from the store.
-//! For example, these could copy to / from main memory, "write to" / "read from"
-//! disk, etc. Copying these functions is assumed to have no side effects and they
-//! are assumed to be thread safe. If they have shared state the implementation
-//! must ensure that access to this is synchronized.
+//! The store format is determined by the user implementing functionality to
+//! read and write state from the store. For example, these could copy to /
+//! from main memory, "write to" / "read from" disk, etc. A factory function
+//! must be provided to the constructor which effectively that determines the
+//! type of storage used. It is assumed that copying this has no side effects.
 //!
 //! The data frame is divided into slices each of which represent a number of
 //! contiguous rows. The idea is that they contain a reasonable amount of memory
@@ -144,6 +143,9 @@ private:
 //! "reads from" disk (a whole slice being written or read in one go), mean we'll
 //! get good locality of reference and mean there is minimal book keeping overhead
 //! (such as state for vector sizes, pointers to starts of memory blocks, etc).
+//! In addition, it is assumed that access to the individual slices is thread
+//! safe. If they share state the implementation must ensure that access to this
+//! is synchronized.
 //!
 //! Reads and writes of a single row are also done via call backs supplied to the
 //! readRows and writeRow functions. This is to achieve maximum decoupling from
@@ -169,8 +171,6 @@ public:
     using TRowSlicePtrVec = std::vector<TRowSlicePtr>;
     using TSizeRowSliceHandlePr = std::pair<std::size_t, CDataFrameRowSliceHandle>;
     using TWriteSliceToStoreFunc = std::function<TRowSlicePtr(std::size_t, TFloatVec)>;
-    using TReadSliceFromStoreFunc =
-        std::function<TSizeRowSliceHandlePr(const TRowSlicePtr&)>;
     using TSizeFloatVecPrQueue = CConcurrentQueue<TSizeFloatVecPr, 1>;
 
     //! Controls whether to read and write to storage asynchronously.
@@ -183,7 +183,6 @@ public:
     //! \param[in] asyncReadAndWriteToStore Controls whether reads and writes
     //! from slice storage are synchronous or asynchronous.
     //! \param[in] writeSliceToStore The callback to write a slice to storage.
-    //! \param[in] readSliceFromStore The callback to read a slice from storage.
     //!
     //! \warning This requires that \p writeSliceToStore and \p readSliceFromStore
     //! can be copied and are thread safe. If they are not stateless then it is
@@ -191,14 +190,12 @@ public:
     CDataFrame(std::size_t numberColumns,
                std::size_t sliceCapacityInRows,
                EReadWriteToStorage asyncReadAndWriteToStore,
-               const TWriteSliceToStoreFunc& writeSliceToStore,
-               const TReadSliceFromStoreFunc& readSliceFromStore);
+               const TWriteSliceToStoreFunc& writeSliceToStore);
 
     //! Overload which manages the setting of slice capacity to a sensible default.
     CDataFrame(std::size_t numberColumns,
                EReadWriteToStorage asyncReadAndWriteToStore,
-               const TWriteSliceToStoreFunc& writeSliceToStore,
-               const TReadSliceFromStoreFunc& readSliceFromStore);
+               const TWriteSliceToStoreFunc& writeSliceToStore);
 
     CDataFrame(const CDataFrame&) = delete;
     CDataFrame& operator=(const CDataFrame&) = delete;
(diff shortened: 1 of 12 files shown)


lib/core/CDataFrame.cc
```diff
@@ -0,0 +1,513 @@
...
+    COnDiskDataFrameRowSlice::TTemporaryDirectoryPtr directory{
+        std::make_shared<COnDiskDataFrameRowSlice::CTemporaryDirectory>(
+            rootDirectory, minimumSpace)};
+
+    // Note the writer lambda holds a reference to the directory shared
+    // pointer is copied to the data frame. So this isn't destroyed, and
+    // the folder cleaned up, until the data frame itself is destroyed.
+
+    auto writer = [directory](std::size_t firstRow, TFloatVec slice) {
+        return boost::make_unique<COnDiskDataFrameRowSlice>(directory, firstRow,
```
droberts195: If this is never going to be backported to 6.x you can use `std::make_unique`.  If it _is_ going to be backported to 6.x then there are potential problems with `std::async` not conforming to standard in VS2013: https://stackoverflow.com/questions/44829115/from-vs2013-to-vs2017-stdasync-does-not-start-a-new-thread

lib/core/CDataFrame.cc
```diff
@@ -0,0 +1,513 @@
...
+        std::size_t j{0};
+        for (std::size_t i = 0; i + 1 < numberThreads; ++i, j += stride) {
+            auto begin = m_Slices.begin() + j;
+            auto end = m_Slices.begin() + j + stride;
+            reserves.push_back(std::async(std::launch::async, reserver, begin, end));
+        }
+        auto begin = m_Slices.begin() + j;
+        auto end = m_Slices.end();
+        if (begin != end) {
+            reserves.push_back(std::async(std::launch::async, reserver, begin, end));
```
droberts195: Using `std::launch::async` may introduce headaches with platform differences, thus increasing the testing burden.  From https://docs.microsoft.com/en-gb/cpp/standard-library/future-functions?view=vs-2017:

> The C++ standard states that if policy is launch::async, the function creates a new thread. However the Microsoft implementation is currently non-conforming. It obtains its threads from the Windows ThreadPool, which in some cases may provide a recycled thread rather than a new one. This means that the launch::async policy is actually implemented as launch::async|launch::deferred. Another implication of the ThreadPool-based implementation is that there is no guarantee that thread-local variables will be destroyed when the thread completes. If the thread is recycled and provided to a new call to async, the old variables will still exist. We therefore recommend that you do not use thread-local variables with async.

lib/core/CDataFrame.cc
```diff
@@ -0,0 +1,513 @@
...
+        CDataFrameRowSliceReserver reserver{
+            m_NumberColumns, m_RowCapacity - m_NumberColumns};
+
+        std::vector<std::future<bool>> reserves;
+        reserves.reserve(numberThreads);
+        std::size_t j{0};
+        for (std::size_t i = 0; i + 1 < numberThreads; ++i, j += stride) {
+            auto begin = m_Slices.begin() + j;
+            auto end = m_Slices.begin() + j + stride;
+            reserves.push_back(std::async(std::launch::async, reserver, begin, end));
```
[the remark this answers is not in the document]
tveasey: I think only once, at least for the foreseeable future. For all analyses I have planned at the moment we'd know exactly how many columns we need to reserve at the very outset.

lib/core/CDataFrame.cc
tveasey: I thought about this. I've used a couple of other C++14 features, for example capture with an initialiser so I can move variables into a lambda. However, at this point I suspect that we may want to back port this, so I wanted to keep non-portable features to a minimum and only where there was a real advantage.

I've been thinking in general it would be good to do a few more boost to std tidy ups in the code, particularly when we move to C++17 on the master. What I might do is raise an issue to cover this, but leave this as is now.

lib/core/CDataFrame.cc
```diff
@@ -0,0 +1,513 @@
...
+
+    std::vector<std::future<bool>> reads;
+    reads.reserve(numberThreads);
+    std::size_t j{0};
+    for (std::size_t i = 0; i + 1 < numberThreads; ++i, j += stride) {
+        auto begin = m_Slices.begin() + j;
+        auto end = m_Slices.begin() + j + stride;
+        CDataFrameRowSliceReader sliceReader{readers[i], m_NumberColumns,
+                                             m_RowCapacity, m_AsyncReadAndWriteToStore};
+        reads.push_back(std::async(std::launch::async, sliceReader, begin, end));
```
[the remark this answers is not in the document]
tveasey: By way of context, I expect there to be fairly few full reads, i.e. there will typically be a lot of work performed for each read. Although if there was a very low memory constraint and we were working on a large data frame then we would typically need 10-100s of reads for a full analysis (this would be a perverse configuration).

Reading around beforehand, Windows seems to have the best implementation of std::async at present (which it does back with a thread pool).

My rationale for not using a thread pool from the outset was:
1) Threading support is getting a lot of attention from the standards committee and I expect support to improve with new std library releases. So if we can get away with std::async at present, we'll get all these benefits for free going forwards.
2) I expect the work performed per read to be non-trivial and I don't have a good intuition for how this will stack up against the cost of any overhead of launching a thread.

That said I agree it is worth profiling and if thread creation proves to be a bottleneck *once we have a full analysis implemented* then we should think about using a thread pool library. Based on the above considerations though, I think I'd prefer to wait and see how the numbers stack up when this work is done. How about for the time being I leave some TODOs in the code to this effect?

lib/core/CDataFrame.cc
tveasey: I agree the C++11 thread support was rushed in and doesn't feel fully thought out. Also, that is a very interesting point about clean up.

The thing I have in the back of my mind is that there are lots of features coming in the near future to the C++ standard to improve thread support. For example, once executors arrive I think std::async is going to be the right model for threading a lot of our tasks related to data frames. I don't want to do work to bring in a different library if we can avoid it.

tveasey:
Thanks for the review @droberts195!

Given your comments, I'll create an issue to audit uses of std::async I've introduced. I'm coming round to thinking perhaps the best option will be to create some stop gap wrappers around an open source thread pool library which mimics what we can expect from the standard library in upcoming releases.

Since this is a currently on a feature branch, I think I'd prefer to delay any further work on this until we are in the position to test one analysis type end-to-end.

include/core/CDataFrame.h
```diff
@@ -0,0 +1,406 @@
...
+    //! The slice writer which is currently active.
+    TRowSliceWriterPtr m_Writer;
+};
+
+//! Make a data frame which uses main memory storage for its slices.
+//!
+//! \param[in] numberColumns The number of columns in the data frame created.
+//! \param[in] sliceCapacity If none null this overrides the default slice
+//! capacity in rows.
+//! \param[in] readWriteToStoreAsync Controls whether reads and writes from
```
hendrikmuhs: maybe `readWriteToStoreSyncStrategy` or `readWriteToStoreSyncOptions`? - right now it sounds like a bool.

PR #230, opened 2018-10-03
[ML] Add multi_bucket_impact label to anomalies

edsavage:
Add a label indicating the impact of the multi bucket analysis on the
overall probability. The value is in the range -5 to 5 where -5
indicates a wholly single bucket contribution and 5 a wholly multi
bucket contribution to the final probability.

### DIFF
--- a/lib/model/CProbabilityAndInfluenceCalculator.cc
+++ b/lib/model/CProbabilityAndInfluenceCalculator.cc
@@ -576,8 +576,9 @@ CProbabilityAndInfluenceCalculator::CProbabilityAndInfluenceCalculator(double cu
     : m_Cutoff(cutoff), m_InfluenceCalculator(nullptr),
       m_ProbabilityTemplate(CModelTools::CProbabilityAggregator::E_Min),
       m_Probability(CModelTools::CProbabilityAggregator::E_Min),
-	  m_ExplainingProbabilities{{SINGLE_BUCKET_FEATURE_LABEL, {CModelTools::CProbabilityAggregator::E_Min}},
-								{MULTI_BUCKET_FEATURE_LABEL, {CModelTools::CProbabilityAggregator::E_Min}}},
+      m_ExplainingProbabilities{
+          {SINGLE_BUCKET_FEATURE_LABEL, {CModelTools::CProbabilityAggregator::E_Min}},
+          {MULTI_BUCKET_FEATURE_LABEL, {CModelTools::CProbabilityAggregator::E_Min}}},
       m_ProbabilityCache(nullptr) {
 }
 
@@ -598,7 +599,7 @@ void CProbabilityAndInfluenceCalculator::addAggregator(
     m_ProbabilityTemplate.add(aggregator);
     m_Probability.add(aggregator);
     for (auto& ep : m_ExplainingProbabilities) {
-    	ep.second.add(aggregator);
+        ep.second.add(aggregator);
     }
 }
 
@@ -606,7 +607,7 @@ void CProbabilityAndInfluenceCalculator::addAggregator(const maths::CProbability
     m_ProbabilityTemplate.add(aggregator);
     m_Probability.add(aggregator);
     for (auto& ep : m_ExplainingProbabilities) {
-    	ep.second.add(aggregator);
+        ep.second.add(aggregator);
     }
 }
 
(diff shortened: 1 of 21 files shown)


lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -827,6 +859,42 @@ bool CProbabilityAndInfluenceCalculator::calculate(double& probability) const {
...
+        LOG_INFO(<< "Failed to compute explaining probabilities");
+        return false;
+    } else {
+        double sbProbability = explainingProbabilities[SINGLE_BUCKET_FEATURE_LABEL];
+        double mbProbability = explainingProbabilities[MULTI_BUCKET_FEATURE_LABEL];
+
+        double ls = std::log(std::max(sbProbability, ml::maths::MINUSCULE_PROBABILITY));
+        double lm = std::log(mbProbability);
+
+        double scale = 5.0 * std::min(ls, lm) /
```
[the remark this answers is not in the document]
tveasey: +1 and in various other files above. I'd propose adding something like MAXIMUM_MULTI_BUCKET_IMPACT CModelConfig.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -49,6 +49,9 @@ using TProbabilityCalculation2Vec = core::CSmallVector<maths_t::EProbabilityCalc
 using TSizeDoublePr = std::pair<std::size_t, double>;
 using TSizeDoublePr1Vec = core::CSmallVector<TSizeDoublePr, 1>;
 
+const std::string SINGLE_BUCKET_FEATURE_LABEL{"single_bucket"};
+const std::string MULTI_BUCKET_FEATURE_LABEL{"multi_bucket"};
```
tveasey: I think we'd be better off defining these once as static constants in `maths::CModel`. As it is these have to match the locally declared strings in `CTimeSeriesModel.cc`.

Better still would be to use an enum and update the relevant types to be defined w.r.t. this enum.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -612,6 +624,16 @@ void CProbabilityAndInfluenceCalculator::add(const CProbabilityAndInfluenceCalcu
     if (!other.m_Probability.empty()) {
         m_Probability.add(p, weight);
     }
+
+    for (const auto& ep : other.m_ExplainingProbabilities) {
+        if (ep.second.calculate(p) && !ep.second.empty()) {
+            auto itr = m_ExplainingProbabilities.find(ep.first);
```
tveasey: This should insert if missing, i.e.
```
std::tie(itr, added) = m_ExplainingProbabilities.insert(ep);
if (added == false) {
    itr->second.add(p, weight);
}
```

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -693,18 +715,34 @@ bool CProbabilityAndInfluenceCalculator::addProbability(model_t::EFeature featur
         return false;
     }
 
+    auto readResult = [&](const maths::SModelProbabilityResult& result) {
+        for (auto fp : result.s_FeatureProbabilities) {
```
tveasey: const auto&

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -827,6 +859,42 @@ bool CProbabilityAndInfluenceCalculator::calculate(double& probability) const {
     return m_Probability.calculate(probability);
 }
 
+bool CProbabilityAndInfluenceCalculator::calculateExplainingProbabilities(
+    TStrDoubleUMap& explainingProbabilities) const {
```
tveasey: It seems wasteful copy the strings around here. This would also be fixed if you migrated to using an enum.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -827,6 +859,42 @@ bool CProbabilityAndInfluenceCalculator::calculate(double& probability) const {
...
+    TStrDoubleUMap explainingProbabilities;
+    if (!this->calculateExplainingProbabilities(explainingProbabilities)) {
+        LOG_INFO(<< "Failed to compute explaining probabilities");
+        return false;
+    } else {
+        double sbProbability = explainingProbabilities[SINGLE_BUCKET_FEATURE_LABEL];
+        double mbProbability = explainingProbabilities[MULTI_BUCKET_FEATURE_LABEL];
+
+        double ls = std::log(std::max(sbProbability, ml::maths::MINUSCULE_PROBABILITY));
+        double lm = std::log(mbProbability);
```
tveasey: This should be bounded as well, i.e. `std::log(std::max(mbProbability, ...))`, in case of underflow. Also these should use `CTools::smallestProbability()`.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -827,6 +859,42 @@ bool CProbabilityAndInfluenceCalculator::calculate(double& probability) const {
...
+            return false;
+        } else {
+            explainingProbabilities.emplace(ep.first, probability);
+        }
+    }
+
+    return true;
+}
+
+bool CProbabilityAndInfluenceCalculator::calculateMultiBucketImpact(double& multiBucketImpact) const {
```
tveasey: I think we should add some explanation for this function, i.e. something along the lines of "we choose a function s.t. the impact saturates when one "probability" < "other probability" / 1000 or one probability is close to one, i.e. one factor is not at all anomalous".

PR #243, opened 2018-10-05
[ML] hard_limit memory error fix

edsavage:
Changed the calculation to decrease the byte limit margin to account
for bucket spans greater than 1 day

### DIFF
--- a/lib/api/unittest/CAnomalyJobLimitTest.cc
+++ b/lib/api/unittest/CAnomalyJobLimitTest.cc
@@ -344,147 +344,173 @@ void CAnomalyJobLimitTest::testModelledEntityCountForFixedMemoryLimit() {
     // create for a small(ish) memory limit to catch large changes in
     // the memory used per partition of the data.
 
-    core_t::TTime bucketLength{600};
-    TGeneratorVec generators{periodic, tradingDays, level, ramp, sparse};
-    std::stringstream outputStrm;
-    core::CJsonOutputStreamWrapper wrappedOutputStream(outputStrm);
-    api::CAnomalyJob::TStrStrUMap dataRows;
-    TSizeVec generator;
+    struct STestParams {
+        core_t::TTime s_BucketLength;
+        std::size_t s_ExpectedByFields;
+        std::size_t s_ExpectedOverFields;
+        std::size_t s_ExpectedPartitionFields;
+        std::size_t s_ExpectedByLowerMemoryLimit;
+        std::size_t s_ExpectedPartitionLowerMemoryLimit;
+        std::size_t s_ExpectedOverLowerMemoryLimit;
+    } testParams[] = {{600, 550, 6000, 300, 33, 40, 40},
+                      {3600, 550, 5800, 300, 30, 25, 20},
+                      {7200, 550, 5000, 300, 30, 26, 10},
+                      {172800, 150, 850, 120, 6, 6, 3},
+                      {604800, 150, 850, 120, 6, 6, 3}};
 
-    LOG_DEBUG(<< "**** Test by ****");
-    {
-        std::size_t memoryLimit{10 /*MB*/};
-        model::CLimits limits;
-        limits.resourceMonitor().memoryLimit(memoryLimit);
-        api::CFieldConfig fieldConfig;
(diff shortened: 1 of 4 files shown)


lib/model/CResourceMonitor.cc
```diff
@@ -311,9 +313,14 @@ void CResourceMonitor::decreaseMargin(core_t::TTime elapsedTime) {
     // will be the overwhelmingly common source of additional memory
     // so the model memory should be accurate (on average) in this
     // time frame.
-    double scale{1.0 - static_cast<double>(elapsedTime) /
-                           static_cast<double>(core::constants::DAY)};
-    m_ByteLimitMargin = 1.0 - scale * (1.0 - m_ByteLimitMargin);
+    if (elapsedTime <= core::constants::DAY) {
+        double scale{1.0 - static_cast<double>(elapsedTime) /
+                               static_cast<double>(core::constants::DAY)};
```
tveasey: Good catch!

I think a cleaner fix is just to lower bound `elapsedTime`, i.e. something like `static_cast<double>(std::min(elapsedTime, x)) / static_cast<double>(core::constants::DAY)`. This way the rate at which the margin is decreased is a smooth function of the bucket length. As it is now there is a very big discontinuity when the bucket length is changed from DAY to DAY + eps.

The I'd recommended choosing a value for `x` around 2 hours. For longer bucket lengths the common source of extra memory will be weekly periodicity. This takes longer to detect. We want the`m_ByteLimitMargin` to approach 1 on this time scale for long bucket lengths and this limit would achieves this.

Also, I think the comment could be clarified a bit. We'll aim for `m_ByteLimitMargin` * target memory. So whilst we do want to increase `m_ByteLimitMargin` by doing this we're effectively decreasing the safety margin we're applying to the process memory. The function name refers to this safety margin whilst the comment refers to `m_ByteLimitMargin`.

droberts195:
jenkins test this please

lib/model/CResourceMonitor.cc
```diff
@@ -24,6 +24,7 @@ namespace model {
 const core_t::TTime CResourceMonitor::MINIMUM_PRUNE_FREQUENCY(60 * 60);
 const std::size_t CResourceMonitor::DEFAULT_MEMORY_LIMIT_MB(4096);
 const double CResourceMonitor::DEFAULT_BYTE_LIMIT_MARGIN(0.7);
+const core_t::TTime CResourceMonitor::MAXIMUM_BYTE_LIMIT_MARGIN_PERIOD(172800); // 2 hours
```
tveasey: This is 2 days. I think this should actually be 2 hours no?

lib/api/unittest/CAnomalyJobLimitTest.cc
```diff
@@ -344,147 +344,173 @@ void CAnomalyJobLimitTest::testModelledEntityCountForFixedMemoryLimit() {
...
+        std::size_t s_ExpectedOverFields;
+        std::size_t s_ExpectedPartitionFields;
+        std::size_t s_ExpectedByLowerMemoryLimit;
+        std::size_t s_ExpectedPartitionLowerMemoryLimit;
+        std::size_t s_ExpectedOverLowerMemoryLimit;
+    } testParams[] = {{600, 550, 6000, 300, 33, 40, 40},
+                      {3600, 550, 5800, 300, 30, 25, 20},
+                      {7200, 550, 5000, 300, 30, 26, 10},
+                      {172800, 150, 850, 120, 6, 6, 3},
+                      {604800, 150, 850, 120, 6, 6, 3}};
```
tveasey: This is great, but I wonder how long this unit test runs for now? This was already pretty long, it may be we want to cut the cases, or only optionally run some of them, if the total runtime is very high.

lib/model/CResourceMonitor.cc
tveasey: Also, maybe use core/Constants.h `HOUR` variable.

lib/api/unittest/CAnomalyJobLimitTest.cc
edsavage: I've trimmed a few of the cases - runtime is now ~35s (was originally ~30s).

PR #249, opened 2018-10-10
[DO NOT MERGE] Test PR CI finds Windows build problem

droberts195:
Should fail on Windows only due to missing export

PR #250, opened 2018-10-11
[ML] Consolidate CI scripts

droberts195:
This change combines the `ci` and `jenkins-ci` scripts, as there's
no good reason for splitting the functionality.

The `ci.bat` script is converted to Powershell, and the exit code
of every invocation of external programs is checked so that
failures can immediately fail the overall CI run.

Additionally there are some style changes to the Powershell syntax
to match standard practice.


droberts195:
The refactoring of the Windows CI scripts in this PR also fix the problem identified in #249.  Following the changes in this PR, failures that occur in the Gradle invocation on Windows propagate back to the CI status reported by the PR on GitHub.

PR #260, opened 2018-10-17
[ML] Account for step discontinuities when reinitialising the residual model after a detecting a change or new decomposition components

tveasey:
This further improves the reinitialisation of the time series residual model after detecting either a change or news component of the decomposition for some edge cases. In particular, as a result of this change we remove any piecewise linear trend and better account for variable weights of the window values.

I've marked this as a non-issue since it is fall out from the change to use the decomposition window to reinitialise the residual model, i.e. #218.

### DIFF
--- a/include/maths/CTimeSeriesDecomposition.h
+++ b/include/maths/CTimeSeriesDecomposition.h
@@ -170,7 +170,7 @@ public:
     virtual bool mightAddComponents(core_t::TTime time) const;
 
     //! Get the values in a recent time window.
-    virtual TTimeDoublePrVec windowValues() const;
+    virtual TTimeFloatMeanAccumulatorPrVec windowValues() const;
 
     //! Roll time forwards by \p skipInterval.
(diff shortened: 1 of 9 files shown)


droberts195:
The unit tests don't compile after this change.  So far no unit test files have been altered for this PR.  Is it possible to add or change a test to validate the code change?

tveasey:
This is in a bit of a grey area between unit testing and integration testing IMO. It is in fact "covered" by a QA data set, which is what motivated the change. I can add a unit test using similar data characteristics and assert something about the overall modelling behaviour which would probably pick up degradations.

tveasey:
@droberts195 and @edsavage, [this](https://github.com/elastic/ml-cpp/pull/260/commits/e4ba9747a21ed3d3dc2a302df7fff82c5f584612) change fixes unit test fallout  and also adds a new check for model confidence after initialisation (which would have failed before this change).

PR #157, opened 2018-07-18
[ML] build boost test framework in buildenv

hendrikmuhs:
enable build of boost test framework in all provision steps and change documentation accordingly

Relates to #156

Issue #156, opened 2018-07-18
[ML] Switch unit testing framework from cppunit to boost


hendrikmuhs:
## Motivation

 - moving away from cppunit:
   - quite outdated and on low maintenance (but still used by e.g. libreoffice)
   - requires a lot of boiler plate code, easy to be forgotten
 - boost unit test framework
   - is well maintained
   - already part of the build env (almost, see [*])
   - is simpler to maintain:
       - better macros
       - automatic test case collection
        - much less boilerplate (no header files, no suite classes required)
        - => less error prone
   - is  more feature rich: more and better assertions
  
## Alternatives

Apart from staying with cppunit the most obvious alternative is `gtest`:

  - equally feature rich to boost, at least for the _core_ features we need
  - would be a new dependency

On the long run we might want to make use of mocking libraries which can be combined with both `boost` and `gtest`

## Con's

 [*] After investigation it turned out that we skip building the boost unit test libraries, therefore the build environment isn't ready yet. Rebuilding the images is substantial work.

## Tasks

 - [x] rebuild build environment images
 - [x] migrate tests to boost framework
 - [ ] ensure CI readiness: timings, reporting
 - [x] cleanup/remove cppunit

droberts195:
The reason this is labeled `v7.0.0` is that 6.x is using different compiler versions and we'd have to update the build machine images twice if we were to backport this to 6.x.  It saves some time to only update the build images for the current master branch environment.  However, if the changes to the test suites were made in master now then they would cause considerable backporting pain for the remaining development of 6.x.  Therefore it would be best to do the majority of the changes after feature freeze for 6.6 (ideally while master is still for 7.0 but even if it's after the 7.0 branch has been created the backport of the test framework changes will still be a trivial cherry-pick at that early stage).

droberts195:
We discussed this again in the EAH last week.

We decided that since 6.7 will be maintained for a long time we'd like to backport the changes to the 6.7 branch.  This implies that the best time to make the change would be between 6.6 feature freeze and 6.7 feature freeze.

This also implies that the Boost unit test library needs to be added to the Boost build for the 6.7 toolchain as well as later versions.

droberts195:
The first task is completed by #263 (6.x) and #264 (master)

droberts195:
cppunit uses `std::auto_ptr` internally, and that is removed in C++17.  So this refactoring will be a prerequisite for upgrading to C++17.

PR #263, opened 2018-10-18
[ML] Add Boost Test to build environment

droberts195:
Relates #156

This change includes changes similar to those made in #157,
but for the 6.x branch, and also incorporates rebuilds of
VM setup bundles and Docker image updates.

### DIFF
--- a/dev-tools/docker/build_linux_build_image.sh
+++ b/dev-tools/docker/build_linux_build_image.sh
@@ -16,13 +16,13 @@
 
 ACCOUNT=droberts195
 REPOSITORY=ml-linux-build
-VERSION=2
+VERSION=4
 
 set -e
 
 cd `dirname $0`
 
-docker build -t $ACCOUNT/$REPOSITORY:$VERSION linux_image
+docker build --no-cache -t $ACCOUNT/$REPOSITORY:$VERSION linux_image
 docker login
 docker push $ACCOUNT/$REPOSITORY:$VERSION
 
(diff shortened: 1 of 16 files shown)


dev-tools/docker/build_linux_build_image.sh
```diff
@@ -16,13 +16,13 @@
...
 REPOSITORY=ml-linux-build
-VERSION=2
+VERSION=4
 
 set -e
 
 cd `dirname $0`
 
-docker build -t $ACCOUNT/$REPOSITORY:$VERSION linux_image
+docker build --no-cache -t $ACCOUNT/$REPOSITORY:$VERSION linux_image
```
hendrikmuhs: nit: just out of curiosity, would be nice to add a comment why `--no-cache`

dev-tools/docker/build_linux_test_image.sh
```diff
@@ -16,7 +16,7 @@
 
 ACCOUNT=droberts195
 REPOSITORY=ml-linux-test
-VERSION=1
+VERSION=3
```
hendrikmuhs: If I get this right,  with -e you can pass environment variables down:

```
docker build .... -e VERSION
```

and use it in `dev-tools/docker/linux_tester/Dockerfile` instead of requiring changing it there in sync with this.

droberts195: I didn't realise you could use a variable in the `FROM` clause in a `Dockerfile`, but you're right that for about a year it's been possible: moby/moby#31352

I don't think I should change how everything fits together in this PR though.  Any changes to how the different images are linked should be separate.

PR #267, opened 2018-10-19
[ML] Change Mac compiler version in build setup docs

droberts195:
Newer macOS versions can use newer Xcode versions.


droberts195:
retest

PR #269, opened 2018-10-20
[DO NOT MERGE] Debugging uncommitted change found by lib/ver/Makefile

droberts195:
Causes version to say "based on"


droberts195:
The problem this PR was used to debug is fixed by #270

PR #270, opened 2018-10-22
[ML] Manually do what dissociate flag does in build script

dimitris-athanasiou:


dimitris-athanasiou:
jenkins please test this

dev-tools/jenkins_ci.sh
```diff
@@ -56,6 +56,10 @@ fi
 # Remove any old builds
 rm -rf ../builds
 
+# Disassociate from reference repo
+git repack -a -d
+rm -rf ../.git/objects/info/alternates
```
fatmcgav: @dimitris-athanasiou This might be a bit safer by doing something like:

```bash
readonly GIT_TOPLEVEL=$(git rev-parse --show-toplevel 2> /dev/null)

rm -f "${GIT_TOPLEVEL}/.git/objects/info/alternates"
```

dimitris-athanasiou:
jenkins please test this

dev-tools/jenkins_ci.sh
dimitris-athanasiou: Done! Thanks for the suggestion!

dimitris-athanasiou:
I have merged this in master so I can check it fixes the issue in question. I'll be back once I have confirmed (or not) the fix.

dimitris-athanasiou:
This has done the job! I'll now backport to `6.x` and `6.4`

dimitris-athanasiou:
OK, backporting done.

Issue #271, opened 2018-10-23
[ML] Error - Random Variable parameter was -nan, but must be > 0


dolaru:
Spotted in 6.5.0

The following error messages were found in ES logs while running the `gallery1` QA dataset:
```
[907][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [7DdayTu] [gallery1_20181023-1031_650_2275] [autodetect/30069] [CStatisticalTests.cc@99] Failed to compute significance Error in function boost::math::cdf(fisher_f_distribution const&, double): Random Variable parameter was -nan, but must be > 0 ! d1 = 335, d2 = 333, x = -nan
[910][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [7DdayTu] [gallery1_20181023-1031_650_2275] [autodetect/30069] [CStatisticalTests.cc@99] Failed to compute significance Error in function boost::math::cdf(fisher_f_distribution const&, double): Random Variable parameter was -nan, but must be > 0 ! d1 = 280, d2 = 278, x = -nan
[911][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [7DdayTu] [gallery1_20181023-1031_650_2275] [autodetect/30069] [CStatisticalTests.cc@99] Failed to compute significance Error in function boost::math::cdf(fisher_f_distribution const&, double): Random Variable parameter was -nan, but must be > 0 ! d1 = 271, d2 = 269, x = -nan
[913][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [7DdayTu] [gallery1_20181023-1031_650_2275] [autodetect/30069] [CStatisticalTests.cc@99] Failed to compute significance Error in function boost::math::cdf(fisher_f_distribution const&, double): Random Variable parameter was -nan, but must be > 0 ! d1 = 42, d2 = 40, x = -nan | repeated [4]
[914][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [7DdayTu] [gallery1_20181023-1031_650_2275] [autodetect/30069] [CStatisticalTests.cc@99] Failed to compute significance Error in function boost::math::cdf(fisher_f_distribution const&, double): Random Variable parameter was -nan, but must be > 0 ! d1 = 21, d2 = 19, x = -nan
[...]

PR #272, opened 2018-10-23
[ML] Do isnan check when computing significance

edsavage:
A quick-fix for #271 - Perform a check for `isnan` before attempting to compute significance.
A more in-depth analysis of the cause of #271 is required in due course

### DIFF
--- a/lib/maths/CStatisticalTests.cc
+++ b/lib/maths/CStatisticalTests.cc
@@ -74,6 +74,9 @@ double CStatisticalTests::leftTailFTest(double x, double d1, double d2) {
     if (std::isinf(x)) {
         return 1.0;
     }
+    if (std::isnan(x)) {
+        return 0.0;
+    }
     try {
         boost::math::fisher_f_distribution<> F(d1, d2);
         return boost::math::cdf(F, x);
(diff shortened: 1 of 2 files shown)


lib/maths/CStatisticalTests.cc
```diff
@@ -74,6 +74,9 @@ double CStatisticalTests::leftTailFTest(double x, double d1, double d2) {
     if (std::isinf(x)) {
         return 1.0;
     }
+    if (std::isnan(x)) {
+        return 0.0;
```
hendrikmuhs: I would expect this to return the same value as for `isinf(x)` -> `1.0`?

edsavage: Yes, it makes sense to have them on the same footing

lib/maths/CStatisticalTests.cc
```diff
@@ -91,6 +94,9 @@ double CStatisticalTests::rightTailFTest(double x, double d1, double d2) {
     if (std::isinf(x)) {
         return 0.0;
     }
+    if (std::isnan(x)) {
+        return 0.0;
```
tveasey: I think it would be safer to make this `1.0`. This is likely due to a 0/0 problem in the calling code. Returning `1.0` here means we shouldn't accept any decision using the F statistic.

edsavage: Good advice, thanks Tom

PR #287, opened 2018-10-30
[ML] Prevent trend detection during a change in a time series creating a poorly reinitialised model

tveasey:
Our QA suite showed up an issue where detecting a trend interferes with detecting a change. The resulting model is poorly reinitialised in this case which damages anomaly detection for some time.

This PR prevents us adding a trend component whilst change detection is occurring. If no change is subsequently detected we'll add a trend component as soon as change detection is abandoned, so this simply marginally delays when we first detect a trend.

Since this is covered by a regression test I haven't duplicated as a unit test.


droberts195:
Since we are close to the release of 6.5 I think it would be good to leave a day to double check that this only affects the results of the expected QA jobs before backporting.  Then if it has unexpected side effects we can reconsider whether to backport to 6.5.

PR #289, opened 2018-10-30
[ML][6.6] Switch Docker registry to docker.elastic.co

droberts195:
Previously the Docker images were hosted on Docker Hub,
i.e. registry-1.docker.io.  This change moves them to
Elastic's Docker registry at docker.elastic.co.

(I did not actually rebuild all the images using the
updated scripts.  I merely re-tagged the already-built
images and pushed them to the new registry.  But it's
good to update the image build scripts so that the next
change pushes directly to the correct registry.)

Backport of #288


droberts195:
retest

PR #298, opened 2018-11-01
[FEATURE][ML] Introduce data_frame_analyzer process

dimitris-athanasiou:
This is introducing the process that will perform analytics
on the data frames. This first commit contains the bare minimum
to allow records to be passed in.


bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+    ml::core::CProcessPriority::reducePriority();
+
+    if (ioMgr.initIo() == false) {
+        LOG_FATAL("Failed to initialise IO");
+        return EXIT_FAILURE;
+    }
+
+    typedef boost::scoped_ptr<ml::api::CInputParser> TScopedInputParserP;
+    TScopedInputParserP inputParser;
+    inputParser.reset(new ml::api::CLengthEncodedInputParser(ioMgr.inputStream()));
```
benwtrent: Two things I am confused about:

* Why the `typedef` ? Are you expecting to use this type again later or is this strictly for line length purposes?
* Why not use `std::unique_ptr`? I don't understand why `boost::scoped_ptr` is being used in this instance.

dimitris-athanasiou: All very good questions! This is copy-pasta in all our `Main.cc` :P. I'll look into your points and refactor if possible

droberts195: The code this was copied from is using `boost::scoped_ptr` because it predates move semantics. You can use `boost::make_unique` to make a `std::unique_ptr` and store the result in an `auto` variable.

benwtrent: So why wouldn't the following work? I suppose it could be considered harder to read, but it does not use any move semantics to my knowledge. 
```
const std::unique_ptr<ml::api::CInputParser> inputParser(new ml::api::CLengthEncodedInputParser(ioMgr.inputStream()));
```

tveasey: It would work, but there are [reasons](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3588.txt) to prefer make_unique.

tveasey: (It is planned to gradually phase out direct calls to new from the code base now we can have factory methods return unique_ptr.)

bin/Makefile
```diff
@@ -17,6 +17,7 @@ COMPONENTS= \
             categorize \
             controller \
             normalize \
+            data_frame_analyzer \
```
droberts195: nit: it would be nice to keep the list in alphabetical order

bin/controller/Main.cc
```diff
@@ -149,6 +149,7 @@ int main(int argc, char **argv)
     permittedProcessPaths.push_back("./autodetect");
     permittedProcessPaths.push_back("./categorize");
     permittedProcessPaths.push_back("./normalize");
+    permittedProcessPaths.push_back("./data_frame_analyzer");
```
droberts195: nit: it would be nice to keep the list in alphabetical order

bin/data_frame_analyzer/CCmdLineParser.cc
```diff
@@ -0,0 +1,87 @@
...
+                           std::string& configFile,
+                           std::string& logProperties,
+                           std::string& logPipe,
+                           std::string& inputFileName,
+                           bool& isInputFileNamedPipe,
+                           std::string& outputFileName,
+                           bool& isOutputFileNamedPipe) {
+    try {
+        boost::program_options::options_description desc(DESCRIPTION);
+        desc.add_options()("help", "Display this information and exit")(
```
droberts195: You can make this list of options look a lot nicer by putting `// clang-format off` at the beginning and `// clang-format on` at the end.  All the other `CCmdLineParser.cc` files do this.

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+
+    ml::core::CJsonOutputStreamWrapper wrappedOutputStream(ioMgr.outputStream());
+
+    ml::api::CDataFrameAnalyzer dataFrameAnalyzer;
+
+    if (inputParser->readStream(boost::bind(&ml::api::CDataFrameAnalyzer::handleRecord,
```
droberts195: I know this is another copy-and-paste, but the latest style guidelines say to prefer lambdas to `bind`.

include/api/CDataFrameAnalyzer.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#ifndef INCLUDED_ml_api_CDataFrameAnalyzer_h
+#define INCLUDED_ml_api_CDataFrameAnalyzer_h
+
+#include <core/CLogger.h>
```
droberts195: You don't need the logger in the header file - this line should be in the `.cc` file instead.

bin/data_frame_analyzer/CCmdLineParser.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#ifndef INCLUDED_ml_data_frame_analyzer_CCmdLineParser_h
+#define INCLUDED_ml_data_frame_analyzer_CCmdLineParser_h
+
+#include <core/CoreTypes.h>
```
droberts195: I don't think this is needed, as there's nothing in the `core_t` namespace.

dimitris-athanasiou:
I have pushed a commit that addresses all review points thus far.

bin/data_frame_analyzer/Main.cc
benwtrent: OK, cool, wanted to make sure I was not crazy :).

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+
+    ml::core::CProcessPriority::reducePriority();
+
+    if (ioMgr.initIo() == false) {
+        LOG_FATAL("Failed to initialise IO");
+        return EXIT_FAILURE;
+    }
+
+    auto inputParser =
+        std::make_unique<ml::api::CLengthEncodedInputParser>(ioMgr.inputStream());
```
droberts195: `std::make_unique` will break if this is backported to 6.7.  You could use `boost::make_unique` instead, which will work in both master and 6.x.  Or just change it for the backport if we decide to backport.

If you stick with `std::make_unique` then you should `#include <memory>`.

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+//! \brief
+//! Applies a range of ML analyses on a data frame.
+//!
+//! DESCRIPTION:\n
+//! Expects to be streamed CSV or length encoded data on STDIN or a named pipe,
```
droberts195: I agree it would be nice to have the option to accept CSV.  That would facilitate easier use from the command line, e.g. to find memory bugs with `valgrind`.  However, at the moment the program cannot accept CSV.

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#include <api/CDataFrameAnalyzer.h>
+#include <api/CIoManager.h>
+#include <api/CLengthEncodedInputParser.h>
+#include <api/COutputChainer.h>
+
+#include "CCmdLineParser.h"
+
+#include <boost/bind.hpp>
+#include <boost/scoped_ptr.hpp>
```
droberts195: These two headers are redundant now.

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+//! Expects to be streamed CSV or length encoded data on STDIN or a named pipe,
+//! and sends its JSON results to STDOUT or another named pipe.
+//!
+//! IMPLEMENTATION DECISIONS:\n
+//! Standalone program.
+//!
+#include <core/CJsonOutputStreamWrapper.h>
+#include <core/CLogger.h>
+#include <core/CProcessPriority.h>
+#include <core/CoreTypes.h>
```
droberts195: Nothing from `core_t` is used

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#include <ver/CBuildInfo.h>
+
+#include <api/CDataFrameAnalyzer.h>
+#include <api/CIoManager.h>
+#include <api/CLengthEncodedInputParser.h>
+#include <api/COutputChainer.h>
```
droberts195: I don't think we'll use the output chainer in this program

bin/data_frame_analyzer/Main.cc
```diff
@@ -0,0 +1,94 @@
...
+        return EXIT_FAILURE;
+    }
+
+    // Construct the IO manager before reconfiguring the logger, as it performs
+    // std::ios actions that only work before first use
+    ml::api::CIoManager ioMgr(inputFileName, isInputFileNamedPipe,
+                              outputFileName, isOutputFileNamedPipe);
+
+    if (ml::core::CLogger::instance().reconfigure(logPipe, logProperties) == false) {
+        LOG_FATAL("Could not reconfigure logging");
```
tveasey: We prefer leading << for all log messages even where not strictly needed for consistency. There are a few discrepancies in this file.

bin/data_frame_analyzer/Makefile
```diff
@@ -0,0 +1,29 @@
...
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+include $(CPP_SRC_HOME)/mk/defines.mk
+
+TARGET=$(OBJS_DIR)/data_frame_analyzer$(EXE_EXT)
+
+INSTALL_DIR=$(CPP_PLATFORM_HOME)/bin
+
+ML_LIBS=$(LIB_ML_CORE) $(LIB_ML_API)
```
tveasey: I suspect that this will create link issues on linux. You'll probably find you need to add in $(LIB_ML_MODEL) as a minimum.

include/api/CDataFrameAnalyzer.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <boost/unordered_map.hpp>
+
+#include <string>
+
+namespace ml {
+namespace api {
+
+class API_EXPORT CDataFrameAnalyzer {
+
```
tveasey: nit unnecessary new line

dimitris-athanasiou:
I have now added support for CSV input and I've addressed all additional comments.

PR #299, opened 2018-11-02
[FEATURE][ML] Linux and Windows build fixes

tveasey:
I've factored some build fixes out of #297 since these will also be useful for #298.

Note unit tests are failing. I'll fix those in a separate PR.


include/maths/CInformationCriteria.h
```diff
@@ -42,9 +42,11 @@ LOG_DETERMINANT(5);
 #undef LOG_DETERMINANT
 
 //! The log determinant of our internal heap symmetric matrix.
+MATHS_EXPORT
```
dimitris-athanasiou: Why do we need the export in individual functions?

tveasey: Because these are free functions. See [here](https://docs.microsoft.com/en-gb/cpp/cpp/dllexport-dllimport?view=vs-2017) for more info.

PR #300, opened 2018-11-02
[ML] Preserve test report if unit tests fail inside Docker

droberts195:
If the only thing that goes wrong with a Docker build is unit test
failures then we want to keep the cppunit_results.xml files that
store the details of the failure.  Prior to this change these
files were getting lost because the Docker container was discarded
on the first non-zero exit code.  Following this change the status
of the unit tests is stored in a status file and a non-zero exit
code of the unit tests suppressed within the Docker container.
Then the outer script checks the status file after copying out the
cppunit_results.xml files.

### DIFF
--- a/dev-tools/docker/docker_entrypoint.sh
+++ b/dev-tools/docker/docker_entrypoint.sh
@@ -25,8 +25,6 @@ cd "$MY_DIR/../.."
 # Build the code
 make -j`grep -c '^processor' /proc/cpuinfo`
 
-# Note: there is no testing so this is not suitable as a CI script!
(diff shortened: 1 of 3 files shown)


droberts195:
The build with the deliberate error shows a test report now: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+nix/79/

droberts195:
And the test report is still present when all tests succeed: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+nix/85/

PR #304, opened 2018-11-05
[FEATURE][ML] Fix failing unit tests

tveasey:
This reworks the outlier detection unit test given changes to functionality to score the outliers, i.e. we now test some points on the P-R curve. Also, the test rng was changed on master and this caused a perturbation in one of the test thresholds.


tveasey:
retest

PR #306, opened 2018-11-08
[FEATURE][ML] Infrastructure for reading a data frame analysis specification

tveasey:
This is groundwork for wiring in different analyses to the data_frame_analyzer command. In particular, this provides an initial schema for passing a JSON specification object to the process which configures the analysis, and also code to parse this and run the corresponding analysis. (Note the schema isn't set in stone, but is something to get the ball rolling so we can get the outlier detection endpoint working asap.)

All analyses will be launched in the background and it is anticipated that the main thread will be mostly idle and just used to monitor and send progress reports and errors as they are emitted.

I have left some stubs for running outlier detection on a data frame. I need to do a bit more work before this is ready to go and don't want this PR to grow any more.

### DIFF
--- a/include/maths/CLassoLogisticRegression.h
+++ b/include/maths/CLassoLogisticRegression.h
@@ -37,7 +37,7 @@ public:
 
 public:
     CLrDenseMatrix();
-    explicit CLrDenseMatrix(TDoubleVecVec& elements);
+    CLrDenseMatrix(TDoubleVecVec& elements);
 
     //! Efficiently swap the contents of two matrices.
     void swap(CLrDenseMatrix& other);


include/api/CDataFrameAnalysisRunner.h
```diff
@@ -0,0 +1,133 @@
...
+private:
+    const CDataFrameAnalysisSpecification& m_Spec;
+
+    bool m_Bad = false;
+    std::atomic_bool m_Finished;
+    std::atomic<double> m_FractionalProgress;
+    TStrVec m_Errors;
+
+    std::thread m_Runner;
+    static core::CFastMutex m_Mutex;
```
droberts195: If this is `static` then the name should be `ms_Mutex` (per the [style guide](https://github.com/elastic/ml-cpp/blob/master/STYLEGUIDE.md#naming-conventions)).

But does it need to be `static`?  It's only protecting non-`static` members of this class, so a separate lock per object would work.

include/api/CDataFrameAnalysisSpecification.h
```diff
@@ -0,0 +1,117 @@
...
+    //! {
+    //!   "rows": <integer>,
+    //!   "cols": <integer>,
+    //!   "memory_limit": <integer>,
+    //!   "threads": <integer>,
+    //!   "analysis": {
+    //!     "name": <string>,
+    //!     "parameters": <object>
+    //!   }
+    //! }
```
droberts195: We might want to rename some of these fields to match what will be used in the Java side configuration objects.  That will reduce confusion in the future.  But as far as this PR is concerned it can be merged with temporary names and they can be renamed in a future PR.

/cc @dimitris-athanasiou

include/api/CDataFrameAnalysisSpecification.h
```diff
@@ -0,0 +1,117 @@
...
+//! This manages extracting all configuration for a particular analysis from a JSON
+//! header which is passed to the data_frame_analyzer command before any data. This
+//! creates and owns an analysis runner object which is also configured by the header.
+//! The analysis is run asynchronously via the CDataFrameAnalysisSpecification::run
+//! method which returns a handle to the runner to retrieve progress and errors.
+class API_EXPORT CDataFrameAnalysisSpecification {
+public:
+    using TStrVec = std::vector<std::string>;
+    using TRunnerPtr = std::unique_ptr<CDataFrameAnalysisRunner>;
+    using TRunnerFactoryPtr = std::unique_ptr<CDataFrameAnalysisRunnerFactory>;
```
droberts195: There are many places in the code where the shortened form `UPtr` has been used for `unique_ptr`.  I'm not sure if this is consistent at the moment but it might be nice to standardise on `Ptr` for shared pointers and `UPtr` for unique pointers (unless we're so far from consistency at the moment that that would be futile).

include/api/CDataFrameAnalysisRunner.h
tveasey: Actually, you're right. This no longer needs to be static, this is a hang over from a refactor.

include/api/CDataFrameAnalysisSpecification.h
tveasey: I'll leave a TODO in the code. This is really a placeholder and we should definitely standardise naming.

include/api/CDataFrameAnalysisSpecification.h
tveasey: Agreed.

tveasey: [e6e7868](https://github.com/elastic/ml-cpp/pull/306/commits/e6e7868318e3b332fd9dcb92f786aed30255a02e)

include/api/CDataFrameAnalysisRunner.h
tveasey: [80d78e4](https://github.com/elastic/ml-cpp/pull/306/commits/80d78e45462740adaad5fc10318ff4074c388b18)

include/api/CDataFrameAnalysisSpecification.h
tveasey: [0839984](https://github.com/elastic/ml-cpp/pull/306/commits/0839984d906801c86b4079f03d89ad4d39ab15de)

PR #312, opened 2018-11-12
[ML] Deemphasizing X-Pack

droberts195:
ML is now a feature of the Elastic Stack

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-"# Machine Learning for the Elastic Stac2
+# Machine Learning for the Elastic Stack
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,16 +1,16 @@
-# Machine Learning for the Elastic Stack
+"# Machine Learning for the Elastic Stac2
```
benwtrent: `Stac2` ?

droberts195: Thanks.  Also there's the `"` at the beginning.  It's due to https://github.com/vim/vim/issues/2238.  Unless there's a workaround I might have to stop using `vim`!

Issue #315, opened 2018-11-13
[ML] Decisions relating to new analytics process input


droberts195:
We have a [stub](https://github.com/elastic/ml-cpp/pull/298/files#diff-740e94a98f6c2a36c617970df2bc2c81R14) for handling data in the new `data_frame_analyzer` analytics process.  Currently this is receiving input in the form of an unordered map of named fields.

We've said we want to have the option to use either our "length encoded" input parser (for production) or a CSV parser (for standalone testing and debugging).

We have also previously decided that we'd like the input parsers to directly populate the dataframe to avoid any unnecessary copying.  However, this raises a number of questions:

1. How will we handle control fields in the input stream sent from Java?
2. The output from the analytics process needs to know the field names, but the data frame doesn't store them.  How will we deal with this?
3. We have a requirement to map string inputs to numbers for the data frame, with each unique value of the string being mapped to a unique number, and then these numbers being mapped back to the original strings in the output.  Where will this mapping take place?

@dimitris-athanasiou and I talked this over earlier today and this is what we came up with.

Considering the first question, of how to handle control fields:

* If the input parser directly populated the data frame then the control field would have to be the first field on the line, because otherwise the input parser wouldn't realise that there was a control message until after it had already sent a row of blank values to the data frame.  Currently the Java side puts it last on the line, but this wouldn't be too hard to change.
* For testing it's nice if the input file doesn't _have_ to contain a control field.

The field names are not a big problem because the `CInputParser` already provides a [method to get them](https://github.com/elastic/ml-cpp/blob/e32b8890648ee4abf81e99809c61e745d3ec7ad4/include/api/CInputParser.h#L69).

For mapping the strings to numbers we'd previously assumed the `CStringStore` class would be a good way to do this.  However, it's not really, as it maps strings to strings wrapped in shared pointers that aren't guaranteed to be unique.  We should create a new string store class that uses a Boost multi-index with hash and ordered lookups to allow mapping from string to ordinal and back again.

[...]

PR #316, opened 2018-11-19
[FEATURE][ML] Add facility to read input as vectors instead of map

droberts195:
The original anomaly detection and categorization received input
as a map of field name to field value.  For the new data frame
analytics, it is desirable to preserve the order of fields in
the input, so this change makes it possible for the handler
function to instead take vectors of field names and field values
(in the same order as the raw input).

To make this change easier the XML input parser is removed, as
it is unlikely to ever be used.  Also, the term "lineified JSON"
has been replaced with the more standard ND-JSON.

Relates #315


droberts195:
> When using vectors: Will the field names be read from the first row? It would be nice to document this.

Where the field names come from is defined by the input format (which is one of CSV, length-encoded or ND-JSON), not the signature of the handler function.

There is already documentation about where the field names come from for [CSV](https://github.com/elastic/ml-cpp/blob/master/include/api/CCsvInputParser.h#L27) and [length-encoded](https://github.com/elastic/ml-cpp/blob/master/include/api/CLengthEncodedInputParser.h#L28).

For the ND-JSON parser things are more complex because it has the option to either parse each document fully or assume all documents have the same structure.  I clarified that in https://github.com/elastic/ml-cpp/pull/316/commits/0b74bcec20be2598fefbb438e63e5b3c7f83cc51

hendrikmuhs:
CSV and length-encoded is clear, I was talking about (ND)-json. But maybe I missunderstand the new feature. I was expecting that I now can feed something like:

```
["id", "count", "name"]
[1, 33, "bee"]
[2, 22, "flower"]
[3, 11, "honey"]
```

instead of:
```
{"id": 1, "count": 33, "name":"bee"}
{"id": 2, "count": 22, "name":"flower"}
{"id": 3, "count": 11, "name":"honey"}
```

hendrikmuhs:
ok, sorry, than I got it wrong. Would it be ok to rename `readStreamAs...` to `readStreamInto...`, I think that would make it clearer that it's about the output, not the input.

PR #317, opened 2018-11-20
[FEATURE][ML] Outliers for memory mapped vectors

tveasey:
This makes necessary changes to support outlier detection on memory mapped vector types. I need to separate code paths (because we need to use different types for projected vectors) and this leads to some refactoring of `CLocalOutlierFactors`.

I also migrate to passing vector collections by value to the top-level APIs. This is important because it means we can, as will be the case for data frame analysis, move these in to place in the calling code. The result is we never need to duplicate the collection, so halve our memory overhead.

Finally, as well as testing with memory mapped vectors, I've expanded the unit testing, in particular, adding some unit testing for the case we're (randomly) projecting vectors.

### DIFF
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -181,6 +181,7 @@ public:
     //! \note The \p points are moved into place.
     void build(TPointVec&& points) {
         this->build(points.begin(), points.end(), True());
+        points.clear();
     }
 
     //! Build from a pair of output random access iterators.
@@ -210,7 +211,7 @@ public:
             TCoordinatePrecise distanceToNearest{
                 std::numeric_limits<TCoordinatePrecise>::max()};
             return this->nearestNeighbour(point, m_Nodes[0],
-                                          0, // Split coordinate,
+                                          0, // Split coordinate
                                           nearest, distanceToNearest);
         }
         return nearest;
@@ -221,14 +222,16 @@ public:
         result.clear();
 
         if (n > 0 && n < m_Nodes.size()) {
-            TNearestAccumulator neighbours(n);
             const TCoordinatePrecise inf{
                 boost::numeric::bounds<TCoordinatePrecise>::highest()};
-            for (std::size_t i = 0u; i < n; ++i) {
-                neighbours.add({inf, las::zero(point)});
-            }
+
+            // These neighbour points will be completely replaced by the call
+            // to nearestNeighbours, but we need the list to be full so we get
+            // the correct value for the furthest nearest neighbour in the set
+            // "found to date" branch and bound search.
+            TNearestAccumulator neighbours(n, {inf, m_Nodes[0].s_Point});
             this->nearestNeighbours(point, m_Nodes[0],
-                                    0, // Split coordinate,
+                                    0, // Split coordinate
                                     neighbours);
 
             result.reserve(neighbours.count());
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -225,9 +225,9 @@ public:
                 boost::numeric::bounds<TCoordinatePrecise>::highest()};
 
             // These neighbour points will be completely replaced by the call
-            // to nearestNeighbours, but we need the list to be full so we get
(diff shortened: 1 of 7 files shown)


include/maths/CLocalOutlierFactors.h
```diff
@@ -94,12 +106,18 @@ class MATHS_EXPORT CLocalOutlierFactors {
...
     //! \param[out] scores The scores of \p points.
     template<typename POINT>
-    static void normalizedLdof(std::size_t k,
-                               bool project,
-                               const std::vector<POINT>& points,
-                               TDoubleVec& scores) {
-        CLdof<POINT, TKdTree<POINT>> ldof(k);
-        compute(ldof, project, points, scores);
+    static void
+    normalizedLdof(std::size_t k, bool project, std::vector<POINT> points, TDoubleVec& scores) {
```
dimitris-athanasiou: These all look the same except for the types that are in {CLof, CLdof, CDistancekNN, etc.} Good candidates for templating? Just a suggestion.

include/maths/CKdTree.h
```diff
@@ -181,6 +181,7 @@ class CKdTree {
     //! \note The \p points are moved into place.
     void build(TPointVec&& points) {
         this->build(points.begin(), points.end(), True());
+        points.clear();
```
hendrikmuhs: I think you uncovered a bug in the usage of rvalues, `points`is not really moved until you call `std::move()` on it before usage, points is still a reference. That's why you had to add `clear()` to actually free memory. In code:

```
void build(TPointVec&& points) {
        TPointVec movedPoints = std::move(points);
        this->build(movedPoints.begin(), movedPoints.end(), True());
}
```

should be the way to do it, keeping the existing interface, but I think the code would be easier to understand if you change points to a real reference `TPointVec&` and call `clear()` on it like in this fix.

Note: Maybe better to fix in a separate PR?

include/maths/CLocalOutlierFactors.h
```diff
@@ -51,8 +51,10 @@ class MATHS_EXPORT CLocalOutlierFactors {
     //! the number of nearest neighbours to use based on the data
     //! characteristics.
     template<typename POINT>
-    static void normalizedLof(const std::vector<POINT>& points, TDoubleVec& scores) {
-        normalizedLof(numberNeighbours(E_Lof, points), project(points), points, scores);
+    static void normalizedLof(std::vector<POINT> points, TDoubleVec& scores) {
```
hendrikmuhs: if you want to move points, shouldn't this be `std::vector<POINT>&& points` ?

include/maths/CKdTree.h
tveasey: Ok I agree, this is really superfluous to this PR and is aimed at transient memory usage in outlier detection code. I have a PR (upcoming) which addresses some other inefficiencies and I need to add memory instrumentation. I'll back out this one line and leave for later.

include/maths/CLocalOutlierFactors.h
tveasey: The reason is I want to call:
```
CLocalOutlierFactors::normalizedLof(std::move(mypoints), scores);
```
in code running outlier detection on some collection `mypoints` derived from a data frame, which I don't subsequently need to preserve in the calling context. This will move the container into this local variable (very nearly as efficient), which is later on moved into an annotated collection. However, there seems no reason to force that the calling code always to have to move.

include/maths/CKdTree.h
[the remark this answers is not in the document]
tveasey: @dimitris-athanasiou, this is referring explicitly to an omission in the standard so that, for example, iterators which are held for a container which is subsequently moved remain valid, i.e. this is saying that the following is well formed:
```
std::vector<int> foo{1, 2, 3};
auto i = foo.begin();
std::vector<b> bar(std::move(foo));
*i = 2;
std::cout << bar[0] << std::endl;
```
output: 2. One is perfectly at liberty to call functions on rvalue references.

Issue #302, opened 2018-11-02
`CStringStore` Unit tests write way too many logs


benwtrent:
CStringStore unit tests write thousands of logs like the following: 

```
2018-11-02 15:49:24,988 UTC [3922] ERROR CBucketGatherer.cc@451 Sample start time 166605000 is not bucket aligned
2018-11-02 15:49:24,988 UTC [3922] ERROR CBucketGatherer.cc@451 Sample start time 166605000 is not bucket aligned
2018-11-02 15:49:24,988 UTC [3922] ERROR CBucketGatherer.cc@451 Sample start time 166605000 is not bucket aligned
2018-11-02 15:49:24,988 UTC [3922] ERROR CBucketGatherer.cc@451 Sample start time 166615000 is not bucket aligned
2018-11-02 15:49:24,988 UTC [3922] ERROR CBucketGatherer.cc@451 Sample start time 166615000 is not bucket aligned
```

We need to either fix the test or mute these particular logs when being ran from a unittest.

benwtrent:
Found the path that is causing this:

In all of the tests a non-zero bucket delay is added: https://github.com/elastic/ml-cpp/blob/0328bdb64942271b17986f2755020ed676bce06a/lib/api/unittest/CStringStoreTest.cc#L131

This trips this conditional: https://github.com/elastic/ml-cpp/blob/0328bdb64942271b17986f2755020ed676bce06a/lib/model/CAnomalyDetector.cc#L370-L372

Which cuts the bucketLength in half. Then the following start/ends are off by half a bucket length and when:

https://github.com/elastic/ml-cpp/blob/0328bdb64942271b17986f2755020ed676bce06a/lib/model/CAnomalyDetector.cc#L388-L390 

https://github.com/elastic/ml-cpp/blob/0328bdb64942271b17986f2755020ed676bce06a/lib/model/CAnomalyDetector.cc#L398-L400

is called, certain models validate the sample times

e.g `CCountingModel::sample`, 

https://github.com/elastic/ml-cpp/blob/4dd90fa93338667b681364657222715f81c9868a/lib/model/CCountingModel.cc#L210-L219

Which calls:

https://github.com/elastic/ml-cpp/blob/0328bdb64942271b17986f2755020ed676bce06a/lib/model/CBucketGatherer.cc#L450-L453

I can make this stop by having CStringStoreTest not set the bucketDelay and let it use the default of 0. 

What say y'all @tveasey @droberts195 ?

[...]

PR #318, opened 2018-11-20
[ML] Remove out-of-phase buckets feature

dimitris-athanasiou:
This feature was never fully completed and in fact
we no longer need it as the multibucket feature
covers the benefits from supporting out-of-phase buckets.

closes https://github.com/elastic/ml-cpp/issues/302


include/api/CAnomalyJob.h
```diff
@@ -224,14 +216,11 @@ class API_EXPORT CAnomalyJob : public CDataProcessor {
     void outputInterimResults(core_t::TTime bucketStartTime);
 
     //! Helper function for outputResults.
-    //! \p processingTimer is the processing time can be written to the bucket
-    //! \p sumPastProcessingTime is the total time previously spent processing
-    //! but resulted in no bucket being outputted.
+    //! \p processingTime is the processing time can be written to the bucket
```
tveasey: is the processing time can be written to the bucket |-> is the processing time of the bucket

lib/model/CDataGatherer.cc
```diff
@@ -877,43 +818,39 @@ bool CDataGatherer::restoreBucketGatherer(const std::string& summaryCountFieldNa
...
-            inserter.insertLevel(tag, boost::bind(&CEventRateBucketGatherer::acceptPersistInserter,
-                                                  boost::cref(gatherer_), _1));
-        } else if (tag == CBucketGatherer::METRIC_BUCKET_GATHERER_TAG) {
-            const CMetricBucketGatherer* gatherer_ =
-                dynamic_cast<const CMetricBucketGatherer*>(gatherer.get());
-            inserter.insertLevel(tag, boost::bind(&CMetricBucketGatherer::acceptPersistInserter,
-                                                  boost::cref(gatherer_), _1));
-        }
+    const std::string& tag = m_BucketGatherer->persistenceTag();
+    if (tag == CBucketGatherer::EVENTRATE_BUCKET_GATHERER_TAG) {
```
tveasey: This probably doesn't need to cast and use separate code paths since `bind` behaves polymorphically, i.e.
I think
```
inserter.insertLevel(m_BucketGatherer->persistenceTag(),
                     boost::bind(&CBucketGatherer::acceptPersistInserter, m_BucketGatherer.get(), _1));
```
works. Persist and restore unit tests will fail if it doesn't so you can try it out safely.

dimitris-athanasiou: Done!

PR #321, opened 2018-11-27
[FEATURE][ML] Infrastructure for parallelising data frame analyses

tveasey:
I've broken this down into two commits:
1. A [change](https://github.com/elastic/ml-cpp/commit/854dc7f9bbb82e73527b5d8823a11af732e25e5f) to introduce a new threading primitives library, [transwarp](https://github.com/bloomen/transwarp), which supersedes our old thread_pool library. (We reviewed a number of alternative open source libraries, our previous thread pool predated C++11. I won't repeat the conclusions here in detail; however, this one is relatively lightweight, header only, well tested and anticipates various aspects of the upcoming executors standard library extensions which we will migrate to when it becomes available.)
2. A [change](https://github.com/elastic/ml-cpp/commit/fa11b7431cd8033736d2573e1e47cb5e1c56c110) to add some higher level threading primitives (e.g. parallel_for_each), to migrate the data frame to use a thread pool and to parallelise outlier detection.

The design is based on a singleton thread pool which is started when the analysis specification is read and is used as the executor for all parallelised code. In the end, I decided that this is preferable to passing around thread pool references. This will fit our requirements for data frame analysis well, we will start processes for specific analyses and shut them down when finished. It also means that a lot of code can more or less ignore threading and we can work in terms of primitives which make use of the global thread pool. Finally, we get the advantage that there is very little overhead with minimal special code if the process is run single threaded, since the default executor is simply set to be "sequential in the same thread" in this case.

For example, I've been able to effectively thread outlier detection using only parallel_for_each with fairly minimal code changes. This implementation is close to final, i.e. for large(ish) point sets, I'm getting close to perfect speedup: "run time serial" / "number threads". The following are measured runtimes in ms for 100000 inliers 1000 outliers 6d points for one to six threads using an ensemble of all four measures:
outliers_threading

(The gap is not due to contention, since the code is very nearly lock free. It appears to be related to worse cache performance in the k-d tree due to concurrent look ups with different points.)

Note, for standard anticipating functionality I've used snake_case.


3rd_party/include/transwarp/transwarp.h
```diff
@@ -0,0 +1,3064 @@
+/// @mainpage transwarp is a header-only C++ library for task concurrency
```
droberts195: 3 files need adding for this in https://github.com/elastic/ml-cpp/tree/master/3rd_party/licenses:

1. transwarp-INFO.csv (with the following columns: name,version,revision,url,license,copyright)
2. transwarp-NOTICE.txt (zero length file)
3. transwarp-LICENSE.txt (plain text copy of https://github.com/bloomen/transwarp/blob/master/LICENSE.md)

lib/core/Concurrency.cc
```diff
@@ -0,0 +1,65 @@
...
+namespace core {
+namespace {
+std::size_t THREAD_POOL_SIZE{0};
+
+std::unique_ptr<executor> makeThreadPool(std::size_t threads) {
+    if (threads == 0) {
+        threads = std::thread::hardware_concurrency();
+    }
+
+    THREAD_POOL_SIZE = threads;
```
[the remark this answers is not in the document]
tveasey: I agree, this is definitely cleaner.

include/maths/CLocalOutlierFactors.h
```diff
@@ -502,6 +520,7 @@ class MATHS_EXPORT CLocalOutlierFactors {
             this->run(m_K, std::move(points), scores);
         }
 
+        //! \warning This must be thread safe.
```
hendrikmuhs: I wonder: for this implementation of `add` and others all over the place in this change, `add` could be _const_, that would make the thread safeness a bit clearer.

(I know you can always construct a case where this is not true or vice versa)

I am not familiar with the code but maybe: is there a chance of making `add` _const_ in the base class and or have a _const_ version of `CMethod`?

lib/core/Concurrency.cc
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        if (threadPoolSize > 0) {
+            try {
+                return CExecutorHolder{threadPoolSize};
+            } catch (const std::exception& e) {
+                LOG_ERROR(<< "Failed to create thread pool: " << e.what());
+            }
+        } else {
+            LOG_WARN(<< "Failed to determine hardware concurrency");
```
hendrikmuhs: the choice of log level is not clear to me, why isn't it error in both cases?

I would also add the consequence in the log message: "Failed to ... , fallback to ..."

lib/core/Concurrency.cc
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        // We'll just have to execute in the main thread.
```
hendrikmuhs: do we really want that? Doesn't that break the whole architecture and causes problems?

Maybe failing instead is the better option, otherwise we need to support this case, which also means we have to test it, etc.

lib/core/Concurrency.cc
```diff
@@ -0,0 +1,78 @@
...
+private:
+    std::size_t m_ThreadPoolSize;
+    std::unique_ptr<core::executor> m_Executor;
+};
+
+CExecutorHolder singletonExecutor;
+}
+
+void startDefaultAsyncExecutor(std::size_t threadPoolSize) {
+    singletonExecutor = CExecutorHolder::makeThreadPool(threadPoolSize);
```
hendrikmuhs: looks unsafe to me, what if this is called more than once and worse, while it the old executor is in use.

droberts195: The assignment will destroy the old executor, which will wait for it to finish processing.

I agree the method is not thread safe though (nor is the corresponding `stop`).  If the work of these two methods were locked then they would be.

I think there's an assumption these methods will only be used in `main()` and unit tests.  It's probably worth documenting this.

include/maths/CLocalOutlierFactors.h
tveasey: The reason that this code is thread safe is slightly subtle and this comment was not really helpful on reflection. Basically, different threads only read/write distinct elements in pre-sized vectors. I'm going to comment this better in the code which calls the add functions.

lib/core/Concurrency.cc
tveasey: I originally had them as both warn (because of the fallback) and I think I agree they probably should be the same. ++ on improving the message.

lib/core/Concurrency.cc
tveasey: I mean we could. The thing is there is nothing to stop someone telling the process to run single threaded because of hardware constraints, so we at least need to deal with this case anyway.

Regarding testing, if the program has a bug for a sequential executor it would also have it with a parallel executor, because any sequential run of the program could happen (by chance) when run in parallel. I guess that is the advantage of the executor model: it means we have very little code which differs between the two execution models.

lib/core/Concurrency.cc
tveasey: It is commented as such in the header. I thought about maybe making this call_once, but it makes writing test code hard.

One definitely shouldn't be calling this throughout a running data frame analysis and I think locking this makes the intention ambiguous.

Maybe production code could use call_once and we have "test only" functions to start and stop.

droberts195: I'm happy to leave these as not thread safe and just have comments to say only use in `main()` and unit tests.

Even though it's documented in the header it might be worth adding a comment in the `.cc` file to say something like "not thread safe - see header for reasons" just to avoid any danger of someone deciding to make them thread safe in the future.

tveasey:
Thanks for the feedback so far.

Unfortunately, transwarp crashes the Windows compiler we're using. I've tried paring the header back to the bare minimum functionality we use, but it still crashes. I don't really want this change to force a compiler upgrade, so I'm going to think about alternatives. I'll close this PR off and open a new one when I've resolved this issue.

I'll incorporate existing feedback in the new PR.

lib/core/Concurrency.cc
hendrikmuhs: what I meant with unsafe: 
```
startDefaultAsyncExecutor(10);
e_ref = executor(); // gets a reference to the executor
e_ref.doSomething() // ok, this is safe
startDefaultAsyncExecutor(10); // invalidates e, but waits for the e to finish
e_ref.doSomething() // the ref is invalid, therefore this is doomed!
```

I thought of an atomic bool as guard in ExecutorHolder, on the other side, this doesn't solve the problem completely. The only way to make it bullet proof would be to proxy the executor and get rid of `executor()` - which makes it super complex. "test only" sounds good to me.

PR #322, opened 2018-11-29
[FEATURE][ML] Infrastructure for parallelising data frame analyses

tveasey:
This is a second attempt at #321. See that PR for additional background.

After trying and failing to get transwarp to compile successfully with the Windows compiler we're using and investigating a couple of other alternatives, I decided in the end it was easier to just implement our own cutdown primitives for a thread pool and executors. These were all I was actually using from that library in the first instance.

Note this thread pool significantly outperforms our old implementation (for example CJsonOutputStreamWrapperTest::testConcurrentWrite drops from around 450ms to around 160ms). I think this is mainly due to a better strategy for avoiding contention between worker threads, although it also moves objects around.

(@droberts195 and @hendrikmuhs I believe I captured all your comments from the last PR.)


lib/core/unittest/CConcurrentWrapperTest.cc
```diff
@@ -9,13 +9,14 @@
...
 #include <core/CMemoryUsage.h>
+#include <core/CStaticThreadPool.h>
 
 #include <boost/bind.hpp>
 #include <boost/ref.hpp>
-#include <boost/threadpool.hpp>
 
 #include <algorithm>
 #include <chrono>
+#include <iomanip>
```
droberts195: Is this needed?

lib/core/unittest/CConcurrencyTest.h
```diff
@@ -3,16 +3,18 @@
...
 public:
-    void testPool();
+    void testAsyncWithExecutors();
+    void testParallelForEach();
 
     static CppUnit::Test* suite();
 };
 
-#endif // INCLUDED_CThreadPoolTest_h
+#endif // INCLUDED_CDataFrameTest_h
```
droberts195: `INCLUDED_CConcurrencyTest_h`

bin/data_frame_analyzer/Main.cc
```diff
@@ -101,6 +103,14 @@ int main(int argc, char** argv) {
     // TODO Actually use the specification to create the data frame and run the analysis
     ml::api::CDataFrameAnalysisSpecification dataFrameAnalysisSpecification{
         makeDataFrameAnalysisSpecification(configFile)};
+    if (dataFrameAnalysisSpecification.bad()) {
+        LOG_FATAL("Failed to parse analysis specification");
+        return EXIT_FAILURE;
+    }
+    if (dataFrameAnalysisSpecification.threads() > 1) {
+        ml::core::startDefaultAsyncExecutor(dataFrameAnalysisSpecification.threads() - 1);
```
droberts195: I'm not sure about the `- 1` here.  I get that it's for the main thread, but if we expect _either_ the main thread to be active _or_ the thread pool threads then it's not against the spirit of the `threads` argument for the thread pool to use as many threads as were specified.  Or am I missing something?

tveasey: The analyses are always launched in the background so the main thread stays live for control messages, monitoring, etc. This is managed by `CDataFrameAnalysisRunner`.

The rationale is if we've been told to use one thread (I don't count the mainly idle main thread) we might as well execute everything sequentially, since we don't then have any overhead of passing values or blocking for thread communication, i.e. see `CImmediateExecutor`.

droberts195: Makes sense.  Thanks.

droberts195: Actually, that makes sense for the `> 1` on line 110, but not the `- 1` on line 111.  If we're told to use 2 threads then as it stands we'll create a thread pool with 1 thread.

lib/core/CStaticThreadPool.cc
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void CStaticThreadPool::worker(std::size_t id) {
+
+    auto noThrowExecute = [](TOptionalTask& task) {
+        try {
+            (*task)();
+        } catch (const std::future_error& e) {
+            LOG_ERROR(<< "Failed executing packaged task: " << e.code() << " "
+                      << e.what());
```
hendrikmuhs: if you get a task failure like this: what does it mean for the overall process? Except from the logline, I see no error being populated back. Maybe `schedule(...)` could take a callback parameter which will be executed in an error situation?

lib/core/unittest/CConcurrencyTest.cc
```diff
@@ -0,0 +1,127 @@
...
+        result = core::async(core::defaultAsyncExecutor(),
+                             [](int i, int j) { return i + j; }, 22, 22);
+        CPPUNIT_ASSERT_EQUAL(44, result.get());
+
+        core::startDefaultAsyncExecutor();
+    }
+
+    core::stopDefaultAsyncExecutor();
+}
+
```
hendrikmuhs: would be nice to have a test case that shows an error situation, a task that simply throws. I also wonder what `result` should provide in this case, e.g. there could be `result.success()` returning `true` or `false` and `result.error()` which gives back the error message.

lib/core/unittest/CConcurrencyTest.cc
```diff
@@ -0,0 +1,127 @@
...
+void CConcurrencyTest::testParallelForEach() {
+
+    core::stopDefaultAsyncExecutor();
+
+    std::vector<int> values(10000);
+
+    LOG_DEBUG(<< "Test empty");
```
hendrikmuhs: nit: could be split into separate test cases?

lib/core/unittest/CConcurrencyTest.cc
tveasey: Yes I was reflecting that I should also add a thread pool test, because even though it is being tested indirectly in quite a few places I also want to check the balancing fallbacks work properly  with very uneven task durations. I'll beef up the testing a bit.

lib/core/CStaticThreadPool.cc
tveasey: Basically, I wanted to avoid this causing the worker thread to exit. I mean this really shouldn't happen, I was being paranoid because the tasks are passed in. I think in both cases it throws: task already executed (the thread pool really should ensure this doesn't happen if being used in a sane way) and not initialised the right behaviour is to just ignore, but felt uncomfortable silently swallowing an exception.

This goes to our discussions earlier about overall monitoring strategy, because I think it is unlikely the calling code can act on this failure, it really means that there was probably a logic failure somewhere in the program. Having a variable which has to be dealt with in every calling context creates a lot of debt.

lib/core/unittest/CConcurrentWrapperTest.cc
tveasey: It's needed for e.g. [this](https://github.com/elastic/ml-cpp/blob/master/lib/core/unittest/CConcurrentWrapperTest.cc#L58). I think it must have been coming in via the old thread pool header in the past.

lib/core/unittest/CConcurrencyTest.cc
tveasey: I added [testing](https://github.com/elastic/ml-cpp/pull/322/commits/780ebb0177036c0ba47990fc9cf44283ad7e8406) along the lines you suggest. Interestingly, it shows up an undesirable feature of futures which is there appears to be no way to tell if an exception has been caught with std::future<void> because you can't call get and this is the only method which propagates this information. I've tweaked the parallel_for_each functions accordingly. (Of course we shouldn't be throwing and catching exceptions locally, but this tidies up a loose end.)

lib/core/unittest/CConcurrencyTest.cc
tveasey: [780ebb0](https://github.com/elastic/ml-cpp/pull/322/commits/780ebb0177036c0ba47990fc9cf44283ad7e8406).

tveasey:
@droberts195 and @hendrikmuhs I've now been through all the review comments. I also pushed some direct tests of the thread pool. Can you take another look? Thanks.

lib/core/Concurrency.cc
```diff
@@ -0,0 +1,139 @@
...
+
+private:
+    CStaticThreadPool m_ThreadPool;
+};
+
+class CExecutorHolder {
+public:
+    CExecutorHolder()
+        : m_ThreadPoolSize{0},
+    //    Work around cross compile for OS X from Linux
```
droberts195: I think this comment should be changed, because it's not the cross compilation that causes the problem but the fact that build platform for 7.x on macOS will be El Capitan, and the latest version of Xcode that works on El Capitan is 8.2.  Anyone still using Xcode 8.2 on their laptop would see the same problem.  So I'd prefer the comment to be "Workaround for Xcode 8.2 C++ headers".

Same on line 78.

tveasey: [2ec6fe8](https://github.com/elastic/ml-cpp/pull/322/commits/2ec6fe8e1b708edb10cedfd17ce47985b55f685e)

PR #323, opened 2018-11-30
[FEATURE][ML] Wire in creation of data frame and control messages to data_frame_analyzer

tveasey:
This adds support for control messages, creates a data frame and reads data from the input pipe into it in the data_frame_analyzer command.

The remaining work to run the analysis, and right results back to the output stream will be in a follow up PR.


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -6,15 +6,133 @@
...
+
+bool CDataFrameAnalyzer::handleControlMessage(const TStrVec& fieldValues) {
+    bool unrecognised{fieldValues.size() > 1};
+    if (unrecognised == false) {
+        switch (fieldValues[m_ControlFieldValue][0]) {
+        case RUN_ANALYSIS_CONTROL_MESSAGE_FIELD_VALUE:
+            this->run();
+            break;
+        default:
+            unrecognised = true;
```
droberts195: It should also recognise a value of `.` and silently ignore the row.  (This allows for resending the field names.)

And it should also recognise a value of ` ` (space) and silently ignore the row.  This allows for flushing the input buffers.

To allow for code reuse on the Java side the logic for these should match https://github.com/elastic/ml-cpp/blob/master/lib/api/CAnomalyJob.cc#L314-L323

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -6,15 +6,133 @@
...
+        }
```
droberts195: I think it's nice to have a `break` for the `default` case too, even though it won't change the control flow in this instance.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -6,15 +6,133 @@
...
     return true;
 }
+
+bool CDataFrameAnalyzer::isControlMessage(const TStrVec& fieldValues) const {
+    return m_ControlFieldValue != CONTROL_FIELD_MISSING &&
+           fieldValues[m_ControlFieldValue].size() > 0;
+}
+
+bool CDataFrameAnalyzer::handleControlMessage(const TStrVec& fieldValues) {
+    bool unrecognised{fieldValues.size() > 1};
```
droberts195: The Java _will_ send empty values for the non-control fields, so I think this should be simply `bool unrecognised{false};`

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -6,15 +6,133 @@
...
+        if (pos != fieldNames.end()) {
+            LOG_ERROR(<< "Unexpected possible control field: ignoring");
+            return false;
+        }
+    }
     return true;
 }
+
+bool CDataFrameAnalyzer::isControlMessage(const TStrVec& fieldValues) const {
+    return m_ControlFieldValue != CONTROL_FIELD_MISSING &&
```
dimitris-athanasiou: Shouldn't this be checking `m_ControlFieldValue >= 0 && fieldValues[m_ControlFieldValue].size() > 0`?

lib/api/CDataFrameAnalyzer.cc
tveasey: This was a typo, it should have been fieldValue[0].size() > 0, because we only read the first character and I wanted some diagnostics if someone introduced a string.

lib/api/CDataFrameAnalyzer.cc
tveasey: Good point on flush.

Do we really want to allow for resend of field names via a stop? I can't think of a reason for needing to do this.

droberts195: If we ever keep the C++ process alive for further analysis then it means the code on the Java side that prepares the input doesn't need to know whether it's sending the very first batch to the process or a subsequent batch.  I agree it's not a requirement to add it now though - it can be added later if needed.

tveasey: I think in the context of sending data for multiple analyses the model will be a little different than for autodetect. What will likely happen will be that we'd send all the data, but potentially send subsequent requests to perform specific analyses. I think I'll leave this part for now, as there is no problem introducing later if we change our minds. I addressed the flush issue in [3291b12](https://github.com/elastic/ml-cpp/pull/323/commits/3291b12a9127e463bb43cd48d11c71066e738e41).

lib/api/CDataFrameAnalyzer.cc
droberts195: It looks like this has been changed to `m_ControlFieldValue >= CONTROL_FIELD_MISSING`, which could lead to a SEGV.  It either needs to be `>= 0` or `!= CONTROL_FIELD_MISSING && != CONTROL_FIELD_UNSET`.

Also, it might be worth renaming `m_ControlFieldValue` to `m_ControlFieldIndex`, as "value" is easy to confuse with the control message itself.

lib/api/CDataFrameAnalyzer.cc
tveasey: I should add, there doesn't seem to be a large advantage in keeping a process running for analyses on different data frames to me at present, but we can think about pros and cons offline.

PR #330, opened 2018-12-06
Revert "[ML] Correct query times for model plot and forecast"

droberts195:
Reverts #329

The change that was made in #329 changes the rule that all ML results had timestamps that were the start time of the bucket they relate to.  I don't think this change should be made in a patch.  Further investigation of the side effects is required.


droberts195:
Marked as `>non-issue` to please the changelog-status check.

PR #331, opened 2018-12-06
[FEATURE][ML] Prepare data frame analytics results writing

dimitris-athanasiou:
This fixes a couple of issues and adds writing of fake results
which allows us to test the java side before wiring in outlier
detection.

### DIFF
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -7,7 +7,9 @@
 
 #include <core/CContainerPrinter.h>
 #include <core/CFloatStorage.h>
+#include <core/CJsonOutputStreamWrapper.h>
 #include <core/CLogger.h>
+#include <core/CRapidJsonConcurrentLineWriter.h>
 
 #include <maths/CTools.h>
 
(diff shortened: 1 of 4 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -61,6 +65,20 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec&
 
 void CDataFrameAnalyzer::run() {
     // TODO
+    auto outStream = m_OutStreamSupplier();
+    core::CRapidJsonConcurrentLineWriter outputWriter{*outStream};
+    for (std::size_t i = 0; i < m_AnalysisSpecification.rows(); ++i) {
+        outputWriter.StartObject();
+        outputWriter.String("id");
```
droberts195: It would be nice to define `"id"`, `"results"` and `"outlier_score"` as `std::string` constants.  This would also be more efficient because as it stands _every iteration of the loop_ is constructing temporary `std::string` objects from the `const char*`s to pass to `outputWriter.String()`.

lib/core/CJsonOutputStreamWrapper.cc
```diff
@@ -27,7 +27,7 @@ CJsonOutputStreamWrapper::CJsonOutputStreamWrapper(std::ostream& outStream)
 }
 
 CJsonOutputStreamWrapper::~CJsonOutputStreamWrapper() {
-    m_ConcurrentOutputStream([](std::ostream& o) { o.put(JSON_ARRAY_END); });
+    m_ConcurrentOutputStream([](std::ostream& o) { o.put(JSON_ARRAY_END); o.flush(); });
```
edsavage: nit: we now have multiple statements on one line..

dimitris-athanasiou:
I've addressed review feedback. Note that the bit that writes results is throw-away code. But I extracted the strings to constants for now anyway.

PR #332, opened 2018-12-07
[ML] Write out forecasts predictions at "bucket time" rather than the actual times they are made in the time buckets

tveasey:
The Java notion of "bucket time" is the start of a job's time buckets, i.e. if the job uses 30 minute time buckets then the "bucket time" is the start of each 30 minute time interval (from 00:00:00 in UTC). Currently, all anomaly detection results are written at bucket time for the buckets which contain them.

As a result of #327, we started evaluating forecasts at the sample offset in the time bucket (for good reasons). In addition, we reported their times as at the evaluation time rather than "bucket time" for the bucket which contains them. This change reverts to report them at bucket time.

Going forward we will probably want to revisit this behaviour and decouple the notion of bucket time from forecast documents. They would simply hold a time and the forecast at that time, *whatever* time that may be. However, this requires knock on changes to our UI, which we don't want to make now, and makes more sense in the context of planned changes to the forecast API.

I've marked this as a non-issue, as it is a change to revert behaviour (introduced in #327 which is not yet released).

### DIFF
--- a/lib/model/CForecastDataSink.cc
+++ b/lib/model/CForecastDataSink.cc
@@ -9,6 +9,8 @@
 #include <core/CLogger.h>
 #include <core/CScopedRapidJsonPoolAllocator.h>
 
+#include <maths/CIntegerTools.h>
+
 #include <boost/bind.hpp>
 
 #include <vector>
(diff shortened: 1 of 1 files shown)


tveasey:
Thanks. Will do.

PR #334, opened 2018-12-10
[FEATURE][ML] Wire outlier detection into the data_frame_analyzer command

tveasey:
This finishes up the implementation of `CDataFrameAnalyzer`: wiring in outlier detection and writing results to the output stream.

It also fixes a couple of bugs discovered while testing and renames some functions to be consistent across all functionality related to a data frame.

### DIFF
--- a/include/api/CDataFrameAnalysisRunner.h
+++ b/include/api/CDataFrameAnalysisRunner.h
@@ -23,6 +23,10 @@
 namespace ml {
 namespace core {
 class CDataFrame;
+class CRapidJsonConcurrentLineWriter;
+namespace data_frame_detail {
+class CRowRef;
+}
 }
 namespace api {
 class CDataFrameAnalysisSpecification;
@@ -53,6 +57,7 @@ class CDataFrameAnalysisSpecification;
 class API_EXPORT CDataFrameAnalysisRunner {
 public:
     using TStrVec = std::vector<std::string>;
+    using TRowRef = core::data_frame_detail::CRowRef;
 
 public:
     CDataFrameAnalysisRunner(const CDataFrameAnalysisSpecification& spec);
@@ -67,9 +72,25 @@ public:
     //! into main memory during an analysis.
     virtual std::size_t numberOfPartitions() const = 0;
 
-    //! \return The number of columns this analysis requires. This includes
-    //! the columns of the input frame plus any that the analysis will append.
-    virtual std::size_t requiredFrameColumns() const = 0;
+    //! \return The number of columns this analysis appends.
+    virtual std::size_t numberExtraColumns() const = 0;
+
+    //! Write the extra columns of \p row added by the analysis to \p writer.
+    //!
+    //! This should create a new object of the form:
+    //! <pre>
+    //! {
+    //!   "name of column n":   "value of column n",
+    //!   "name of column n+1": "value of column n+1",
+    //!   ...
+    //! }
+    //! </pre>
+    //! with one named member for each column added.
+    //!
+    //! \param[in] row The row to write the columns added by this analysis.
+    //! \param[in,out] writer The stream to which to write the extra columns.
+    virtual void writeOneRow(TRowRef row,
+                             core::CRapidJsonConcurrentLineWriter& writer) const = 0;
 
     //! Checks whether the analysis is already running and if not launches it
     //! in the background.
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -63,6 +63,8 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec& 
     // be triggered to run by calling run explicitly.
 
     if (m_AnalysisSpecification == nullptr || m_AnalysisSpecification->bad()) {
+        // TODO We need to communicate an error but don't want one for each row.
+        // Revisit this when we have finalized our monitoring strategy.
         LOG_TRACE(<< "Specification is bad");
         return false;
     }
@@ -73,6 +75,8 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec& 
     }
 
     if (this->sufficientFieldValues(fieldValues) == false) {
+        // TODO We need to communicate an error but don't want one for each row.
+        // Revisit this when we have finalized our monitoring strategy.
         LOG_TRACE(<< "Expected " << m_AnalysisSpecification->numberColumns()
                   << " field values and got " << fieldValues.size());
         return false;
@@ -88,7 +92,7 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec& 
 
 void CDataFrameAnalyzer::receivedAllRows() {
     m_DataFrame->finishWritingRows();
-    LOG_TRACE(<< "Received " << m_DataFrame->numberRows() << " rows");
+    LOG_DEBUG(<< "Received " << m_DataFrame->numberRows() << " rows");
 }
 
 void CDataFrameAnalyzer::run() {
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -110,27 +110,7 @@ void CDataFrameAnalyzer::run() {
 
     analysis->waitToFinish();
 
-    auto outStream = m_OutStreamSupplier();
-    core::CRapidJsonConcurrentLineWriter outputWriter{*outStream};
-
-    // We write results single threaded because we need to write the rows to
-    // Java in the order they were written to the data_frame_analyzer so it
-    // can join the extra columns with the original data frame.
-    std::size_t numberThreads{1};
-
-    using TRowItr = core::CDataFrame::TRowItr;
-    m_DataFrame->readRows(numberThreads, [&](TRowItr beginRows, TRowItr endRows) {
-        for (auto row = beginRows; row != endRows; ++row) {
-            outputWriter.StartObject();
-            outputWriter.String(ID_HASH);
-            outputWriter.String(ID_HASH); // TODO this should be the actual hash
-            outputWriter.String(RESULTS);
-            analysis->writeOneRow(*row, outputWriter);
-            outputWriter.EndObject();
-        }
-    });
-
-    outputWriter.flush();
+    this->writeResultsOf(*analysis);
 }
 
 bool CDataFrameAnalyzer::readyToReceiveControlMessages() const {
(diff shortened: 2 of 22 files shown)


tveasey:
Regarding [bf892f9](https://github.com/elastic/ml-cpp/pull/334/commits/bf892f9ba5ae10ca9989c470c7e2ef4284ac7609), I talked this through a bit more with @dimitris-athanasiou. We decided in the end it was cleanest thing is to trigger run from an end of data. Down-the-line, as we add the capability to compose analyses, we will support receiving new specifications as control messages. But the basic idea is that the data_frame_analyzer will switch to holding a queue of analyses to run and just start draining this queue as soon as end of data has been triggered.  If new analyses arrive we will run them immediately.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -111,38 +156,51 @@ bool CDataFrameAnalyzer::prepareToReceiveControlMessages(const TStrVec& fieldNam
...
+
+bool CDataFrameAnalyzer::sufficientFieldValues(const TStrVec& fieldValues) const {
+    if (m_ControlFieldIndex >= 0) {
+        return fieldValues.size() == m_AnalysisSpecification->numberColumns() + 1;
+    }
+    return fieldValues.size() == m_AnalysisSpecification->numberColumns();
 }
 
 bool CDataFrameAnalyzer::handleControlMessage(const TStrVec& fieldValues) {
+    LOG_TRACE(<< "Control message: '" << fieldValues[m_ControlFieldIndex] << "'");
```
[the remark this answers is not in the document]
droberts195: We already have an error for unrecognised messages.  To avoid logging masses of spaces when the log level is `DEBUG`, it might be best to put this as `LOG_DEBUG` in the `FINISHED_DATA_CONTROL_MESSAGE_FIELD_VALUE` case.  Each case can separately decide which level to log at.

lib/maths/CLocalOutlierFactors.cc
```diff
@@ -80,7 +80,7 @@ bool computeOutliersNoPartitions(std::size_t numberThreads, core::CDataFrame& fr
 
     auto rowsToPoints = [&points](TRowItr beginRows, TRowItr endRows) {
         for (auto row = beginRows; row != endRows; ++row) {
-            points[row->index()] = TVector{row->data(), row->numberColumns()};
+            new (&points[row->index()]) TVector{row->data(), row->numberColumns()};
```
droberts195: Is it really necessary to use placement new here?  It could cause some subtle bug to be introduced in the future if a future version of one of the classes involved here ever has a non-trivial destructor.

Is there a way of achieving similar efficiency that doesn't involve placement new?  If assigning the whole `TVector` has proved to be highly inefficient for some reason, how about assigning the two members individually?

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -53,54 +62,90 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec&
...
+        return false;
+    }
+
+    if (this->readyToReceiveControlMessages() == false &&
+        this->prepareToReceiveControlMessages(fieldNames) == false) {
+        return false;
+    }
+
+    if (this->sufficientFieldValues(fieldValues) == false) {
+        LOG_TRACE(<< "Expected " << m_AnalysisSpecification->numberColumns()
```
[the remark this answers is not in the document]
tveasey: I agree these should be recorded at error level somehow. However, the question is do we want to generate one error for every single row or a summary.

This is wrapped up with what we're planning to do for monitoring, but the consensus last time we discussed was we should maintain counters for problems which could occur many times.

lib/maths/CLocalOutlierFactors.cc
tveasey: This is the only way you can change the value underlying values referenced by an Eigen::Map: see https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html.

tveasey: I suspect that it is overloaded for this class although haven't looked at the details

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -53,54 +62,90 @@ bool CDataFrameAnalyzer::handleRecord(const TStrVec& fieldNames, const TStrVec&
...
-        outputWriter.EndObject();
-        outputWriter.EndObject();
-    }
+
+    // We write results single threaded because we need to write the rows to
+    // Java in the order they were written to the data_frame_analyzer so it
+    // can join the extra columns with the original data frame.
+    std::size_t numberThreads{1};
+
+    using TRowItr = core::CDataFrame::TRowItr;
```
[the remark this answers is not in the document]
tveasey: [d618268](https://github.com/elastic/ml-cpp/pull/334/commits/d6182689bf7a4af225f84e5e62fd4f9de1b4206f)

lib/api/CDataFrameAnalyzer.cc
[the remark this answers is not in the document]
tveasey: I think I'll keep as is: we will no doubt create new functions for each new significant control message and these can log appropriate information, such as `receivedAllRows` does now. I think it is quite nice having a log line here though which we can can use for ad hoc debugging.

tveasey:
I've left some TODOs regarding logging levels. I think I'll revisit in a PR which improves monitoring in general.

lib/maths/CLocalOutlierFactors.cc
tveasey: By way of background, I should add that I want to reference memory in the `CDataFrame` directly, so no copying is wanted. I also want to pre-size the vector for Eigen::Map objects, so I don't need to lock on push_buck. This seems to be the only/designed way of reseating such objects.

droberts195: Based on the docs you could change lines 78-79 from:

```
    CFloatStorage initial;
    TVectorVec points(frame.numberRows(), TVector{&initial, 1});
```

to:

```
    TVectorVec points(frame.numberRows(), TVector{nullptr});
```

That might help avoid any side effects of not calling a destructor of an object overwritten by the placement new after some future change.

include/api/CDataFrameAnalysisRunner.h
```diff
@@ -67,9 +72,25 @@ class API_EXPORT CDataFrameAnalysisRunner {
...
+    //!   "name of column n+1": "value of column n+1",
+    //!   ...
+    //! }
+    //! </pre>
+    //! with one named member for each column added.
+    //!
+    //! \param[in] row The row to write the columns added by this analysis.
+    //! \param[in,out] writer The stream to which to write the extra columns.
+    virtual void writeOneRow(TRowRef row,
+                             core::CRapidJsonConcurrentLineWriter& writer) const = 0;
```
hendrikmuhs: can the writer change per call? if not, can it be set at construction?

tveasey: This is the object created [here](https://github.com/elastic/ml-cpp/pull/334/files#diff-740e94a98f6c2a36c617970df2bc2c81R213) so not currently available at construction time of `CDataFrameAnalysisRunner`. We could refactor so it is a member of `CDataFrameAnalyzer` and only created once (rather than once per analysis, although currently this is equivalent). I think this would be a good idea, but I don't want to make that refactor now.

That aside I'm not sure I'd want to make this a member of the `CDataFrameAnalysisRunner`. Really, this function shouldn't require state from any implementation object. It isn't static in order to be virtual, because different analyses will want to extract different column values. However, all the information is coming from the row passed in and some local statics. So this feels extrinsic to a particular instance of the class and being forced to always create a row writer whenever you create an object feels undesirable to me.

lib/api/CDataFrameAnalyzer.cc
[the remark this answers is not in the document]
hendrikmuhs: apart from the log level, i think it is interesting how we transport the message back to the user, logs are overwhelming and most people do not read them until they are aware of some big trouble. So somehow the job status should indicate a problem/error and show a message (doing it like forecast is an option, but it could also be something else)

lib/maths/CLocalOutlierFactors.cc
tveasey: Yes, this is definitely better and should mean we'll immediately get a fault as soon as an object is used if it wasn't reset rather than unpredictable behaviour. (I'm pretty sure this should be future safe since Eigen::Map is essentially a pointer.)

tveasey:
I think I've now captured all comments one way or another. Can people have another look?

lib/api/CDataFrameAnalyzer.cc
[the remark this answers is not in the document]
tveasey: I agree. Some messages definitely need to be displayed more visibly to the user. However, IMO some messages are going to remain very difficult for a user to act on and I think these appearing in logs is appropriate. Remember as well we have two use cases: interactive and realtime. In real time we're going to need a different mechanism for surfacing important errors. Let's continue to work on the design for monitoring offline.

include/api/CDataFrameAnalysisRunner.h
hendrikmuhs: understood, I am asking because this looks like the other extreme: supplying the write for _every_ row. Looks like the reason for this is the extra wrapping required in `writeResultsOf`.

What about just accessing the data and doing the json'ification in `writeResultsOf` , which would also get rid of forward declarations.

Just suggestions.

tveasey: The problem is we'd then need logic in this function for every single type of analysis. I quite like that the CDataFrameAnalyzer class is agnostic to the type of analyses it is running. To keep this it has to somehow pass off writing analysis specific parts of the results to the `CDataFrameAnalysisRunner`.

If I understand your suggestion correctly, we'd have to have a something like a big switch statement in writeResultsOf and some sort of  identifier returned by each analyser.

Equally if all the logic in `writeResultsOf` gets pushed down into `CDataFrameAnalysisRunner` implementations we'd duplicate it for each different analysis.

hendrikmuhs: > If I understand your suggestion correctly, we'd have to have a something like a big switch statement in writeResultsOf and some sort of identifier returned by each analyser.

sorry, I do not get that. All I see at the moment is writing 1 key and 1 value. No need for a big switch statement.

tveasey: I mean down the line when we have additional analysis types (regression, etc) and these will all have their own runners (say CDataFrameXGBoostRunner, etc). Each type of analysis will have its own column names and numbers of columns they want to write. This information can all be in the `writeResultsOf`, but not without doing something like a switch statement. Additional analysis types are definitely coming, so this isn't premature generalisation.

lib/maths/CLocalOutlierFactors.cc
tveasey: [bfd91e3](https://github.com/elastic/ml-cpp/pull/334/commits/bfd91e33b4f550238e4a395573434748d6a9a49d)

PR #335, opened 2018-12-11
[ML] Fix for CPoissonMeanConjugate sampling error.

edsavage:
Fix for "sample out of bounds" error generated from
CPoissonMeanConjugate

### DIFF
--- a/lib/maths/CPoissonMeanConjugate.cc
+++ b/lib/maths/CPoissonMeanConjugate.cc
@@ -642,6 +642,11 @@ void CPoissonMeanConjugate::sampleMarginalLikelihood(std::size_t numberSamples,
                                 m_Offset;
 
                 LOG_TRACE(<< "sample = " << sample);
+                if (sample < 0) {
+                    // We discard negative samples as they are logically incongruous for a counting distribution.
+                    // These are extremely rare but can occur when the variance is large and the offset is 0, for example.
+                    continue;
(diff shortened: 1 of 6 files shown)


lib/maths/CPoissonMeanConjugate.cc
```diff
@@ -642,6 +642,11 @@ void CPoissonMeanConjugate::sampleMarginalLikelihood(std::size_t numberSamples,
                                 m_Offset;
 
                 LOG_TRACE(<< "sample = " << sample);
+                if (sample < 0) {
+                    // We discard negative samples as they are logically incongruous for a counting distribution.
+                    // These are extremely rare but can occur when the variance is large and the offset is 0, for example.
+                    continue;
+                }
```
tveasey: I don't think this is quite right. I think `lastPartialExpectation = partialExpectation;` does need executing at the end of the loop in this case. I would propose instead changing the following lines to:
```
if (sample >= support.first && sample <= support.second) {
    samples.push_back(sample);
} else if (sample < support.first) {
    ...
    // fall through
} else {
    LOG_ERROR(<< "Sample out of bounds: sample = " << sample << ", support = ["
                           << support.first << "," << support.second << "]"
                           << ", mean = " << mean << ", variance = " << variance
                           << ", q = " << q << ", x(q) = " << xq);
}
```

PR #336, opened 2018-12-11
[FEATURE][ML] Migrate data frame background write to storage to use thread pool

tveasey:
This reworks the data frame writer to use the thread pool rather than launching its own threads to write to storage in the background. Aside from simplifying the code, this also means that we properly enforce the number of threads constraint supplied to the data_frame_analyzer command. I also improved a couple of comments from #322.

### DIFF
--- a/include/core/CDataFrame.h
+++ b/include/core/CDataFrame.h
@@ -7,22 +7,22 @@
 #ifndef INCLUDED_ml_core_CDataFrame_h
 #define INCLUDED_ml_core_CDataFrame_h
 
-#include <core/CConcurrentQueue.h>
-#include <core/CDataFrameRowSlice.h>
 #include <core/CFloatStorage.h>
+#include <core/Concurrency.h>
 #include <core/ImportExport.h>
 
 #include <boost/optional.hpp>
 
 #include <algorithm>
 #include <functional>
-#include <future>
 #include <iterator>
 #include <memory>
 #include <vector>
 
 namespace ml {
 namespace core {
+class CDataFrameRowSlice;
+class CDataFrameRowSliceHandle;
 
 namespace data_frame_detail {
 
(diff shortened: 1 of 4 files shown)


include/core/CDataFrame.h
```diff
@@ -175,11 +175,10 @@ class CORE_EXPORT CDataFrame final {
     using TRowFuncVec = std::vector<TRowFunc>;
     using TRowFuncVecBoolPr = std::pair<TRowFuncVec, bool>;
     using TWriteFunc = std::function<void(TFloatVecItr)>;
-    using TRowSlicePtr = std::unique_ptr<CDataFrameRowSlice>;
+    using TRowSlicePtr = std::shared_ptr<CDataFrameRowSlice>;
```
droberts195: Just noticed this breaks rule 2 from https://github.com/elastic/ml-cpp/blob/master/STYLEGUIDE.md#language-extensions because we're still using `boost::shared_ptr` almost everywhere else.  It doesn't matter so much here because there's probably nowhere where `TRowSlicePtr` is ever a `boost::shared_ptr`.  But `using TStrCPtr = std::shared_ptr<const std::string>` would create exactly the risk that that rule was designed to avoid.  So it would be good to replace _all_ the `boost::shared_ptr`s in the codebase with `std::shared_ptr` at some point.  Soon would be a good time to do it as the new analytics won't be backported to 6.x and if it's done in master now it will be in 7.x by default when that's branched.

tveasey: I think the style guide is now out-of-date. We migrated in #52. But let's migrate to std for all C++14 supported functionality on master now. There are a few other things in a backlog we can switch over, std::ref, std::bind, std::make_unique, etc. I'll do a quick audit and raise an issue.

droberts195: Doh.  I was looking at the 5.6 branch.  Yes, we should be using `std` for pointer types.  I think the rule is still relevant because `unordered_map` and `optional` are still from boost at the moment.  But there are only two stray boost pointer types left in the master branch.

tveasey: Ok cool. Yes, I was thinking the guidelines explicitly referenced pointer types without rereading (I agree, it is fine as it is worded). I'd still be in favour of migrating other boost usage where we can (aside from hash maps for reasons we discussed previously).

tveasey:
retest

PR #339, opened 2018-12-12
[6.5][ML] Correct query times for model plot and forecast

tveasey:
Backport #327.


docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,14 @@
 
 //=== Regressions
 
+ == {es} version 6.5.3
```
droberts195: Please change to 6.5.4

PR #344, opened 2018-12-14
[FEATURE][ML] change artifactname in order to create a CI build job

hendrikmuhs:
changes the artifact name to  `ml-cpp-df` in order setup a CI build that uploads builds without clashing with master builds

Before the feature branch is merged, this change must be reverted.

@dimitris-athanasiou

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,4 +2,4 @@ org.gradle.daemon=false
 
 elasticsearchVersion=7.0.0
 
-artifactName=ml-cpp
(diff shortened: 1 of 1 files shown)


tveasey:
Could we leave a TODO so this isn't missed

dimitris-athanasiou:
I have raised an issue to track those and I added it there

tveasey:
Ignore the nix failure: it is #343

PR #345, opened 2018-12-17
[FEATURE][ML] Groundwork for partitioning data frame analyses

tveasey:
This is some of the groundwork for partitioning data frame analyses in general and outlier detection in particular.

It adds functionality to compute the expected memory usage for a `CDataFrame` and for bookkeeping for the algorithms which run on it. This is used to estimate the size of the partition, the maximum number of rows per partition subset and whether to store the data frame in or out of main memory. I've also added a placeholder for passing the directory to use for out-of-core storage from Java.

It isn't fully testable yet because not all the associated functionality is implemented, so I've left some todos in the unit tests.

### DIFF
--- a/include/maths/CLocalOutlierFactors.h
+++ b/include/maths/CLocalOutlierFactors.h
@@ -45,7 +45,13 @@ public:
     using TDoubleVecVec = std::vector<TDoubleVec>;
 
     //! The available algorithms.
-    enum EAlgorithm { E_Lof, E_Ldof, E_DistancekNN, E_TotalDistancekNN };
+    enum EAlgorithm {
+        E_Lof,
+        E_Ldof,
+        E_DistancekNN,
+        E_TotalDistancekNN,
+        E_Ensemble
+    };
 
 public:
     //! Compute the normalized LOF scores for \p points.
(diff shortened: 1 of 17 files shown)


include/maths/CLocalOutlierFactors.h
```diff
@@ -184,6 +190,47 @@ class MATHS_EXPORT CLocalOutlierFactors {
...
+    static std::size_t estimateMemoryUsage(EAlgorithm algorithm,
+                                           std::size_t k,
+                                           std::size_t numberPoints,
+                                           std::size_t dimension) {
+        std::size_t result{TKdTree<POINT>::estimateMemoryUsage(numberPoints, dimension)};
+        switch (algorithm) {
+        case E_Ensemble:
+            result += CLof<POINT, TKdTree<POINT>>::estimateMemoryUsage(k, numberPoints) +
+                      CEnsemble<POINT, TKdTree<POINT>>::estimateMemoryUsage(
+                          numberPoints, dimension);
```
benwtrent: Since Ensemble could include `lof` why do we add the size of `lof` here as well? Does ensemble not care?

tveasey: As it stands each class only accounts for their own memory usage in estimateMemoryUsage, so `CEnsemble` is only accounting for its own state and not that of `CLof`.

There is a bit of a rough edge here in that the ensemble can have any collection of methods added to it at run time, but this is not configured anywhere, and this code just assumes that all of them will be included.

So, the reason I did this rather than have CEnsemble::estimateMemoryUsage account for the memory usage of the underlying methods as well, was that I was thinking down the line, it might be cleaner to use a bit mask for selecting subsets of the methods to use as an ensemble. In this case, we'd change the switch to something like:
```
if (algorithm & E_Lof) {
}
...
```
and it seemed cleaner to then have each class **only** account for own memory overheads.

Given the confusion, I think it is fair to say that I should make the naming of this method clearer; how about `estimateMemoryDirectlyUsed`?

benwtrent: Ah, I see now. Yeah, a rename would be nice as `estimateMemoryUsage` does seem to imply "The memory I am using" would could easily encompass the underlying methodologies.

tveasey: [972d15b](https://github.com/elastic/ml-cpp/pull/345/commits/972d15b0e70e9742b314c15133beb59bfb9c8713).

This also fixes a couple of silly bugs. These would get caught when we can test running memory usage vs estimated memory usage, which needs the monitoring framework in place.

PR #359, opened 2019-01-08
[FEATURE][ML] Reduce bookkeeping memory usage for outlier detection

tveasey:
This revisits the storage format for some bookkeeping state used during outlier detection. It also makes sure to reclaim memory allocated for the vector buffer which is moved into the k-d tree.

In particular, this will save us at least `# rows * sizeof(point type)` + `# rows * sizeof(std::vector)` + `# rows * 8 * k` bytes, for k nearest neighbours, during a run of ensemble or LOF outlier detection and `# rows * sizeof(point type)` for all other methods.

### DIFF
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -20,6 +20,8 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <limits>
+#include <type_traits>
 #include <vector>
 
 namespace ml {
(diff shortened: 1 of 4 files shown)


lib/maths/unittest/CLocalOutlierFactorsTest.cc
```diff
@@ -8,6 +8,7 @@
 
 #include <core/CContainerPrinter.h>
 #include <core/CLogger.h>
+#include <core/CStopWatch.h>
```
edsavage: Doesn't appear to be used?

include/maths/CKdTree.h
```diff
@@ -306,12 +307,12 @@ class CKdTree {
 
 private:
     //! Append a node moving \p point into place.
-    void append(True /*move*/, SNode* parent, POINT& point) {
+    void append(std::true_type, SNode* parent, POINT& point) {
         m_Nodes.emplace_back(parent, std::move(point));
     }
 
     //! Append a node coping \p point into place.
```
edsavage: nit: copying

PR #360, opened 2019-01-09
Persist static statistics in a consistent manner

edsavage:
Uses a transient cache that is cleaned up after persistence occurs

### DIFF
--- a/include/core/CStatistics.h
+++ b/include/core/CStatistics.h
@@ -13,9 +13,11 @@
 #include <boost/array.hpp>
 
 #include <iosfwd>
-#include <map>
+#include <vector>
 
 #include <stdint.h>
+
+class CStatisticsTest;
 
 namespace ml {
 namespace stat_t {
@@ -109,12 +111,25 @@ class CStatePersistInserter;
 //! A singleton class: there should only be one collection of global stats
 //!
 class CORE_EXPORT CStatistics : private CNonCopyable {
+private:
+    class CStatsCache;
+
+private:
+    using TStatArray = boost::array<CStat, stat_t::E_LastEnumStat>;
+    using TStatsCacheUPtr = std::unique_ptr<CStatsCache>;
+
 public:
     //! Singleton pattern
     static CStatistics& instance();
 
     //! Provide access to the relevant stat from the collection
     static CStat& stat(int index);
+
+    //! copy the collection of live statistics to a cache
+    static void cacheStats();
+
+    //! Transfer ownership of the cached statistics to the caller
+    static TStatsCacheUPtr transferCachedStats();
 
     //! \name Persistence
     //@{
@@ -126,9 +141,6 @@ public:
     //@}
 
 private:
-    using TStatArray = boost::array<CStat, stat_t::E_LastEnumStat>;
-
-private:
     //! Constructor of a Singleton is private
     CStatistics();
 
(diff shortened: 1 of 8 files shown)


lib/core/CStatistics.cc
```diff
@@ -196,5 +204,21 @@ std::ostream& operator<<(std::ostream& o, const CStatistics& /*stats*/) {
     return o;
 }
 
+CStatistics::CStatsCache::CStatsCache() {
+}
+
+void CStatistics::CStatsCache::populate(const CStatistics::TStatArray& statArray) {
```
droberts195: It would be safer to have the constructor do the population.  If this `populate()` method is accidentally called more than once for the same object then it just appends, resulting in twice as many statistics as expected.

lib/core/CStatistics.cc
```diff
@@ -67,23 +67,31 @@ CStat& CStatistics::stat(int index) {
     return ms_Instance.m_Stats[index];
 }
 
+void CStatistics::cacheStats() {
+    ms_Instance.m_Cache = std::make_unique<CStatsCache>();
+    ms_Instance.m_Cache->populate(ms_Instance.m_Stats);
```
droberts195: I left a comment elsewhere that the work of `populate()` should be done in the constructor.  That has a spin-off benefit here too, because then if this method is called simultaneously from two threads it probably won't cause a core dump - whichever assigns `m_Cache` last wins, but at least you get a single consistent cache.

lib/core/CStatistics.cc
```diff
@@ -67,23 +67,31 @@ CStat& CStatistics::stat(int index) {
     return ms_Instance.m_Stats[index];
 }
 
+void CStatistics::cacheStats() {
```
droberts195: Is there ever a valid situation when we expect this to be called when `m_Cache` is not null?  If not, it would be good to log a warning as the first operation if `m_Cache != nullptr`.

lib/core/unittest/CStatisticsTest.cc
```diff
@@ -97,6 +99,55 @@ void CStatisticsTest::testStatistics() {
     CPPUNIT_ASSERT_EQUAL(uint64_t(0x1000000), stats.stat(TEST_STAT).value());
 }
 
+void CStatisticsTest::testCacheStatistics() {
+    ml::core::CStatistics& stats = ml::core::CStatistics::instance();
+
+    {
+        //! obtain ownership of the stats cache
```
tveasey: nit //! -> //

include/core/CStatistics.h
```diff
@@ -109,13 +112,26 @@ class CStatePersistInserter;
 //! A singleton class: there should only be one collection of global stats
 //!
 class CORE_EXPORT CStatistics : private CNonCopyable {
+private:
+    class CStatsCache;
+
+private:
+    using TStatArray = boost::array<CStat, stat_t::E_LastEnumStat>;
```
tveasey: nit: std::array.

include/core/CStatistics.h
```diff
@@ -138,8 +151,40 @@ class CORE_EXPORT CStatistics : private CNonCopyable {
     //! Collection of statistics
     TStatArray m_Stats;
 
+    //! cache of stats for use in persistence
+    TStatsCacheUPtr m_Cache;
```
tveasey: Couldn't this just be `std::unique_ptr<TStatsArray>`? I'm not sure CStatsCache is adding anything.

lib/core/CStatistics.cc
tveasey: If as above this issue goes away, i.e. `ms_Instance.m_Cache = std::make_unique<TStatsArray>(ms_Instance.m_Stats);`

PR #362, opened 2019-01-09
[FEATURE][ML] Progress monitoring

tveasey:
This wires in progress monitoring to the data frame analysis framework. I've discretised the progress so I can efficiently update atomically. The actual progress is updated via a callback to maximally decouple the "observer" from the algorithms whose progress is being observed. This also makes it trivial to stub out progress monitoring.

Note I haven't yet done anything to pass this back during a long running analysis, because the Java side isn't expecting anything yet.

### DIFF
--- a/include/api/CDataFrameAnalysisRunner.h
+++ b/include/api/CDataFrameAnalysisRunner.h
@@ -132,7 +132,17 @@ protected:
 
     void setToBad();
     void setToFinished();
-    void updateProgress(double fractionalProgress);
+
+    //! This adds \p fractionalProgess to the current progress.
+    //!
+    //! \note The caller should try to ensure that the sum of the values added
+    //! at the end of the analysis is equal to one.
+    //! \note This is converted to an integer - so we can atomically add - by
+    //! scaling by 1024. Therefore, this shouldn't be called with values less
+    //! than 0.001. In fact, it is unlikely that such high resolution is needed
+    //! and typically this would be called significantly less frequently.
+    void recordProgress(double fractionalProgress);
+
     void addError(const std::string& error);
 
 private:
@@ -150,7 +160,7 @@ private:
 
     bool m_Bad = false;
     std::atomic_bool m_Finished;
-    std::atomic<double> m_FractionalProgress;
+    std::atomic_int m_FractionalProgress;
     TStrVec m_Errors;
 
     std::thread m_Runner;
--- a/lib/core/unittest/CLoopProgressTest.cc
+++ b/lib/core/unittest/CLoopProgressTest.cc
@@ -19,7 +19,6 @@ void CLoopProgressTest::testShort() {
 
     double progress{0.0};
 
-    LOG_DEBUG(<< "Test short loops");
     for (std::size_t n = 1; n < 16; ++n) {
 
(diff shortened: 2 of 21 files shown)


lib/core/unittest/CLoopProgressTest.cc
```diff
@@ -0,0 +1,142 @@
...
+#include <core/CLogger.h>
+#include <core/CLoopProgress.h>
+
+#include <test/CRandomNumbers.h>
+
+using namespace ml;
+
+using TSizeVec = std::vector<std::size_t>;
+
+void CLoopProgressTest::testShort() {
```
dimitris-athanasiou: I think it'd be be nice to add a simple test (say 4 steps) that also verifies the exact progress in after each increment.

lib/core/unittest/CLoopProgressTest.cc
```diff
@@ -0,0 +1,142 @@
...
+        LOG_DEBUG(<< "Testing " << n);
+
+        progress = 0.0;
+        core::CLoopProgress loopProgress{n,
+                                         [&progress](double p) { progress += p; }};
+
+        for (std::size_t i = 0; i < n; i += 2, loopProgress.increment(2)) {
+            // Empty
+        }
+    }
```
dimitris-athanasiou: missing assertion?

include/api/CDataFrameAnalysisRunner.h
```diff
@@ -132,7 +132,17 @@ class API_EXPORT CDataFrameAnalysisRunner {
 
     void setToBad();
     void setToFinished();
-    void updateProgress(double fractionalProgress);
+
+    //! This adds \p fractionalProgess to the current progress.
+    //!
+    //! \note The caller should try to ensure that the sum of the values added
```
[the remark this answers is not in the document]
tveasey: I think this is still valid. The overall progress monitoring might use multiple loop progress monitors with weights dependent on their proportion of the run time. For example, imagine we have at the top level, schematically

...
Loop A which takes 20% of the run time
...
Loop B which takes 80% of the run time
...

When the caller instruments this code for progress monitoring they have to scale the progress for loop A by w(A) = 0.2 and for loop B by w(B) = 0.8, for example by wrapping the callback recordProgress passed to the function invoking Loop A as follows `[&recordProgress](double p) { recordProgress(0.2 * p); }`.

It the responsibility of the caller to ensure that w(A) + w(B) = 1 to get decent progress monitoring, which is what this comment states.

lib/core/unittest/CLoopProgressTest.cc
tveasey: I've added some incremental progress assertions: [f1d9ea1](https://github.com/elastic/ml-cpp/pull/362/commits/f1d9ea129087137db34347f269dd3c8e7242e3b3). These cover your other comment I think.

lib/core/unittest/CLoopProgressTest.cc
tveasey: See below.

tveasey:
Thanks @dimitris-athanasiou. I've addressed your comment on test coverage, can you take another look.

PR #365, opened 2019-01-16
[FEATURE][ML] Respect user defined overrides for data frame outliers

tveasey:
This wires in user defined overrides for outlier detection.


tveasey:
We're going to merge to master. I also have changes to the outlier detection methodology which impact this PR. I'm going to close for the time being and rework.

PR #366, opened 2019-01-16
[DOCS] Updates changelog for 6.6.0

lcawl:
This PR adds the following PRs to the changelog for 6.6.0:
https://github.com/elastic/ml-cpp/pull/338
https://github.com/elastic/ml-cpp/pull/357

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -43,10 +43,7 @@
 Fix cause of "Sample out of bounds" error message. {ml-pull}335[335]
 
 Fix hanging autodetect process after it receives a warning about failed
-cleanup of forecasts. {ml-pull}357[#357] (issue: {ml-issue}350[#350])
-
-Prevent trend detection during a change in a time series that creates a poorly
-reinitialized model. {ml-pull}292[#292]
+cleanup of forecasts. {ml-pull}352[#352] (issue: {ml-issue}350[#350])
 
 === Regressions
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,7 +28,7 @@
 
 //=== Regressions
 
- == {es} version 6.6.0
+== {es} version 6.6.0
 
 === Breaking Changes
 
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -40,10 +40,17 @@
 
 === Bug Fixes
 
-Fix cause of "Sample out of bounds" error message (See {ml-pull}355[355].}
+Fix cause of "Sample out of bounds" error message. {ml-pull}335[335]
```
droberts195: Should there be a `#` before the `335` in square brackets?

docs/CHANGELOG.asciidoc
```diff
@@ -40,10 +40,17 @@
...
+
+Fix hanging autodetect process after it receives a warning about failed
+cleanup of forecasts. {ml-pull}352[#352] (issue: {ml-issue}350[#350])
```
droberts195: Maybe something more like this?

> Fix hang during job close after failure to clean up temporary storage used by forecasting.

Closing the job is where the end user suffers from the problem.

hendrikmuhs: New forecast requests are also not executed after this happened, plus it requires a large job to run into this, maybe:

Fix hang during job close and forecast execution after temporary storage couldn't be cleaned up when forecasting a large job.

lcawl: Thanks, I've drafted another version of that description.

hendrikmuhs: I suggest to move the entry to 6.6.0, it's listed for 6.5.5 but 6.5.5 has been cancelled (or postponed) and to me it seems it is unlikely that there will be a 6.5.5 before 6.6.0. 

This wasn't foreseeable when I backported the fix.

@droberts195, @lcawl  what do you think?

droberts195: The PR listed in the 6.5.5 section is the exact same one that is already listed on lines 45/46 in the 6.6.0 section, so it's probably easiest to just delete the 6.5.5 section.

Or you could take inspiration from the current 6.5.5 description to adjust the 6.6.0 description.  They're both intending to say the same thing but feel free to adjust the one we keep if you think one way of saying it is clearer.

lcawl: Thanks!  I've removed the duplicate and will now copy this 6.6 information to the Elasticsearch Release Notes.

PR #367, opened 2019-01-17
[FEATURE][ML] Rework fatal error handling during a data frame analysis

tveasey:
We need some explicit feedback to the user if the process fails to produce any results at all for a given data frame analysis, even if it is just that there was some internal / logic error and to report the problem.

~~The intention is to introduce a result "type" member, which we can use for both progress monitoring and error reporting as types of result. This change lays the foundation for the error reporting aspect of this by introducing a common mechanism for returning fatal errors to the calling process.~~

~~An error handler is registered with the data frame analysis classes with the intention that this is a thread safe, non throwing wrapper around push_back on a vector of fatal errors. Since fatal errors halt further processing this collection should always be small.~~

~~At the end of the analysis we can then return any errors in this collection when we have updated the result object. That change will be made in a separate PR.~~

Our plan is to simply exit with a failure status whenever we hit a problem during a run of the process which means we will not be able to produce any results. This allows us to simplify code around error handling in the data frame and means we can also have a single simple mechanism in the Java for providing feedback to the user in this case (see [this](https://github.com/elastic/ml-cpp/pull/367#pullrequestreview-195064103) comment).

This change also tightens up the error handling slightly, for example checking for failure to reserve/resize the data frame during an analysis, which can happen if the process is oom.

### DIFF
--- a/lib/core/CLogger.cc
+++ b/lib/core/CLogger.cc
@@ -182,7 +182,7 @@ const CLogger::TFatalErrorHandler& CLogger::fatalErrorHandler() const {
 }
 
 void CLogger::handleFatal(std::string message) {
-    m_FatalErrorHandler(message);
+    m_FatalErrorHandler(std::move(message));
 }
 
 bool CLogger::setLoggingLevel(ELevel level) {
--- a/lib/test/CTestRunner.cc
+++ b/lib/test/CTestRunner.cc
@@ -56,8 +56,7 @@ const std::string CTestRunner::XML_RESULT_FILE_NAME("cppunit_results.xml");
 
 CTestRunner::CTestRunner(int argc, const char** argv) {
     ml::core::CLogger::instance().fatalErrorHandler([](std::string message) {
-        LOG_FATAL(<< message);
-        CPPUNIT_ASSERT(false);
+        CPPUNIT_FAIL(message);
     });
     this->processCmdLine(argc, argv);
 }


lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -158,19 +153,30 @@ std::size_t CDataFrameAnalysisRunner::estimateMemoryUsage(std::size_t numberRows
            this->estimateBookkeepingMemoryUsage(m_NumberPartitions, numberRows, numberColumns);
 }
 
+void CDataFrameAnalysisRunner::recordProgress(double fractionalProgress) {
+    m_FractionalProgress.fetch_add(std::max(
+        static_cast<int>(MAXIMUM_FRACTIONAL_PROGRESS * fractionalProgress + 0.5), 1));
```
droberts195: Where we've had progress indicators previously we've rounded down rather than to the nearest integer, because we don't want the progress to go to 100% when it's completely finished.

bin/data_frame_analyzer/Main.cc
```diff
@@ -33,13 +33,33 @@
...
 namespace {
-std::pair<std::string, bool> readFileToString(const std::string& fileName) {
+auto makeErrorHandler(std::vector<std::string>& userErrors, std::mutex& userErrorsMutex) {
+    return [&userErrors, &userErrorsMutex](const std::string& error) {
+        try {
+            std::lock_guard<std::mutex> lock{userErrorsMutex};
+            userErrors.push_back(error);
+        } catch (std::exception& e) {
+            // We want this to be nothrow. Copying strings can fail for example with
+            // with std::bad_alloc so we catch and swallow any exceptions here.
```
droberts195: But logging an error could also fail with `std::bad_alloc` (and almost certainly will if copying the exact same string just failed due to lack of memory), so it won't be possible to log an error here if the intention is to make this lambda a nothrow.

I don't think eating `bad_alloc` exceptions is a good idea anyway.  We decided a long time ago to just let `bad_alloc` exceptions propagate to the top level of the stack and crash the program.  I have tried in the past to write handlers for out of memory and they never work as well in practice as the theory suggests.

If you leave `bad_alloc` unhandled then the C runtime eventually intervenes and prints an error to `stderr` in a way that's more reliable than higher level code.  The we have code in our `CppLogMessageHandler` Java class that picks up plain text messages from the `stderr` and logs them as fatal errors.

So I think the exception handler should just be removed here and we let any exception crash the process.

lib/core/unittest/CDataFrameTest.cc
```diff
@@ -510,7 +520,7 @@ void CDataFrameTest::testDocHashes() {
                               << " expected " << expectedDocHash);
                     passed = false;
                 }
-                expectedDocHash += cols + extraCols;
+                expectedDocHash += static_cast<int>(cols + extraCols);
```
droberts195: nit: `int` should be `std::int32_t`

tveasey:
These don't cause the process die, by fatal I just mean not able to produce an any results. I mainly followed the forecasting paradigm of explicitly returning the error messages in this case. I think the requirement is that we produce some user visible message in this case.

If we think we can achieve this just as well/reliably by reading errors from our normal process logging on the Java side, then I'm happy to migrate this PR to simply tightening up error handling.

I'm guessing as a minimum though we need to somehow signal that the analytics has failed. It seems like if we communicate when an analysis has finished the Java could know if it has received any results from it and so deduce whether it failed.

@droberts195 does this seems like a sufficient strategy to meet the main requirement?

droberts195:
> These don't cause the process die, by fatal I just mean not able to produce an any results

I think it boils down to whether this is a good idea.  If the analysis is not able to produce any results because of a fatal environmental, configuration or internal error, is it worth keeping the process running?

If not then the process can exit with a non-zero status code, and that can trigger the Java to report the last few error messages that came via the logger.  (And these error messages will finish up with the exact same ones that would come via this new framework.)

Forecasting was a special case in this respect because forecasting was a short-lived analysis stowing away inside a potentially long-lived process (real-time anomaly detection).  So an error that was only fatal for forecasting could not cause the entire anomaly detection process to exit.

But for the new analytics it seems much cleaner that errors that mean the analysis cannot do anything like what it was asked to do simply exit the process.  Then a new process can be started once the error has been fixed.  I agree that in the case where the input data was fine and just the analysis configuration had a problem this may be inefficient as it will cause the input data to have to be reloaded when technically the user could correct the analysis configuration and we'd just send the process that already had the data in memory that fixed config to try again.  But this is immensely complex to test.  It seems to me that having the C++ process exit if it encounters an error reduces the complexity of the system considerably.

bin/data_frame_analyzer/Main.cc
tveasey: Yes. My rationale was don't increase the chance of the process dying (although perhaps the comment is misleading). If we can't log without dying then fine go ahead and die, but don't add a possibly throwing operation in this context without at least trying to catch. However, this is all moot depending on how we want to proceed on communicating failure back to the calling process.

tveasey:
> I agree that in the case where the input data was fine and just the analysis configuration had a problem this may be inefficient as it will cause the input data to have to be reloaded when technically the user could correct the analysis configuration and we'd just send the process that already had the data in memory that fixed config to try again. But this is immensely complex to test.

This was one case that was concerning me at the moment. Although we can likely trap this before any data are sent, at for the time being.

Longer term I had in the back of my mind, how will this interact with "continuous" operation? That said I think there are too many unknowns related to the details of how we're going to implement this at present to build in premature complexity: we may well simply spin up a task each time.

> It seems to me that having the C++ process exit if it encounters an error reduces the complexity of the system considerably.

Ok I like this suggestion too. Particularly, if we can (re)use one simple mechanism for reporting problems on the Java side. We can always revisit if this will causes us a problem down-the-line. I'll respin this PR along those lines.

include/api/CDataFrameAnalysisSpecification.h
```diff
@@ -15,14 +15,26 @@
...
+#include <functional>
 #include <memory>
 #include <string>
 #include <thread>
 #include <vector>
 
+#ifdef HANDLE_FATAL_ERROR
+#undef HANDLE_FATAL_ERROR
+#endif
+#define HANDLE_FATAL_ERROR(handler, message)                                   \
```
droberts195: It may be better to merge this functionality into `LOG_AND_EXIT` in `CLogger.h` so that a customised exit call can be used _everywhere_ where either `LOG_AND_EXIT` or `HANDLE_FATAL_ERROR` is currently used.

I think it's a bit dodgy at the moment that there are some parts of the code where `LOG_AND_EXIT` is called in code that could be run by unit tests.  If one of these errors occurs in a unit test then we'll exit the test suite without writing the test results file and won't run any subsequent tests.

If `CLogger` had a method to register the default exit function then it would be possible to change the constructor of `ml::test::CTestRunner` so that it registered an exit function that did a `CPPUNIT_FAIL` instead of `std::exit`.  The default default exit function on construction of `CLogger` could still be `std::exit`.

We'd want the individual tests that override the fatal error handler to just collect the fatal messages, so maybe there is a need to keep `LOG_AND_EXIT` and `HANDLE_FATAL_ERROR` separate, but either way, I think a way should be found to make `LOG_AND_EXIT` call `CPPUNIT_FAIL` rather than `std::exit` if a fatal error occurs in a unit test.

tveasey: That sounds sensible. The remaining calls really shouldn't crop up in the unit tests, but your suggestion is neater and I like the idea of having a global override tied to the logger anyway.

tveasey: I've made [this](https://github.com/elastic/ml-cpp/pull/367/commits/52b23d77bcc92724b07e4c3c60c5068994c18e0a) change. This really opens up the follow on task to unify this with LOG_ABORT, but I don't want to make that change now. I'll raise an issue for that work.

tveasey:
@droberts195 I have now addressed all review comments, can you take another look?

lib/test/CTestRunner.cc
```diff
@@ -54,6 +55,10 @@ const std::string CTestRunner::SKIP_FILE_NAME("unit_test_skip_dirs.csv");
 const std::string CTestRunner::XML_RESULT_FILE_NAME("cppunit_results.xml");
 
 CTestRunner::CTestRunner(int argc, const char** argv) {
+    ml::core::CLogger::instance().fatalErrorHandler([](std::string message) {
+        LOG_FATAL(<< message);
+        CPPUNIT_ASSERT(false);
```
droberts195: You could use `CPPUNIT_FAIL(message)`.  Then the message would be part of the test results reported through the framework, so would make it quicker to see what happened from a Jenkins test report for example.

include/api/CDataFrameAnalysisSpecification.h
```diff
@@ -40,9 +42,13 @@ class API_EXPORT CDataFrameAnalysisSpecification {
 public:
     using TStrVec = std::vector<std::string>;
     using TDataFrameUPtr = std::unique_ptr<core::CDataFrame>;
+    using TTemporaryDirectoryPtr = std::shared_ptr<core::CTemporaryDirectory>;
+    using TDataFrameUPtrTemporaryDirectoryPtrPr =
+        std::pair<TDataFrameUPtr, TTemporaryDirectoryPtr>;
     using TRunnerUPtr = std::unique_ptr<CDataFrameAnalysisRunner>;
     using TRunnerFactoryUPtr = std::unique_ptr<CDataFrameAnalysisRunnerFactory>;
     using TRunnerFactoryUPtrVec = std::vector<TRunnerFactoryUPtr>;
+    using TFatalErrorHandler = std::function<void(std::string)>;
```
droberts195: I think the argument here could be `const std::string&` rather than a value.

lib/core/CLogger.cc
```diff
@@ -167,6 +170,21 @@ void CLogger::fatal() {
...
+    // test code.
+    m_FatalErrorHandler = handler;
+}
+
+const CLogger::TFatalErrorHandler& CLogger::fatalErrorHandler() const {
+    return m_FatalErrorHandler;
+}
+
+void CLogger::handleFatal(std::string message) {
+    m_FatalErrorHandler(message);
```
droberts195: This is double copying the string at the moment.  If you don't like my suggestion to make the error handler function take a `const std::string&` then this could move `message`.

include/api/CDataFrameAnalysisSpecification.h
tveasey: The reason for this was that `std::ostringstream::str` (used in `HANDLE_FATAL`) returns by value so this gets copied when bound to the constant reference. I think the best solution is to move the string as below, which is an oversight, although this would never be a bottleneck.

PR #368, opened 2019-01-24
[FEATURE][ML] Ensure progress is less than one for unfinished data frame analysis tasks

tveasey:
This ensures the data_frame_analyzer command never reports progress as greater than or equal to one until the analysis task is complete. It also uses much greater precision for progress monitoring since (provided the sum of values added is close to one as it should be) this won't overflow the integer representation.

I also factored out the mock data frame. I had planned to use this in `CDataFrameAnalysisRunnerTest` but decided the existing test coverage of progress monitoring was good enough. However, I think it is useful to have this in its own file for reusability.

@droberts195 this address your comment in #367.

### DIFF
--- a/lib/api/CDataFrameAnalysisRunner.cc
+++ b/lib/api/CDataFrameAnalysisRunner.cc
@@ -24,7 +24,8 @@ std::size_t memoryLimitWithSafetyMargin(const CDataFrameAnalysisSpecification& s
     return static_cast<std::size_t>(0.9 * static_cast<double>(spec.memoryLimit()) + 0.5);
 }
 
-const std::size_t MAXIMUM_FRACTIONAL_PROGRESS{1ull << ((sizeof(std::size_t) - 2) * 8)};
+const std::size_t MAXIMUM_FRACTIONAL_PROGRESS{std::size_t{1}
+                                              << ((sizeof(std::size_t) - 2) * 8)};
 }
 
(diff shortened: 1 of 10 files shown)


lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -23,7 +24,7 @@ std::size_t memoryLimitWithSafetyMargin(const CDataFrameAnalysisSpecification& s
     return static_cast<std::size_t>(0.9 * static_cast<double>(spec.memoryLimit()) + 0.5);
 }
 
-const double MAXIMUM_FRACTIONAL_PROGRESS{1024.0};
+const std::size_t MAXIMUM_FRACTIONAL_PROGRESS{1ull << ((sizeof(std::size_t) - 2) * 8)};
```
droberts195: Instead of `1ull` I would use `std::size_t(1)`.  On some platforms `std::size_t` is not `unsigned long long`, nor even the same size as it.

tveasey: My thinking was the key requirement was that whatever type the shift operation didn't overflow, but I agree that is probably the cleanest.

tveasey: [a0c6874](https://github.com/elastic/ml-cpp/pull/368/commits/a0c6874634ced3ca48446c87852028de81725b22)

PR #369, opened 2019-01-30
[ML] Refactor CStatistics

edsavage:
Refactor CStatistics to simplify the implementation and to open the way
for it be used by applications other than just the current autodetect.

- Rename CStatistics -> CProgramCounters
- Remove the unneeded CStat wrapper layer
- Dynamic registration of counter types, allowing for separate sets of counters to be used and reported on in separate programs


lib/core/unittest/CProgramCountersTest.cc
```diff
@@ -0,0 +1,386 @@
...
+
+    // populate non-zero live counters
+    for (int i = 0; i < ml::counter_t::E_LastEnumCounter; i++) {
+        counters.counter(i) = (i + 1);
+    }
+
+    // copy the live values to the cache
+    counters.cacheCounters();
+
+    CPPUNIT_ASSERT(!counters.m_Cache.empty());
```
tveasey: nit: `CPPUNIT_ASSERT_EQUAL(false, counters.m_Cache.empty());`

lib/core/unittest/CProgramCountersTest.cc
```diff
@@ -0,0 +1,386 @@
...
+            inserter.toXml(origStaticsXmlSubset);
+        }
+
+        {
+            ml::core::CRapidXmlParser parser;
+            CPPUNIT_ASSERT(parser.parseStringIgnoreCdata(origStaticsXmlSubset));
+            ml::core::CRapidXmlStateRestoreTraverser traverser(parser);
+            CPPUNIT_ASSERT(traverser.traverseSubLevel(
+                &ml::core::CProgramCounters::staticsAcceptRestoreTraverser));
+        }
```
tveasey: This and previous block get repeated quite a lot, maybe factor out into local persist and restore functions?

lib/core/unittest/CProgramCountersTest.cc
```diff
@@ -0,0 +1,386 @@
...
+    LOG_DEBUG(<< output);
+
+    // Test persistence/restoration of a subset  of counters
+    auto testCounterSubset = [&](const ml::counter_t::TCounterTypeSet& counterSet) {
+
+        // Register interest in a subset of counters
+        ml::core::CProgramCounters::registerProgramCounterTypes(counterSet);
+
+        // Check that a save/restore with all zeros is Ok
+        for (int i = 0; i < ml::counter_t::E_LastEnumCounter; i++) {
```
tveasey: nit: coding guidelines are we should always use prefix form unless you really need the returned value (and elsewhere).

include/core/CProgramCounters.h
```diff
@@ -0,0 +1,224 @@
...
+#include <stdint.h>
+
+class CProgramCountersTest;
+
+namespace ml {
+namespace counter_t {
+
+//! Changing the order of these enumeration values will corrupt persisted model
+//! state, so don't.  Any new counters should be added in the penultimate
+//! position in the enum, immediately before E_LastEnumCounter.
```
hendrikmuhs: this can be avoided by explicitly assigning values to the enum values, especially if the enum is used for persistence/communication

Also think about using a scoped enum.

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,192 @@
...
+void CProgramCounters::staticsAcceptPersistInserter(CStatePersistInserter& inserter) {
+    // As the analytics thread could be updating counters while this method is running, in order
+    // to guarantee that consistent counters get persisted for a background persistence task we must not persist
+    // live values. Instead, we access a cached set of self consistent counters for persistence.
+
+    auto staticsAcceptPersistInserter = [&](const auto& container) {
+        int i = 0;
+        for (const auto& counter : container) {
+            inserter.insertValue(KEY_TAG, i);
+            inserter.insertValue(VALUE_TAG, counter);
```
hendrikmuhs: if the counter is `0` - could we skip it?

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,192 @@
...
+                LOG_ERROR(<< "Invalid key value in " << traverser.value());
+                return false;
+            }
+        } else if (name == VALUE_TAG) {
+            if (CStringUtils::stringToType(traverser.value(), value) == false) {
+                LOG_ERROR(<< "Invalid counter value in " << traverser.value());
+                return false;
+            }
+            // Only restore counter values the current program has registered interest in. Set all other values to zero.
+            // If no such limited set has been registered the all values are restored.
```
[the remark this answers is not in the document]
hendrikmuhs: I think it makes sense to skip them. Assuming in version X you have counter A, you upgrade to X+1 but X+1 has removed counter A and replaced it with B and C, we would forever carry counter A with the model in all future versions. That effectively also disallows reusing the spot of A (after 2 major version cycles).

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,192 @@
...
+}
+
+std::ostream& operator<<(std::ostream& o, const CProgramCounters& counters) {
+    rapidjson::OStreamWrapper writeStream(o);
+    TGenericLineWriter writer(writeStream);
+
+    writer.StartArray();
+
+    // If the application has not specified a limited set of (counters using registerProgramCounterTypes) then print the entire set
+    // Take care to print in definition order
```
hendrikmuhs: could we skip counters with value `0`?

lib/core/unittest/CProgramCountersTest.cc
```diff
@@ -0,0 +1,386 @@
...
+
+    virtual void shutdown() {}
+
+    int m_I;
+    int m_N;
+};
+
+} // namespace
+
+CppUnit::Test* CProgramCountersTest::suite() {
```
hendrikmuhs: would be good to test some malicious activity[*], counters that do not exist etc, depends of course on removing the abort. 

[*] future versions, mixed version clusters, ... there might be valid reasons counters disappear or show up

include/core/CProgramCounters.h
```diff
@@ -0,0 +1,224 @@
...
+class CORE_EXPORT CProgramCounters : private CNonCopyable {
+private:
+    using TCounter = std::atomic_uint_fast64_t;
+    using TCounterArray = std::array<TCounter, counter_t::E_LastEnumCounter>;
+    using TCounterDefinitionArray = std::array<SCounterDefinition, counter_t::E_LastEnumCounter>;
+    using TUInt64Vec = std::vector<uint64_t>;
+
+public:
+    //! Singleton pattern
+    static CProgramCounters& instance();
```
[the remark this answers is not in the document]
droberts195: The `m_Counters` member is not static, so the class does need to be a singleton rather than a non-instantiatable grouping of all `static` functionality.

I think the lack of usage of `instance()` outside of the class itself is to reduce verbosity when incrementing the counters, so I would be in favour of dropping this method.  But then it's weird to have a singleton without an `instance()` method, so there should be a comment to say that it's to reduce verbosity.

lib/core/CProgramCounters.cc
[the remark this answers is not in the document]
tveasey: I'm not sure about repurposing counters: there is only one atomic int per counter for the program so we would have to add an awful lot for this to be any sort of overhead. If we just append new counters and never remove then we have no state upgrade issues around this. If we really don't want to round trip to state then we can only write out the values the program registers.

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,192 @@
...
+CProgramCounters::TCounter& CProgramCounters::counter(counter_t::ECounterTypes counterType) {
+    if (counterType >= counter_t::E_LastEnumCounter) {
+        LOG_ABORT(<< "Bad index " << counterType);
+    }
+    return ms_Instance.m_Counters[counterType];
+}
+
+CProgramCounters::TCounter& CProgramCounters::counter(size_t index) {
+    if (index >= ms_Instance.m_Counters.size()) {
+        LOG_ABORT(<< "Bad index " << index);
```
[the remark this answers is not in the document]
tveasey: This can't be completely ignored: this has to return a writable reference, but a catch all counter for this case should do the trick. Alternatively, this could be reworked as an increment function and then truly ignored, but that is a bigger change.

lib/core/CProgramCounters.cc
tveasey: Also, perhaps only write out counter the program has registered?

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,192 @@
...
+CProgramCounters::CProgramCounters()
+    : m_Counters(), m_Cache(), m_ProgramCounterTypes() {
+}
+
+CProgramCounters& CProgramCounters::instance() {
+    return ms_Instance;
+}
+
+CProgramCounters::TCounter& CProgramCounters::counter(counter_t::ECounterTypes counterType) {
+    if (counterType >= counter_t::E_LastEnumCounter) {
```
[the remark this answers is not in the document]
droberts195: You can still have values that don't correspond to a single enum value - if a scoped enum only has constants `e1 = 2` and `e2 = 4` then 6 will be considered valid even though it doesn't correspond to `e1` or `e2` and is greater than the highest single enum value.  So I don't think scoped enums completely solve this problem.

My suggestion to leave gaps for future expansion also makes this validation harder, so that's an argument against leaving gaps.

lib/core/CProgramCounters.cc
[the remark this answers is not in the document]
droberts195: Since the method returns `CProgramCounters::TCounter&` we'll have to return _something_ if we ignore out-of-bounds errors.  Similar to what @tveasey said in https://github.com/elastic/ml-cpp/pull/369/files#r252367258 I would suggest having a "bad counter" member of the class that's separate to the array of valid counters.  We can log a warning whenever this is accessed, but not abort the process.

lib/core/CProgramCounters.cc
droberts195: One thing to confirm before doing this is what the affect on BWC would be.  Would 6.7 tolerate restoring an incomplete set of counters?

lib/core/CProgramCounters.cc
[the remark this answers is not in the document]
droberts195: > after 2 major version cycles

There's currently nothing to stop someone opening a 5.5 job in 8.0 without having opened it in 6.x or 7.x in between.  We should probably add this (in a separate PR).

include/core/CProgramCounters.h
tveasey: The only slight downside of this is it forces us to have a much larger array at the start since we use direct addressing based on the enum value. I think we have to keep this so the lookup remains cheap.

hendrikmuhs: +1,

An argument against reserving ranges is the use of an array for the counters, so we would space for unused counters between the ranges.

So simple increment could work and as we explicitly set the value, you can group the code the way you like it. The highest value could be a constant outside of the enum. Programming errors like forgetting to update this constant for the highest enum, double assigning the same value, could be mitigated with a unit test which loops over the definitions. Maybe thats even an argument to not allow any holes.

lib/core/CProgramCounters.cc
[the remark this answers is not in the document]
hendrikmuhs: what I meant was that scoped enums force you to be explicit and disallow some of the implicit int behavior, but they are not perfect, I agree.

include/core/CProgramCounters.h
droberts195: True, that makes two downsides to leaving gaps then, as validation complexity is another problem then.  So maybe we should stick with the way it's done now.  We just need a policy on whether the names are ordered numerically or grouped by the process they go with.  For example, if we add a new `TSAD` counter after we've added the counters for "new analytics", should it go next to the existing `TSAD` names or at the end of the list?

Either way, I think it's safest to assign every single name an explicit value.

lib/core/unittest/CProgramCountersTest.cc
```diff
@@ -65,181 +37,187 @@ void CProgramCountersTest::testCounters() {
...
+    for (int i = 0; i < 0x1000000; ++i) {
+        ++counters.counter(TEST_COUNTER);
     }
     CPPUNIT_ASSERT_EQUAL(TCounter(0x1000000), counters.counter(TEST_COUNTER));
 }
 
+void CProgramCountersTest::testUnknownCounter() {
+    ml::core::CProgramCounters& counters = ml::core::CProgramCounters::instance();
+
+    const char* logFile = "test.log";
```
hendrikmuhs: as far as I remember there is some temporary directory to be used for unit tests: `ml::test::CTestTmpDir::tmpDir()`

edsavage: Thanks Hendrik. I'm not sure that making use of such a temporary directory is entirely feasible in this case. The logging configuration file would also require updating with the temporary path name which would only be possible by way of an environment variable. I think on the face of it that would be less attractive than the current solution.

There are a couple of other precedents setting up the logging in this way, see e.g. CAnomalyJobTest and PoissonMeanConjugateTest.

include/core/CProgramCounters.h
```diff
@@ -0,0 +1,288 @@
...
+    //! Post increment is not supported - to enforce best practice
+    //! Implicit conversion to/from uint64_t is also allowed.
+    //!
+    class CORE_EXPORT CCounter {
+    public:
+        CCounter() : m_Counter(0) {}
+
+        explicit CCounter(std::uint64_t counter) : m_Counter(counter) {}
+
+        CCounter& operator=(uint64_t counter) {
```
tveasey: nit: std::uint64_t and others

lib/core/CProgramCounters.cc
```diff
@@ -0,0 +1,213 @@
...
+        // can only happen when restoring state that contains one or more counters that are unknown to this version of the application.
+        // As this doesn't indicate a problem with the analytics in the running application we simply log a warning.
+        // A dummy counter is returned in which to store the unknown counter.
+        LOG_WARN(<< "Bad index " << counterPos);
+        return ms_Instance.m_DummyCounter;
+    }
+    return ms_Instance.m_Counters[counterPos];
+}
+
+CProgramCounters::TCounter& CProgramCounters::counter(size_t index) {
```
tveasey: nit: couldn't this just call overload with enum or vice verse?

lib/core/unittest/CProgramCountersTest.cc
hendrikmuhs: nit: maybe you can add a code comment that `test.log` is linked to `testLogErrorsLog4cxx.properties` to document the reason for later

lib/core/CProgramCounters.cc
edsavage: Will do, thanks Tom.

PR #371, opened 2019-02-04
[FEATURE][ML] Per column standardisation prior to outlier detection

tveasey:
This adds in standardisation of column(feature) values prior to outlier detection (we prefer standardisation to min-max normalisation because it is more resilient to extreme outliers). This will be configurable for the power user (that change will be made in a separate PR focusing on user configuration).

The default behaviour will be to standardise since it guards against obvious misuse when mixing features with very different value ranges.

### DIFF
--- a/lib/maths/unittest/CDataFrameUtilsTest.cc
+++ b/lib/maths/unittest/CDataFrameUtilsTest.cc
@@ -1,0 +1,109 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#include "CDataFrameUtilsTest.h"
+
+#include <maths/CBasicStatistics.h>
+#include <maths/CDataFrameUtils.h>
+
+#include <test/CRandomNumbers.h>
+
+#include <boost/filesystem.hpp>
+
+#include <vector>
+
+using namespace ml;
+
+void CDataFrameUtilsTest::testStandardizeColumns() {
+
+    using TDoubleVec = std::vector<double>;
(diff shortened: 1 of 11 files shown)


lib/maths/unittest/CDataFrameUtilsTest.cc
```diff
@@ -0,0 +1,109 @@
...
+            frame->finishWritingRows();
+
+            CPPUNIT_ASSERT(maths::CDataFrameUtils::standardizeColumns(threads, *frame));
+
+            bool passed{true};
+            frame->readRows(1, [&](core::CDataFrame::TRowItr beginRows,
+                                   core::CDataFrame::TRowItr endRows) {
+                for (auto row = beginRows; row != endRows; ++row) {
+                    for (std::size_t j = 0; j < row->numberColumns(); ++j) {
+                        double mean{maths::CBasicStatistics::mean(moments[j])};
```
dimitris-athanasiou: I wonder if we should also have a test with few, simple numbers so we can assert on the result directly instead of using the same code the function we're testing is using.

tveasey: I see this as more testing that the operations applied directly to the values we expect to be in the data frame are equivalent. The test code is not going via the data frame so all code related to interacting with the data frame is tested, including threading and partitioning the values up to support this.

The question is then do we also want to retest `maths::CBasicStatistics` here, since (amongst other things) it is unit tested with basic cases? I'm not sure this is worthwhile.

However, there is also a couple of invariants which could be checked, i.e. that the mean should be zero and the variance should be one, when computed on the standardised column values. These could be included as another source of "independent" asserts.

tveasey: [f94cfa7](https://github.com/elastic/ml-cpp/pull/371/commits/f94cfa7c5ab563740d07e8472deea1f687a457d5)

PR #372, opened 2019-02-05
Add type to analytics results

dimitris-athanasiou:


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -257,12 +261,15 @@ void CDataFrameAnalyzer::writeResultsOf(const CDataFrameAnalysisRunner& analysis
     using TRowItr = core::CDataFrame::TRowItr;
     m_DataFrame->readRows(numberThreads, [&](TRowItr beginRows, TRowItr endRows) {
         for (auto row = beginRows; row != endRows; ++row) {
+            outputWriter.StartObject();
+            outputWriter.String(ROW_RESULTS);
```
tveasey: What's the rationale for nesting here? It seems more natural to have this as an extra member of the existing object to me.

dimitris-athanasiou: The reason is that the parser on the java side needs to know the field name to know which parser to use. Thus, c++ outputs results that may look like this:

```
{
  "row_results": {
    "checksum": 341343,
   "results": {"outlier_score": 0.04}
  }
}
```

or this:

```
{
  "progress": 0.42
}
```

and java can know reading the field name which parser to use (as these will be parsed into different objects).

We employ the same solution for autodetect.

PR #382, opened 2019-02-11
[ML] Correctly write field names for data frame analytics results

dimitris-athanasiou:
The function `Key()` was used accidentally on the JSON writer when
`String()` should be used for field names. This was causing an
assertion to trigger.


dimitris-athanasiou:
Closing as the actual problem ended up being different. Will raise in a new PR.

PR #384, opened 2019-02-11
[ML] Defend against data frame spec not being an object

dimitris-athanasiou:
There is a test that passes input to the data frame
specification that is not a JSON object. The constructor
is not guarding against that and proceeds into calling
`HasMember` which then triggers the assertion that the
document is not an object.


droberts195:
Actually the format checker found a problem:

```
A format error has been detected within the following files:
lib/api/CDataFrameAnalysisSpecification.cc
```

Please run `clang-format` on this file.

PR #388, opened 2019-02-12
[ML] Fix race condition in test assertion

tveasey:
We could conceivably set the task to finished between checking the loop exit condition and checking the progress is less than one. This would cause the assertion to trigger. This has been observed in CI.


lib/api/unittest/CDataFrameAnalysisSpecificationTest.cc
```diff
@@ -229,16 +229,12 @@ void CDataFrameAnalysisSpecificationTest::testRunAnalysis() {
         CPPUNIT_ASSERT(runner != nullptr);
 
         double lastProgress{runner->progress()};
-        for (;;) {
+        while (runner->finished()) {
```
droberts195: Should it be `== false`?

tveasey: doh! Committing without running the test

tveasey:
retest

PR #389, opened 2019-02-12
[ML] Output data frame analytics progress

dimitris-athanasiou:
This commit implements progress reporting for the analysis.
We poll the runner for progress and we output progress as
a result type.

### DIFF
--- a/include/api/CDataFrameAnalyzer.h
+++ b/include/api/CDataFrameAnalyzer.h
@@ -9,6 +9,8 @@
 
 #include <api/ImportExport.h>
 
+#include <core/CRapidJsonConcurrentLineWriter.h>
+
 #include <cinttypes>
 #include <functional>
 #include <memory>
@@ -69,7 +71,11 @@ private:
     bool isControlMessage(const TStrVec& fieldValues) const;
     bool handleControlMessage(const TStrVec& fieldValues);
     void addRowToDataFrame(const TStrVec& fieldValues);
-    void writeResultsOf(const CDataFrameAnalysisRunner& analysis) const;
+    void reportProgress(const CDataFrameAnalysisRunner& analysis,
+                        core::CRapidJsonConcurrentLineWriter& writer) const;
+    void writeProgress(double progress, core::CRapidJsonConcurrentLineWriter& writer) const;
+    void writeResultsOf(const CDataFrameAnalysisRunner& analysis,
+                        core::CRapidJsonConcurrentLineWriter& writer) const;
 
 private:
     // This has values: -2 (unset), -1 (missing), >= 0 (control field index).
--- a/include/api/CDataFrameAnalyzer.h
+++ b/include/api/CDataFrameAnalyzer.h
@@ -71,9 +71,9 @@ private:
     bool isControlMessage(const TStrVec& fieldValues) const;
     bool handleControlMessage(const TStrVec& fieldValues);
     void addRowToDataFrame(const TStrVec& fieldValues);
-    void reportProgress(const CDataFrameAnalysisRunner& analysis,
-                        core::CRapidJsonConcurrentLineWriter& writer) const;
-    void writeProgress(double progress, core::CRapidJsonConcurrentLineWriter& writer) const;
+    void monitorProgress(const CDataFrameAnalysisRunner& analysis,
+                         core::CRapidJsonConcurrentLineWriter& writer) const;
+    void writeProgress(int progress, core::CRapidJsonConcurrentLineWriter& writer) const;
     void writeResultsOf(const CDataFrameAnalysisRunner& analysis,
                         core::CRapidJsonConcurrentLineWriter& writer) const;
 
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -37,7 +37,7 @@ const char FINISHED_DATA_CONTROL_MESSAGE_FIELD_VALUE{'$'};
 
 // Result types
 const std::string ROW_RESULTS{"row_results"};
-const std::string PROGRESS{"progress"};
+const std::string PROGRESS_PERCENT{"progress_percent"};
 
 // Row result fields
 const std::string CHECKSUM{"checksum"};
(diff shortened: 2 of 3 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -247,12 +254,32 @@ void CDataFrameAnalyzer::addRowToDataFrame(const TStrVec& fieldValues) {
...
-    // TODO This should probably be a member variable so it is only created once.
-    auto outStream = m_OutStreamSupplier();
-    core::CRapidJsonConcurrentLineWriter outputWriter{*outStream};
+void CDataFrameAnalyzer::reportProgress(const CDataFrameAnalysisRunner& analysis,
+                                        core::CRapidJsonConcurrentLineWriter& writer) const {
+    double progress{0.0};
+    while (analysis.finished() == false) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        double latestProgress{analysis.progress()};
+        if (latestProgress > progress) {
```
tveasey: This could end up with very fine grained progress reporting for long running analyses. Imagine if an analysis takes 1hr reporting every 100ms you're going to get 36000 progress reports in increments of 0.003% of complete depending on the granularity of the progress monitoring in low level code. I'd make this something like if `latestProgress > progress + 0.01`.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -247,12 +254,32 @@ void CDataFrameAnalyzer::addRowToDataFrame(const TStrVec& fieldValues) {
...
+    if (progress < 1.0) {
+        this->writeProgress(1.0, writer);
+    }
+}
+
+void CDataFrameAnalyzer::writeProgress(double progress,
+                                       core::CRapidJsonConcurrentLineWriter& writer) const {
+    writer.StartObject();
+    writer.Key(PROGRESS);
+    writer.Double(progress);
```
tveasey: Maybe round to nearest 0.01? High precision isn't needed here and tends to imply we're measuring more accurately than we are.

include/api/CDataFrameAnalyzer.h
```diff
@@ -69,7 +71,11 @@ class API_EXPORT CDataFrameAnalyzer {
     bool isControlMessage(const TStrVec& fieldValues) const;
     bool handleControlMessage(const TStrVec& fieldValues);
     void addRowToDataFrame(const TStrVec& fieldValues);
-    void writeResultsOf(const CDataFrameAnalysisRunner& analysis) const;
+    void reportProgress(const CDataFrameAnalysisRunner& analysis,
+                        core::CRapidJsonConcurrentLineWriter& writer) const;
+    void writeProgress(double progress, core::CRapidJsonConcurrentLineWriter& writer) const;
```
tveasey: nit: `writeProgress` and `reportProgress` seem very similar to me. How about `reportProgress` |-> `monitorProgress`.

lib/api/CDataFrameAnalyzer.cc
dimitris-athanasiou: I am rounding in the java side. I wasn't sure whether I should write out progress as an integer (percentage) already from c++ or whether it is more flexible to let java decide what to do with it.

tveasey: I don't think it matters too much. The one small advantage I see in doing it all on the C++ side is you can tie the rounding here to the granularity at which you report above so you always get non-zero progress for each progress result written.

If we reckon reporting every 1% is suitable (seems reasonable to me) then as you say you rounding to an integer in the output seems like a nice choice.

dimitris-athanasiou: True, that would minimise the overhead. Perhaps that's simpler and makes more sense for now. We can always change it later. I'll do that.

droberts195: We have a precedent for reporting progress.  It's in `CForecastDataSink::writeStats`, and it writes progress as a proportion rather than a percentage.  So I think that for consistency it would be best if the `progress` field in the JSON doc is still a proportion rather than a percentage.  But there's definitely no need to report increments of less than 1%.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -254,28 +255,30 @@ void CDataFrameAnalyzer::addRowToDataFrame(const TStrVec& fieldValues) {
...
-                                        core::CRapidJsonConcurrentLineWriter& writer) const {
-    double progress{0.0};
+void CDataFrameAnalyzer::monitorProgress(const CDataFrameAnalysisRunner& analysis,
+                                         core::CRapidJsonConcurrentLineWriter& writer) const {
+    // Progress as percentage
+    int progress{0};
     while (analysis.finished() == false) {
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
-        double latestProgress{analysis.progress()};
+        int latestProgress{static_cast<int>(floor(100.0 * analysis.progress()))};
```
tveasey: nit: std::floor

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -254,28 +255,30 @@ void CDataFrameAnalyzer::addRowToDataFrame(const TStrVec& fieldValues) {
...
         if (latestProgress > progress) {
             progress = latestProgress;
             this->writeProgress(progress, writer);
         }
     }
-    if (progress < 1.0) {
-        this->writeProgress(1.0, writer);
+    if (progress < 100) {
```
tveasey: nit: this if block isn't strictly needed. The managing of setting to 1.0 is handled in the exit condition of task runner, see [here](https://github.com/elastic/ml-cpp/blob/master/lib/api/CDataFrameAnalysisRunner.cc#L150) and its call [here](https://github.com/elastic/ml-cpp/blob/master/lib/api/CDataFrameAnalysisRunner.cc#L103).

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -35,6 +37,7 @@ const char FINISHED_DATA_CONTROL_MESSAGE_FIELD_VALUE{'$'};
 
 // Result types
 const std::string ROW_RESULTS{"row_results"};
+const std::string PROGRESS{"progress"};
```
droberts195: Since this is now inconsistent with `progress` from forecasting, maybe call it `progressPercent`.

PR #395, opened 2019-02-18
[ML] Improve performance of k-d tree nearest neighbour search

tveasey:
There was a silly copy of the point class into a temporary in the `CKdTree::nearestNeighbours`. After fixing that creating temporary pairs to (maybe) push onto the heap still added 25% to its runtime so I've refactored to avoid this. Improving the lower bound on the search (to be the distance to point at the intersection of all hyperplanes above the current node in the recursion) gets a further 20% improvement. Finally, the modulo turned out much slower than a (predictable) branch for getting the next coordinate on which to compare and I've made some tweaks to take a few operations off the critical path. All-in-all this gives us around a 6x performance improvement of outlier detection.

### DIFF
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -18,6 +18,7 @@
 #include <maths/CTypeTraits.h>
 
 #include <boost/operators.hpp>
+#include <boost/ref.hpp>
 
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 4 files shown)


tveasey:
retest (failure caused by #396)

tveasey:
retest

include/maths/CKdTree.h
```diff
@@ -442,6 +484,13 @@ class CKdTree {
         f(node);
     }
 
+    //! Get the next coordinate.
+    inline std::size_t nextCoordinate(std::size_t coordinate) const {
```
droberts195: I'm not convinced it's worth specifying `inline` here.  The method is small so the compiler will almost certainly inline it anyway since it has the definition.

tveasey: Equally I don't think it hurts and expresses intention clearly that this shouldn't incur a function call.

tveasey: My inclination is to leave this as annotated inline unless you feel strongly.

PR #396, opened 2019-02-18
[ML] Prevent potential invalid read if worker throws an exception in parallel_for_each

tveasey:
We need to wait on the result of all futures before eventually propagating the (last) exception. Carrying on straightaway meant other threads could still be working on state which had been cleaned up by the stack unwinding.

This caused a rare SIGSEGV in CConcurrencyTest.

### DIFF
--- a/lib/core/Concurrency.cc
+++ b/lib/core/Concurrency.cc
@@ -116,12 +116,28 @@ CExecutor& defaultAsyncExecutor() {
 }
 
 bool get_conjunction_of_all(std::vector<future<bool>>& futures) {
-    return std::accumulate(futures.begin(), futures.end(), true,
-                           [](bool conjunction, future<bool>& future) {
-                               // Don't shortcircuit
-                               bool value = future.get();
-                               return conjunction && value;
-                           });
+
+    // This waits until results are present. If we get an exception we still want
+    // to wait until all results are ready in case continuing destroys state access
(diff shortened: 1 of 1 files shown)


lib/core/Concurrency.cc
```diff
@@ -116,12 +116,28 @@ CExecutor& defaultAsyncExecutor() {
...
-                               bool value = future.get();
-                               return conjunction && value;
-                           });
+
+    // This waits until results are present. If we get an exception we still want
+    // to wait until all results are ready in case continuing destroys state access
+    // which a worker thread reads. We just rethrow the _last_ exception we received.
+    std::exception_ptr e;
+    bool result{std::accumulate(futures.begin(), futures.end(), true,
+                                [&](bool conjunction, future<bool>& future) {
```
droberts195: The [style guide](https://github.com/elastic/ml-cpp/blob/master/STYLEGUIDE.md) says "Default lambda capture modes SHOULD be avoided" so maybe `&e`?  But I'm not sure what we've been doing elsewhere - do we already have lots of places that use just `&`?

tveasey: We do have quite a few. I personally find it worse having a very long capture list than just & (or =). In this case, because it is just this one value I guess maybe it is a little cleaner to capture explicitly. I'm happy to switch this.

PR #397, opened 2019-02-18
[7.1][ML] Output data frame analytics progress (#389)

dimitris-athanasiou:
This commit implements progress reporting for the analysis.
We poll the runner for progress and we output progress as
a result type.


dimitris-athanasiou:
please test this

PR #405, opened 2019-02-20
[ML] Features influencing outliers

tveasey:
This wires in a calculation of which features are most responsible for a point being an outlier. Very briefly, this is based on understanding how the outlier score changes as we change each feature value independently. The greater the change in the score the greater the feature's influence. I'll prepare some separate detailed documentation.

The extra values are written with the label `feature_influence.<feature_name>` in the result document.

Although this wires it through to result writing it is default disabled and not currently possible to enable. I need to finish up supporting user configuration and will make the change to allow it to be optionally enabled and unit test the result writing in a separate PR.

### DIFF
--- a/lib/maths/unittest/COutliersTest.cc
+++ b/lib/maths/unittest/COutliersTest.cc
@@ -114,14 +114,14 @@ void outlierErrorStatisticsForEnsemble(std::size_t numberThreads,
     for (std::size_t t = 0; t < 100; ++t) {
         gaussianWithUniformNoise(rng, numberInliers, numberOutliers, points);
 
-        auto dataFrame = pointsToDataFrame(points);
+        auto frame = pointsToDataFrame(points);
 
         maths::COutliers::SComputeParameters params{numberThreads,
                                                     numberPartitions, false, 0.05};
-        maths::COutliers::compute(params, *dataFrame);
+        maths::COutliers::compute(params, *frame);
 
-        dataFrame->readRows(1, [&scores](core::CDataFrame::TRowItr beginRows,
-                                         core::CDataFrame::TRowItr endRows) {
+        frame->readRows(1, [&scores](core::CDataFrame::TRowItr beginRows,
+                                     core::CDataFrame::TRowItr endRows) {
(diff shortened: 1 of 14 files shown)


lib/maths/unittest/COutliersTest.cc
```diff
@@ -389,6 +392,100 @@ void COutliersTest::testEnsemble() {
...
+            maths::COutliers::compute(params, *frame);
+
+            bool passed{true};
+            frame->readRows(1, [&](core::CDataFrame::TRowItr beginRows,
+                                   core::CDataFrame::TRowItr endRows) {
+                for (auto row = beginRows; row != endRows; ++row) {
+                    if (row->index() == outlierIndexes[0]) {
+                        LOG_DEBUG(<< "x-significance = " << (*row)[3]
+                                  << ", y-significance = " << (*row)[4]);
+                        passed &= (1.0 - (*row)[4] < 0.005);
```
dimitris-athanasiou: Should we also check the influence of the feature which is not outlying is a low score?

tveasey: Currently, I don't actually control the feature influence of inliers.

Given the nature of the calculation these could still be dominated by one feature whose influence is close to 1. This is because we're computing relative influence so renormalise by the sum of individual feature influences. The intention is we don't output the feature influence at all if the outlier score is low.

I could rejig the calculation to impose this constraint, but my inclination is to leave it. As a result this sort of assertion is not possible.

tveasey: That said, it may be possible to make a meaningful assertion about the average behaviour of inlier influences. I'll ponder

tveasey: Discussed this further with @dimitris-athanasiou, his suggestion was to check the other coordinates in these cases, which is sensible. In fact, we can do even better and check that the scores are normalised, which I do as part of [a4f6481](https://github.com/elastic/ml-cpp/pull/405/commits/a4f6481a96f2db3041c5cd48c68d45366d47a32d). This also checks to see that the influences don't show evidence of bias.

Issue #364, opened 2019-01-16
[ML] Error - CTimeSeriesModel@990 Failed to compute P([nan] | weight = [[nan, 0.7091527]], time = 1547594505)


edsavage:
Found in build: 6.3.2

With job config:

```
      "analysis_config" : {
        "bucket_span" : "1h",
        "detectors" : [
          { 
            "detector_description" : "mean(tiemposaltos)",
            "function" : "mean",
            "field_name" : "tiemposaltos",
            "partition_field_name" : "sistema.keyword",
            "detector_index" : 0
          },
          { 
            "detector_description" : "mean(errores)",
            "function" : "mean",
            "field_name" : "errores",
            "partition_field_name" : "sistema.keyword",
            "detector_index" : 1
          },
          { 
            "detector_description" : "mean(transacciones)",
            "function" : "mean",
            "field_name" : "transacciones",
            "partition_field_name" : "sistema.keyword",
            "detector_index" : 2
          }
        ],
        "influencers" : [
          "sistema.keyword"
        ]
      }
```

Errors, that repeat every 10 minutes, have been reported in the logs:

```
[2019-01-15T17:51:46,344][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [monitores] [autodetect/8947] [CDecompositionComponent.cc@147] Failed calculating confidence interval: Error in function boost::math::normal_distribution::normal_distribution: Scale parameter is nan, but must be > 0 !, n = 20.0602, m = nan, sd = nan, confidence = 50
[2019-01-15T17:51:46,344][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [monitores] [autodetect/8947] [CDecompositionComponent.cc@147] Failed calculating confidence interval: Error in function boost::math::normal_distribution::normal_distribution: Scale parameter is nan, but must be > 0 !, n = 13.5703, m = nan, sd = nan, confidence = 50
[2019-01-15T17:51:46,347][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [monitores] [autodetect/8947] [CNormalMeanPrecConjugate.cc@192] Error calculating joint distribution: Bad variance scale nan
[2019-01-15T17:51:46,347][ERROR][o.e.x.m.j.p.l.CppLogMessageHandler] [monitores] [autodetect/8947] [CNormalMeanPrecConjugate.cc@274] Failed to compute probability of less likely samples
[...]

PR #408, opened 2019-02-21
[ML] Handle NaNs when detrending seasonal components

edsavage:
While it is as yet unknown how it can occur, NaN values have been
observed to be present in seasonal components when detrending, resulting
in a chain of errors in the logs as the NaN values percolate up the call
stack.

This PR attempts to detect the NaN values as close as possible to their
source and ensure that they do not unduly affect the analysis.

Relates to #364

### DIFF
--- a/lib/maths/CDecompositionComponent.cc
+++ b/lib/maths/CDecompositionComponent.cc
@@ -15,7 +15,6 @@
 
 #include <maths/CChecksum.h>
 #include <maths/CIntegerTools.h>
-#include <maths/CMathsFuncs.h>
 #include <maths/CSampling.h>
 #include <maths/CSeasonalTime.h>
 
@@ -70,17 +69,8 @@ bool CDecompositionComponent::acceptRestoreTraverser(core::CStateRestoreTraverse
     } while (traverser.next());
 
     if (this->initialized()) {
-        if (CMathsFuncs::isNan(this->valueSpline().mean())) {
-            LOG_ERROR(<< "Calculated mean of value spline is not a number.");
-        } else {
-            m_MeanValue = this->valueSpline().mean();
-        }
-
-        if (CMathsFuncs::isNan(this->varianceSpline().mean())) {
-            LOG_ERROR(<< "Calculated mean of variance spline is not a number.");
-        } else {
-            m_MeanVariance = this->varianceSpline().mean();
-        }
+        m_MeanValue = this->valueSpline().mean();
+        m_MeanVariance = this->varianceSpline().mean();
     }
 
     return true;
@@ -118,17 +108,8 @@ void CDecompositionComponent::interpolate(const TDoubleVec& knots,
                                           const TDoubleVec& variances) {
     m_Splines.interpolate(knots, values, variances, m_BoundaryCondition);
 
-    if (maths::CMathsFuncs::isNan(this->valueSpline().mean())) {
-        LOG_ERROR(<< "Calculated mean of value spline is not a number.");
-    } else {
-        m_MeanValue = this->valueSpline().mean();
-    }
-
-    if (maths::CMathsFuncs::isNan(this->varianceSpline().mean())) {
-        LOG_ERROR(<< "Calculated mean of variance spline is not a number.");
-    } else {
-        m_MeanVariance = this->varianceSpline().mean();
-    }
+    m_MeanValue = this->valueSpline().mean();
+    m_MeanVariance = this->varianceSpline().mean();
 }
 
 void CDecompositionComponent::shiftLevel(double shift) {
@@ -151,16 +132,8 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
 
     // Also check that the provided offset is valid as this is a potential
     // source of NaNs being generated.
-    if (maths::CMathsFuncs::isNan(offset) == false) {
-        LOG_WARN(<< "Input value offset is not a number.");
-        return {0.0, 0.0};
-    }
     if (this->initialized()) {
         double m{this->valueSpline().value(offset)};
-        if (maths::CMathsFuncs::isNan(m)) {
-            LOG_WARN(<< "Calculated mean value is not a number.");
-            return {m_MeanValue, m_MeanValue};
-        }
 
         if (confidence == 0.0) {
             return {m, m};
@@ -169,11 +142,6 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
         n = std::max(n, 1.0);
         double sd{::sqrt(std::max(this->varianceSpline().value(offset), 0.0) / n)};
         if (sd == 0.0) {
-            return {m, m};
-        }
-
-        if (maths::CMathsFuncs::isNan(sd)) {
-            LOG_WARN(<< "Calculated standard deviation value is not a number.");
             return {m, m};
         }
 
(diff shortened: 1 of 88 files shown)


lib/maths/CDecompositionComponent.cc
```diff
@@ -136,6 +137,21 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
             return {m, m};
         }
 
+        // Sanity check that neither m nor sd are NaN values
+        // This is an extremely rare edge case as checks elsewhere prevent NaNs from
+        // being added to the models. As yet it is unknown how NaNs could be returned
+        // from the splines, only that they do occur.
+        // The returned interval will not contribute to the baseline.
+        if (maths::CMathsFuncs::isNan(m)) {
```
droberts195: Is there any chance that `m` could be `nan` when `confidence == 0.0`?  If so then this test would be better done above `if (confidence == 0.0) {` above, to prevent returning `{m, m}` there when `m` is `nan`.

lib/maths/CDecompositionComponent.cc
```diff
@@ -136,6 +137,21 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
...
+            LOG_WARN(<< "Calculated mean value is not a number.");
+            return {0.0, 0.0};
+        }
+
+        if (maths::CMathsFuncs::isNan(sd)) {
+            LOG_WARN(<< "Calculated standard deviation value is not a number.");
+            return {0.0, 0.0};
```
droberts195: `{m, m}` might be a better fallback here than `{0.0, 0.0}` - that's what we're returning elsewhere for edge cases.

lib/maths/CDecompositionComponent.cc
```diff
@@ -136,6 +137,21 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
             return {m, m};
         }
```
droberts195: One way for `m` to be `nan` is if `offset` is `nan` when the function is called.  It could be worth checking if `offset` is `nan` before the line that checks `if (this->initialized()) {`.

lib/maths/CDecompositionComponent.cc
```diff
@@ -136,6 +137,21 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
...
 
+        // Sanity check that neither m nor sd are NaN values
+        // This is an extremely rare edge case as checks elsewhere prevent NaNs from
+        // being added to the models. As yet it is unknown how NaNs could be returned
+        // from the splines, only that they do occur.
+        // The returned interval will not contribute to the baseline.
+        if (maths::CMathsFuncs::isNan(m)) {
+            LOG_WARN(<< "Calculated mean value is not a number.");
+            return {0.0, 0.0};
```
droberts195: Maybe `{m_MeanValue, m_MeanValue}` would be a better fallback here.  That's what's returned when the spline is not initialized.

That implies it would be necessary to defend `m_MeanValue`, but that's probably a good thing anyway.  So maybe search for `m_MeanValue = ` in this file and log and error and refuse to set if the right hand side is `nan`.

lib/maths/CDecompositionComponent.cc
edsavage: Yes, that was my first thought here too @droberts195, but it is possible that both `m` and `sd` are NaN values (which in itself is interesting as it indicates that the NaN values stem from both the linear and the cubic spline types). Returning `{0.0, 0.0}` here effectively means that we return `value` from `CTimeSeriesDecomposition::detrend` - the same as we would if not initialised.

tveasey:
My reservation with this approach is that if a NaN gets into the state then I think as stands we're done for because we get stuck with bad state that we don't remove and NaN cascade. In other areas of the code we trap this and reinitialise as a last resort to flush bad values. It might be that we'd be better off checking for NaN at certain update points and reinitialising in this case. I'd like to think a bit more about this as a possible alternative.

tveasey:
So I think this can be made to work by checking for NaNs in the decomposition components in the following functions:
1. `CTimeSeriesDecompositionDetail::CComponents::addSeasonalComponents`
2. `CTimeSeriesDecompositionDetail::CComponents::addCalendarComponent`
3. `CTimeSeriesDecompositionDetail::CComponents::interpolate`

These would also need to signal to the rest of the modelling infrastructure if a component is removed. We should also discard NaNs passed in `CTimeSeriesDecomposition::addPoint`. For the reasons outlined above, I'd prefer this approach. Let's discuss this a bit offline @edsavage.

edsavage:
> So I think this can be made to work by checking for NaNs in the decomposition components in the following functions:
> 
> 1. `CTimeSeriesDecompositionDetail::CComponents::addSeasonalComponents`
> 2. `CTimeSeriesDecompositionDetail::CComponents::addCalendarComponent`
> 3. `CTimeSeriesDecompositionDetail::CComponents::interpolate`
> 
> These would also need to signal to the rest of the modelling infrastructure if a component is removed. We should also discard NaNs passed in `CTimeSeriesDecomposition::addPoint`. For the reasons outlined above, I'd prefer this approach. Let's discuss this a bit offline @edsavage.

I discussed these comments in detail with Tom and agree that it does seem to be a better approach to fully purge components if NaNs are detected in them. Doing this also precludes the necessity of checking for NaN's on output (as has been done in this PR)

lib/maths/CDecompositionComponent.cc
```diff
@@ -245,6 +254,10 @@ bool CDecompositionComponent::CPackedSplines::acceptRestoreTraverser(
         this->interpolate(knots, values, variances, boundary);
     }
 
+    LOG_DEBUG(<< "knots = " << core::CContainerPrinter::print(knots));
+    LOG_DEBUG(<< "values = " << core::CContainerPrinter::print(values));
+    LOG_DEBUG(<< "variances = " << core::CContainerPrinter::print(variances));
```
tveasey: Do we need this? There is trace logging in `this->interpolate` above and I don't think we want this extra debug logging.

lib/maths/CDecompositionComponent.cc
```diff
@@ -124,8 +125,16 @@ TDoubleDoublePr CDecompositionComponent::value(double offset, double n, double c
     // and variance equal to the sample variance divided by root
     // of the number of samples.
 
+    // We sanity check that neither m nor sd are NaN values
+    // This is an extremely rare edge case as checks elsewhere prevent NaNs from
+    // being added to the models. As yet it is unknown how NaNs could be returned
+    // from the splines, only that they do occur.
+
+    // Also check that the provided offset is valid as this is a potential
+    // source of NaNs being generated.
```
tveasey: This comment appears to be out-of-date.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -2519,6 +2553,32 @@ bool CTimeSeriesDecompositionDetail::CComponents::CCalendar::prune(core_t::TTime
     return m_Components.empty();
 }
 
+bool CTimeSeriesDecompositionDetail::CComponents::CCalendar::removeComponentsWithBadValues(core_t::TTime time) {
+    std::size_t n{m_Components.size()};
+
+    if (n > 1) {
```
tveasey: This condition isn't wanted here.

lib/maths/CDecompositionComponent.cc
tveasey: In fact, are changes to this file left over from before.

PR #415, opened 2019-02-25
[ML] Rework user overrides for data frame analytics plus support overriding additional parameters for outlier detection

tveasey:
This factors out the validation of the data_frame_analyzer command parameters into a standalone parameter reader class.

I've also added in some additional parameters for outlier detection, allowing the user to opt out of computing feature influences, adjust the minimum outlier score at which we'll write influences out and support disabling column standardisation as part of outlier detection.

As a result I've also been able to finish up unit testing the API with feature influence.

### DIFF
--- a/include/api/CDataFrameAnalysisConfigReader.h
+++ b/include/api/CDataFrameAnalysisConfigReader.h
@@ -88,7 +88,7 @@ public:
         //! Get an enum point parameter.
         template<typename ENUM>
         ENUM fallback(ENUM value) const {
-            static_assert(std::is_enum<ENUM>::value, "Type must be an enumeration");
+            static_assert(std::is_enum<ENUM>::value, "ENUM must be an enumeration");
             if (m_Value == nullptr) {
                 return value;
             }
--- a/include/api/CDataFrameAnalysisConfigReader.h
+++ b/include/api/CDataFrameAnalysisConfigReader.h
@@ -1,0 +1,166 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#ifndef INCLUDED_ml_api_CDataFrameAnalysisConfigReader_h
+#define INCLUDED_ml_api_CDataFrameAnalysisConfigReader_h
+
+#include <core/CLogger.h>
+
+#include <api/ImportExport.h>
+
+#include <rapidjson/document.h>
+
+#include <map>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+namespace ml {
+namespace api {
+
+//! \brief Reads and validates parameters for a data frame analysis.
+//!
+//! DESCRIPTION:\n
+//! This wraps up extracting parameter values from a JSON configuration object.
+//! It supports predefining a set of expected parameters. It is expected that
+//! there will be one static reader object per analysis type.
+//!
+//! The read method extracts all the parameters the JSON contains and returns
+//! a collection to get them by name.
+//!
+//! Expected usage is:
+//! \code
+//! static const CDataFrameAnalysisConfigReader reader{[] {
+//!     CDataFrameAnalysisConfigReader theReader;
+//!     theReader.addParameter("foo", CDataFrameAnalysisConfigReader::E_OptionalParameter);
+//!     theReader.addParameter("bar", CDataFrameAnalysisConfigReader::E_RequiredParameter);
+//! }()};
+//!
+//! auto parameters = reader.read(json);
+//! bool foo{parameters["foo"].fallback(true)};
+//! double bar{parameters["bar"].as<double>()};
+//! \endcode
+//!
+//! IMPLEMENTATION:\n
+//! Not understanding a parameter is likely to result in the wrong analysis being
+//! performed so any errors are treated as fatal causing the process to terminate
+//! immediately. For the same reason this treats unexpected parameters as an error.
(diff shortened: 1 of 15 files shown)


lib/api/CDataFrameOutliersRunner.cc
```diff
@@ -123,10 +99,13 @@ void CDataFrameOutliersRunner::writeOneRow(const TStrVec& featureNames,
...
-        this->spec().numberThreads(), this->numberPartitions(),
-        m_ComputeFeatureInfluence, m_ProbabilityOutlier};
-
+    maths::COutliers::SComputeParameters params{this->spec().numberThreads(),
+                                                this->numberPartitions(),
+                                                m_StandardizeColumns,
+                                                static_cast<maths::COutliers::EMethod>(m_Method),
+                                                m_NumberNeighbours,
+                                                m_ComputeFeatureInfluence,
+                                                m_ProbabilityOutlier};
```
benwtrent: `m_ProbabilityOutlier` it seems to me that this would be good to have as a parameter as well. Do we actually use this in our `lof` calculation?

lib/api/CDataFrameAnalysisSpecification.cc
```diff
@@ -70,82 +73,30 @@ CDataFrameAnalysisSpecification::CDataFrameAnalysisSpecification(const std::stri
...
-        for (auto i = document.MemberBegin(); i != document.MemberEnd(); ++i) {
-            if (isValidMember(*i) == false) {
-                HANDLE_FATAL(<< "Input error: unexpected member '" << i->name.GetString()
-                             << "' of analysis specification. Please report this problem.");
-            }
+        m_NumberRows = parameters[ROWS].as<std::size_t>();
+        m_NumberColumns = parameters[COLS].as<std::size_t>();
+        m_MemoryLimit = parameters[MEMORY_LIMIT].as<std::size_t>();
+        m_NumberThreads = parameters[THREADS].as<std::size_t>();
+        m_TemporaryDirectory = parameters[TEMPORARY_DIRECTORY].fallback(boost::filesystem::current_path().string());
```
hendrikmuhs: looks dangerous to me, probably better to enforce it to not accidentally write somewhere.

(FWIW for forecast it only leads to an error if tmp storage is required)

include/api/CDataFrameAnalysisConfigReader.h
```diff
@@ -0,0 +1,166 @@
...
+//! Expected usage is:
+//! \code
+//! static const CDataFrameAnalysisConfigReader reader{[] {
+//!     CDataFrameAnalysisConfigReader theReader;
+//!     theReader.addParameter("foo", CDataFrameAnalysisConfigReader::E_OptionalParameter);
+//!     theReader.addParameter("bar", CDataFrameAnalysisConfigReader::E_RequiredParameter);
+//! }()};
+//!
+//! auto parameters = reader.read(json);
+//! bool foo{parameters["foo"].fallback(true)};
```
hendrikmuhs: nit: what about "default" instead of "fallback"?

lib/api/CDataFrameAnalysisConfigReader.cc
```diff
@@ -0,0 +1,141 @@
...
+        this->handleFatal();
+    }
+    return m_Value->GetBool();
+}
+
+std::size_t CDataFrameAnalysisConfigReader::CParameter::fallback(std::size_t value) const {
+    if (m_Value == nullptr) {
+        return value;
+    }
+    if (m_Value->IsUint() == false) {
```
hendrikmuhs: why not `IsUint64()`?

lib/api/CDataFrameAnalysisConfigReader.cc
```diff
@@ -0,0 +1,141 @@
...
+        this->handleFatal();
+    }
+    return m_Value->GetUint();
+}
+
+double CDataFrameAnalysisConfigReader::CParameter::fallback(double value) const {
+    if (m_Value == nullptr) {
+        return value;
+    }
+    if (m_Value->IsDouble() == false) {
```
hendrikmuhs: according to https://github.com/Tencent/rapidjson/blob/f5f6052c5682b0932da3704059892dc1405f3607/test/unittest/valuetest.cpp#L404 this fails for integers

lib/api/CDataFrameOutliersRunner.cc
tveasey: I agree. We use it in the ensemble as the prior probability something is an outlier.

(It doesn't really enter directly into LOF as a parameter. The sklearn contamination parameter to LOF acts (somewhat equivalently) on a decision_function and not the raw calculation of outlier factors. There are potentially other ways we could use it, but some of these would assume that it is reasonably accurate and it is unlikely that this quantity is accurately known in advance. This might be something we revisit though.)

lib/api/CDataFrameAnalysisSpecification.cc
tveasey: I agree. I've left a TODO to make this a required parameter. We need to update the Java to pass this as soon as that change is made I'll make this change.

include/api/CDataFrameAnalysisConfigReader.h
tveasey: I'd have preferred that, but it is a reserved word.

lib/api/CDataFrameAnalysisConfigReader.cc
tveasey: Oh, that's a pain I'll check for Int64 as well and coerce the type.

tveasey:
Thanks for the reviews. I've addressed or commented on all points now I think. Can you take another look @benwtrent and @hendrikmuhs?

PR #416, opened 2019-02-26
[ML] Ensure simple count detector is updated first in interim result …

### LINKED ISSUE #324: [ML] Unexpected interim results after advancing time into new empty bucket
**How to reproduce**

1. Create a job with simple count detector and a bucket span of 5m
2. Run some data through the job up to the end of a bucket (using the `end` parameter of the start datafeed API)
3. Open the job again (it should have been auto-closed from step 2)
4. Call the flush API:

```
POST _xpack/anomaly_detectors/{job_id}/flush?advance_time={time}&calc_interim=true
```

where {time} should be a timestamp into the current bucket. E.g., if `end` was `2018-12-01T00:00:00Z`, {time} should be `2018-12-01T00:00:01Z`

**Observed Behaviour**

If you get the anomaly records, you should see a record which is interim and has an actual value of `0.0`. This shouldn't have been created. Interestingly, calling step 4 with {time} being one millisecond forward makes that record disappear.

Also, this is broken since version `6.4.`

dimitris-athanasiou:
…calc

This commit sorts detectors before caclulating interim results making sure
that the simple count detector goes first. We rely on that being the case
as the simple count model updates the interim results corrector with the
current bucket count. If it does not get updated first, then the bucket
completeness will be thought to be full and interim results may be
created when they should not.

Closes #324

### DIFF
--- a/lib/model/unittest/CSearchKeyTest.cc
+++ b/lib/model/unittest/CSearchKeyTest.cc
@@ -1,0 +1,28 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#include "CSearchKeyTest.h"
+
+#include <model/CSearchKey.h>
+
+using namespace ml;
+using namespace model;
+
+void CSearchKeyTest::testSimpleCountComesFirst() {
+    CSearchKey key1{1};
+    CSearchKey key2{2};
+    CPPUNIT_ASSERT(CSearchKey::simpleCountKey() < key1);
+    CPPUNIT_ASSERT(CSearchKey::simpleCountKey() < key2);
+}
+
+CppUnit::Test* CSearchKeyTest::suite() {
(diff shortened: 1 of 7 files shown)


dimitris-athanasiou:
I have prepared an integration test that captures this scenario on the java side. I'll raise a PR to merge it in after this one is merged.

lib/model/unittest/CSearchKeyTest.cc
```diff
@@ -0,0 +1,28 @@
...
+#include <model/CSearchKey.h>
+
+using namespace ml;
+using namespace model;
+
+void CSearchKeyTest::testSimpleCountComesFirst() {
+    CSearchKey key1{1};
+    CSearchKey key2{2};
+    CPPUNIT_ASSERT(CSearchKey::simpleCountKey() < key1);
+    CPPUNIT_ASSERT(CSearchKey::simpleCountKey() < key2);
```
tveasey: nit: this seems a bit on the weak side. I mean the sort implementation change is clearly correct, but maybe randomly choose from multiple different input parameter values if you're going to unit test, i.e. different `function_t::EFunction`, `useNull`, `model_t::EExcludeFrequent` and field name strings. `test::CRandomNumbers::generateWords` could be used for the last part and the other.

PR #417, opened 2019-02-26
[ML] Syntax-check log trace even when optimised out

droberts195:
We have had a few situations where LOG_TRACE macros
contained syntax errors which were only found by the
nightly debug build because most builds optimise
away these trace macros.

This commit changes the approach for optimising away
LOG_TRACE.  Instead of completely ignoring the contents
of the macro they are used within an anonymous lambda
that is immediately discarded.  The hope then is that
the compiler's dead-code elimination optimisation will
completely remove any generated code from the final
program, leading to the same shipped product we had
before but without the pain of having to fix syntax
errors when building debug binaries.

### DIFF
--- a/include/core/LogMacros.h
+++ b/include/core/LogMacros.h
@@ -21,9 +21,12 @@
 #define LOG_TRACE(message)                                                     \
     LOG4CXX_INFO(ml::core::CLogger::instance().logger(), "" message)
 #elif defined(EXCLUDE_TRACE_LOGGING)
-// When this option is set TRACE logging is expanded to nothing - this avoids
-// the overhead of checking the logging level at all for this low level logging
-#define LOG_TRACE(message)
+// When this option is set TRACE logging is expanded to dummy code that can be
+// eliminated from the compiled program (certainly when optimisation is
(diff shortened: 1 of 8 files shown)


droberts195:
For size of the maths library (optimised but unstripped) we have:

* Before this change: 137301640 bytes
* After this change: 137685848 bytes
* Before this change but with a hack to ignore the `EXCLUDE_TRACE_LOGGING` macro: 143675456 bytes

So it's not a complete no-op, but trace logging is certainly _almost_ entirely being optimised away.

droberts195:
Also the test timings on an old/slow CentOS 7 machine show that there is little/no impact of this change on performance:

* Before this change: 8m49s
* After this change: 8m50s
* Before this change but with a hack to ignore the `EXCLUDE_TRACE_LOGGING` macro: 9m25s

The 9m25s is _not_ with trace logging enabled, just with the effect of all the checks of whether or not trace logging is enabled (which demonstrates why we decided to compile it out for production in the first place).

These timings are not particularly scientific - we'd need multiple runs of each scenario for that - but they were taken on a bare metal machine with no other major activity going on which at least rules out the vagaries of VM performance.

PR #427, opened 2019-02-27
[ML] Remove CRegression and move current implementation to a more specific header

tveasey:
This is in preparation for new regression models for data frame analyses because I don't want a single header file for all this functionality. I also took the opportunity to delete some functionality we no longer use.

Note that I haven't modified the implementation of `CRegression::CLeastSquaresOnline` aside from the indentation and name change.


tveasey:
retest

PR #432, opened 2019-03-05
[FEATURE][ML] Data frame column value quantile estimates

tveasey:
Adds the capability to compute approximate column(/feature) quantile values for a data frame. (This is a useful building block for certain types of regression model.)

### DIFF
--- a/lib/maths/CDataFrameUtils.cc
+++ b/lib/maths/CDataFrameUtils.cc
@@ -12,15 +12,16 @@
 
 #include <maths/CBasicStatistics.h>
 #include <maths/CMathsFuncs.h>
+#include <maths/CQuantileSketch.h>
 
 #include <vector>
 
 namespace ml {
 namespace maths {
+using TRowItr = core::CDataFrame::TRowItr;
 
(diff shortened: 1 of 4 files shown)


lib/maths/CDataFrameUtils.cc
```diff
@@ -74,5 +75,50 @@ bool CDataFrameUtils::standardizeColumns(std::size_t numberThreads, core::CDataF
...
     return frame.writeColumns(numberThreads, standardiseColumns);
 }
+
+bool CDataFrameUtils::columnQuantiles(std::size_t numberThreads,
+                                      const core::CDataFrame& frame,
+                                      const TSizeVec& columnMask,
+                                      TQuantileSketchVec& result,
+                                      TWeightFunction weight) {
+    if (result.size() != columnMask.size()) {
+        return false;
```
droberts195: It's probably a good idea to log an error with the 2 different sizes here (a) to give more information to debug it if it does ever happen in production (I assume it should be impossible if the code works as intended) and (b) to distinguish from the other code path that returns `false` (which I assume is more likely).

PR #433, opened 2019-03-05
[ML] Threading optimisations

tveasey:
Wrapping function calls in a `std::packaged_task` was slightly more than doubling the runtime overhead per task in the thread pool. This moves to using a standard function pointer which can (optionally) hold the promise to which results are written. A nice side effect is we don't actually need a custom wrapper around std::future to unpack the result of function passed to the thread pool, so the code is slightly cleaner.

### DIFF
--- a/include/core/Concurrency.h
+++ b/include/core/Concurrency.h
@@ -11,8 +11,6 @@
 #include <core/CLoopProgress.h>
 #include <core/ImportExport.h>
 
-#include <boost/any.hpp>
-
 #include <algorithm>
 #include <functional>
 #include <future>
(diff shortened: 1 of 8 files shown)


include/core/Concurrency.h
```diff
@@ -146,62 +130,66 @@ using future = concurrency_detail::CTypedFutureAnyWrapper<R>;
...
+}
+};
 
-    // Schedule the task to compute the result.
-    executor.schedule(std::move(task));
+// Schedule the task to compute the result.
+executor.schedule(std::move(task));
 
-    return std::move(result);
+return std::move(result);
```
droberts195: The indenting looks a bit off here and in the 4 lines above - is what clang-format decided?

tveasey: Yes. I've seen this before when using brace initialiser with a lambda. I'll rejig and see if it improves the formatting.

PR #437, opened 2019-03-08
[ML] Improve autodetect logic for persistence

### LINKED ISSUE #393: [ML] Flawed logic for when autodetect writes state/quantiles on graceful close
When the `autodetect` process is shut down gracefully it does the following:

1. Writes the latest quantiles to its output unconditionally.
2. Persists state if at least one input record was observed while it was running.

Both these conditions have problems:

1. If a job is opened and closed with nothing else in between then it unnecessarily writes quantiles on close, causing a renormalization on the Java side.  This leads to problems like elastic/elasticsearch#30300.
2. If a job is opened, time is advanced causing results to be generated, then it is closed, then it does not persist state, thus losing the knowledge that time was advanced over a period in which there was no input data.

Both quantiles and state should be output on graceful close if and only if one or more of the following conditions is true:

1. An input record has been processed.
2. Time has been advanced.

Once this is fixed, state will be output more eagerly than it is now and quantiles less eagerly.

edsavage:
Changed the logic surrounding persistence of both state and quantiles on
graceful shutdown so that persistence only occurs if and only if at
least one input record has been processed or time has been advanced.

closes #393

### DIFF
--- a/lib/api/CFieldDataTyper.cc
+++ b/lib/api/CFieldDataTyper.cc
@@ -336,6 +336,11 @@ bool CFieldDataTyper::persistState(core::CDataAdder& persister) {
 }
 
 bool CFieldDataTyper::isPersistenceNeeded(const std::string& description) const {
+    // Pass on the request in case we're chained
+    if (m_OutputHandler.isPersistenceNeeded(description) == true) {
+        return true;
+    }
+
     if (m_NumRecordsHandled == 0) {
         LOG_DEBUG(<< "Zero records were handled - will not attempt to persist "
                   << description << ".");
--- a/lib/api/CFieldDataTyper.cc
+++ b/lib/api/CFieldDataTyper.cc
@@ -337,7 +337,7 @@ bool CFieldDataTyper::persistState(core::CDataAdder& persister) {
 
 bool CFieldDataTyper::isPersistenceNeeded(const std::string& description) const {
     // Pass on the request in case we're chained
-    if (m_OutputHandler.isPersistenceNeeded(description) == true) {
+    if (m_OutputHandler.isPersistenceNeeded(description)) {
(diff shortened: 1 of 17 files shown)


lib/api/CFieldDataTyper.cc
```diff
@@ -335,6 +335,15 @@ bool CFieldDataTyper::persistState(core::CDataAdder& persister) {
     return this->doPersistState(m_DataTyper->makePersistFunc(), m_ExamplesCollector, persister);
 }
 
+bool CFieldDataTyper::isPersistenceNeeded(const std::string& description) const {
+    if (m_NumRecordsHandled == 0) {
+        LOG_DEBUG(<< "Zero records were handled - will not attempt to persist "
+                  << description << ".");
+        return false;
```
droberts195: In the case of a categorizer chained to an anomaly detector this might not be correct - it's possible that time has been advanced on the anomaly detector.

So this method needs a `// Pass on the request in case we're chained` step like some of the other methods in this class have.  If the chained processor's `isPersistenceNeeded()` returns `true` then this method should return `true`, otherwise it should make its own decision using the code that's there now.

lib/api/CFieldDataTyper.cc
```diff
@@ -336,6 +336,11 @@ bool CFieldDataTyper::persistState(core::CDataAdder& persister) {
 }
 
 bool CFieldDataTyper::isPersistenceNeeded(const std::string& description) const {
+    // Pass on the request in case we're chained
+    if (m_OutputHandler.isPersistenceNeeded(description) == true) {
```
droberts195: nit: although we prefer `== false` to `!`, I think `== true` is unnecessary for methods that clearly return a `bool`

PR #438, opened 2019-03-08
[FEATURE][ML] Support reading/writing a mask of the rows in a data frame

tveasey:
This also moves our bit vector to core and extends it to support `std::bitset` like operations and provide an efficient iterator over one bits. I separated out the change to move to core from the change to extend it, so those changes can be reviewed separately if desired.

[For regression tree training we will be recursively partitioning the training data and we need to maintain a bit set of the rows in each leaf of the tree so we can read them efficiently from the data frame. We need this mask to be space efficient (since we don't want to store them out-of-core). Our packed bit vector implementation is pretty well suited to this task. As the number of leaves increases, which is the worst case for memory usage, the one bits will get sparser and it should approach its optimum compression, i.e. 32 rows per bit.]


include/core/CPackedBitVector.h
```diff
@@ -0,0 +1,232 @@
...
+#include <core/ImportExport.h>
+
+#include <boost/operators.hpp>
+
+#include <cmath>
+#include <cstddef>
+#include <string>
+#include <vector>
+
+#include <cstdint>
```
droberts195: nit: move into alphabetical order with the other C++ headers

PR #447, opened 2019-03-22
[ML] Overhaul concurrent queue to mostly avoid blocking

tveasey:
I've been continuing to look at performance of the new data frame analytics as a background task. This reworks concurrent queue so that it uses an atomic front and back position to pop and push the next item, respectively. Contention only happens when same the position is simultaneously acted on. We use a simple state machine to ensure reads and writes at a position are interleaved. If a write is waiting for a read to finish or vice versa the thread spins, otherwise it waits. There is one modest complication, which is that we must ensure if a thread is going to wait on a read (or write) then it receives a notification from a write (or read) in progress only after it starts to wait. We use the double-checked locking pattern to ensure this.

As a result of this change I've been able to significantly simplify `CStaticThreadPool` since it is as efficient to just use a single queue. I also did some work on `CStaticThreadPoolTest::testSchedulingOverhead` since the work loop was actually getting optimised away. This now tests performance across a range of task sizes.

What does this gain us? The overhead per task for `CStaticThreadPool` for our typical usage pattern drops by a factor of around 7 (testing on bare metal). We also get significant speedups in some other functionality built on top of `CConcurrentQueue`. For example, `CConcurrentWrapperTest::testThreads` and `CConcurrentWrapperTest::testThreadsLowCapacity` speed up by a factor of around 4 and 2 respectively and `CJsonOutputStreamWrapperTest::testConcurrentWrites` speeds up by a factor of 4.


tveasey:
I take your point and I was anticipating some push back on this.

So here is my rationale: 1) I've run into a couple of cases in the code already where I'd like to parallelise something, but doing so requires finer grained tasks than we can currently achieve. This tends to force ones hand so you have to look for parallelism not at a natural level, but at a level which is coarse enough. This creates more complexity in other parts of the code. 2) We pay penalties in complexity in the thread pool to work around the deficiencies of the queue (this all goes). 3) This is a foundational piece which we should invest some effort in.

We should consider this change carefully and I should do some more work to demonstrate correctness. I'll add a discuss for the time being and don't plan to rush this change in.

stevedodson:
The benefit of this seems good long term, but a concern is that we need to spend more time testing and understanding the behaviour of this new concurrent queue before pushing this. 

Currently, there are is a huge development push in many areas to get core features in, and IMO it would be better to implement an optimization (particularly a complex one) once we get the core features in - i.e. postpone pushing until we have evaluated the benefits more. 

Is this a possibility?

tveasey:
Ok, I agree this shouldn't hold up other work, so I'd be happy to delay this until we feel we're ready.

I should note that I added a unit test for correctness with fuzzing of the push and pop schedule which should over multiple integration test runs give us high confidence of correctness. Also, I'd worked through slightly more formally showing correctness after Dave's original comment, which I'd shared offline.

I think at some point it would make sense to cut over to this new version, if we're happy there are no errors. I'd propose committing to master at this point and then leaving the change there for a while to be happy that there are no side effects. I think, given our needs, this will be a positive change and in the short term will open up some additional parallelisation opportunities.

sophiec20:
I would prefer master and the other branches remain more closely aligned. If this is pushed to master now, this change cannot be anything other than short term. After hearing team comments, please postpone.

tveasey:
I'd like to think about reviving this. The last round of experiments I ran on boosted tree training, just switching to this queue saved between 15% and 25% across a range of realistic problems: 100k to 500k examples and 5 to 25 features. I also think it opens up some avenues for finer grained threading that don't work well with the current implementation. One possibility would be to commit this shortly after 7.4 is released. This gives us a reasonably long time for it to be tested internally before the next release. Thoughts?

PR #448, opened 2019-03-25
[FEATURE][ML] Fix a bug in bit vector line scan

tveasey:
We were incorrectly inverting the bits of one of the vectors when performing a line scan on our packed bit vector in an edge case. This is a fix and unit test.


tveasey:
retest

PR #450, opened 2019-04-04
[FEATURE][ML] Framework for regression model hyperparameter optimisation

tveasey:
This implements Bayesian optimisation based on a Gaussian Process. We'll use this to optimise most of the hyperparameters for the regression tree implementation. I plan to wire in the use of this in a separate PR to keep this change smaller. Also, although restarts are trivially parallel, I haven't yet threaded this code because I want to do this in the context of end-to-end performance tuning. There are likely to be a few more tweaks down-the-line, but I think this is ready for review.

### DIFF
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -710,8 +710,8 @@ private:
                   << ", sum square weights = " << W[1]);
 
         double scale{static_cast<double>(m_NumberFolds - 1) / static_cast<double>(m_NumberFolds)};
-        double lambda{scale * std::max(0.5 * (L[0] - L[1]) / (W[1] - W[0]), 0.0)};
-        double gamma{scale * std::max(0.5 * (L[0] - L[1]) / (T[1] - T[0]), 0.0)};
+        double lambda{scale * std::max((L[0] - L[1]) / (W[1] - W[0]), 0.0)};
+        double gamma{scale * std::max((L[0] - L[1]) / (T[1] - T[0]), 0.0)};
 
         if (computeLambda) {
             m_Lambda = computeGamma ? 0.5 * lambda : lambda;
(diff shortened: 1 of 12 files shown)


lib/maths/CBoostedTree.cc
```diff
@@ -693,14 +709,17 @@ class CBoostedTree::CImpl final {
...
-            m_Gamma = std::max((L[0] - L[1]) / (T[1] - T[0]), 0.0);
+        double scale{static_cast<double>(m_NumberFolds - 1) / static_cast<double>(m_NumberFolds)};
+        double lambda{scale * std::max(0.5 * (L[0] - L[1]) / (W[1] - W[0]), 0.0)};
+        double gamma{scale * std::max(0.5 * (L[0] - L[1]) / (T[1] - T[0]), 0.0)};
+
+        if (computeLambda) {
+            m_Lambda = computeGamma ? 0.5 * lambda : lambda;
+        }
+        if (computeGamma) {
+            m_Gamma = computeLambda ? 0.5 * gamma : gamma;
```
benwtrent: This change does not result in the same functionality as before, not sure if this is intentional or not. 

```
if computeLambda == true:
    double lambda{scale * std::max(0.25 * (L[0] - L[1]) / (W[1] - W[0]), 0.0)};
```
Similar situation with `computerGamma` as it will now multiple `0.25` instead of just `0.5` when the respective `*compute` variable is `true`.

tveasey: Thanks Ben, good spot! This was unintended. In practice this probably wouldn't matter, it is just used as a start point for the search for the "best" values for these parameters, but there is no reason for the extra factor of 0.5 so I'll revert.

(This whole area needs revisiting in the context of testing automatic hyperparameter tuning. I suspect it will change somewhat at that point. I had started wiring this in, hence the changes to this class, but stopped because this change was getting too large. Since this is still all on a branch I decided to just leave them in this PR.)

tveasey:
retest

tveasey:
@benwtrent I reverted the change you spotted. I also made a couple of small additional changes after investigating instability of test results on different platforms. Could you take another look, thanks!

PR #451, opened 2019-04-05
[ML] Stop using the ml-linux-test Docker image

droberts195:
This image was pointless (doh!), because the ml-linux-build
image already includes CppUnit.  Since libMlTest.so links
to the test framework the build image will always need to
include the test framework libraries.


droberts195:
The reason I am making this change now is that we need to rebuild our 3rd party libraries with extra hardening flags, and removing this superfluous image first will reduce the number of images that need to be updated.

PR #458, opened 2019-04-10
[FEATURE][ML] Boosted tree hyperparameter tuning

tveasey:
This is a first pass wiring hyperparameter tuning for the principle regularisation parameters of the regression tree.

### DIFF
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -11,6 +11,7 @@
 #include <core/CPackedBitVector.h>
 
 #include <maths/CBasicStatistics.h>
+#include <maths/CBayesianOptimisation.h>
 #include <maths/CDataFrameUtils.h>
 #include <maths/CQuantileSketch.h>
 #include <maths/CSampling.h>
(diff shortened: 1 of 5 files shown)


lib/maths/CBoostedTree.cc
```diff
@@ -1082,11 +1120,76 @@ class CBoostedTree::CImpl final {
...
     //! Capture the current hyperparameter values.
-    SHyperparameters captureHyperparameters() {
-        return {*m_Lambda, *m_Gamma, m_FeatureBagFraction, m_ShrinkageFactor,
-                m_FeatureSampleProbabilities};
+    void captureBestHyperparameters(double loss) {
+        if (loss < m_BestForestTestLoss) {
+            m_BestForestTestLoss = loss;
+            m_BestHyperparameters =
+                SHyperparameters{m_Lambda, m_Gamma, m_FeatureBagFraction,
+                                 m_ShrinkageFactor, m_FeatureSampleProbabilities};
```
benwtrent: Why isn't this using `currentHyperparameters()`?

benwtrent: I think I see the reason why it is not used. `currentHyperparameters()` is the current `tunable` parameters, not the absolute current hyperparameters to be used when building the forest. 

I think it should probably be renamed, it is not obvious that it will not include certain hyperparameters if they are overridden by the user.

tveasey: Yes, I think I agree. I still haven't fully decided what to do with feature weight tuning, although these will be using a different strategy to values returned by `currentHyperparameters()`. I'll try and come up with better naming.

tveasey: I couldn't think of good names for these (bad sign) so I refactored to do the hyperparameter update for a single round into one function; I think this is cleaner. I also factored out cross validation to make the train function as clean as possible. Can you take another look @benwtrent.

PR #459, opened 2019-04-11
[ML] Add the "time" syscall to the syscall filter

droberts195:
This is required with the Oracle Unbreakable
Enterprise Kernel (UEK).  The UEK is not
supported but it's possible some other
supported Linux distro will require the "time"
syscall at some point, so it doesn't hurt to
permit it in our filter.

Fixes elastic/elasticsearch#41101

### DIFF
--- a/lib/seccomp/CSystemCallFilter_Linux.cc
+++ b/lib/seccomp/CSystemCallFilter_Linux.cc
@@ -35,6 +35,7 @@ const std::uint32_t SECCOMP_DATA_NR_OFFSET = 0x00;
 #define SECCOMP_RET_ERRNO 0x00050000U
 #define SECCOMP_RET_ALLOW 0x7fff0000U
 #define SECCOMP_RET_DATA 0x0000ffffU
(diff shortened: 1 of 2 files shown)


lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -35,6 +35,7 @@ const std::uint32_t SECCOMP_DATA_NR_OFFSET = 0x00;
 #define SECCOMP_RET_ERRNO 0x00050000U
 #define SECCOMP_RET_ALLOW 0x7fff0000U
 #define SECCOMP_RET_DATA 0x0000ffffU
+#define SECCOMP_RET_KILL 0x00000000U
```
davidkyle: Remove this please

droberts195: I left it in to avoid having to look it up next time we need to debug.  It doesn't really hurt, and can be removed when all the constants are removed once our minimum build machine has the header containing the constants.

Issue #124, opened 2018-06-13
[ML] Forecast results show non intuitive trends


pheyos:
**Found in version**
- 6.3.0-bc11

**Steps to reproduce**
1. Create a multi metric ML job (e.g. for the farequote dataset with 15m buckets and `mean(responsetime) partition_field_name=airline`)
2. Create a `28d` forecast on the results

**Actual result**
The forecast shows a trend to infinity or zero in an exponential manner, which is sometimes unexpected when looking at the actual data the forecast is based on. And it is a significant change compared to the results of the same forecast in version 6.2.5:

![forecast_comparison_630_625](https://user-images.githubusercontent.com/1945390/41347046-ca5b6842-6f08-11e8-8202-1515a63be36f.png)

**Additional information**
In other datasets with more base data for the forecast (e.g. `cloudwatch`) the same trend can be seen:
![forecast_05](https://user-images.githubusercontent.com/1945390/41347145-0b451b28-6f09-11e8-9216-25797e23ba55.png)

tveasey:
This is definitely undesirable. The usage is a bit unusual: i.e. forecasting multiples of the full data range used to generate the model. Also, these signals are not really interesting from a forecasting perspective: basically constant + (white) noise. However, I'd expect much closer to fixed level extrapolation in this case. This needs understanding and likely indicates a bug. (In 6.2 we wouldn't in fact model a trend at all for such signals, which is causing the difference.)

tsviel:
**Elasticsearch version** (Elastic cloud deployment v6.6.0):

**Description of the problem including expected versus actual behavior**:
Irrational forecasts were made to for given time series data.
Given the data it's easy for the naked eye to see that the forecasts are irrational given the history in the following examples:

Series 1: https://drive.google.com/file/d/1kOwXnQaYp1AcuC9SlZoe_LyQsVryLQd8/view?usp=sharing
Series 2: https://drive.google.com/file/d/1bWqHNXjF_UdLWcWeOYmpCyLkBYgABoK9/view?usp=sharing
Series 3: https://drive.google.com/file/d/1tq6prsKzqhsKQNZai7kIkG4mIqSagjhw/view?usp=sharing

**Steps to reproduce**:
 1. Create index

PUT example_data/
{
  "mappings": {
    "doc": {
      "properties": {
        "ts":{
          "type": "date",
          "format": "epoch_second"
        },
        "val": {
          "type": "long"
        },
        "series_id": {
          "type": "keyword"
[...]

PR #463, opened 2019-04-13
[ML] Select trend model complexity for forecasting based on evidence

tveasey:
Check the residual variance for candidate trend models and only select more complex models if there is significant evidence they are needed.

This either fixes or improves many of the problem results described in #124. One data set still has issues. This is related to the interaction between detecting level shifts and the trend modelling. I'll look into that as a separate task.

This only affects forecasting.


tveasey:
@droberts195 thanks for the review. I've fixed the failing unit tests and checked integration tests. Can you take another look.

PR #464, opened 2019-04-15
[FEATURE][ML] Rename number_neighours to k

dimitris-athanasiou:


lib/api/CDataFrameOutliersRunner.cc
```diff
@@ -66,7 +66,7 @@ CDataFrameOutliersRunner::CDataFrameOutliersRunner(const CDataFrameAnalysisSpeci
 
     auto parameters = PARAMETER_READER.read(jsonParameters);
     m_StandardizeColumns = parameters[STANDARDIZE_COLUMNS].fallback(true);
-    m_NumberNeighbours = parameters[NUMBER_NEIGHBOURS].fallback(std::size_t{0});
+    m_NumberNeighbours = parameters[K].fallback(std::size_t{0});
```
tveasey: I wonder if it would be better to rename `m_NumberNeighbours` to `m_K` as well for consistency.

dimitris-athanasiou: I think it should be your call! If you think the current name makes more sense for the rest of the code let's leave it as is. If not, I'll change it to `m_K` as well.

tveasey: I was somewhat conflicted, I was hoping to pass the buck :). I quite like having a descriptive name here, but I also think it is good to be consistent and not rename variables if possible.

Looking at `maths::COutliers` I've called it `s_NumberNeighbours` in the parameters data structure, but actually mostly use k or variants.

Although it is a slightly bigger change, I therefore propose we swap to `m_K` here and `s_K` in `COutliers::SComputeParameters`. Then we're more consistent throughout the code base and it matches the API which I also think is good.

PR #467, opened 2019-04-18
[ML] Rename outlier detection param k to n_neighbors

dimitris-athanasiou:


lib/api/CDataFrameOutliersRunner.cc
```diff
@@ -27,7 +27,7 @@ namespace api {
 namespace {
 // Configuration
 const char* const STANDARDIZE_COLUMNS{"standardize_columns"};
-const char* const K{"k"};
+const char* const N_NEIGHBORS{"n_neighbors"};
```
edsavage: For consistency I would have preferred that this was named `NUMBER_NEIGHBOURS` and the associated text was `"number_neighbours"`. In any case the spelling of `neighbour` needs to be consistent, you've used British spelling here and US above.

PR #468, opened 2019-04-23
[ML] Improve residual model selection

tveasey:
This addresses the remaining issues from #124, which were in fact related to poor model selection. These data sets pose problems because none of candidate residual models are a particularly good fit for all values. The outcome is we end up choosing a model with undesirable characteristics. This also interferes with detecting change points correctly.

This PR makes two changes:
1. It limits the amount a model is penalised with values which are not well fit by any of the candidate models,
2. It additionally penalises models whose predicted variance is much larger than the data variance: anomaly detection is sensitive to this parameter and because we consider heavy-tailed distributions we are susceptible to this sort of error.

I've reviewed the result changes across a range of our QA data sets and where it has affected results it is doing better job.


lib/maths/COneOfNPrior.cc
```diff
@@ -322,83 +363,92 @@ void COneOfNPrior::addSamples(const TDouble1Vec& samples,
...
+        minusBics.push_back((status & maths_t::E_FpOverflowed) ? MINUS_INF : minusBic);
+        failed |= (status & maths_t::E_FpFailed);
+
         // Update the component prior distribution.
         model.second->addSamples(samples, weights);
 
-        used.push_back(use);
+        varianceMismatchPenalties.push_back(
+            -m * MAXIMUM_LOG_BAYES_FACTOR *
+            std::max(1.0 - 9.0 * CBasicStatistics::variance(m_SampleMoments) /
```
edsavage: I'm curious about the use of the `9.0` factor here - perhaps replacing with a named constant would aid understanding?

tveasey: This is the maximum relative error in the estimated variance of the model (vs the sample variance) for which we will not penalise it at all. I named this constant in [7c76c7f](https://github.com/elastic/ml-cpp/pull/468/commits/7c76c7f376c367944cea4778e61d7c28f1a9ff59).

tveasey:
retest

PR #469, opened 2019-04-24
[FEATURE][ML] Compute loss improvement in the training loop

tveasey:
This computes the change in training data loss in the training loop (when writing new predictions to the data frame). This will be useful to report progress down-the-line.

Also improve some naming and fix an occasional off-by-one error in training.


tveasey:
retest

PR #470, opened 2019-04-24
[ML] Use hardening options for gcc 7.3 build

droberts195:
This is a follow on to #453

The stack protector, relro and fortify source options
need to be used when building gcc, because we redistribute
two libraries that are built as part of it, namely
libgcc_s.so.1 and libstdc++.so.6.

### DIFF
--- a/build-setup/linux.md
+++ b/build-setup/linux.md
@@ -43,6 +43,7 @@ export CFLAGS='-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2'
 export CXX='g++ -std=gnu++14'
 export CXXFLAGS='-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2'
 export LDFLAGS='-Wl,-z,relro -Wl,-z,now'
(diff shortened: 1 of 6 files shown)


droberts195:
Marked as `>non-issue` because it doesn't need to be release-noted separately to #453.

build-setup/linux.md
```diff
@@ -60,6 +61,7 @@ Unlike most automake-based tools, gcc must be built in a directory adjacent to t
 tar zxvf gcc-7.3.0.tar.gz
 cd gcc-7.3.0
 contrib/download_prerequisites
+sed -i -e 's/$(SHLIB_LDFLAGS)/$(LDFLAGS) $(SHLIB_LDFLAGS)/' libgcc/config/t-slibgcc
```
droberts195: This is the change from [this gcc patch](https://github.com/alibaba/AliOS-Things-Linux-Edition/blob/master/meta-bsp/meta-via/meta-via-arm/recipes-devtools/gcc/gcc-6.3/0046-Link-libgcc-using-LDFLAGS-not-just-SHLIB_LDFLAGS.patch).

PR #472, opened 2019-04-26
[ML] Improve behaviour of outlier detection with when nearly all points are duplicates

tveasey:
We were hitting issues when the sample of points used for computing nearest neighbours contained only duplicates. In particular, we weren't initialising the local reachability density (lrd) for LOF properly and we weren't computing the cumulative density function (cdf) value for the scores properly when aggregating. This fixes both these problems by using the usual lower bound for lrd and lower bounding the scores' coefficient of variation used to compute the cdf.

### DIFF
--- a/include/maths/COutliers.h
+++ b/include/maths/COutliers.h
@@ -276,11 +276,10 @@ private:
             min += result.s_FunctionState;
         }
         if (min.count() > 0) {
-            // Use twice times the maximum "density" at any other point if there
-            // are k-fold duplicates. Note it is possible that all reference
-            // points are duplicates, in which case we need to set their local
-            // reachability density in this loop too. The overwritten densities
-            // are reset in setup.
+            // Use twice the maximum "density" at any other point if there are
+            // k-fold duplicates. Note it is possible that all lookup points are
+            // duplicates, in which case we need to set their local reachability
+            // density in this loop. The overwritten densities are reset in setup.
             for (std::size_t i = 0; i < m_EndAddresses; ++i) {
                 if (m_Lrd[i] == UNSET_DISTANCE) {
                     m_Lrd[i] = 2.0 / min[0];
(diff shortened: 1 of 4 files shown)


include/maths/COutliers.h
```diff
@@ -274,9 +276,12 @@ class CLof final : public CNearestNeighbourMethod<POINT, NEAREST_NEIGHBOURS> {
             min += result.s_FunctionState;
         }
         if (min.count() > 0) {
-            // Use twice the maximum "density" at any other point if there
-            // are k-fold duplicates.
-            for (std::size_t i = m_StartAddresses; i < m_EndAddresses; ++i) {
+            // Use twice times the maximum "density" at any other point if there
```
droberts195: typo: "twice times" -> "two times" or just "twice"

lib/maths/COutliers.cc
```diff
@@ -556,23 +556,26 @@ void CEnsemble<POINT>::CScorer::add(const TMeanVarAccumulator2Vec& logScoreMomen
     // of benchmark sets to make this effective a priori. We also account for the
     // fact that the scores will be somewhat correlated.
 
+    // We lower bound the coefficient of variation (to 1e-4) of the log-normal we
+    // use to compute the score c.d.f.
+    static const double MINIMUM_VARIANCE{CTools::fastLog(1.0 + 1e-8)};
```
droberts195: It might be better to put this outside the method.  It's a non-trivial static initialisation, so will automatically get protected by some locking mechanism on the compilers that support that.  So putting it in the method will cause some locking overhead each time the method is called.

Alternatively, is `CTools::fastLog(1.0 + 1e-8)` so close to `1e-8` that incorporating the call to `CTools::fastLog()` is unnecessary and the line can become simply `static const double MINIMUM_VARIANCE{1e-8}`?

tveasey: I could use the approximation. I left it like this so it is clear(er) where the value comes from, but maybe I could just add to the comment. This shouldn't be a bottleneck, but better not to be needlessly slow.

PR #485, opened 2019-05-10
[FEATURE][ML] Boosted tree improvements plus more unit testing

tveasey:
This improves various parameter initialisations. It also improves weight shrinkage. Finally, it switches to consider both the mean loss and its variation across the folds when choosing the best hyperparameter values. I've also done a lot more work on unit testing. Although there is still some more to do I think this represents a reasonable point to review.

### DIFF
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -82,7 +82,9 @@ public:
         if (this->isLeaf()) {
             return index;
         }
-        return row[m_SplitFeature] < m_SplitValue
+        double value{row[m_SplitFeature]};
+        bool missing{CDataFrameUtils::isMissing(value)};
+        return (missing && m_AssignMissingToLeft) || (missing == false && value < m_SplitValue)
                    ? tree[m_LeftChild].leafIndex(row, tree, m_LeftChild)
                    : tree[m_RightChild].leafIndex(row, tree, m_RightChild);
     }
@@ -100,9 +102,10 @@ public:
 
     //! Split this node and add its child nodes to \p tree.
     std::pair<std::size_t, std::size_t>
-    split(std::size_t splitFeature, double splitValue, TNodeVec& tree) {
+    split(std::size_t splitFeature, double splitValue, bool assignMissingToLeft, TNodeVec& tree) {
         m_SplitFeature = splitFeature;
         m_SplitValue = splitValue;
+        m_AssignMissingToLeft = assignMissingToLeft;
         m_LeftChild = static_cast<std::int32_t>(tree.size());
         m_RightChild = static_cast<std::int32_t>(tree.size() + 1);
         tree.resize(tree.size() + 2);
@@ -112,8 +115,7 @@ public:
     //! Get the row masks of the left and right children of this node.
     auto rowMasks(std::size_t numberThreads,
                   const core::CDataFrame& frame,
-                  core::CPackedBitVector rowMask,
-                  bool assignMissingToLeft) const {
+                  core::CPackedBitVector rowMask) const {
 
         LOG_TRACE(<< "Splitting feature '" << m_SplitFeature << "' @ " << m_SplitValue);
         LOG_TRACE(<< "# rows in node = " << rowMask.manhattan());
@@ -127,7 +129,7 @@ public:
                         std::size_t index{row->index()};
                         double value{(*row)[m_SplitFeature]};
                         bool missing{CDataFrameUtils::isMissing(value)};
-                        if ((missing && assignMissingToLeft) ||
+                        if ((missing && m_AssignMissingToLeft) ||
                             (missing == false && value < m_SplitValue)) {
                             leftRowMask.extend(false, index - leftRowMask.size());
                             leftRowMask.extend(true);
@@ -180,6 +182,7 @@ private:
 private:
     std::size_t m_SplitFeature = 0;
     double m_SplitValue = 0.0;
+    bool m_AssignMissingToLeft = true;
     std::int32_t m_LeftChild = -1;
     std::int32_t m_RightChild = -1;
     double m_NodeValue = 0.0;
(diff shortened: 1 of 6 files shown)


lib/maths/CBoostedTree.cc
```diff
@@ -498,7 +501,7 @@ class CBoostedTree::CImpl final {
     }
 
     void maximumHyperparameterOptimisationRounds(std::size_t rounds) {
-        m_MaximumHyperparameterOptimisationRounds = rounds;
+        m_MaximumOptimisationRoundsPerHyperparameter = rounds;
```
dimitris-athanasiou: Should we also rename the function?

lib/maths/CBoostedTree.cc
```diff
@@ -531,15 +534,19 @@ class CBoostedTree::CImpl final {
...
+        for (std::size_t round = 0; round < numberRounds; ++round) {
             LOG_TRACE(<< "Optimisation round = " << round + 1);
 
             TMeanVarAccumulator lossMoments{this->crossValidateForest(
                 frame, trainingRowMasks, testingRowMasks, recordProgress)};
 
-            this->captureBestHyperparameters(CBasicStatistics::mean(lossMoments));
+            this->captureBestHyperparameters(lossMoments);
 
+            if (std::sqrt(CBasicStatistics::variance(lossMoments)) <
```
dimitris-athanasiou: Could you add a comment why we break here?

lib/maths/CBoostedTree.cc
```diff
@@ -708,9 +719,15 @@ class CBoostedTree::CImpl final {
         m_Gamma = m_GammaOverride.value_or(0.0);
         if (m_EtaOverride) {
             m_Eta = *m_EtaOverride;
+        } else {
+            m_Eta = 1.0 / std::max(10.0, std::sqrt(static_cast<double>(
```
dimitris-athanasiou: A comment that explains why we init those the way we do would be great

lib/maths/CBoostedTree.cc
```diff
@@ -708,9 +719,15 @@ class CBoostedTree::CImpl final {
...
+                                             frame.numberColumns() - 4)));
+            m_EtaGrowthRatePerTree = 1.0 + m_Eta / 2.0;
         }
         if (m_MaximumNumberTreesOverride) {
             m_MaximumNumberTrees = *m_MaximumNumberTreesOverride;
+        } else {
+            m_MaximumNumberTrees = static_cast<std::size_t>(2.0 / m_Eta + 0.5);
```
dimitris-athanasiou: ditto

dimitris-athanasiou: Also, this seems like it could be zero. Would that be fine?

lib/maths/CBoostedTree.cc
```diff
@@ -757,7 +774,7 @@ class CBoostedTree::CImpl final {
         // We allow a large number of trees by default in the main parameter
         // optimisation loop. In practice, we should use many fewer if they
         // don't significantly improve test error.
-        m_MaximumNumberTrees = 500;
+        m_MaximumNumberTrees *= 10;
```
dimitris-athanasiou: would it make sense to bound this to avoid very large numbers?

lib/maths/CBoostedTree.cc
```diff
@@ -1154,14 +1177,23 @@ class CBoostedTree::CImpl final {
             result.emplace_back(std::log(m_Gamma / 10.0), std::log(10.0 * m_Gamma));
         }
         if (m_EtaOverride == boost::none) {
-            result.emplace_back(0.8, 1.0);
+            double rate{m_EtaGrowthRatePerTree - 1.0};
```
dimitris-athanasiou: Could you add a comment to explain what happens here (and all the magic numbers)?

lib/maths/CBoostedTree.cc
```diff
@@ -1226,14 +1263,18 @@ class CBoostedTree::CImpl final {
...
+        LOG_TRACE(<< "lambda* = " << m_Lambda << ", gamma* = " << m_Gamma
+                  << ", eta* = " << m_Eta << ", eta growth rate per tree* = " << m_EtaGrowthRatePerTree
+                  << ", feature bag fraction* = " << m_FeatureBagFraction);
     }
 
     //! Get the number of hyperparameters to tune.
     std::size_t numberHyperparametersToTune() const {
         return (m_LambdaOverride ? 0 : 1) + (m_GammaOverride ? 0 : 1) +
-               (m_EtaOverride ? 0 : 1) + (m_FeatureBagFractionOverride ? 0 : 1);
+               (m_EtaOverride ? 0 : 2) + (m_FeatureBagFractionOverride ? 0 : 1);
```
dimitris-athanasiou: For eta, is the second hyperparam  its growth rate?

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -114,33 +101,385 @@ void CBoostedTreeTest::testPiecewiseConstant() {
             LOG_DEBUG(<< "function moments = " << functionMoments);
             LOG_DEBUG(<< "model prediction error moments = " << modelPredictionErrorMoments);
 
-            // Unbiased...
-            CPPUNIT_ASSERT(maths::CBasicStatistics::mean(modelPredictionErrorMoments) < 0.01);
```
dimitris-athanasiou: Are we now asserting something here?

lib/maths/CBoostedTree.cc
tveasey: It is implicitly bounded by the value of the learn rate. But I think the observation does then apply to the learn rate override, i.e. we should clearly trap that this is greater than zero. I'll apply some sanity checks on the overrides.

lib/maths/CBoostedTree.cc
tveasey: Yes

lib/maths/unittest/CBoostedTreeTest.cc
tveasey: This is an artefact of reworking the tests, i.e. this `CPPUNIT_ASSERT` is compared against a helper function which extracts the statistics the tests assert on.

tveasey:
Thanks for the review @dimitris-athanasiou. Regarding bounding the number of trees I wasn't totally sure whether to impose any sanity checking on the parameter overrides at this level, but in the end I decided it is  probably worth it. This addresses, indirectly, all routes by which this could become v.large. I also added comments where you requested I think. Can you take another look?

PR #489, opened 2019-05-31
[ML] Fix regression in QA data set

tveasey:
One of our QA sets has shown a regression associated with #468. This was triggered by selecting our Poisson process model in conjunction with a periodic event rate. Including the Poisson process at all in such cases is low value and since we don't properly deal with the change in variance associated with rate changes in this case it has undesirable side effects. This change simply removes it from model selection if we start modelling seasonality (we can revisit this if we do any work to better deal with the variance).

I also increased the parameter introduced in #468 (this means we revert slightly more to our old behaviour). Finally, I've rejigged reinitialisation of the multi-bucket model decay rate, which should have been matching the single bucket model.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -1619,6 +1619,9 @@ void CUnivariateTimeSeriesModel::reinitializeStateGivenNewComponent(
     // re-weight so that the total sample weight corresponds to the sample
     // weight the model receives from a fixed (shortish) time interval.
 
+    // We can't properly handle periodicity in the variance of the rate if
+    // using a Poisson process so remove it from model detectio if we detect
+    // seasonality.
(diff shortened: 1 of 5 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -1619,6 +1619,8 @@ void CUnivariateTimeSeriesModel::reinitializeStateGivenNewComponent(
     // re-weight so that the total sample weight corresponds to the sample
     // weight the model receives from a fixed (shortish) time interval.
 
+    m_ResidualModel->removeModels(
```
edsavage: Perhaps add a comment to explain why we're removing Poisson from model selection in this case?

tveasey: I added this comment

Issue #477, opened 2019-05-02
[ML] False positives associated with multi-bucket features


tveasey:
We've had a number of support tickets raised related to false positives associated with multi-bucket features. Some of these are related to not having appropriate visuals. However, it also seems like the default parametrisation may be too sensitive.

There are a couple of avenues (which are not mutually exclusive) to explore:
1. User facing sensitivity controls.
2. A different parameterisation for multi-bucket detection.
3. More aggressively weighting the single vs multi-bucket contribution to the anomaly score.

tveasey:
Item 2 was addresses in #491. Item 1 is an interesting issue in its own right and something which warrants its own issue, as it affects more than just multi-bucket analysis. Item 3 could be another form of user configuration. We already label results with the impact from multiple buckets, so I think this is lower priority. I'm going to raise a separate issue for sensitivity control and close this.

PR #491, opened 2019-06-07
[ML] Reduce false positives associated with the multi-bucket feature

tveasey:
This improves the interaction between multi-bucket feature and the model we maintain attending specifically to periods flagged as anomalous. In particular, we were suppressing updates of this anomaly model if the bucket probability was high but continuing to adjust the overall probability if the multi-bucket feature probability was low. The upshot our scores were inappropriately high.

This change pegs the time used to compute the anomaly duration to the last update time for the anomaly model and also smoothly reduces the impact on our overall result if the bucket value itself is normal.

This should help with #477.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -34,6 +34,7 @@
 
 * Fix an edge case causing spurious anomalies (false positives) if the variance in the count of events
 changed significantly throughout the period of a seasonal quantity. (See {ml-pull}489[#489].)
+* Reduce false positives associated with the multi-bucket feature. (See {ml-pull}491[#491].)
 
 == {es} version 7.2.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,13 +28,18 @@
 
 //=== Regressions
 
+== {es} version 7.3.0
+
+=== Bug Fixes
+
+* Reduce false positives associated with the multi-bucket feature. (See {ml-pull}491[#491].)
+
 == {es} version 7.2.1
 
 === Bug Fixes
 
 * Fix an edge case causing spurious anomalies (false positives) if the variance in the count of events
 changed significantly throughout the period of a seasonal quantity. (See {ml-pull}489[#489].)
-* Reduce false positives associated with the multi-bucket feature. (See {ml-pull}491[#491].)
 
 == {es} version 7.2.0
 
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -112,6 +112,7 @@ double aggregateFeatureProbabilities(const TDouble4Vec& probabilities, double co
 
 const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_5_TAG("6.5");
+const std::string VERSION_7_2_TAG("7.2");
 
 // Models
 // Version >= 6.3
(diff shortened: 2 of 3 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -387,10 +399,15 @@ class CTimeSeriesAnomalyModel {
...
     //! Update the anomaly model with a sample of the current feature vector.
-    void sample(const CModelProbabilityParams& params, core_t::TTime time, double weight);
+    void sample(const CModelProbabilityParams& params, double weight);
 
     //! Compute the probability of the anomaly feature vector.
-    bool anomalyProbability(core_t::TTime time, double& result) const;
+    bool anomalyProbability(double& result) const;
+
+    //! Get the largest probability the model counts as anomalous.
+    double largestAnomalyProbability() const {
```
edsavage: A bit of a nit pick - feel free to disregard! - Rename this method to say `twiceLargestSignificantProbability`?

lib/maths/CTimeSeriesModel.cc
```diff
@@ -573,11 +598,13 @@ bool CTimeSeriesAnomalyModel::acceptRestoreTraverser(const SModelRestoreParams&
...
+    // else we can't upgrade the state of the anomaly model pre 6.5.
+
     return true;
 }
 
 void CTimeSeriesAnomalyModel::acceptPersistInserter(core::CStatePersistInserter& inserter) const {
-    inserter.insertValue(VERSION_6_5_TAG, "");
+    inserter.insertValue(VERSION_7_2_TAG, "");
     if (m_Anomaly) {
         inserter.insertLevel(ANOMALY_6_5_TAG, boost::bind(&CAnomaly::acceptPersistInserter,
```
droberts195: This is no longer restored, so is it worth persisting it?  Or is it a mistake not to restore it?

droberts195:
Also, I'm not convinced it's a good idea to put this into 7.2.1.  This change affects results and 7.2.1 is going to be a patch release that people will upgrade to expecting no change to functionality.  Potentially Cloud clusters could be force-upgraded to 7.2.1 if there turns out to be a security bug in 7.2.0.  Technically this is a bug fix, but if there's some set of circumstances where the anomaly detection is clearly worse after this change then we'll still have a very annoyed user.

Therefore I would say that either we prioritise testing this next Monday and Tuesday and get it into 7.2.0 or else it waits for 7.3.0.

And if it waits for 7.3.0 then the 7.2 constant in the code should be changed.

lib/maths/CTimeSeriesModel.cc
tveasey: I think I'll leave this. It is the probability which is deemed sufficiently anomalous to update this model and although it is currently twice the cutoff probability to output a result there is nothing that requires this, it was just a reasonable value.

lib/maths/CTimeSeriesModel.cc
tveasey: I'd deleted the restore code from the wrong branch. Good catch.

tveasey:
Ok I agree this is on the fence regarding bug fix vs enhancement. I'm happy to delay until 7.3 and we can test more. I've updated the naming accordingly.

docs/CHANGELOG.asciidoc
```diff
@@ -34,6 +34,7 @@
 
 * Fix an edge case causing spurious anomalies (false positives) if the variance in the count of events
 changed significantly throughout the period of a seasonal quantity. (See {ml-pull}489[#489].)
+* Reduce false positives associated with the multi-bucket feature. (See {ml-pull}491[#491].)
```
droberts195: Please move this to the 7.3.0 section.

Crazybus:
jenkins test this please

PR #492, opened 2019-06-07
[ML] Reduce false positives for sum and count functions on sparse data

tveasey:
For count/sum functions for sparse data we automatically switch to modelling the count/sum given the bucket is not empty when the proportion of empty buckets becomes too high. In such cases, we account for the chance a bucket is empty, but we applied the correction to both the single and multi-bucket feature probabilities. The upshot was we were accounting for this term twice in the overall bucket probability and non-empty buckets for sparse data were generating false positives.

This is another change aimed at #477.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -109,10 +109,10 @@ double featureProbabilityGivenBucket(TBool2Vec empty,
     // is non-null for an empty bucket it's value is defined to be zero. This is
     // true for all our current features, i.e. counts and sums.
 
-    auto pOneSide = [&](double pBucketEmpty) {
+    auto pOneSide = [probability](double pBucketEmpty) {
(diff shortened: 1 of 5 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -96,6 +98,51 @@ TCalculation2Vec expand(maths_t::EProbabilityCalculation calculation) {
...
+                                     maths_t::EProbabilityCalculation calculation,
+                                     TDouble2Vec value,
+                                     TDouble2Vec probabilityBucketEmpty,
+                                     double probability) {
+
+    // The value is tested against zero because it is assumed that if a feature
+    // is non-null for an empty bucket it's value is defined to be zero. This is
+    // true for all our current features, i.e. counts and sums.
+
+    auto pOneSide = [&](double pBucketEmpty) {
```
edsavage: nit: perhaps just capture `probability` in this lambda (and the next) rather than using the default lambda capture mode?

PR #495, opened 2019-06-10
[ML] Upgrade the Apache Portable Runtime libraries

droberts195:
In this change:

- libapr is upgraded to version 1.7.0 from version 1.5.2
- libapriconv is upgraded to version 1.2.2 from version 1.2.1
- libaprutil is upgraded to version 1.6.1 from version 1.5.4
- libexpat is upgraded to version 2.2.6 from the (ancient) version that used to be embedded into libaprutil prior to version 1.6.0

libaprutil no longer embeds libexpat, so it has to be built separately.


droberts195:
Jenkins test this please

PR #500, opened 2019-06-13
[FEATURE][ML] Read and store categorical fields

tveasey:
Proper encoding will happen later.

### DIFF
--- a/lib/api/unittest/CDataFrameAnalyzerTest.cc
+++ b/lib/api/unittest/CDataFrameAnalyzerTest.cc
@@ -40,6 +40,20 @@ using TRowItr = core::CDataFrame::TRowItr;
 using TPoint = maths::CDenseVector<maths::CFloatStorage>;
 using TPointVec = std::vector<TPoint>;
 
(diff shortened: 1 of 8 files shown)


lib/api/unittest/CDataFrameAnalyzerTest.cc
```diff
@@ -745,6 +763,80 @@ void CDataFrameAnalyzerTest::testRoundTripDocHashes() {
...
+                    passed &= (expected == (*row)[0]);
+                } else {
+                    core::CFloatStorage expected{static_cast<double>(
+                        api::CDataFrameAnalyzer::MAX_CATEGORICAL_CARDINALITY)};
+                    passed &= (expected == (*row)[0]);
+                }
+            }
+        });
+
+        CPPUNIT_ASSERT(passed);
```
droberts195: If this assertion ever fails it won't give any indication of what went wrong.  Therefore it might be worth logging an error at the point(s) where `expected != (*row)[0]`.

Same in the method above too.

tveasey:
@droberts195 I followed your suggestion. Can you have another look.

PR #502, opened 2019-06-17
[ML] Fix a SIGSEGV in outlier detection

tveasey:
The special case handling for when the requested number of neighbours *exactly* equals the number of points in the k-d tree was broken: it wasn't returning any points. This could cause an invalid read in outlier detection.

This PR fixes the bug in the k-d tree and makes the outlier code defensive if the neighbour set is ever ~~empty~~ too small.

The code isn't released so marking as a non-issue.

### DIFF
--- a/include/maths/COutliers.h
+++ b/include/maths/COutliers.h
@@ -224,6 +224,9 @@ private:
     void add(const POINT& point, const TPointVec& neighbours, TDouble1VecVec&) override {
         // This is called exactly once for each point therefore an element
         // of m_KDistances is only ever written by one thread.
+        if (neighbours.empty()) {
+            return;
+        }
         std::size_t i{point.annotation()};
         std::size_t a(point == neighbours[0] ? 1 : 0);
         std::size_t b{std::min(this->k() + a - 1, neighbours.size() + a - 2)};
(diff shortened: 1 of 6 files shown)


include/maths/COutliers.h
```diff
@@ -529,12 +540,16 @@ class CTotalDistancekNN final
...
-
         auto& score = scores[point.annotation()];
         score.assign(this->computeFeatureInfluence() ? las::dimension(point) + 1 : 1, 0.0);
 
+        if (neighbours.empty()) {
+            return;
+        }
+
+        std::size_t a(point == neighbours[0] ? 1 : 0);
+        std::size_t b{std::min(this->k() + a - 1, neighbours.size() + a - 2)};
```
droberts195: Could `a` ever be 0 when `neighbours.size()` is 1?  If so `neighbours.size() + a - 2` will wrap to 2 ^ 64 - 1.  It looks like this is a possibility that needs to be defended against.  If there's some subtle reason why this can never happen it would be good to add a comment to say why.

tveasey: In practice, `neighbours.size() >= 2`. Of course, this invariant was violated because of a bug in the k-d tree, so if there is some other bug maybe it could be size 1. (Although, I don't think there is currently any bug of that sort.)

In this case, since `this->k() >= 1` the overflow is invisible to the rest of the code since `k  + a - 1` would be small. Handling the empty case seemed like a natural defence to make, since it results in undefined behaviour. My initial reaction is I'm less sure that defending this case is important, but I'll ponder it a bit more.

tveasey: Ok, so thinking about this a bit more I decided to check `neighbours.size() < 2` and let the outlier score be set to zero if this doesn't hold. This is an error case since `neighbours` should contain at least `k + 1` points and `k` should be greater than 0. Defining the score to be zero in this case seems reasonable. There was also inconsistency between whether the score vector was reset to zeros or zeros were appended. In practice the score vectors should be empty when these methods are called so the alternatives should be equivalent, but the former is the more correct. I've made these changes in [f4f6149](https://github.com/elastic/ml-cpp/pull/502/commits/f4f6149117ce2d067c7d61b6df48387b844ec4d3).

PR #504, opened 2019-06-18
[ML] Fix out-of-bounds read

tveasey:
This fixes an oob read causing `COutliersTest::testFewPoints` to fail in a debug build.


tveasey:
I've independently verified this passes on all platforms with #505 merged.

PR #508, opened 2019-06-19
[ML] Replace boost functions by their std equivalents

### LINKED ISSUE #337: [ML] Migrate boost functionality to std alternatives
We have some more functionality we can migrate on master from boost to std:
- [x] boost::array
- [ ] boost::size
- [x] boost::begin, boost::end (boost/range.hpp)
- [x] boost::is_pod
- [x] boost::ref, boost::cref, etc (boost/ref.hpp)
- [x] boost::make_unique
- [x] boost::algorithm::is_sorted
- [x] boost::bind

mlocs:
Following boost functions have been raplaced with their std equivalents:  `boost::array`, `boost::begin`, `boost::end`, `boost::is_pod`, `boost::ref`, `boost::cref`, `boost::reference_wrapper`, `boost::make_unique`, `boost::algorithm::is_sorted`, `boost::bind`.

Due to the limitations of the std::ref implementation, boost::ref cannot be removed where it is used in Input and Output filters (e.g. `lib/core/CCompressOStream.cc` and `lib/core/CStateDecompressor.cc`).

Fixes #337


mlocs:
recheck

elasticcla:
Hi @mlocs, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ml-cpp/pull/508.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

include/core/UnwrapRef.h
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#ifndef INCLUDE_CORE_UNWRAPREF_H_
+#define INCLUDE_CORE_UNWRAPREF_H_
```
droberts195: nit: following the pattern of all our other guards this would be `INCLUDED_ml_core_UnwrapRef_h`

(Note especially `INCLUDED` ending with `D`.)

include/core/UnwrapRef.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <functional>
+
+namespace ml {
```
droberts195: Please can you add a Doxygen comment.  Something along the lines of:

```
//! Temporary substitute for std::unwrap_ref. This will be replaced by
//! std::unwrap_ref once all our compilers support it. The names in this
//! file do not conform to the coding style to ease the eventual transition
//! to std::unwrap_ref.
```

lib/core/unittest/CMemoryUsageTest.cc
```diff
@@ -785,7 +786,7 @@ void CMemoryUsageTest::testDynamicSizeAlwaysZero() {
     CPPUNIT_ASSERT_EQUAL(haveStructPodCompilerSupport, test);
     test = core::memory_detail::SDynamicSizeAlwaysZero<std::pair<int, int>>::value();
     CPPUNIT_ASSERT_EQUAL(true, test);
-    test = boost::is_pod<SFoo>::value;
+    test = std::is_pod<SFoo>::value;
```
droberts195: You can now delete the comment on lines 762 to 770.

Also I imagine now `haveStructPodCompilerSupport` should always be `true`, so you could get rid of that too and replace with literal `true`s.

droberts195:
> Due to the limitations of the std::ref implementation, boost::ref cannot be removed where it is used in Input and Output filters (e.g. lib/core/CCompressOStream.cc and lib/core/CStateDecompressor.cc)

Sorry, I just realised this is why most of the extra places my first `grep` found still exist.  But I think most of the headers that are still included that the second one found don't need to be included now.

mlocs:
recheck

lib/core/unittest/CMemoryUsageTest.cc
mlocs: done

include/core/UnwrapRef.h
mlocs: done

PR #512, opened 2019-06-27
[ML] Do not write model size stats after autodetect no-op

### LINKED ISSUE #394: [ML] Model should defend against times earlier than epoch 0
I saw this in a Java test log:

```
[2019-02-13T11:42:03,268][WARN ][o.e.x.m.d.DatafeedManager] [node-2] [test-lookback-only-given-empty-index] Datafeed lookback retrieved no data
... SKIPPED 9 LINES ...
[2019-02-13T11:42:03,648][WARN ][o.e.t.OutboundHandler    ] [node-2] send message failed [channel: Netty4TcpChannel{localAddress=/127.0.0.1:34349, remoteAddress=/127.0.0.1:57094}]
java.lang.IllegalStateException: Negative longs unsupported, use writeLong or writeZLong for negative numbers [-3600000]
	at org.elasticsearch.common.io.stream.StreamOutput.writeVLong(StreamOutput.java:274) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.xpack.core.ml.job.process.autodetect.state.ModelSizeStats.writeTo(ModelSizeStats.java:156) ~[?:?]
	at org.elasticsearch.common.io.stream.StreamOutput.writeOptionalWriteable(StreamOutput.java:854) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.xpack.core.ml.action.GetJobsStatsAction$Response$JobStats.writeTo(GetJobsStatsAction.java:277) ~[?:?]
	at org.elasticsearch.common.io.stream.StreamOutput.lambda$writeCollection$28(StreamOutput.java:1056) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.common.io.stream.StreamOutput.writeCollection(StreamOutput.java:1075) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.common.io.stream.StreamOutput.writeCollection(StreamOutput.java:1056) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.common.io.stream.StreamOutput.writeList(StreamOutput.java:1063) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
	at org.elasticsearch.xpack.core.ml.action.util.QueryPage.writeTo(QueryPage.java:57) ~[?:?]
	at org.elasticsearch.action.support.tasks.TransportTasksAction$NodeTasksResponse.writeTo(TransportTasksAction.java:441) ~[elasticsearch-6.7.0-SNAPSHOT.jar:6.7.0-SNAPSHOT]
```

However, this did not actually fail any test.

It shows that a model size stats document contained an epoch time of -3600 seconds = -3600000 milliseconds.  The Java side cannot cope with this.  We should change the C++ code so that it does not create such documents.

A request to generate results for times before the epoch could fail the job, which would then cause a test failure on the Java side that we could fix.

droberts195:
Similar to the fix for state and quantiles in #437,
if no input is received and time is not advanced then
there is no need to write model size stats when the
autodetect process exits. Doing this can actually
cause a problem for a job that has never ever seen
any input, as the unnecessary model size stats were
written with a negative timestamp. This change also
adds an extra defensive check to prevent that ever
happening, although the only situation when it is
thought to be possible should be prevented by the
first change.

Fixes #394

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -80,6 +80,12 @@ to the model. (See {ml-pull}214[#214].)
 
 == {es} version 7.0.0-alpha1
 
+== {es} version 6.8.2
+
+=== Bug Fixes
+
+* Don't write model size stats when job is closed without any input {ml-pull}512[#512] (issue: {ml-issue}394[#394])
+
 == {es} version 6.7.2
 
 === Enhancements
--- a/lib/api/CAnomalyJob.cc
+++ b/lib/api/CAnomalyJob.cc
@@ -221,14 +221,14 @@ bool CAnomalyJob::handleRecord(const TStrStrUMap& dataRowFields) {
 
 void CAnomalyJob::finalise() {
     // Persist final state of normalizer iff an input record has been handled or time has been advanced.
-    if (this->isPersistenceNeeded("quantiles state")) {
+    if (this->isPersistenceNeeded("quantiles state and model size stats")) {
         m_JsonOutputWriter.persistNormalizer(m_Normalizer, m_LastNormalizerPersistTime);
+
+        // Prune the models so that the final persisted state is as neat as possible
+        this->pruneAllModels();
+
+        this->refreshMemoryAndReport();
     }
-
-    // Prune the models so that the final persisted state is as neat as possible
-    this->pruneAllModels();
-
-    this->refreshMemoryAndReport();
 
     // Wait for any ongoing periodic persist to complete, so that the data adder
     // is not used by both a periodic periodic persist and final persist at the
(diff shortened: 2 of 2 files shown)


droberts195:
The test for this is in elastic/elasticsearch#43681

droberts195:
retest

droberts195:
retest

PR #514, opened 2019-06-27
[ML] Workaround a bug with std::bind by changing to lambda

valeriy42:
For some reason, if code is compiled with `-g -O0` using g++ 7.3, then several fields within the const reference to `params` get modified to random values. This problem does not occur when using `boost::bind` or replacing `std::bind` with an equivalent lambda expression.  I could not find any relevant reference to an official bug report in `stdlib`.

### DIFF
--- a/lib/model/CForecastModelPersist.cc
+++ b/lib/model/CForecastModelPersist.cc
@@ -16,6 +16,8 @@
 
 #include <model/CAnomalyDetectorModelConfig.h>
 
+#include <boost/bind.hpp>
+
 namespace ml {
 namespace model {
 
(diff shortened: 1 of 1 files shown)


lib/model/CForecastModelPersist.cc
```diff
@@ -128,9 +128,9 @@ bool CForecastModelPersist::CRestore::nextModel(TMathsModelPtr& model,
                         m_ModelParams.distributionRestoreParams(dataType)},
                     m_ModelParams.distributionRestoreParams(dataType)};
 
-                if (traverser.traverseSubLevel(std::bind<bool>(
-                        maths::CModelStateSerialiser(), std::cref(params),
-                        std::ref(model_), std::placeholders::_1)) == false) {
+                auto serialiser_operator = [&params = static_cast<const maths::SModelRestoreParams&>(params), &model_](core::CStateRestoreTraverser& traverser){
```
droberts195: Please use camelCase for local variables, so `serialiserOperator`.

Also, is the cast needed?  I tried building on a Mac with clang++ and it wasn't.

I don't think the format checker will like the long line.  This worked for me with clang++ on Mac (after running through clang-format):

```
                auto serialiserOperator =
                    [&params, &model_](core::CStateRestoreTraverser& traverser_) {
                        return maths::CModelStateSerialiser()(params, model_, traverser_);
                    };
```

I also renamed the argument from `traverser` to `traverser_` to avoid a shadow warning from clang++.

If the cast is necessary with g++ then it might be better to use `const_cast` instead of `static_cast` and put it inside the lambda.  But usually a cast is not necessary to _add_ constness, only to discard constness.

valeriy42: 1) Interestingly, clang-format did not break the line. I did it manually now. 
2) I renamed `traverser` as you suggested. 
3) `const_cast` is used to remove `const`, not to add it, hence use of `static_cast`. I was replacing `std:cref`, hence I performed casting explicitly, but afaik it is not absolutely necessary at this point. So, I removed it.

PR #518, opened 2019-06-27
Model extractor

edsavage:
Added a new executable in `devbin/model_extractor` that is designed to restore part of the model state in a more human digestible format (un-compressed / meaningful tag names) than the standard persistence format used by `autodetect`. This new tool is intended to be used for debug / educational purposes in general but specifically as a means of automating several manual steps key to the python notebooks residing in https://github.com/elastic/ml-cpp-data (which should make production of these notebooks a simpler task).

I've created this PR as a draft initially as this is still very much a work-in-progress but am keen to canvas opinion of the approach taken.

The main points to note are:

 * A new command line option added to autodetect to trigger persistence every N buckets (rather than an interval of time)
 * New persistence tags with meaningful strings have been added into a large proportion of the existing model classes. These are optionally chosen based on the type of inserter passed to the `acceptPersistInserter` methods.
 * The `model_extractor` executable restores an entire `autodetect` persistence dump (or sequence of such dumps) from file or pipe and re-persists the part of the hierarchy of interest in human readable format. Currently we're only interested in the residual models
 * Extra statistics are generated during the re-persistence of residual model state for convenience.
 * Currently the sequence of steps required to perform model extraction is entirely based around the command line. For example:
 ```
 $ ./build/distribution/platform/darwin-x86_64/bin/autodetect --jobid=test --bucketspan=60 --summarycountfield=count --timefield=time --delimiter=, --modelplotconfig=modelplotconfig.conf --fieldconfig=fieldconfig.conf --persist normal_named_pipe --persistIsPipe --bucketPersistInterval=1 < normal.csv >  normal.log
$ ./devbin/model_extractor --input normal_named_pipe --inputIsPipe --output normal_residual_models.json
```
* Some post-processing of the extracted model documents is required to transform them into a more valid format acceptable by e.g. the standard python json module. This is due to the presence of duplicate object names at the same level of the hierarchy. Currently this is done with a `jq` snippet obtained from https://github.com/stedolan/jq/issues/1795#issuecomment-456985766
* The issue with the non-standards compliant JSON output format can now be bypassed entirely by specifying  '--outputFormat=XML'  in the command line arguments. This generates the model state as an XML document wrapped in a JSON object named 'xml' in the output. The convenience of parsing the XML document comes at the expense of verbosity.

### CODE REVIEW

File: lib/api/CAnomalyJob.cc [hunk 1]
```diff
@@ -972,6 +980,19 @@ bool CAnomalyJob::restoreDetectorState(const model::CSearchKey& key,
     return true;
 }
 
+bool CAnomalyJob::persistResidualModelsState(core::CDataAdder& persister,
+                                             ml::core_t::TTime timestamp) {
```
droberts195: We don't usually use the `ml::` qualifier in code that's in the `ml` namespace.

There are many other places where this could be removed.  It's generally only the stuff in `Main.cc` that's not in the `ml::` namespace.

lib/maths/CGammaRateConjugate.cc
```diff
@@ -691,6 +691,17 @@ class CLogMarginalLikelihood : core::CNonCopyable {
 
 } // detail::
 
+const std::string READABLE_OFFSET_TAG("offset");
```
droberts195: It might be less error prone to define the readable and minified tags on the same line.  This could be something like:

```
const TStrStrPr OFFSET_TAG("a", "offset");
```

(where `TStrStrPr` is a `std::pair<std::string, std::string>`)

Then overload `insertValue` so that the first argument can be either a `std::string` or `TStrStrPr`.  This would avoid the need to have all the `readableTags ? READABLE_FOO_TAG : FOO_TAG` constructs as the test for readable tags would move into the `TStrStrPr` overload of `insertValue`.

Alternatively a custom class could be used:

```
const core::CPersistTag OFFSET_TAG("a", "offset");
```

To start with it would basically just be a pair of strings, but would be more future proof if we ever wanted to add another way of creating tags in the future, as we wouldn't need to search and replace a load of `TStrStrPr`s with `core::CPersistTag` at that future time.

lib/maths/CGammaRateConjugate.cc
```diff
@@ -1439,6 +1450,21 @@ void CGammaRateConjugate::print(const std::string& indent, std::string& result)
         return;
     }
 
+    std::string meanStr{"<unknown>"};
+    std::string sdStr{"<unknown>"};
```
droberts195: It might be worth considering if `<unknown>` is the best string to use to represent unknown values.  Is the likely client for reading the values Python?  If so, would some other string be easier for Python to interpret as meaning unknown?

Same in the other places where `<unknown>` is used.

The other thing about this is that using the string literal `<unknown>` in all the places means that:

1. If we ever want to change it we've got to grep for it and
2. There's a call to `strlen` on every initialisation

You could solve these problems by declaring a `const std::string` containing the unknown value.  That will also store the length making copying more efficient and can easily be changed if we decide the primary client is, say, JavaScript rather than Python.

devbin/model_extractor/model_state_parser.py
```diff
@@ -0,0 +1,78 @@
+import json
+import sys
+import sh
+
```
droberts195: I think this file should have:

1. A copyright header.
2. A comment saying what the prerequisites are.  For example, it requires `jq` be installed, which it isn't by default on macOS.  And does it require Python 2 or Python 3 or does it work in both?

devbin/model_extractor/model_state_parser.py
```diff
@@ -0,0 +1,78 @@
...
+                if name != 'multimodal':
+                    if prior['mean'][0] == '<unknown>' or prior['standard_deviation'][0] == '<unknown>':
+                        continue
+                    mean = float(prior['mean'][0])
+                    sd = float(prior['standard_deviation'][0])
+                    print("{name}: mean = {mean:f}, sd = {sd:f}".format(name=name, mean=mean, sd=sd))
+        else:
+            pass
+    except:
+        sys.exit("Error: Cannot parse JSON document '" + str(json_string) + "' Encountered " + str(sys.exc_info()[0]))
```
droberts195: Is this error going to get too big to be useful?  If the JSON is 2MB then this is a 2MB error message.

devbin/model_extractor/model_state_parser.py
```diff
@@ -0,0 +1,78 @@
...
+
+        print ("Reading items from {}".format(fileName))
+
+        with open(fileName) as json_file:  
+            data=json_file.read()
+
+    reformatted_json = jq_reformat(data)
+
+    for line in reformatted_json.splitlines():
+        if line != '0':
```
droberts195: It's probably worth adding a comment to say what the lines consisting of just `0` mean.

bin/autodetect/CCmdLineParser.cc
```diff
@@ -103,7 +104,9 @@ bool CCmdLineParser::parse(int argc,
                         "Optional file to persist state to - not present means no state persistence")
             ("persistIsPipe", "Specified persist file is a named pipe")
             ("persistInterval", boost::program_options::value<core_t::TTime>(),
-                        "Optional interval at which to periodically persist model state - if not specified then models will only be persisted at program exit")
+                        "Optional time interval at which to periodically persist model state - if not specified then models will only be persisted at program exit. (Mutually exclusive with bucketPersistInterval)")
```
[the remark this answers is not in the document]
droberts195: > Tying to the bucket length makes more sense to me since the models change on this time scale.

But remember that this is not greenfield functionality in this PR.  We've done background persistence based on wall clock time for about 5 years now.  The rationale for using wall clock time was that if a lookback churns through 1000 half hour buckets in 20 minutes of wall clock time we probably don't need a background persistence in the middle of that work, but if running in real time then we want lots of periodic persists during the 500 hours it takes to get through 1000 half hour buckets.

This really leads on to the question of whether this functionality is only ever going to be used as an internal debugging tool.  At one time the thinking was that it would be internal only.  End users don't call the C++ process directly, so it doesn't matter if the options are esoteric (although it's still best for the descriptions to be factually accurate).  But if the thinking on exposing this externally has now changed then the external user interface needs more careful thought, including how to protect people from accidentally doing a denial-of-service attack on themselves by for example persisting huge model state every 2 buckets during a 10000 bucket lookback.  Also before exposing this externally things like using `jq` from a Python script would need revisiting...

lib/maths/CPrior.cc
```diff
@@ -304,6 +305,19 @@ std::string CPrior::debug() const {
...
+CPrior::TStrStrPr CPrior::printMarginalLikelihoodStatistics() const {
+    TStrStrPr unknownValuePr{UNKNOWN_VALUE_STRING, UNKNOWN_VALUE_STRING};
+
+    if (this->isNonInformative()) {
+        return unknownValuePr;
+    }
+
+    try {
+        return this->doPrintMarginalLikelihoodStatistics();
+    } catch (...) {}
```
droberts195: I don't think it's good practice to use `catch (...)` except where it's effectively a `finally` clause and the caught exception is rethrown or otherwise propagated.

lib/maths/CPrior.cc
```diff
@@ -304,6 +305,19 @@ std::string CPrior::debug() const {
     return std::string();
 }
 
+CPrior::TStrStrPr CPrior::printMarginalLikelihoodStatistics() const {
+    TStrStrPr unknownValuePr{UNKNOWN_VALUE_STRING, UNKNOWN_VALUE_STRING};
+
+    if (this->isNonInformative()) {
+        return unknownValuePr;
```
droberts195: You can construct a pair from an intializer list, so this line and the other one that returns `unknownValuePr` could be `return {UNKNOWN_VALUE_STRING, UNKNOWN_VALUE_STRING};`.  Then you wouldn't have to construct `unknownValuePr` in the happy day case of a useful mean and standard deviation being available.

include/maths/CClusterer.h
```diff
@@ -147,7 +147,7 @@ class CClusterer : public CClustererTypes {
     //! \name Clusterer Contract
     //@{
     //! Get the tag name for this clusterer.
-    virtual std::string persistenceTag() const = 0;
+    virtual core::TPersistenceTag persistenceTag() const = 0;
```
droberts195: Is there ever a need for this method to return a dynamically constructed tag rather than a static constant tag?  If not then returning a const reference would avoid the need to copy _two_ strings into the return value.

lib/maths/CPrior.cc
tveasey: Also, this method catches all its exceptions so I don't think we need a try here. (Unless you really can't avoid it, say callback into 3rd party code, none of our functions should be throwing exceptions.)

devbin/model_extractor/model_state_parser.py
```diff
@@ -0,0 +1,78 @@
...
+
+    return
+
+if __name__ == '__main__':
+
+    data=''
+    if len(sys.argv) < 2:
+        data = sys.stdin.read();
+    else:
+        fileName = sys.argv[1]
```
Winterflower: Not entirely familliar with the usecase here, so this could be a misplaced suggestion but `argparse` is a builti-in lib in Py3 that can be used to create parsers for cli arguments.

PR #525, opened 2019-06-28
[ML] Add a docker image to perform clang-format in a correct version

valeriy42:
This PR changes the base of the `ml-check-style` docker image from `ubuntu:16.04` to `alpine:3.8`, which already comes with clang in `version 5.0.1` out of the box. This reduces the image size from 1.56GB to 214MB. The version of `ml-check-style` is bumped from 1 to 2.

PR also introduce the script `run_docker_clang_format` which mounts the `$CPP_SRC_HOME` director into the container and performs formatting. Running docker with `-u` option should solve file ownership issues on the mounted directories. I did not test it on macOS though.


droberts195:
We already have a Docker image and scripting for doing this:

* https://github.com/elastic/ml-cpp/blob/master/dev-tools/docker/check_style_image/Dockerfile
* https://github.com/elastic/ml-cpp/blob/master/dev-tools/docker/build_check_style_image.sh
* https://github.com/elastic/ml-cpp/blob/master/dev-tools/docker_check_style.sh

That is then called by the Jenkins job from https://github.com/elastic/ml-cpp/blob/master/dev-tools/jenkins_ci.sh#L66

I don't think you should introduce a parallel system for doing the same thing.  If it's simpler and more performant to use Alpine for this purpose then I don't mind if you create a version 2 of `ml-dev/ml-check-style` image that's based on Alpine.

valeriy42:
I updated your `Dockerfile` and bumped the version.

PR #526, opened 2019-06-28
[7.x][ML] Ignore "persist now" control message if no prior input

droberts195:
Persisting state when there has been no input data:

1. Is a waste of time
2. Causes the model snapshot document to embed an invalid
   model size stats document which causes an error in Elasticsearch

So we should not do this.

Backport of #521


droberts195:
retest

PR #529, opened 2019-06-28
[7.3][ML] Workaround a bug with std::bind by changing to lambda (#514)

valeriy42:
For some reason, if code is compiled with -g -O0 using g++ 7.3, then several fields within the const reference to params get modified to random values. This problem does not occur when using boost::bind or replacing std::bind with an equivalent lambda expression. I could not find any relevant reference to an official bug report in stdlib.


valeriy42:
retest

droberts195:
retest

PR #530, opened 2019-06-28
[7.3][ML] Add a docker image to perform clang-format in a correct version …

valeriy42:
…(#525)

This PR changes the base of the ml-check-style docker image from ubuntu:16.04 to alpine:3.8, which already comes with clang in version 5.0.1 out of the box. This reduces the image size from 1.56GB to 214MB. The version of ml-check-style is bumped from 1 to 2.

PR also introduce the script run_docker_clang_format which mounts the $CPP_SRC_HOME director into the container and performs formatting. Running docker with -u option should solve file ownership issues on the mounted directories. I did not test it on macOS though.


valeriy42:
recheck

valeriy42:
retest

PR #544, opened 2019-07-12
[FEATURE][ML] Boosted tree training checkpointing

valeriy42:
Persist and restore the state of bayesian optimization class.

Is related to elastic/ml-team#130


lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+            newTVector(i) = newVector[i];
+        }
+        vector = std::move(newTVector);
+        return true;
+    };
+    return restorationFunction;
+}
+
+std::function<bool(core::CStateRestoreTraverser&)>
+CBayesianOptimisation::restoreParameterValuePair(TVector& parameters, double& functionValue) {
```
tveasey: If you implement `acceptRestoreTraverser` on `maths::CDenseVector` then you can use `core::CPersistUtils::restore` for this and it should automatically find and use that function.

lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+    auto restoreParameterValuePair = [&parameters, &functionValue,
+                                      this](core::CStateRestoreTraverser& traverser) {
+        do {
+            if (traverser.name() == FUNCTION_PARAMETERS_TAG) {
```
tveasey: In `core/RestoreMacros.h` there are macros for a lot of the boilerplate code around writing restore functions.

lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+bool CBayesianOptimisation::acceptRestoreTraverser(core::CStateRestoreTraverser& traverser) {
+    try {
+
+        // Call name() to prime the traverser if it hasn't started
+        traverser.name();
+        if (traverser.isEof()) {
+            //            m_RestoredStateDetail.s_RestoredStateStatus = E_NoDetectorsRecovered;
+            LOG_ERROR(<< "Expected persisted state but no state exists");
+            return false;
+        }
```
tveasey: I'm not sure this is needed. Basically, the first restore you do inside the loop will catch this error and fail.

lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+    }
+
+    return true;
+}
+
+bool CBayesianOptimisation::restoreSubLevelVector(const std::string& tag,
+                                                  const std::string& name,
+                                                  TVector& vector,
+                                                  core::CStateRestoreTraverser& traverser) {
+    if (traverser.hasSubLevel()) {
```
tveasey: In general, I'm not sure it is worth doing the checking for sublevels explicitly in the code: we don't in any other restore code and it bloats the code a bit. The error will already be detected by `traverser.traverseSubLevel(restoreVector(tag, vector))` which will return false. I know you could argue that it is useful to have the separate error message, but I think on balance it is overkill. From experience, if you get a problem with restore you really need to inspect the whole state document to work out what's wrong.

To me this points to just ditching these wrapper `restoreSubLevel*` functions.

lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+        inserter.insertLevel(FUNCTION_MEAN_VALUES_TAG,
+                             persistFunctionMeanValues(m_FunctionMeanValues));
+    } catch (std::exception& e) {
+        LOG_ERROR(<< "Failed to persist state! " << e.what());
+    }
+}
+
+std::function<void(core::CStatePersistInserter&)>
+CBayesianOptimisation::persistVector(const std::string& tag,
+                                     const CBayesianOptimisation::TVector& vector) const {
```
tveasey: As above have a look at uses of `CPersistUtils`.

lib/maths/CBayesianOptimisation.cc
```diff
@@ -372,6 +398,268 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+            }
+        } while (traverser.next());
+        return true;
+    };
+    return restoreParameterValuePair;
+}
+
+bool CBayesianOptimisation::restoreFunctionMeanValues(CBayesianOptimisation::TVectorDoublePrVec& functionMeanValues,
+                                                      core::CStateRestoreTraverser& traverser) {
+    if (traverser.hasSubLevel()) {
```
tveasey: I wouldn't bother with this (more on rationale in [this](TODO) comment).

include/maths/CBayesianOptimisation.h
```diff
@@ -101,14 +119,44 @@ class MATHS_EXPORT CBayesianOptimisation {
     TVectorDoublePr kernelCovariates(const TVector& a, const TVector& x, double vx) const;
     double kernel(const TVector& a, const TVector& x, const TVector& y) const;
 
+    std::function<void(core::CStatePersistInserter&)>
+    persistFunctionMeanValues(const CBayesianOptimisation::TVectorDoublePrVec& functionMeanValues) const;
+    std::function<void(core::CStatePersistInserter&)>
+    persistVector(const std::string& tag, const CBayesianOptimisation::TVector& vector) const;
+
+    std::function<void(core::CStatePersistInserter&)>
+    persistVector(const std::string& tag, const CBayesianOptimisation::TDoubleVec& vector) const;
```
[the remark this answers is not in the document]
valeriy42: Changed in 4257c26be799234a50e275e76996944026a098c7

lib/maths/CBayesianOptimisation.cc
```diff
@@ -6,20 +6,47 @@
...
 #include <maths/CLbfgs.h>
 #include <maths/CLinearAlgebraEigen.h>
 #include <maths/CLinearAlgebraShims.h>
 #include <maths/CSampling.h>
-#include <maths/CTools.h>
 
 #include <boost/math/distributions/normal.hpp>
+#include <core/CJsonStateRestoreTraverser.h>
+#include <core/CStateDecompressor.h>
+#include <maths/CTools.h>
```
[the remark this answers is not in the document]
valeriy42: Changed in 4257c26be799234a50e275e76996944026a098c7

include/maths/CBayesianOptimisation.h
```diff
@@ -101,14 +119,44 @@ class MATHS_EXPORT CBayesianOptimisation {
     TVectorDoublePr kernelCovariates(const TVector& a, const TVector& x, double vx) const;
     double kernel(const TVector& a, const TVector& x, const TVector& y) const;
 
+    std::function<void(core::CStatePersistInserter&)>
+    persistFunctionMeanValues(const CBayesianOptimisation::TVectorDoublePrVec& functionMeanValues) const;
```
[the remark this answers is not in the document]
valeriy42: Deleted in 912241d2484e7d564c25f3f0d01cd3727c95ee01

valeriy42:
|`include/maths/CBayesianOptimisation.h:59`|
|---|
This is unnecessary

valeriy42:
|`include/maths/CBayesianOptimisation.h:65`|
|---|
Line comment?

valeriy42:
|`lib/maths/CBayesianOptimisation.cc:36`|
|---|
Is this a line comment?

lib/maths/CBayesianOptimisation.cc
valeriy42: I removed the checks and `restoreSubLevel` functions in all places but one, where I need it to have a correct restoration behavior if no points were added to the optimizer yet.

lib/maths/CBayesianOptimisation.cc
valeriy42: Changed in a1f55433acfe120c4b58ba7836f64fc6b2cf535c

lib/maths/CBayesianOptimisation.cc
valeriy42: Here I need this check to make sure that restoring optimizer without any function values works correctly. Do you see any other solution?

lib/maths/CBayesianOptimisation.cc
valeriy42: I copied it from `CAnomalyJob.cc`. It isn't needed here?

lib/maths/CBayesianOptimisation.cc
valeriy42: Good tip! I resolved it in 0685d44abbb315b6fd227460c7dcb9abaf02df1b

include/maths/CBayesianOptimisation.h
```diff
@@ -101,14 +119,44 @@ class MATHS_EXPORT CBayesianOptimisation {
...
+    std::function<void(core::CStatePersistInserter&)>
+    persistVector(const std::string& tag, const CBayesianOptimisation::TVector& vector) const;
```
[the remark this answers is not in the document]
valeriy42: Done in a1f55433acfe120c4b58ba7836f64fc6b2cf535c

valeriy42:
Thank you @tveasey! I implemented you suggestions an this significantly simpified the code. At one place, however, I kept the level checks to be resilient to corner cases. Please let me know what you think.

lib/maths/CBayesianOptimisation.cc
tveasey: My thinking was that this method shouldn't have this sort of top-level functionality since one might expect it to be called as part of restoring state comprising multiple objects.

As with `CAnomalyJob` my inclination would be a `restore` function on `CBoostedTree`, or as we discussed offline, moving this to the factory object.

lib/maths/CBayesianOptimisation.cc
tveasey: I think if this is written/read using core::CPersistUtils it handles empty collections by adding a "dummy" value to the state document (which is the alternative route).

This comment preceded the observation that that functionality could be used instead.

include/maths/CBayesianOptimisation.h
```diff
@@ -8,6 +8,11 @@
 #include <maths/CPRNG.h>
 #include <maths/ImportExport.h>
 
+#include <core/CDataSearcher.h>
+#include <core/CStatePersistInserter.h>
+#include <core/CStateRestoreTraverser.h>
+#include <core/CoreTypes.h>
```
tveasey: nit: as with other code can we keep includes in library dependency order, i.e. core first.

include/maths/CLinearAlgebraEigen.h
```diff
@@ -271,8 +275,58 @@ class CDenseVector : public Eigen::Matrix<SCALAR, Eigen::Dynamic, 1> {
         }
         return seed;
     }
+
+    std::vector<SCALAR> toStdVector() const {
+        std::vector<SCALAR> result;
```
tveasey: Can we reserve here

include/maths/CLinearAlgebraEigen.h
```diff
@@ -271,8 +275,58 @@ class CDenseVector : public Eigen::Matrix<SCALAR, Eigen::Dynamic, 1> {
...
+        CDenseVector<SCALAR> result(vector.size());
+        for (auto element : vector) {
+            result << element;
+        }
+        return result;
+    }
+
+    //! Persist by passing information to \p inserter.
+    static void persist(const CDenseVector<SCALAR>& vector,
+                        core::CStatePersistInserter& inserter) {
```
tveasey: If you implement as non-static method called `acceptPersistInserter` then this magically works with `core::CPersistUtils`, i.e. the following would just work:
```
std::vector<CDenseVector<double>> foo;
core::CPersistUtils::persist(FOO_TAG, foo, inserter);
```

include/maths/CLinearAlgebraEigen.h
```diff
@@ -271,8 +275,58 @@ class CDenseVector : public Eigen::Matrix<SCALAR, Eigen::Dynamic, 1> {
...
+        std::vector<SCALAR> temporaryVector;
+        for (int i = 0; i < vector.size(); ++i) {
+            temporaryVector.push_back(vector(i));
+        }
+        inserter.insertValue(DENSE_VECTOR_TAG, core::CPersistUtils::toString(temporaryVector));
```
tveasey: This should be using toStdVector no?

include/maths/CLinearAlgebraEigen.h
```diff
@@ -271,8 +275,58 @@ class CDenseVector : public Eigen::Matrix<SCALAR, Eigen::Dynamic, 1> {
...
+    }
+
+    //! Populate the object from serialized data
+    static bool restore(const std::string& tag,
+                        CDenseVector<SCALAR>& vector,
+                        core::CStateRestoreTraverser& traverser) {
```
tveasey: As above, if you implement this as non-static acceptRestoreTraverser this magically works with `core::CPersistUtils` also again this could use `fromStdVector`.

include/maths/CLinearAlgebraEigen.h
```diff
@@ -9,6 +9,10 @@
 
 #include <core/CMemory.h>
 
+#include <core/CPersistUtils.h>
```
tveasey: nit: keep with core include block.

tveasey:
I should add, if you follow the suggestion in [this](https://github.com/elastic/ml-cpp/pull/544#discussion_r303444868) comment I think you should be able to implement all these methods with calls to `core::CPersistUtils`:
```
std::function<void(core::CStatePersistInserter&)>
persistFunctionMeanValues(const TVectorDoublePrVec& functionMeanValues) const;

bool restoreFunctionMeanValues(TVectorDoublePrVec& functionMeanValues,
                                   core::CStateRestoreTraverser& traverser);

std::function<bool(core::CStateRestoreTraverser&)>
restoreParameterValuePair(TVector& parameters, double& functionValue);
```

include/maths/CBayesianOptimisation.h
valeriy42: Done

include/maths/CLinearAlgebraEigen.h
valeriy42: Done

include/maths/CLinearAlgebraEigen.h
valeriy42: Done

include/maths/CLinearAlgebraEigen.h
valeriy42: Done

lib/maths/CBayesianOptimisation.cc
valeriy42: I agree. I removed the the lines from the code.

lib/maths/CBayesianOptimisation.cc
valeriy42: Ok, I changed it to `core::CPersistUtils` now the methods become obsolete.

include/maths/CLinearAlgebraEigen.h
valeriy42: The method was replaced by `acceptPersistInserter` which uses `toStdVector` internally.

valeriy42: The method was replaced by `acceptPersistInserter` which uses `toStdVector` internally.

include/maths/CLinearAlgebraEigen.h
valeriy42: The method was replaced by `acceptRestoreTraverser`, however I am not using `fromStdVector` on `this` to prevent memory leakage.

valeriy42:
Thank you @tveasey for suggestions. The code is much cleaner now. I could also remove most of the code I wrote initially, so less bugs too ;)

include/maths/CLinearAlgebraEigen.h
tveasey: It shouldn't leak memory. You'd need to do something like `*this = fromStdVector(vector)`. The assignment operator would do the clean up any memory held by `this`.

valeriy42: I tried it, but I've got an Eigen asserting error during test execution. Therefore, I defaulted to the explicit copy.

tveasey: Hmm interesting. It says shouldn't do that unless the type has explicit sizes in the template declaration. This should be using dynamic so that ought not to be an issue.

PR #546, opened 2019-07-15
Updates changelog for 7.2.1

lcawl:
This PR adds descriptions for the following PRs in the changelog:
* https://github.com/elastic/ml-cpp/pull/527
* https://github.com/elastic/ml-cpp/pull/516
https://github.com/elastic/ml-cpp/pull/489

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -35,10 +35,7 @@
 * Don't persist model state at the end of lookback if the lookback did not
 generate any input {ml-pull}527[#527] (issue: {ml-issue}519[#519])
 * Don't write model size stats when job is closed without any input
-{ml-pull}516[#512] (issue: {ml-issue}394[#394])
-* Fix an edge case causing spurious anomalies (false positives) if the variance
-in the count of events changed significantly throughout the period of a seasonal
-quantity. {ml-pull}489[#489]
+{ml-pull}516[#516] (issue: {ml-issue}394[#394])
 
 == {es} version 7.2.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,18 @@
 
 //=== Regressions
 
+== {es} version 7.2.1
+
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,18 @@
...
 //=== Regressions
 
+== {es} version 7.2.1
+
+=== Bug Fixes
+
+* Don't persist model state at the end of lookback if the lookback did not
+generate any input {ml-pull}527[#527] (issue: {ml-issue}519[#519])
+* Don't write model size stats when job is closed without any input
+{ml-pull}516[#512] (issue: {ml-issue}394[#394])
```
droberts195: 516 and 512 are inconsistent.  512 is the master branch PR for this change and 516 is the 7.2 branch backport PR.  I'm not sure which one is standard to link to but I guess whichever is preferred it should be consistent for the link and text.

docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,18 @@
...
+* Fix an edge case causing spurious anomalies (false positives) if the variance
+in the count of events changed significantly throughout the period of a seasonal
+quantity. {ml-pull}489[#489]
```
droberts195: I think #489 was mislabelled.  It had the `v7.2.1` label but was only backported as far as the `7.3` branch.  I just fixed the labelling - sorry about that.

tveasey: Yes I just saw that. I think the original intention was that it would go into 7.2.1, but we decided to just release in 7.3. Looking at the PR I now wonder if the change log is wrong for 7.3. I'll raise a separate PR for that.

docs/CHANGELOG.asciidoc
lcawl: Good catch, thanks!

docs/CHANGELOG.asciidoc
lcawl: Okay thanks, I've removed it from the 7.2.1 changelog

PR #549, opened 2019-07-16
[FEATURE][ML] Fix CDenseVector::fromStdVector

tveasey:
[This](https://github.com/elastic/ml-cpp/pull/544/files/64cdffd6576bb0ab40b86555635877824a901337#r303851555) comment intrigued me. It turns out the underlying reason is the Eigen operator<< doesn't behave like one might expect: it works by overwriting the elements from the start of the array representation with the comma separated list supplied. Therefore, as it stood `fromStdVector` wrote all the values into the first element of the vector and left the remaining ones uninitialised.

This change also makes a couple of tweaks to conform to coding standards (public static members first, etc) and makes the to and from std vector methods private for the time being since they're only used by persistence.


tveasey:
retest

PR #550, opened 2019-07-16
[ML] Add option to persist in foreground

edsavage:
Added command line option to both `autodetect` and `categorizer`
indicating that persistence should be performed in the foreground.
Default behaviour remains to perform persistence in a background thread.

Foreground persistence has the advantage that it does not require data
to be copied and hence the memory requirements are half that of background
persistence. This is accounted for in the resource manager.

Relates to elastic/elasticsearch#29770

### DIFF
--- a/include/api/CAnomalyJob.h
+++ b/include/api/CAnomalyJob.h
@@ -171,11 +171,10 @@ public:
 
     //! Restore previously saved state
     bool restoreState(core::CDataSearcher& restoreSearcher,
-                              core_t::TTime& completeToTime) override;
+                      core_t::TTime& completeToTime) override;
 
     //! Persist current state
-    bool persistState(core::CDataAdder& persister,
-                              const std::string& descriptionPrefix) override;
+    bool persistState(core::CDataAdder& persister, const std::string& descriptionPrefix) override;
 
     //! Initialise normalizer from quantiles state
     virtual bool initNormalizer(const std::string& quantilesStateFile);
--- a/include/api/CAnomalyJob.h
+++ b/include/api/CAnomalyJob.h
@@ -255,14 +255,14 @@ private:
     //! Persist the detectors to a stream.
     bool persistCopiedState(const std::string& descriptionPrefix,
                             core_t::TTime time,
-                      const TKeyCRefAnomalyDetectorPtrPrVec& detectors,
-                      const model::CResourceMonitor::SResults& modelSizeStats,
-                      const model::CInterimBucketCorrector& interimBucketCorrector,
-                      const model::CHierarchicalResultsAggregator& aggregator,
-                      const std::string& normalizerState,
-                      core_t::TTime latestRecordTime,
-                      core_t::TTime lastResultsTime,
-                      core::CDataAdder& persister);
+                            const TKeyCRefAnomalyDetectorPtrPrVec& detectors,
+                            const model::CResourceMonitor::SResults& modelSizeStats,
+                            const model::CInterimBucketCorrector& interimBucketCorrector,
+                            const model::CHierarchicalResultsAggregator& aggregator,
+                            const std::string& normalizerState,
+                            core_t::TTime latestRecordTime,
+                            core_t::TTime lastResultsTime,
+                            core::CDataAdder& persister);
 
     //! Persist current state due to the periodic persistence being triggered.
     bool periodicPersistStateInBackground() override;
--- a/include/api/CAnomalyJob.h
+++ b/include/api/CAnomalyJob.h
@@ -36,7 +36,7 @@
 
 #include <stdint.h>
 
-class CBackgroundPersisterTest;
+class CPersistenceManagerTest;
 class CAnomalyJobTest;
 
 namespace ml {
@@ -50,7 +50,7 @@ class CHierarchicalResults;
 class CLimits;
 }
 namespace api {
-class CBackgroundPersister;
+class CPersistenceManager;
 class CModelPlotDataJsonWriter;
 class CFieldConfig;
 
@@ -148,42 +148,43 @@ public:
                 model::CAnomalyDetectorModelConfig& modelConfig,
                 core::CJsonOutputStreamWrapper& outputBuffer,
                 const TPersistCompleteFunc& persistCompleteFunc = TPersistCompleteFunc(),
-                CBackgroundPersister* periodicPersister = nullptr,
+                CPersistenceManager* periodicPersister = nullptr,
                 core_t::TTime maxQuantileInterval = -1,
                 const std::string& timeFieldName = DEFAULT_TIME_FIELD_NAME,
                 const std::string& timeFieldFormat = EMPTY_STRING,
                 size_t maxAnomalyRecords = 0u);
 
-    virtual ~CAnomalyJob();
+    virtual ~CAnomalyJob() override;
 
     //! We're going to be writing to a new output stream
-    virtual void newOutputStream();
+    virtual void newOutputStream() override;
 
     //! Access the output handler
-    virtual COutputHandler& outputHandler();
+    virtual COutputHandler& outputHandler() override;
 
     //! Receive a single record to be processed, and produce output
     //! with any required modifications
-    virtual bool handleRecord(const TStrStrUMap& dataRowFields);
+    virtual bool handleRecord(const TStrStrUMap& dataRowFields) override;
 
     //! Perform any final processing once all input data has been seen.
-    virtual void finalise();
+    virtual void finalise() override;
 
     //! Restore previously saved state
     virtual bool restoreState(core::CDataSearcher& restoreSearcher,
-                              core_t::TTime& completeToTime);
+                              core_t::TTime& completeToTime) override;
 
     //! Persist current state
-    virtual bool persistState(core::CDataAdder& persister);
+    virtual bool persistState(core::CDataAdder& persister,
+                              const std::string& descriptionPrefix) override;
 
     //! Initialise normalizer from quantiles state
     virtual bool initNormalizer(const std::string& quantilesStateFile);
 
     //! How many records did we handle?
-    virtual uint64_t numRecordsHandled() const;
+    virtual uint64_t numRecordsHandled() const override;
 
     //! Is persistence needed?
-    virtual bool isPersistenceNeeded(const std::string& description) const;
+    virtual bool isPersistenceNeeded(const std::string& description) const override;
 
     //! Log a list of the detectors and keys
     void description() const;
@@ -243,11 +244,13 @@ private:
                               core::CStateRestoreTraverser& traverser);
 
     //! Persist current state in the background
-    bool backgroundPersistState(CBackgroundPersister& backgroundPersister);
+    bool backgroundPersistState(CPersistenceManager& backgroundPersister);
 
     //! This is the function that is called in a different thread to the
     //! main processing when background persistence is triggered.
     bool runBackgroundPersist(TBackgroundPersistArgsPtr args, core::CDataAdder& persister);
+
+    bool runForegroundPersist(core::CDataAdder& persister);
 
     //! Persist the detectors to a stream.
     bool persistState(const std::string& descriptionPrefix,
@@ -262,7 +265,8 @@ private:
                       core::CDataAdder& persister);
 
     //! Persist current state due to the periodic persistence being triggered.
-    virtual bool periodicPersistState(CBackgroundPersister& persister);
+    virtual bool periodicPersistStateInBackground(CPersistenceManager& persistenceManager) override;
+    virtual bool periodicPersistStateInForeground(CPersistenceManager& persistenceManager) override;
 
     //! Acknowledge a flush request
     void acknowledgeFlush(const std::string& flushId);
@@ -429,7 +433,7 @@ private:
     //! Pointer to periodic persister that works in the background.  May be
     //! nullptr if this object is not responsible for starting periodic
     //! persistence.
-    CBackgroundPersister* m_PeriodicPersister;
+    CPersistenceManager* m_PeriodicPersister;
 
     //! If we haven't output quantiles for this long due to a big anomaly
     //! we'll output them to reflect decay.  Non-positive values mean never.
@@ -460,7 +464,7 @@ private:
     //! Flag indicating whether or not time has been advanced.
     bool m_TimeAdvanced{false};
 
-    friend class ::CBackgroundPersisterTest;
+    friend class ::CPersistenceManagerTest;
     friend class ::CAnomalyJobTest;
 };
 }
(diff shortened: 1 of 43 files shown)


droberts195:
I think we should mark this `>non-issue` for the time being, because until it's integrated into user-facing functionality it's just confusing to release note it.

lib/api/CAnomalyJob.cc
```diff
@@ -1060,6 +1060,15 @@ bool CAnomalyJob::backgroundPersistState(CBackgroundPersister& backgroundPersist
     return true;
 }
 
+bool CAnomalyJob::runForegroundPersist(core::CDataAdder& persister) {
+    LOG_INFO(<< "Foreground persist commencing...");
+
+    // Finalise the processor so it gets a chance to write any remaining results
+    this->finalise();
```
droberts195: I'm not sure it's a good idea to call `finalise` more than once.  It might confuse a future maintainer as there are other `finalise` methods in the the code that really must only be called once.

But also this is changing the behaviour for persisting quantiles.  Previously quantiles were persisted:

1. On a significant change to the highest ever score
2. Periodically on the quantile persistence interval
3. On graceful close

But by calling `finalise` here you're changing that to:

1. On a significant change to the highest ever score
2. Periodically on the quantile persistence interval
3. On graceful close
4. Periodically on the state persistence interval

That leads to the question of why we'd have separate persistence intervals for quantiles and state.

To be more consistent with background persist it's probably better to call `pruneAllModels` here instead of `finalise`.

include/api/CDataProcessor.h
```diff
@@ -71,10 +71,14 @@ class API_EXPORT CDataProcessor : private core::CNonCopyable {
                               core_t::TTime& completeToTime) = 0;
 
     //! Persist current state
-    virtual bool persistState(core::CDataAdder& persister) = 0;
+    virtual bool persistState(core::CDataAdder& persister,
+                              const std::string& descriptionPrefix = "") = 0;
```
droberts195: Do we need a default here?  It seems like the default makes the logging far less useful so it would be better to enforce one.

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -3,7 +3,8 @@
  * or more contributor license agreements. Licensed under the Elastic License;
  * you may not use this file except in compliance with the Elastic License.
  */
-#include "CBackgroundPersisterTest.h"
+#include "../../../include/api/CPersistenceManager.h"
```
droberts195: Can this be just `#include <api/CPersistenceManager.h>` and grouped with the other `api` library includes?

lib/model/CResourceMonitor.cc
```diff
@@ -282,21 +285,26 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
...
+        // the models. On top of that, after the persist is done we may not
+        // be able to retrieve that memory back. Thus, we report twice the
+        // memory usage in order to allow for that.
+        // See https://github.com/elastic/x-pack-elasticsearch/issues/1020.
+        // Issue https://github.com/elastic/x-pack-elasticsearch/issues/857
+        // discusses adding an option to perform only foreground persist.
+        // If that gets implemented, we should only double when background
+        // persist is configured.
+
+        adjustedUsage = static_cast<std::size_t>(2 * adjustedUsage);
```
droberts195: I don't think a cast is necessary here: `adjustedUsage *= 2` should do it.

lib/model/unittest/CResourceMonitorTest.cc
```diff
@@ -82,6 +82,25 @@ void CResourceMonitorTest::testMonitor() {
...
+        // Test foreground persistence constructor
+        CResourceMonitor mon(true);
+        CPPUNIT_ASSERT(mon.m_ByteLimitHigh > 0);
+        CPPUNIT_ASSERT_EQUAL((49 * mon.m_ByteLimitHigh) / 50, mon.m_ByteLimitLow);
+        CPPUNIT_ASSERT(mon.m_ByteLimitHigh > mon.m_ByteLimitLow);
+        CPPUNIT_ASSERT(mon.m_AllowAllocations);
+        LOG_DEBUG(<< "Resource limit is: " << mon.m_ByteLimitHigh);
+        if (sizeof(std::size_t) == 4) {
+            // 32-bit platform
+            CPPUNIT_ASSERT_EQUAL(std::size_t(1024ull * 1024 * 1024 / 2), mon.m_ByteLimitHigh);
```
droberts195: We haven't built 32 bit for about 5 years and I bet there are other things that would go wrong if somebody did try a 32 bit build.  So it's probably best to scrap this branch and let `sizeof(std::size_t) == 4` go into the `// Unexpected platform` branch.

lib/api/CFieldDataTyper.cc
```diff
@@ -422,6 +423,19 @@ bool CFieldDataTyper::periodicPersistState(CBackgroundPersister& persister) {
     return true;
 }
 
+bool CFieldDataTyper::periodicPersistStateInForeground(CPersistenceManager& persistenceManager) {
+    LOG_DEBUG(<< "Periodic persist typer state");
```
tveasey: This doesn't have any chaining logic as in
```
if (m_OutputHandler.periodicPersistStateInForeground(persister) == false) {
    return false;
}
```
Is this intentional?

tveasey: Oh I see this is in `this->persistState`. I think this warrants a comment.

tveasey: Also, I don't totally understand the logic here: this object has a pointer to a persist manager, but it is passed in here and for example in `this->persistState` it checks to see if the member variable persist is busy rather than the supplied `persistenceManager` object.

include/api/CAnomalyJob.h
```diff
@@ -148,42 +148,43 @@ class API_EXPORT CAnomalyJob : public CDataProcessor {
...
                 const TPersistCompleteFunc& persistCompleteFunc = TPersistCompleteFunc(),
-                CBackgroundPersister* periodicPersister = nullptr,
+                CPersistenceManager* periodicPersister = nullptr,
                 core_t::TTime maxQuantileInterval = -1,
                 const std::string& timeFieldName = DEFAULT_TIME_FIELD_NAME,
                 const std::string& timeFieldFormat = EMPTY_STRING,
                 size_t maxAnomalyRecords = 0u);
 
-    virtual ~CAnomalyJob();
+    virtual ~CAnomalyJob() override;
```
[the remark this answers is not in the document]
edsavage: > Is it best practice to specify _either_ `virtual` _or_ `override` but not both? (Does Effective Modern C++ have guidance on this?)

Meyers doesn't have explicit guidance on specifying both `virtual` and `override` at the same time but certainly in all his examples in item 12 of Effective Modern C++ both _are_ given.

lib/model/CResourceMonitor.cc
```diff
@@ -282,21 +285,26 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
...
-    // be able to retrieve that memory back. Thus, we report twice the
-    // memory usage in order to allow for that.
-    // See https://github.com/elastic/x-pack-elasticsearch/issues/1020.
-    // Issue https://github.com/elastic/x-pack-elasticsearch/issues/857
-    // discusses adding an option to perform only foreground persist.
-    // If that gets implemented, we should only double when background
-    // persist is configured.
-
-    // We also scale the reported memory usage by the inverse of the byte limit margin.
+    // We  scale the reported memory usage by the inverse of the byte limit margin.
```
tveasey:
```suggestion
    // We scale the reported memory usage by the inverse of the byte limit margin.
```

lib/model/CResourceMonitor.cc
```diff
@@ -282,21 +285,26 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
...
+    if (m_PersistenceInForeground == false) {
+        // Background persist causes the memory size to double due to copying
+        // the models. On top of that, after the persist is done we may not
+        // be able to retrieve that memory back. Thus, we report twice the
+        // memory usage in order to allow for that.
+        // See https://github.com/elastic/x-pack-elasticsearch/issues/1020.
+        // Issue https://github.com/elastic/x-pack-elasticsearch/issues/857
+        // discusses adding an option to perform only foreground persist.
+        // If that gets implemented, we should only double when background
+        // persist is configured.
```
tveasey: I'd be inclined to have a function something like `std::size_t persistenceMemoryIncreaseFactor() { return m_PersistenceInForeground ? 1 : 2 }`. Then you don't need to duplicate this comment.

lib/model/unittest/CResourceLimitTest.cc
```diff
@@ -389,61 +389,121 @@ void CResourceLimitTest::testLargeAllocations() {
...
-        ::addPersonData(3000, 3010, time, *gatherer, resourceMonitor);
-
-        LOG_DEBUG(<< "Testing for 3rd time");
-        model.test(time);
-        CPPUNIT_ASSERT_EQUAL(oldNumberPeople + 10, model.getNewPeople());
-        CPPUNIT_ASSERT_EQUAL(std::size_t(0), model.getNewAttributes());
-        CPPUNIT_ASSERT_EQUAL(model.getNewPeople(), gatherer->numberActivePeople());
+
+        {
+            CModelFactory::TDataGathererPtr gatherer(factory.makeDataGatherer(FIRST_TIME));
```
tveasey: How about splitting this test into `testLargeAllocations[Foreground|Background]Persist`? Also, can we share the code better? Seems like we just need to supply different assertion values.

include/api/CAnomalyJob.h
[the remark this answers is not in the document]
tveasey: This seems odd to me though because you can only override virtual functions, so the virtual keyword carries no information in this case.

lib/api/CAnomalyJob.cc
```diff
@@ -972,18 +972,18 @@ bool CAnomalyJob::restoreDetectorState(const model::CSearchKey& key,
...
 
-bool CAnomalyJob::persistState(core::CDataAdder& persister) {
+bool CAnomalyJob::persistState(core::CDataAdder& persister, const std::string& descriptionPrefix) {
     if (m_PeriodicPersister != nullptr) {
         // This will not happen if finalise() was called before persisting state
         if (m_PeriodicPersister->isBusy()) {
-            LOG_ERROR(<< "Cannot do final persistence of state - periodic "
-                         "persister still busy");
+            LOG_ERROR(<< "Cannot perform foreground persistence of state - periodic "
+                         "background persistence is still in progress");
```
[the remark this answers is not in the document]
droberts195: The message is correct for `finalise` because on graceful close we do a foreground persist just before the process exits.  We never do background persist during a graceful close because duplicating all the models at this point would just waste time.

But this ties in with my comment about calling `finalise` at times other than immediately before shutting down.  I don't think we should do that.

include/api/CAnomalyJob.h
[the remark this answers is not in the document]
edsavage: I'll also add an entry regarding `override` to the style guide...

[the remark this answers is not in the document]
edsavage: Scrap that - while specifying `override` without `virtual` in a function declaration is possible it doesn't mean the `virtual` keyword is now superfluous. In fact the declaration results in an ill-formed function, see https://stackoverflow.com/questions/7530616/override-on-non-virtual-functions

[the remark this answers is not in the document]
tveasey: For me the point is that `void foo() override` will fail to compile if `void foo()` isn't virtual no? So you know that `void foo() override` is virtual.

lib/api/CAnomalyJob.cc
```diff
@@ -1051,15 +1055,26 @@ bool CAnomalyJob::backgroundPersistState(CBackgroundPersister& backgroundPersist
...
               maths::COrderings::SFirstLess());
 
-    if (backgroundPersister.addPersistFunc(std::bind(
+    if (m_PeriodicPersister->addPersistFunc(std::bind(
             &CAnomalyJob::runBackgroundPersist, this, args, std::placeholders::_1)) == false) {
         LOG_ERROR(<< "Failed to add anomaly detector background persistence function");
         return false;
     }
 
+    m_PeriodicPersister->persistInForeground(false);
```
droberts195: When I first saw this line I thought it was a bug.  I think it's just that the naming is confusing.  If I'm reading this correctly `persistInForeground(false)` means persist in the background.  Maybe `runPersistFuncs(false)` would be less confusing.  Or maybe there is a name that is better still.  But having `Foreground` in the name seems wrong.

davidkyle: There's acres of debate around boolean parameters on the internet. 

https://martinfowler.com/bliki/FlagArgument.html

lib/api/CAnomalyJob.cc
```diff
@@ -1051,15 +1055,26 @@ bool CAnomalyJob::backgroundPersistState(CBackgroundPersister& backgroundPersist
...
     return true;
 }
 
+bool CAnomalyJob::runForegroundPersist(core::CDataAdder& persister) {
+    LOG_INFO(<< "Foreground persist commencing...");
+
+    // Prune the models so that the persisted state is as neat as possible
+    this->pruneAllModels();
+
+    return this->persistState(persister, "Periodic foreground persist at ");
```
tveasey: This class has a lot of persist methods: `persistState` two overloads, `runForegroundPersist`, `runBackgroundPersist`, `periodicPersistStateInBackground` and `periodicPersistStateInBackground`. I find as a minimum having overloads of `persistState` a bit confusing: is one `persistCopiedState` or something similar?

lib/api/CPersistenceManager.cc
```diff
@@ -172,13 +178,24 @@ bool CBackgroundPersister::startBackgroundPersistIfAppropriate() {
...
+    if (this->isBusy()) {
+        LOG_WARN(<< "Cannot start persist as a previous "
+                    "background persist is still in progress");
+        return false;
+    }
+
+    auto firstProcessorPeriodicPersistFunc = [&]() {
+        return m_PersistInForeground ? m_FirstProcessorForegroundPeriodicPersistFunc
+                                     : m_FirstProcessorBackgroundPeriodicPersistFunc;
+    };
```
tveasey: nit: why wrap in a lambda? Surely this could just be
```
auto firstProcessorPeriodicPersistFunc = m_PersistInForeground
                                     ? m_FirstProcessorForegroundPeriodicPersistFunc
                                     : m_FirstProcessorBackgroundPeriodicPersistFunc
```

PR #551, opened 2019-07-17
[FEATURE][ML] Correct data generation in CDataFrameUtilsTest

tveasey:
A SIGSEGV testing #549 showed up an error in the way we were generating the test data for `CDataFrameUtilsTest::testCategoryMicWithColumn`. We weren't generating enough and getting out-of-bounds reads as a result. This change fixes this issue. The code can also be shared across multiple tests, so I factored it out. Finally, there was an unwanted source of correlation between the shuffling of categories for different columns which I fixed at the same time. This allowed me to tighten up the test thresholds.

### DIFF
--- a/lib/maths/unittest/CDataFrameUtilsTest.cc
+++ b/lib/maths/unittest/CDataFrameUtilsTest.cc
@@ -26,6 +26,7 @@
 
 using namespace ml;
 
+namespace {
 using TDoubleVec = std::vector<double>;
 using TDoubleVecVec = std::vector<TDoubleVec>;
 using TSizeVec = std::vector<std::size_t>;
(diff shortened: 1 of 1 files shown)


lib/maths/unittest/CDataFrameUtilsTest.cc
```diff
@@ -34,6 +35,30 @@ using TMeanAccumulator = maths::CBasicStatistics::SSampleMean<double>::TAccumula
...
+    rng.generateDirichletSamples(expectedFrequencies, cols, frequencies);
+
+    TDoubleVecVec values(cols);
+    for (std::size_t i = 0; i < frequencies.size(); ++i) {
+        for (std::size_t j = 0; j < frequencies[i].size(); ++j) {
+            std::size_t target{static_cast<std::size_t>(
+                static_cast<double>(rows) * frequencies[i][j] + 0.5)};
+            values[i].resize(values[i].size() + target, static_cast<double>(j));
+        }
+        values[i].resize(rows, values[i].back());
```
tveasey: This is the condition which was not guaranteed, in particular, it was possible that `values[i].size() < rows` before.

PR #538, opened 2019-07-05
[ML] Boosted tree memory estimation

valeriy42:
Upfront memory estimation for boosted regression trees.

Closes elastic/ml-team#127

PR #552, opened 2019-07-18
[FEATURE][ML] Introducing factory for boosted tree creation

valeriy42:
This PR refactors the implementation of boosted trees, splitting it into initialization logic and execution logic. Initialization is moved to the `CBoostedTreeFactory` class. 

The next step will be to enable creation of the `CBoostedTree` object from a serialized string instead of parameters.


include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,113 @@
...
+//! SimpleFactory for CBoostedTree object
+class CBoostedTreeFactory final {
+public:
+    using TCBoostedTreeUPtr = std::unique_ptr<CBoostedTree>;
+
+public:
+    //! Construct a boosted tree object from parameters
+    static CBoostedTreeFactory constructFromParameters(std::size_t numberThreads,
+                                                       std::size_t dependentVariable,
+                                                       CBoostedTree::TLossFunctionUPtr loss);
```
tveasey: nit: why not just a constructor?

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,113 @@
...
+#include <vector>
+
+namespace ml {
+namespace maths {
+
+class CNode;
+class CBoostedTreeImpl;
+
+//! SimpleFactory for CBoostedTree object
+class CBoostedTreeFactory final {
```
tveasey: I think the natural thing to do if we're introducing a factory and given `CBoostedTree` inherits from `CDataFrameRegressionModel` would be to have a base class for the factory type as well and build method which returns `TDataFrameRegressionModelUPtr`. Since we don't have other implementations at present this isn't strictly needed, but it would be another advantage of having the factory.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -0,0 +1,738 @@
...
+        featureBagFraction = CTools::truncate(featureBagFraction, 0.0, 1.0);
+    }
+    m_FeatureBagFractionOverride = featureBagFraction;
+}
+
+void CBoostedTreeImpl::maximumOptimisationRoundsPerHyperparameter(std::size_t rounds) {
+    m_MaximumOptimisationRoundsPerHyperparameter = rounds;
+}
+
+//! Train the model on the values in \p frame.
```
tveasey: nit: no need to duplicate doxygen comments from header file (and elsewhere).

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,340 @@
...
+#include <maths/CBayesianOptimisation.h>
+#include <maths/CBoostedTreeImpl.h>
+
+namespace ml {
+namespace maths {
+
+CBoostedTreeFactory::TCBoostedTreeUPtr CBoostedTreeFactory::build() {
+    this->initializeMissingFeatureMasks(*m_Frame);
+    std::tie(m_Tree->m_Impl->m_TrainingRowMasks,
+             m_Tree->m_Impl->m_TestingRowMasks) = this->crossValidationRowMasks();
```
tveasey: I think it would be cleaner if you had a constructor on `CBoostedTree` which takes `CBoostedTreeImpl` and have have `m_Tree` be of type `CBoostedTreeImpl` then you can ditch all the `->m_Impl`s in this code.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,340 @@
...
+
+    this->m_Tree->m_Impl->m_CurrentRound = 0; // for first start
+    return std::move(m_Tree);
+}
+
+CBoostedTreeFactory::operator TCBoostedTreeUPtr() {
+    return std::move(this->build());
+}
+
+//! Get the number of hyperparameter tuning rounds to use.
```
tveasey: nit: duplicating doxygen comments throughout this file.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,340 @@
...
+    this->m_ProgressCallback = callback;
+    return *this;
+}
+
+CBoostedTreeFactory& CBoostedTreeFactory::frame(core::CDataFrame& frame) {
+    this->m_Frame = &frame;
+    return *this;
+}
+CBoostedTreeFactory::operator CBoostedTree &&() {
+    return std::move(*(this->build()));
```
tveasey: I'm not sure about this: the unique pointer is moved out of build and then surely dies after this function returns. Even if it is moving does this guaranty that there isn't an invalid read: feels like a dark corner of the language. I think we can just stick with the version which returns a unique pointer can't we?

include/maths/CBoostedTree.h
```diff
@@ -159,12 +161,17 @@ class MATHS_EXPORT CBoostedTree final : public CDataFrameRegressionModel {
     //! Get the column containing the model's prediction for the dependent variable.
     std::size_t columnHoldingPrediction(std::size_t numberColumns) const override;
 
+    //! Estimate the maximum booking memory that training the boosted tree on a data
+    //! frame with \p numberRows row and \p numberColumns columns will use.
+    std::size_t estimateMemoryUsage(std::size_t numberRows, std::size_t numberColumns) const;
```
tveasey: This is on the feature/regression branch. Is this from an earlier revision?

include/maths/CBoostedTreeImpl.h
```diff
@@ -0,0 +1,698 @@
...
+    //! Set the maximum number of trees in the ensemble.
+    void maximumNumberTrees(std::size_t maximumNumberTrees);
+    //! Set the fraction of features we'll use in the bag to build a tree.
+    void featureBagFraction(double featureBagFraction);
+    //! Set the maximum number of optimisation rounds we'll use for hyperparameter
+    //! optimisation per parameter.
+    void maximumOptimisationRoundsPerHyperparameter(std::size_t rounds);
+
+    //! Train the model on the values in \p frame.
+    void train(core::CDataFrame& frame, CBoostedTree::TProgressCallback recordProgress);
```
tveasey:
```suggestion
    void train(core::CDataFrame& frame, CBoostedTree::TProgressCallback recordProgress);
    
```
Also, given you pass frame to the factory it seems slightly odd that you pass it again to these functions. In particular, this has to be the same frame that you built it on so this seems like a recipe for abuse.

tveasey: I'm thinking we should move towards storing a reference to the frame to run on in the base class, having concrete `void train(CBoostedTree::TProgressCallback recordProgress)` and a private `virtual void trainImpl(core::CDataFrame& frame, CBoostedTree::TProgressCallback recordProgress)` which can then be overridden in the derived classes. And similarly for other interface functions which currently take a `frame`.

include/maths/CBoostedTree.h
```diff
@@ -126,6 +127,7 @@ class MATHS_EXPORT CBoostedTree final : public CDataFrameRegressionModel {
     //@{
     //! Set the number of folds to use for estimating the generalisation error.
     CBoostedTree& numberFolds(std::size_t folds);
+
```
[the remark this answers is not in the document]
valeriy42: Agreed. In fact, a clean design would suggest to make the constructor of `CBoostedTree` private.

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,113 @@
...
+public:
+    using TCBoostedTreeUPtr = std::unique_ptr<CBoostedTree>;
```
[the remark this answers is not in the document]
valeriy42: What do you mean?

lib/maths/CBoostedTreeFactory.cc
valeriy42: Ok. I had to for a better integration with the client code. But if you don't mind, I leave only unique pointer option.

include/maths/CBoostedTree.h
valeriy42: This is related to #538 isn't it?

include/maths/CBoostedTreeFactory.h
valeriy42: There can be potentially another method for construction which takes to `size_t`s and a loss function. Hence, I have a descriptive method name to identify the intention.

include/maths/CBoostedTreeFactory.h
valeriy42: I see your point of having abstract factory for symmetry of class hierarchies. However, I would  not introduce this level of abstraction prematuraly. Say, when there is another regression model we need to construct, we introduce polymorphism.

lib/maths/CBoostedTreeFactory.cc
tveasey: Ok, no I don't mind if you need to change a little more calling code. I think we should stick with just the `unique_ptr` version.

lib/api/CDataFrameBoostedTreeRunner.cc
```diff
@@ -128,8 +138,10 @@ std::size_t CDataFrameBoostedTreeRunner::estimateBookkeepingMemoryUsage(
     std::size_t totalNumberRows,
     std::size_t partitionNumberRows,
     std::size_t numberColumns) const {
-    // TODO
-    return 0;
+    return m_BoostedTree
+               ? m_BoostedTree->estimateMemoryUsage(partitionNumberRows, numberColumns)
+               : m_BoostedTreeFactory->incompleteTreeObject().estimateMemoryUsage(
+                     partitionNumberRows, numberColumns);
```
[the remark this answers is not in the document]
valeriy42: Indeed, It is fixed now.

include/maths/CBoostedTree.h
tveasey: I think the confusion is caused by not branching not off `feature/regression` so we're missing some of the committed changes in this branch. Maybe you can completely clear this up with a rebase, or maybe it will be easier to reapply to a branch of `feature/regression`.

include/maths/CBoostedTree.h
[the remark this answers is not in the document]
tveasey: ++

include/maths/CBoostedTreeFactory.h
tveasey: Ok, I understand you intention. My argument against this is to me whatever is passed to this function should be required (not possible to default) and therefore there should only ever be one such function. If some of the values are optional they should be set by setter on the factory as we do for all the other optional parameters. I'm not 100% against this, and it is after all a matter of style, it just feels a little cumbersome to me.

include/maths/CBoostedTreeFactory.h
tveasey: Ok, I'm happy to keep as it is now: it makes the change smaller. (I suspect we will want to provide some additional models down-the-line. At this point if we want to allow the user to manually select the type, then injecting this choice by an appropriate factory object will be in the back pocket :).)

lib/maths/CBoostedTreeFactory.cc
valeriy42: In this case I would need to expose the `Impl` class in method `incompleteTreeObject`, which I cannot do. IMO, either we have to find another way for memory estimation or we have to live with the `->m_Impl`s :man_shrugging:

tveasey: I kind of wonder if `estimateMemoryUsage` should be on the factory object anyway. To me it makes more sense that one asks the factory how big the object it is being asked to build is going to be than get an incompletely built object which can only be used for this purpose. That way it can call the same method on the implementation itself and this never gets exposed to the calling code. What do you think?

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,340 @@
...
+    if (regressors.empty()) {
+        HANDLE_FATAL(<< "Input error: all features constant.");
+    } else {
+        double p{1.0 / static_cast<double>(regressors.size())};
+        for (auto feature : regressors) {
+            m_Tree->m_Impl->m_FeatureSampleProbabilities[feature] = p;
+        }
+    }
+    LOG_TRACE(<< "P(sample) = "
+              << core::CContainerPrinter::print(m_Tree->m_Impl->m_FeatureSampleProbabilities));
```
tveasey: Just spotted that this is out-of-date: predates #488. I'm thinking it might be better to reapply to a clean branch of feature/regression (if it isn't too much trouble).

lib/maths/CBoostedTreeFactory.cc
valeriy42: I think, factory should have a single responsibility of building a `CBoostedTree` objects. My first intention was to introduce a Proxy for `CBoostedTree`, which can handle memory estimation, but initialized the proper boosted tree implementation if called with `train()`. `incompleteTreeObject` is the best I could come up with at the moment, since it still kind-of creating an object and hence does not violate SRP. But the rest of the process is finished, we may see a more elegant solution.

include/maths/CBoostedTree.h
valeriy42: Indeed I recreated all changes from HEAD of `feature/regression` in the local workspace. I am about to create a PR in the next couple of hours :)

lib/maths/CBoostedTreeFactory.cc
tveasey: I guess this comes down to whose concern you think memory estimation is. I'm fine to kick this decision to the end and it might be easiest to discuss a bit offline.

include/maths/CBoostedTree.h
valeriy42: See PR #555

include/maths/CBoostedTreeImpl.h
valeriy42: done see PR #555

lib/maths/CBoostedTreeFactory.cc
valeriy42: I rebased in and created PR #555

valeriy42:
Since this PR was based on the wrong branch, I recreated it in the PR #555

PR #553, opened 2019-07-19
[FEATURE][ML] Fix some merge fallout

tveasey:
Following on from merging #542, there were some more C-style strings to migrate on this branch. There were also some formatting and test failures to fix.

### DIFF
--- a/lib/api/CDataFrameAnalysisRunner.cc
+++ b/lib/api/CDataFrameAnalysisRunner.cc
@@ -73,11 +73,15 @@ void CDataFrameAnalysisRunner::computeAndSaveExecutionStrategy() {
     LOG_TRACE(<< "number partitions = " << m_NumberPartitions);
 
     if (memoryUsage > memoryLimit) {
+        auto roundMb = [](std::size_t memory) {
+            return 0.01 * static_cast<double>((100 * memory) / (1024 * 1024));
+        };
+
         // Report rounded up to the nearest MB.
-        std::size_t memoryUsageMb = (memoryUsage + (1024 * 1024 - 1)) / (1024 * 1024);
-        HANDLE_FATAL(<< "Input error: memory limit " << memoryLimit / (1024 * 1024)
+        HANDLE_FATAL(<< "Input error: memory limit " << roundMb(memoryLimit)
                      << "MB is too low to perform analysis. You need to give the process"
-                     << " at least " << memoryUsageMb << "MB, but preferably more.");
+                     << " at least " << std::ceil(roundMb(memoryUsage))
+                     << "MB, but preferably more.");
 
     } else if (m_NumberPartitions > 1) {
         // The maximum number of rows is found by binary search in the interval
--- a/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
+++ b/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
@@ -8,6 +8,7 @@
 
 #include <core/CContainerPrinter.h>
 #include <core/CLogger.h>
+#include <core/CRegex.h>
 
 #include <api/CDataFrameAnalysisSpecification.h>
 #include <api/CDataFrameAnalysisSpecificationJsonWriter.h>
@@ -16,7 +17,6 @@
 #include <test/CTestTmpDir.h>
 
 #include <mutex>
-#include <regex>
 #include <string>
 #include <vector>
 
@@ -94,9 +94,10 @@ void CDataFrameAnalysisRunnerTest::testComputeAndSaveExecutionStrategyDiskUsageF
 
         // single error is registered that the memory limit is to low
         LOG_DEBUG(<< "errors = " << core::CContainerPrinter::print(errors));
-        std::regex re{"Input error: memory limit.*"};
+        core::CRegex re;
+        re.init("Input error: memory limit.*");
         CPPUNIT_ASSERT_EQUAL(1, static_cast<int>(errors.size()));
-        CPPUNIT_ASSERT(std::regex_match(errors[0], re));
+        CPPUNIT_ASSERT(re.matches(errors[0]));
     }
 
     // Test large memory requirement with disk usage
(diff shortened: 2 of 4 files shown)


lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
```diff
@@ -91,9 +93,10 @@ void CDataFrameAnalysisRunnerTest::testComputeAndSaveExecutionStrategyDiskUsageF
         api::CDataFrameAnalysisSpecification spec{jsonSpec};
 
         // single error is registered that the memory limit is to low
+        LOG_DEBUG(<< "errors = " << core::CContainerPrinter::print(errors));
+        std::regex re{"Input error: memory limit.*"};
```
droberts195: I don't think we should use `std::regex`.  There are numerous reports of it being slow compared to Boost, for example:

* https://stackoverflow.com/questions/51503590/stdregex-is-much-slower-than-boostregex-on-vc2015u3
* https://stackoverflow.com/questions/20942450/why-c11-regex-libc-implementation-is-so-slow
* https://stackoverflow.com/questions/14205096/c11-regex-slower-than-python

I think we should stick with the `core::CRegex` class.

lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -73,11 +73,15 @@ void CDataFrameAnalysisRunner::computeAndSaveExecutionStrategy() {
     LOG_TRACE(<< "number partitions = " << m_NumberPartitions);
 
     if (memoryUsage > memoryLimit) {
+        auto roundMb = [](std::size_t memory) {
+            return 0.01 * static_cast<double>((100 * memory) / (1024 * 1024));
```
droberts195: What is the benefit of switching this code to use floating point rather than doing the calculation using integer arithmetic?

Also, as it stands the lambda doesn't do the rounding, as that's done by the `std::ceil` that's separate.

If it's going to use floating point then can't the formula simply be `return static_cast<double>(memory) / static_cast<double>(1024 * 1024)`?

But I don't really see a need for this whole change.

tveasey: I changed it because if the actual limit is small the reported limit could easily be 0MB: for example you could get weird messages like `"Input error: memory limit 0MB is too low to perform analysis. You need to give the process at least 3MB, but preferably more."` The `0.01 * 100 * ` is just to get it to 2 d.p.

lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
tveasey: This isn't a performance bottleneck: it is test only and instantaneous. I do wonder if performance of regex would really be important for any of our code. If it is there are better libraries than boost from that perspective, e.g. hscan, if it isn't isn't it better to cut down our dependencies. However, I guess the argument would be that we should be consistent with the rest of the code and that probably uses `core::CRegex`, so I'm happy to switch this.

lib/api/CDataFrameAnalysisRunner.cc
droberts195: > The `0.01 * 100 *` is just to get it to 2 d.p.

But the `std::ceil` in the log macro wipes out the decimals.

> if the actual limit is small the reported limit could easily be 0MB

The old formula was rounding up to the next MB, so 1 byte was and still is reported as 1MB.  The combination of converting to MB rounded to 2 d.p. and then taking the `ceil` is equivalent to the old formula.

tveasey: ... so the round in the name applies refers to the rounding to 2d.p. @droberts195 are you happy with this explanation?  Do you have any suggestions for a better alternative?

lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
droberts195: > if it isn't isn't it better to cut down our dependencies

This change doesn't cut the dependencies, because it hasn't removed every single usage of `boost::regex` from the code.

> there are better libraries than boost from that perspective, e.g. hscan

That's a good argument for always using `core::CRegex` because then if we switch over then we just change it in one place.

> This isn't a performance bottleneck

True, but it sets the precedent that it's OK to use `std::regex` directly.

tveasey: [ad16672](https://github.com/elastic/ml-cpp/pull/553/commits/ad16672db1881e47791f94a655386fc24ccf25d1)

lib/api/CDataFrameAnalysisRunner.cc
tveasey: > But the std::ceil in the log macro wipes out the decimals.
...
  The old formula was rounding up to the next MB, so 1 byte was and still is reported as 1MB. The combination of converting to MB rounded to 2 d.p. and then taking the ceil is equivalent to the old formula.

This isn't quite right. The original code was (rounding down):
```
HANDLE_FATAL(<< "Input error: memory limit " << memoryLimit / (1024 * 1024)
```
for reporting the memory limit. There are now two calls: one rounds to 2d.p. (for current limit) one rounds to up to nearest MB (for required memory limit). We could make both round up. My rationale for reporting the actual limit more accurately is that I thought people might find it confusing if this is significantly different from what they know they've set. Either way I think we should avoid the 0MB.

droberts195: Sorry, I missed that that lambda was used in two places.

The Java code already enforces a minimum of 1MB for the model memory limit.  But it's true you could still configure it to something like 1.43MB.  Since we're reporting the limit to 2 d.p. maybe we should just get rid of the `std::ceil` and also report the requirement to 2 d.p.?

(It makes me think we should have just done what we did in anomaly detection and enforce that the model memory limit was a whole number of MB, as I'm sure 99.9% of people will set it this way.)

tveasey: The reason for leaving the round up was to give us some margin given that the required memory is the absolute minimum we can get away with as a limit. We need some pad, but maybe this could be a fixed pad and still report to 2d.p. The one observation, I have is it seems kind of odd to me to be so precise in a required memory statement (given that this includes various things we can only approximately estimate).

> It makes me think we should have just done what we did in anomaly detection and enforce that the model memory limit was a whole number of MB, as I'm sure 99.9% of people will set it this way

This is probably right. The one downside is that it can be wasteful, say if you don't have much data for training a prediction model, where you can get away with significantly less than 1MB. Given where we are now, I'm not sure it is worthwhile making that change.

PR #555, opened 2019-07-23
[FEATURE][ML] Introducing factory for boosted tree creation (second try)

valeriy42:
This PR refactors the implementation of boosted trees, splitting it into initialization logic and execution logic. Initialization is moved to the `CBoostedTreeFactory` class. It replaces the PR #552, which was based on the wrong branch.

The next step will be to enable creation of the `CBoostedTree` object from a serialized string instead of parameters.


include/maths/CBoostedTree.h
```diff
@@ -166,11 +146,38 @@ class MATHS_EXPORT CBoostedTree final : public CDataFrameRegressionModel {
...
-    using TImplUPtr = std::unique_ptr<CImpl>;
+    using TImplUPtr = std::unique_ptr<CBoostedTreeImpl>;
+
+private:
+    CBoostedTree(std::size_t numberThreads, std::size_t dependentVariable, TLossFunctionUPtr loss);
+
+    //! \name Parameter Setters
+    //@{
+    //! Set the number of folds to use for estimating the generalisation error.
+    CBoostedTree& numberFolds(std::size_t folds);
```
tveasey: Are any of these setters actually called? You call m_Tree->m_Impl->numberFolds(folds) from the factory for example. I think these functions can just be dropped.

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,111 @@
...
+#include <utility>
+#include <vector>
+
+namespace ml {
+namespace maths {
+
+class CNode;
+class CBoostedTree;
+
+//! SimpleFactory for CBoostedTree object
```
tveasey:
```suggestion
//! Simple factory for CBoostedTree objects.
```

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,111 @@
...
+    CBoostedTreeFactory& maximumNumberTrees(std::size_t maximumNumberTrees);
+    //! Set the fraction of features we'll use in the bag to build a tree.
+    CBoostedTreeFactory& featureBagFraction(double featureBagFraction);
+    //! Set the maximum number of optimisation rounds we'll use for hyperparameter
+    //! optimisation per parameter.
+    CBoostedTreeFactory& maximumOptimisationRoundsPerHyperparameter(std::size_t rounds);
+    //! Set the callback function for progress monitoring.
+    CBoostedTreeFactory& progressCallback(CBoostedTree::TProgressCallback callback);
+    //! Set the reference to the data frame.
+    CBoostedTreeFactory& frame(core::CDataFrame& frame);
```
tveasey: This underlines why I'm not sold on having memory estimation mixed up with having to build an object: this is a non-optional argument and so code will have to check if it has been specified and then you only get runtime rather than compile time failures for misuse.

I'd much prefer parameters which are not optional, such as this one, to be passed in the `constructFromParameters` method for this reason.

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,111 @@
...
+    //! Returns the reference to the CBoostedTree object before this was completely
+    //! initialized. This is useful since some functions, e.g estimateMemoryUsage has to be called before
+    //! data frame reference was passed
+    const CBoostedTree& incompleteTreeObject() const;
+    //! implicit converter operator
```
tveasey:
```suggestion
    //! Create and return a new boosted tree trainer.
```

include/maths/CBoostedTreeFactory.h
```diff
@@ -0,0 +1,111 @@
...
+    //! optimisation per parameter.
+    CBoostedTreeFactory& maximumOptimisationRoundsPerHyperparameter(std::size_t rounds);
+    //! Set the callback function for progress monitoring.
+    CBoostedTreeFactory& progressCallback(CBoostedTree::TProgressCallback callback);
+    //! Set the reference to the data frame.
+    CBoostedTreeFactory& frame(core::CDataFrame& frame);
+    //! Returns the reference to the CBoostedTree object before this was completely
+    //! initialized. This is useful since some functions, e.g estimateMemoryUsage has to be called before
+    //! data frame reference was passed
+    const CBoostedTree& incompleteTreeObject() const;
```
tveasey: I understand why you want to do this, but I don't think it is a good choice in this case. The only thing this object can do is return the memory usage, so it is like wrapping up that calculation in an object whose interface can't be supported. It presumably also means that we have to be able to create a version of `CBoostedTree` with nullable type for `frame` and then have runtime checks that this exists when for example `train` is called. Let's see if we can come up with a better solution offline.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,325 @@
...
+    this->m_Tree->m_Impl->m_BayesianOptimization =
+        std::make_unique<CBayesianOptimisation>(this->hyperparameterBoundingBox());
+    this->m_Tree->m_Impl->m_NumberRounds = this->numberHyperparameterTuningRounds();
+
+    this->m_Tree->m_Impl->m_CurrentRound = 0; // for first start
+    return std::move(m_Tree);
+}
+
+CBoostedTreeFactory::operator TBoostedTreeUPtr() {
+    return std::move(this->build());
```
tveasey:
```suggestion
    return this->build();
```
This prevents copy elision of temporary by forcing it to use move constructor.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,325 @@
...
+#include <maths/CBoostedTreeFactory.h>
+
+#include <maths/CBayesianOptimisation.h>
+#include <maths/CBoostedTreeImpl.h>
+
+namespace ml {
+namespace maths {
+
+CBoostedTreeFactory::TBoostedTreeUPtr CBoostedTreeFactory::build() {
+    this->initializeMissingFeatureMasks(*m_Frame);
```
tveasey: This code SIGSEGVs if the `m_Frame` wasn't set before build was called. As I mentioned elsewhere, I'd prefer to pass by reference to `constructFromParameters` to ensure this can't happen.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,325 @@
...
+    std::tie(m_Tree->m_Impl->m_TrainingRowMasks,
```
tveasey: I find the `m_Tree->m_Impl->` distracting reading the code in this class. How about wrapping this up in a private function `CBoostedTreeFactory::tree()` which return the implementation by writable reference?

include/maths/CDataFrameRegressionModel.h
```diff
@@ -32,7 +32,7 @@ class MATHS_EXPORT CDataFrameRegressionModel {
     virtual ~CDataFrameRegressionModel() = default;
 
     //! Train the model on the values in \p frame.
-    virtual void train(core::CDataFrame& frame, TProgressCallback recordProgress = noop) = 0;
+    virtual void train(TProgressCallback recordProgress = noop) = 0;
```
tveasey: Also, `predict` still takes a data frame. I feel like this class should enforce that a `CDataFrame` has been supplied by requiring it by *reference* in the constructor. Then it can provide a protected accessor to the `CDataFrame` object. This forces any implementation to ensure that it is available at the point train is run.

include/maths/CBoostedTreeFactory.h
valeriy42: `frame` is  now a required parameter for the constructor of `CBoostedTree`. Unfortunately, it cannot be passed over `constructFromParameters` since `CBoostedTreeRunner` does not have it at the moment when `constructFromParameters` is invoked. I introduced a check int the `CBoostedTreeFactory::build()` to prevent constructing a `CBoostedTree` object without `frame` initialized.

include/maths/CBoostedTreeFactory.h
valeriy42: Changed in 366750d

lib/maths/CBoostedTreeFactory.cc
valeriy42: As described above, now this should be impossible.

include/maths/CDataFrameRegressionModel.h
valeriy42: Done.

lib/api/CDataFrameBoostedTreeRunner.cc
```diff
@@ -106,28 +109,33 @@ CDataFrameBoostedTreeRunner::~CDataFrameBoostedTreeRunner() {
...
+        writer.Double(row[m_BoostedTree->columnHoldingPrediction(row.numberColumns())]);
+        writer.EndObject();
+    } else {
+        HANDLE_FATAL(<< "Boosted tree object was not completely created. Please report this error.");
+    }
 }
 
 void CDataFrameBoostedTreeRunner::runImpl(core::CDataFrame& frame) {
-    m_BoostedTree->train(frame, this->progressRecorder());
+    m_BoostedTree = m_BoostedTreeFactory->frame(frame);
```
tveasey: Suggestion: if we had a public `TBoostedTreeUPtr CBoostedTreeFactory::buildFor(core::CDataFrame& frame)` which replaced both implicit conversion to pointer and `frame` functions then this ensures at compile time that the tree can only be got when frame is non-null and you could ditch the assertion on is `nullptr` in build. I also wonder whether this makes the calling code a little easier to read, i.e. this line would be `m_BoostedTree = m_BoostedTreeFactory->buildFor(frame);` which makes it clearer what is going on IMO.

include/maths/CDataFrameRegressionModel.h
```diff
@@ -50,8 +50,15 @@ class MATHS_EXPORT CDataFrameRegressionModel {
...
 
+    //! Get the reference to the data frame object.
+    core::CDataFrame& frame() { return m_Frame; };
+
 protected:
+    CDataFrameRegressionModel(core::CDataFrame& frame) : m_Frame{frame} {};
     static void noop(double);
+
+protected:
+    core::CDataFrame& m_Frame;
```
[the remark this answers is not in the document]
valeriy42: I created an issue #557 to handle the `operator=`.

lib/api/unittest/CDataFrameAnalyzerTest.cc
```diff
@@ -239,30 +240,32 @@ void addRegressionTestData(TStrVec fieldNames,
 
     auto frame = test::CDataFrameTestUtils::toMainMemoryDataFrame(rows);
 
-    maths::CBoostedTree tree{1, weights.size(),
-                             std::make_unique<maths::boosted_tree::CMse>()};
+    maths::CBoostedTreeFactory treeFactory = maths::CBoostedTreeFactory::constructFromParameters(
```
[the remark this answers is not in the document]
valeriy42: done

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -0,0 +1,329 @@
...
+    std::tie(m_Impl->m_TrainingRowMasks, m_Impl->m_TestingRowMasks) =
+        this->crossValidationRowMasks();
+
+    // We store the gradient and curvature of the loss function and the predicted
+    // value for the dependent variable of the regression.
+    m_Frame->resizeColumns(m_Impl->m_NumberThreads, m_Frame->numberColumns() + 3);
+
+    this->initializeFeatureSampleDistribution(*m_Frame);
+    this->initializeHyperparameters(*m_Frame, m_ProgressCallback);
+    m_Impl->m_BayesianOptimization =
```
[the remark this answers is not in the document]
valeriy42: done

lib/api/CDataFrameBoostedTreeRunner.cc
valeriy42: This is actually a very good suggestions! I implemented it.

PR #574, opened 2019-07-31
[ML] Slow tests speed up

tveasey:
This is a quick win speed up for some long running boosted tree QoR unit tests. We now only check that results are equivalent between in memory and on disk data frames for one data set for each test rather than all of them. The upshot is close to a 50% speed up.


droberts195:
After timings of the tests in the debug build was:

Test name|Duration|Status
---|---|---
testConstantFeatures|1 min 0 sec|Passed
testConstantObjective|0.27 sec|Passed
testErrors|0 ms|Passed
testLinear|7 min 49 sec|Passed
testMissingData|0 ms|Passed
testNonLinear|0 ms|**Regression**
testPersistRestore|55 sec|Passed
testPiecewiseConstant|9 min 14 sec|Passed
testThreading|3 min 3 sec|Passed

PR #577, opened 2019-07-31
[ML] Some corrections to error handling on fail to restore boosted tree training state

tveasey:
Previously, we absorbed any errors emitted restoring boosted tree training state and just returned the bad object. We were also silently swallowing errors in `CBayesianOptimisation` constructor from a state traverser. Finally, we were unnecessarily wrapping individual restore loops in try blocks (this was already done in the `traverseSubLevel` function).

We now ensure all errors are propagated to `CBoostedTreeFactory::constructFromString` where we then propagate errors back to the Java via the usual `HANDLE_FATAL` mechanism. (We need to revisit whether a restore issue should be a fatal error, since we can just resume training from scratch in this case, but I'll defer that decision until it is fully wired in.)

This change also adds some bad state restore unit tests.

### DIFF
--- a/include/core/RestoreMacros.h
+++ b/include/core/RestoreMacros.h
@@ -10,44 +10,60 @@
 namespace ml {
 namespace core {
 
-#define RESTORE(tag, restore)                                                      \
-    if (name == tag) {                                                             \
-        if ((restore) == false) {                                                  \
-            LOG_ERROR(<< "Failed to restore " #tag ", got " << traverser.value()); \
-            return false;                                                          \
-        }                                                                          \
-        continue;                                                                  \
+#define RESTORE(tag, restore)                                                          \
+    if (name == tag) {                                                                 \
+        if ((restore) == false) {                                                      \
+            if (traverser.value().empty()) {                                           \
+                LOG_ERROR(<< "Failed to restore " #tag);                               \
+            } else {                                                                   \
+                LOG_ERROR(<< "Failed to restore " #tag ", got " << traverser.value()); \
+            }                                                                          \
+            return false;                                                              \
+        }                                                                              \
+        continue;                                                                      \
     }
 
-#define RESTORE_BUILT_IN(tag, target)                                               \
-    if (name == tag) {                                                              \
-        if (core::CStringUtils::stringToType(traverser.value(), target) == false) { \
-            LOG_ERROR(<< "Failed to restore " #tag ", got " << traverser.value());  \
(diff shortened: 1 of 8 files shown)


include/core/RestoreMacros.h
```diff
@@ -10,44 +10,60 @@
...
-            return false;                                                          \
-        }                                                                          \
-        continue;                                                                  \
+#define RESTORE(tag, restore)                                                          \
+    if (name == tag) {                                                                 \
+        if ((restore) == false) {                                                      \
+            if (traverser.value().empty()) {                                           \
+                LOG_ERROR(<< "Failed to restore " #tag);                               \
+            } else {                                                                   \
+                LOG_ERROR(<< "Failed to restore " #tag ", got " << traverser.value()); \
```
tveasey: In many cases, i.e. sub-levels, we know that the value will be an empty string. This makes the messages easier to read in this case.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -276,25 +276,32 @@ CBoostedTreeFactory::constructFromParameters(std::size_t numberThreads,
...
-    TBoostedTreeUPtr treePtr{
-        new CBoostedTree{frame, TBoostedTreeImplUPtr{new CBoostedTreeImpl{}}}};
-    core::CJsonStateRestoreTraverser traverser(jsonStringStream);
-    treePtr->acceptRestoreTraverser(traverser);
-    return treePtr;
+    try {
+        TBoostedTreeUPtr treePtr{
+            new CBoostedTree{frame, TBoostedTreeImplUPtr{new CBoostedTreeImpl{}}}};
+        core::CJsonStateRestoreTraverser traverser(jsonStringStream);
+        if (treePtr->acceptRestoreTraverser(traverser) == false) {
```
droberts195: Should probably be:

```
    if (treePtr->acceptRestoreTraverser(traverser) == false ||
        traverser.haveBadState()) {
```

I think this might be needed to pick up every possible error.  It would match: https://github.com/elastic/ml-cpp/blob/55a00f7004a55eca9bf6188190a8125679b08ef6/lib/api/CAnomalyJob.cc#L961-L962

tveasey: Done in [b3f6533](https://github.com/elastic/ml-cpp/pull/577/commits/b3f65336b92343c73d410e44998482efd0840906).

PR #579, opened 2019-08-01
[ML] Fix return type of L-BFGS lambdas

droberts195:
Test failures in the debug build show that a couple
of lambdas in the CLbfgs class were returning Eigen
closure types referencing temporaries that could lead
to invalid memory accesses.

This change makes the lambdas construct self-contained
types before returning to avoid accesses to out-of-scope
temporaries.

(We believe that with optimization enabled the invalid
memory accesses were not occurring as the lambdas were
optimized away.)


droberts195:
retest

PR #584, opened 2019-08-02
Introduce memory usage estimation mode in data_frame_analyzer

przemekwitek:
Introduce a mode in which data_frame_analyzer binary only outputs memory usage estimations but does not perform any analysis. This mode will be used by the new endpoint in Java server.

Relates https://github.com/elastic/elasticsearch/issues/44699

### DIFF
--- a/bin/data_frame_analyzer/Main.cc
+++ b/bin/data_frame_analyzer/Main.cc
@@ -142,12 +142,11 @@ int main(int argc, char** argv) {
         std::make_unique<ml::api::CDataFrameAnalysisSpecification>(analysisSpecificationJson);
 
     if (memoryUsageEstimationOnly) {
-        const auto result = analysisSpecification->estimateMemoryUsage();
         auto outStream = [&ioMgr]() {
             return std::make_unique<ml::core::CJsonOutputStreamWrapper>(ioMgr.outputStream());
         }();
-        ml::api::CMemoryUsageEstimationResultJsonWriter jsonWriter(*outStream);
-        jsonWriter.write(result);
+        ml::api::CMemoryUsageEstimationResultJsonWriter writer(*outStream);
+        analysisSpecification->estimateMemoryUsage(writer);
         return EXIT_SUCCESS;
     }
 
--- a/bin/data_frame_analyzer/Main.cc
+++ b/bin/data_frame_analyzer/Main.cc
@@ -30,6 +30,7 @@
 #include <api/CDataFrameOutliersRunner.h>
 #include <api/CIoManager.h>
 #include <api/CLengthEncodedInputParser.h>
+#include <api/CMemoryUsageEstimationResultJsonWriter.h>
 
 #include "CCmdLineParser.h"
 
@@ -86,6 +87,7 @@ int main(int argc, char** argv) {
 
     // Read command line options
     std::string configFile;
+    bool memoryUsageEstimationOnly(false);
     std::string logProperties;
     std::string logPipe;
     bool lengthEncodedInput(false);
@@ -94,7 +96,7 @@ int main(int argc, char** argv) {
     std::string outputFileName;
     bool isOutputFileNamedPipe(false);
     if (ml::data_frame_analyzer::CCmdLineParser::parse(
-            argc, argv, configFile, logProperties, logPipe, lengthEncodedInput, inputFileName,
+            argc, argv, configFile, memoryUsageEstimationOnly, logProperties, logPipe, lengthEncodedInput, inputFileName,
             isInputFileNamedPipe, outputFileName, isOutputFileNamedPipe) == false) {
         return EXIT_FAILURE;
     }
@@ -127,13 +129,6 @@ int main(int argc, char** argv) {
     }
 
     using TInputParserUPtr = std::unique_ptr<ml::api::CInputParser>;
-    auto inputParser{[lengthEncodedInput, &ioMgr]() -> TInputParserUPtr {
-        if (lengthEncodedInput) {
-            return std::make_unique<ml::api::CLengthEncodedInputParser>(ioMgr.inputStream());
-        }
-        return std::make_unique<ml::api::CCsvInputParser>(
-            ioMgr.inputStream(), ml::api::CCsvInputParser::COMMA);
-    }()};
 
     std::string analysisSpecificationJson;
     bool couldReadConfigFile;
@@ -145,6 +140,17 @@ int main(int argc, char** argv) {
 
     auto analysisSpecification =
         std::make_unique<ml::api::CDataFrameAnalysisSpecification>(analysisSpecificationJson);
+
+    if (memoryUsageEstimationOnly) {
+        const auto result = analysisSpecification->estimateMemoryUsage();
+        auto outStream = [&ioMgr]() {
+            return std::make_unique<ml::core::CJsonOutputStreamWrapper>(ioMgr.outputStream());
+        }();
+        ml::api::CMemoryUsageEstimationResultJsonWriter jsonWriter(*outStream);
+        jsonWriter.write(result);
+        return EXIT_SUCCESS;
+    }
+
     if (analysisSpecification->numberThreads() > 1) {
         ml::core::startDefaultAsyncExecutor(analysisSpecification->numberThreads());
     }
@@ -156,6 +162,13 @@ int main(int argc, char** argv) {
 
     CCleanUpOnExit::add(dataFrameAnalyzer.dataFrameDirectory());
 
+    auto inputParser{[lengthEncodedInput, &ioMgr]() -> TInputParserUPtr {
+        if (lengthEncodedInput) {
+            return std::make_unique<ml::api::CLengthEncodedInputParser>(ioMgr.inputStream());
+        }
+        return std::make_unique<ml::api::CCsvInputParser>(
+            ioMgr.inputStream(), ml::api::CCsvInputParser::COMMA);
+    }()};
     if (inputParser->readStreamIntoVecs(
             [&dataFrameAnalyzer](const auto& fieldNames, const auto& fieldValues) {
                 return dataFrameAnalyzer.handleRecord(fieldNames, fieldValues);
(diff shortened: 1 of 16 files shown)


lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -36,6 +37,19 @@ CDataFrameAnalysisRunner::~CDataFrameAnalysisRunner() {
...
+SMemoryUsageEstimationResult CDataFrameAnalysisRunner::estimateMemoryUsage() const {
+    std::size_t numberRows{m_Spec.numberRows()};
+    std::size_t numberColumns{m_Spec.numberColumns() + this->numberExtraColumns()};
+    std::size_t maximumNumberPartitions{
+        static_cast<std::size_t>(std::sqrt(static_cast<double>(numberRows)) + 0.5)};
+
+    std::size_t memoryUsageWithOnePartition{
+        this->estimateMemoryUsage(numberRows, numberRows, numberColumns)};
+    std::size_t memoryUsageWithMaxPartitions{
+        this->estimateMemoryUsage(numberRows, numberRows / maximumNumberPartitions, numberColumns)};
```
tveasey: Whilst I think it is very likely that the memory usage decreases monotonically with increasing number of partitions I wonder whether we should just factor out the loop from `CDataFrameAnalysisRunner::computeAndSaveExecutionStrategy` into a shared private function, say `estimateMinimumMemoryUsage`, and return the minimum memory limit calculated in the loop. This means that these results keeps step with the logic which is applied there, which it must. Also, the compute cost of the loop should be very low compared to overheads. A related question mark is whether this should respect `m_Spec.diskUsageAllowed()`.

Also this should apply rounding I think, as per [this](https://github.com/elastic/ml-cpp/blob/master/lib/api/CDataFrameAnalysisRunner.cc#L83) line.

bin/data_frame_analyzer/Main.cc
```diff
@@ -145,6 +140,17 @@ int main(int argc, char** argv) {
...
     auto analysisSpecification =
         std::make_unique<ml::api::CDataFrameAnalysisSpecification>(analysisSpecificationJson);
+
+    if (memoryUsageEstimationOnly) {
+        const auto result = analysisSpecification->estimateMemoryUsage();
+        auto outStream = [&ioMgr]() {
+            return std::make_unique<ml::core::CJsonOutputStreamWrapper>(ioMgr.outputStream());
+        }();
+        ml::api::CMemoryUsageEstimationResultJsonWriter jsonWriter(*outStream);
+        jsonWriter.write(result);
```
tveasey: I think it might be nicer to move logic in main for writing the result, i.e.
```
auto outStream = [&ioMgr]() {
            return std::make_unique<ml::core::CJsonOutputStreamWrapper>(ioMgr.outputStream());
        }();
        ml::api::CMemoryUsageEstimationResultJsonWriter jsonWriter(*outStream);
        jsonWriter.write(result);
```
into the function on `CDataFrameAnalysisRunner` then you avoid the need for `SMemoryUsageEstimationResult` altogether. This is also more consistent with the rest of the code which passes off the actual result writing the objects.

przemekwitek:
> I did a quick pass through. Looking pretty good. I know you're working on tests. I've made a couple of suggestions which might have some knock on effects for testing, so thought I'd just go ahead and submit these.

Thanks. I've just sent this PR to regular review and will address your comments shortly.

lib/api/CDataFrameAnalysisRunner.cc
tveasey: Note we discussed this further offline and decided that we don't want to respect `m_Spec.diskUsageAllowed()` here in which case there is less reason to try and share this code with `CDataFrameAnalysisRunner::computeAndSaveExecutionStrategy`.

include/api/CDataFrameAnalysisRunner.h
```diff
@@ -75,6 +76,11 @@ class API_EXPORT CDataFrameAnalysisRunner {
     //! number of rows per subset.
     void computeAndSaveExecutionStrategy();
 
+    //! Estimates memory usage in two cases: one partition (the whole data frame
+    //! fits in main memory) and maximum tolerable number of partitions (only
+    //! one partition needs to be loaded to main memory.
```
tveasey: typo
```suggestion
    //! one partition needs to be loaded to main memory).
```

include/api/CDataFrameAnalysisSpecification.h
```diff
@@ -151,6 +151,11 @@ class API_EXPORT CDataFrameAnalysisSpecification {
     //! calling thread until the runner has finished.
     CDataFrameAnalysisRunner* run(core::CDataFrame& frame) const;
 
+    //! Estimates memory usage in two cases: one partition (the whole data frame
+    //! fits in main memory) and maximum tolerable number of partitions (only
+    //! one partition needs to be loaded to main memory.
```
tveasey: typo
```suggestion
    //! one partition needs to be loaded to main memory).
```

include/api/CMemoryUsageEstimationResultJsonWriter.h
```diff
@@ -0,0 +1,42 @@
...
+
+//! \brief
+//! Write memory usage estimation result in JSON format
+//!
+//! DESCRIPTION:\n
+//! Outputs the memory usage estimation result.
+//!
+class API_EXPORT CMemoryUsageEstimationResultJsonWriter : private core::CNonCopyable {
+public:
+    //! Constructor that causes output to be written to the specified wrapped stream
```
tveasey: This reads slightly wrong to me, i.e. that the writing out happens in the constructor. How about
```suggestion
    //! \param[in] strmOut The specified wrapped stream to which to write output.
```

lib/api/CDataFrameAnalysisSpecification.cc
```diff
@@ -183,6 +184,11 @@ CDataFrameAnalysisRunner* CDataFrameAnalysisSpecification::run(core::CDataFrame&
     return nullptr;
 }
 
+void CDataFrameAnalysisSpecification::estimateMemoryUsage(CMemoryUsageEstimationResultJsonWriter& resultWriter) const {
+    // TODO: What to do if m_Runner is nullptr?
```
tveasey:
```suggestion
    if (m_Runner == nullptr) {
        HANDLE_FATAL("Internal error: no runner available so can't estimate memory. Please report this problem.");
        return;
    }
```

lib/api/CDataFrameAnalysisRunner.cc
tveasey: IMO it is odd to output the estimate to the nearest byte (imagine reporting you need 49991231 bytes to run this analysis). I think it also gives the wrong impression about how accurate the memory is. So, I'd be inclined to round up to the nearest MB/100.

include/api/CDataFrameAnalysisRunner.h
przemekwitek: Done.

include/api/CDataFrameAnalysisSpecification.h
przemekwitek: Done.

include/api/CMemoryUsageEstimationResultJsonWriter.h
przemekwitek: Done.

lib/api/CDataFrameAnalysisSpecification.cc
przemekwitek: Done.

lib/api/CDataFrameAnalysisRunner.cc
droberts195: When `model_memory_limit` is configured via the Elasticsearch `ByteSizeValue` functionality it must be specified as a whole number of some unit.  So something like `"4.56mb"` will not parse using the functionality of `ByteSizeValue`.  Therefore I think we should pick a unit and round up to whole multiples of that unit.  Otherwise we're just creating a need to do extra processing further along the chain.

Anomaly detection jobs have to have `model_memory_limit`s that are whole numbers of megabytes.  We could stick with that for data frame analytics or go to kilobytes I guess.  I am not sure there is _that_ much benefit in going to kilobytes because at such a small level the overhead of running the process dominates the memory required by the analytics.  But if people want something more granular than megabytes then I'm happy with kilobytes.

lib/api/CMemoryUsageEstimationResultJsonWriter.cc
```diff
@@ -0,0 +1,35 @@
...
+    // the moment, the output stream might be redirected after construction
+}
+
+void CMemoryUsageEstimationResultJsonWriter::write(std::size_t expectedMemoryUsageWithOnePartition,
+                                                   std::size_t expectedMemoryUsageWithMaxPartitions) {
+    m_Writer.StartObject();
+    m_Writer.Key(EXPECTED_MEMORY_USAGE_WITH_ONE_PARTITION);
+    m_Writer.Uint64(expectedMemoryUsageWithOnePartition);
+    m_Writer.Key(EXPECTED_MEMORY_USAGE_WITH_MAX_PARTITIONS);
+    m_Writer.Uint64(expectedMemoryUsageWithMaxPartitions);
```
droberts195: If we're going to round to some unit other than bytes then we should return strings here that will be parsed by `ByteSizeValue` on the Elasticsearch side.  For example `"47mb"` or `"7455kb"`.

lib/api/CDataFrameAnalysisRunner.cc
przemekwitek: Done (rounding up to the nearest kilobytes).
Is it ok to still return the results as just bytes? On Java side I can just create ByteSizeValue using public constructor:
    public ByteSizeValue(long bytes) {...}

lib/api/CMemoryUsageEstimationResultJsonWriter.cc
przemekwitek: Please see my other comment.

lib/api/CDataFrameAnalysisRunner.cc
przemekwitek: Scratch that, I'll just use strings in both Java-cpp protocol and public API

przemekwitek: Done.

lib/api/CMemoryUsageEstimationResultJsonWriter.cc
przemekwitek: Done.

PR #587, opened 2019-08-06
[ML] Allow configuration of prediction column name in data_frame_analyzer

tveasey:
Plus a better default.

### DIFF
--- a/lib/api/CDataFrameBoostedTreeRunner.cc
+++ b/lib/api/CDataFrameBoostedTreeRunner.cc
@@ -24,7 +24,7 @@ namespace api {
 namespace {
 // Configuration
 const std::string DEPENDENT_VARIABLE_NAME{"dependent_variable"};
-const std::string PREDICTION_COLUMN_NAME{"prediction_column_name"};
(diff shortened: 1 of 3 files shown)


lib/api/CDataFrameBoostedTreeRunner.cc
```diff
@@ -23,7 +23,8 @@ namespace ml {
 namespace api {
 namespace {
 // Configuration
-const std::string DEPENDENT_VARIABLE{"dependent_variable"};
+const std::string DEPENDENT_VARIABLE_NAME{"dependent_variable"};
+const std::string PREDICTION_COLUMN_NAME{"prediction_column_name"};
```
dimitris-athanasiou: As this is user facing, I think we might want to avoid the use of `column`. How about `prediction_field_name`?

tveasey: Ok, good suggested alternative. Done in [4a92476](https://github.com/elastic/ml-cpp/pull/587/commits/4a924768058c9c92625345135c4d18beb5ad0000).

PR #593, opened 2019-08-14
[ML] Make the ML native code more ".app"-like on macOS

droberts195:
Successful notarization and acceptance by Gatekeeper on macOS
Catalina requires that the ML programs and dynamic libraries be
arranged into the directory structure used by macOS apps.


mk/make_info_plist.sh
```diff
@@ -0,0 +1,24 @@
...
+ML_TARGET=$1
+ML_SHORTENED_TARGET=`echo $ML_TARGET | cut -c1-15`
+
+cat <<EOF
+<plist version="1.0">
+<dict>
+    <key>CFBundleDevelopmentRegion</key>
+    <string>English</string>
+    <key>CFBundleIdentifier</key>
+    <string>co.elastic.ml-cpp.$ML_TARGET</string>
```
droberts195: Exactly what to use here beyond `co.elastic` is not decided.  I've used the repo name followed by the program name as a start.

mk/make_info_plist.sh
```diff
@@ -0,0 +1,24 @@
...
+    <key>CFBundleInfoDictionaryVersion</key>
+    <string>6.0</string>
+    <key>CFBundleName</key>
+    <string>$ML_SHORTENED_TARGET</string>
+    <key>CFBundleDisplayName</key>
+    <string>$ML_TARGET</string>
+    <key>CFBundleVersion</key>
+    <string>$ML_VERSION_NUM</string>
```
droberts195: This deliberately excludes qualifiers like `beta1` or `SNAPSHOT` - the field only tolerates digits and periods.

set_env.sh
```diff
@@ -36,7 +36,7 @@ case `uname` in
             if [ -z "$CPP_CROSS_COMPILE" ] ; then
                 BUNDLE_PLATFORM=linux-x86_64
             else
-                if [ "$CPP_CROSS_COMPILE" = macosx ] ; then
+                if [ "$CPP_CROSS_COMPILE" = macos ] ; then
```
edsavage: This will also require a change to `CPP_CROSS_COMPILE` in `dev-tools//docker/macosx_builder/Dockerfile`

droberts195: There are actually quite a few places I'll have to change to switch `CPP_CROSS_COMPILE` from using `macosx` to `macos`, and it will introduce an annoying discrepancy between branches in the Jenkins infrastructure.  So I think I'll revert this one place in this PR.  Maybe a future PR can do the rename fully.

droberts195:
I think this is ready now, but this PR must not be merged until a few days after elastic/elasticsearch#47013

droberts195:
elastic/elasticsearch#47104 needs to be merged a few days after this

PR #607, opened 2019-08-21
[ML] Instrument the regression code for memory usage and training run time

tveasey:
This instruments the regression code to extract peak memory usage and training run time. This enabled me to test memory usage estimation better and I've fixed a bug where we weren't properly accounting for the maximum tree and forest size.

### DIFF
--- a/include/maths/CBoostedTreeImpl.h
+++ b/include/maths/CBoostedTreeImpl.h
@@ -434,6 +434,11 @@ private:
                 static_cast<std::size_t>(std::ceil(
                     featureBagFraction * static_cast<double>(numberCols - 1))) *
                 sizeof(std::size_t)};
+            // 256 is the number of rows encoded by a single byte in the packed bit
+            // vector assuming best compression. We will typically get this for most
+            // of the leaves when the set of splits becomes large, corresponding to
+            // the worst case for memory usage. This is because the masks will mainly
+            // contain 0 bits in this case.
             std::size_t rowMaskSize{numberRows / 256};
             std::size_t gradientsSize{(numberCols - 1) *
                                       numberSplitsPerFeature * sizeof(double)};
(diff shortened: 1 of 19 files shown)


include/maths/CBoostedTreeImpl.h
```diff
@@ -402,17 +412,29 @@ class MATHS_EXPORT CBoostedTreeImpl final {
...
+                                               std::size_t numberCols,
+                                               double featureBagFraction,
+                                               std::size_t numberSplitsPerFeature) {
             std::size_t featureBagSize{
-                static_cast<std::size_t>(ceil(featureBagFraction) * (numberCols - 1)) *
+                static_cast<std::size_t>(std::ceil(
+                    featureBagFraction * static_cast<double>(numberCols - 1))) *
                 sizeof(std::size_t)};
-            std::size_t rowMaskSize{numberRows / 32};
+            std::size_t rowMaskSize{numberRows / 256};
```
dimitris-athanasiou: Could we document this magic number?

lib/api/CDataFrameBoostedTreeRunner.cc
```diff
@@ -145,10 +156,19 @@ void CDataFrameBoostedTreeRunner::runImpl(const TStrVec& featureNames,
         return;
     }
 
+    core::CProgramCounters::counter(counter_t::E_DFTPMEstimatedPeakMemoryUsage) =
+        this->estimateMemoryUsage(frame.numberRows(),
+                                  frame.numberRows() / this->numberPartitions(),
+                                  frame.numberColumns());
+
+    core::CStopWatch watch{true};
```
dimitris-athanasiou: Should we not also add a watch tracking outlier detection duration?

tveasey: This is done at a lower level for outlier detection: see for example [this](https://github.com/elastic/ml-cpp/blob/master/lib/maths/COutliers.cc#L895) line.

include/maths/CBoostedTreeImpl.h
tveasey: See [68158b7](https://github.com/elastic/ml-cpp/pull/607/commits/68158b7f586e39cdd6e86b205ea9efab69b02d7e).

tveasey:
@dimitris-athanasiou I addressed your two comments. Could you take another look?

PR #609, opened 2019-08-22
[ML] Detect feature data types and refine candidate splits in boosted tree training based on type

tveasey:
This change detects if a feature comprises exclusively integers and also computes the feature range on the training set. For integer valued data we should be splitting at `n + 0.5` for `n` integer valued. Also we should never include splits which are less (greater) than the smallest (largest) feature value. This allows us to significantly prune candidate splits in cases where a feature only takes a small number of distinct integer values.

I also improved the exit condition from the loop which adds trees: it was exiting with significant bias on the training set in some cases.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -292,7 +292,7 @@ CBoostedTreeImpl::trainForest(core::CDataFrame& frame,
     //  4. Update predictions and loss derivatives
 
     double eta{m_Eta};
-    double sumEta{eta};
+    double oneMinusBias{eta};
 
     for (std::size_t retries = 0; forest.size() < m_MaximumNumberTrees; /**/) {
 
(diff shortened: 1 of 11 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -354,23 +359,35 @@ CBoostedTreeImpl::candidateSplits(const core::CDataFrame& frame,
...
         }
 
-        columnSplits.erase(std::unique(columnSplits.begin(), columnSplits.end()),
-                           columnSplits.end());
-        candidateSplits[features[i]] = std::move(columnSplits);
+        const auto& dataType = m_FeatureDataTypes[features[i]];
+
+        if (dataType.s_IsInteger) {
+            std::for_each(featureSplits.begin(), featureSplits.end(),
+                          [](double& split) { split = std::floor(split) + 0.5; });
```
benwtrent: I am not sure why we are doing a floor and then adding `0.5`. Why do we always want to split halfway on an integer? It seems to me that this is the exact same thing as calling `ceil` against the `split`, especially if all the data are integers.

tveasey: The rationale is mainly to snap splits to the same values so we don't needless "duplicate". When we compute quantiles we linearly interpolate so we tend to end up with many distinct values which result in identical partitions of the data set. By snapping to a value (the midpoint is a good as any other) when we subsequently remove duplicates we remove all splits which we know must be equivalent because they fall between two consecutive integers. (I could be cleverer here and actually check to see if there are values between candidate splits, but this requires an additional pass over the data so the runtime tradeoff becomes less clear especially since it depends on the feature distribution.)

tveasey: Do you think it'd be helpful to add a comment @benwtrent?

benwtrent: Yeah, I see that, a `1.1` is the same as a `1.2` as both `1.0` is less than and `2.0` is greater than. Having a comment explaining the "magical number" would be nice. 

Something like
// Since we are splitting Integers, all decimal values incur the same branch path. This allows  duplicate splits to be pruned

tveasey: See [fa42136](https://github.com/elastic/ml-cpp/pull/609/commits/fa421361eca9bf84d83d7cdeaa6859c547ada88a).

PR #610, opened 2019-08-22
[7.4][ML] Detect feature data types and refine candidate splits in boosted tree training based on type

tveasey:
Backport #609.


tveasey:
retest

PR #614, opened 2019-08-23
[ML] Check whether the splits will change before recomputing in loop building the regression forest

tveasey:
This gives us a 30-50% speedup for MSE objective based on the testing I've done.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -294,9 +294,9 @@ CBoostedTreeImpl::trainForest(core::CDataFrame& frame,
     double eta{m_Eta};
     double oneMinusBias{eta};
 
-    for (std::size_t retries = 0; forest.size() < m_MaximumNumberTrees; /**/) {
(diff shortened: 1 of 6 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -318,6 +318,10 @@ CBoostedTreeImpl::trainForest(core::CDataFrame& frame,
             forest.push_back(std::move(tree));
         }
         LOG_TRACE(<< "bias = " << (1.0 - oneMinusBias));
+
+        if (m_Loss->isCurvatureConstant() == false) {
```
benwtrent: Maybe I missed something, but `m_Loss->isCurvatureConstant() == false` will always be `false` since the only implementation has a constant return value of `true`. 

So within `trainForest`, we only ever calculate the candidateSplits once?

tveasey: It is always a constant for MSE. In general, it doesn't have to be. However, in this case there is no need to recompute splits we can get away with using the same candidates. The candidates are chosen to be fairly fine grained so selecting splits when building the tree from this set means the optimal (binary tree) should close to one that is actually available.

PR #622, opened 2019-08-29
[ML] Boosted tree training performance improvements

tveasey:
Profiling showed up a couple of performance improvements:
1. We can calculate the node with the fewest rows more efficiently when refreshing derivative statistics
2. We benefit from inlining and a couple of small tweaks in one of the bottleneck functions `addRowDerivatives`

It also showed it is worthwhile threading the calculation of the leaf values which minimise the prediction errors. This required a refactor of `CArgMinLoss` so that we can get a copyable polymorphic type. I've wrapped up its implementation to achieve this.

### DIFF
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -21,11 +21,6 @@ using namespace boosted_tree_detail;
 
 namespace boosted_tree {
 
-void CArgMinLoss::add(double prediction, double actual) {
-    std::unique_lock<std::mutex> lock{m_Mutex};
-    this->addImpl(prediction, actual);
-}
-
 double CMse::value(double prediction, double actual) const {
     return CTools::pow2(prediction - actual);
 }
@@ -48,7 +43,7 @@ const std::string& CMse::name() const {
     return NAME;
 }
 
-void CArgMinMse::addImpl(double prediction, double actual) {
+void CArgMinMse::add(double prediction, double actual) {
     m_MeanError.add(actual - prediction);
 }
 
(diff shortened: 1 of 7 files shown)


lib/maths/CBoostedTree.cc
```diff
@@ -17,13 +17,59 @@
...
 namespace maths {
-using namespace boosted_tree_detail;
+namespace boosted_tree_detail {
+
+std::unique_ptr<CArgMinLossImpl> CArgMinMseImpl::clone() const {
+    return std::make_unique<CArgMinMseImpl>(*this);
+}
+
+void CArgMinMseImpl::add(double prediction, double actual) {
+    m_MeanError.add(actual - prediction);
```
valeriy42: nit: IMO a better name for `m_MeanError` would be `m_Errors` since actual `mean` is computed only in L38.

tveasey: I take your point. The choice here was that this object is only maintaining (count, mean) statistics. I think if we called it `m_Errors` to me it implies we have a collection which seems misleading. It would probably be more accurate to call it something like `m_MeanErrorAccumulator`. The reason I'm reluctant to do that is I have so many examples of using this class in the code base where I've stuck to the convention of calling it `MeanFoo`. I think on balance I'd prefer to leave as is. WDYT?

tveasey:
Thanks @valeriy42!

lib/maths/CBoostedTree.cc
valeriy42: I see you point. I would probably call it `m_ErrorAccumulator` then, since `mean` is only one function you apply to it. But if this is a convention, then consistency should go first.

Issue #532, opened 2019-07-01
[ML] Replace log4cxx with Boost.Log


droberts195:
Currently we use log4cxx as our logging library.

The most recent release is still version 0.10.0 from 2008.  There is evidence of continued work on the library in its [repo](https://gitbox.apache.org/repos/asf?p=logging-log4cxx.git), but no indication of when a new version will be released.

log4cxx depends on the Apache Portable Runtime (APR), and this in turn depends on the expat XML library.  These libaries _are_ being maintained, and thus there is the requirement to update them (see for example #495).  But we do not use them except indirectly via log4cxx.

Because the released version of log4cxx is so old the source has to be modified to build with modern compilers, for example due to narrowing conversions not being permitted in braced initializers.  This situation will only become more problematic as time goes by, for example when we upgrade to C++17.

A good candidate for a replacement for log4cxx would seem to be [Boost.Log](https://www.boost.org/doc/libs/1_65_1/libs/log/doc/html/index.html).  The dependencies for this library are Boost.Filesystem, Boost.System, Boost.DateTime, Boost.Thread and Boost.Regex, all of which we currently use.  So switching to Boost.Log would get rid of 5 dependencies (log4cxx, apr, apr-iconv, apr-util, expat) whilst adding only one, namely Boost.Log itself.

We will need to write a custom formatter for Boost.Log to output JSON in the format expected by the Java log message handler to the redirected `stderr`.

We currently do not build Boost.Log when we build our Boost dependencies.  This dates back to the early days of Boost.Log when it pushed the limits of C++ compiler standards conformance and didn't build with the compilers we were using.  I imagine this is not a problem now.

So the first step here is to to regenerate our dependencies with Boost.Log included in the Boost build.  However, since we haven't upgraded Boost for a long time we should probably also upgrade to the latest version of Boost while doing this, and it will be best to wait for Boost version 1.71 (due in August) before doing this as it has support for Visual Studio 2019.

- [x] Upgrade the Boost dependency to 1.71 (or later if we don't get around to this soon) for all platforms and adjust build setup instructions
- [ ] Switch logging to use Boost.Log rather than log4cxx
- [ ] Remove all traces of log4cxx and its dependencies

droberts195:
The first step is done by #638

droberts195:
Steps 2 and 3 are done by #655

PR #638, opened 2019-09-03
[ML] Upgrade Boost to version 1.71

droberts195:
We have been using Boost version 1.65.1 for a long time
now, and it's good to upgrade occasionally.

Version 1.71 contains support for Visual Studio 2019,
so upgrading to this version will allow us to upgrade
our Windows compiler (in the future - not in this PR).

Previously we didn't build the Boost.Log library.  The
new instructions for building Boost include Boost.Log,
and this is a step on the way to replacing log4cxx
(see #532).

The final change is to delete the MUSL Docker images
for Linux.  These were not maintained or tested for a
long time and it would be setting a false expectation
that they're expected to work with the Boost upgrade
if they'd been updated in this change.


dev-tools/docker/macosx_image/Dockerfile
```diff
@@ -29,14 +29,15 @@ RUN \
...
-  wget --quiet -O - https://s3-eu-west-1.amazonaws.com/prelert-artifacts/dependencies/usr-x86_64-apple-macosx10.11-2.tar.bz2 | tar jxf - && \
+  wget --quiet -O - https://s3-eu-west-1.amazonaws.com/prelert-artifacts/dependencies/usr-x86_64-apple-macosx10.11-3.tar.bz2 | tar jxf - && \
   wget --quiet -O - https://s3-eu-west-1.amazonaws.com/prelert-artifacts/dependencies/xcode-x86_64-apple-macosx10.11-1.tar.bz2 | tar jxf -
 
 # Build cctools-port
 RUN \
   git clone https://github.com/tpoechtrager/cctools-port.git && \
   cd cctools-port/cctools && \
   git checkout 895-ld64-274.2 && \
+  sed -i -e 's/autoconf/autoreconf -fi/' autogen.sh && \
```
droberts195: This is needed because in https://github.com/tpoechtrager/cctools-port/commit/07619027f8311fa61b4a549c75994b88739a82d8 an obscure version of `ldtools` was used to regenerate the `autoconf` files that are commited to the `895-ld64-274.2` branch in the `cctools-port` repo.

lib/maths/unittest/CBayesianOptimisationTest.cc
```diff
@@ -260,7 +260,7 @@ void CBayesianOptimisationTest::testMaximumExpectedImprovement() {
     }
 
     LOG_DEBUG(<< "mean gain = " << maths::CBasicStatistics::mean(gain));
-    CPPUNIT_ASSERT(maths::CBasicStatistics::mean(gain) > 1.29);
+    CPPUNIT_ASSERT(maths::CBasicStatistics::mean(gain) > 1.27);
```
droberts195: I did a bit of investigation into why this was necessary.  The trace logging from this test running with [Boost 1.65.1](https://github.com/elastic/ml-cpp/files/3573964/1.65.1.txt) and with [Boost 1.71](https://github.com/elastic/ml-cpp/files/3573967/1.71.txt) first shows a discrepancy on line 587 of those files.  For Boost 1.65.1:

```
CBayesianOptimisation.cc@104 x =     0.46252    0.105039   -0.372538 -0.00963219 EI(x) = -0.007086
```

and for Boost 1.71:

```
CBayesianOptimisation.cc@104 x =     0.46252    0.105039   -0.372538 -0.00963219 EI(x) = -0.0070849
```

The difference in `EI(x)` is tiny, but once the difference has occurred it propagates, eventually causing completely different code paths to be taken.

It implies there is a difference in the values being returned for the normal distribution PDF or CDF between the two Boost versions.  There has been no change to `boost/math/distributions/normal.hpp`.  However, that file in turn depends on `boost/math/special_functions/erf.hpp` and `boost/math/special_functions/gamma.hpp`, and there have been changes to those files:

* https://github.com/boostorg/math/commit/e348578b0f70f559ac3692f97aa05ba9b3235449
* https://github.com/boostorg/math/commit/08b7a6145f90eff611a1afbdd21e9c54faebcc27
* https://github.com/boostorg/math/commit/0152cfa2a9f9a586568e1194b125f85d9bc9d870

So it's most likely one of those that causes the difference.

@tveasey are you happy to proceed with the upgrade?

tveasey: Yes, I think this is fine. Qualitatively the difference in the test result is small. The difference is likely due to some improvement to numerics of of one of these functions.

PR #652, opened 2019-09-06
[ML] Improve boosted tree training parameter initialisation 

### LINKED ISSUE #631: [ML] Failure to select initial lambda and gamma for regression if maxNumberOfTrees is set to 1
In `CBoostedTreeFactory::initializeHyperparameters` we want to test if we can improve on the dataset without regularization to make sure if it make sense to do hyperparameter optimization on `lambda` and `gamma`. However, if `maxNumberOfTrees` is set to 1, `trainForest()` will exit without training a single tree, which leads to a wrong impression that training does not improve the results.

tveasey:
We weren't properly centring the data when measuring the change in loss. For data where the variance across the training set is small compared to the mean this meant we could significantly over estimate the initial values for `lambda` and `gamma`.

I've also switched to using a single tree to estimate the magnitude of the change in the loss term of the objective. Using a full forest tests underestimates the change per split by roughly a factor of the shrinkage. Finally, rather than using a fixed scaling of `lambda` and `gamma` I scale them by the shrinkage on initialisation to account for the regularisation effect of smaller shrinkage.

This change also closes #631 since it was useful to be able to train a forest of size one for this change.

### DIFF
--- a/lib/maths/CBoostedTreeFactory.cc
+++ b/lib/maths/CBoostedTreeFactory.cc
@@ -250,18 +250,24 @@ void CBoostedTreeFactory::initializeHyperparameters(core::CDataFrame& frame) con
     } else {
         core::CPackedBitVector trainingRowMask{m_TreeImpl->allTrainingRowsMask()};
 
-        auto vector = m_TreeImpl->initializePredictionsAndLossDerivatives(frame, trainingRowMask);
+        auto tree = m_TreeImpl->initializePredictionsAndLossDerivatives(frame, trainingRowMask);
 
         double L[2];
         double T[2];
         double W[2];
 
         std::tie(L[0], T[0], W[0]) =
-            m_TreeImpl->regularisedLoss(frame, trainingRowMask, {std::move(vector)});
+            m_TreeImpl->regularisedLoss(frame, trainingRowMask, {std::move(tree)});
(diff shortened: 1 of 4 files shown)


lib/maths/CBoostedTreeFactory.cc
```diff
@@ -250,18 +250,24 @@ void CBoostedTreeFactory::initializeHyperparameters(core::CDataFrame& frame) con
...
+            m_TreeImpl->regularisedLoss(frame, trainingRowMask, {std::move(tree)});
         LOG_TRACE(<< "loss = " << L[0] << ", # leaves = " << T[0]
                   << ", sum square weights = " << W[0]);
 
+        double eta{1.0};
+        std::size_t maximumNumberOfTrees{1};
+        std::swap(eta, m_TreeImpl->m_Eta);
+        std::swap(maximumNumberOfTrees, m_TreeImpl->m_MaximumNumberTrees);
         auto forest = m_TreeImpl->trainForest(frame, trainingRowMask, m_RecordMemoryUsage);
+        std::swap(eta, m_TreeImpl->m_Eta);
```
dimitris-athanasiou: Why do we need to do the swaps 2 times?

tveasey: To restore the original values, i.e. first swap forces to 1 the second one restores the original value.

PR #655, opened 2019-09-09
[ML] Replace log4cxx with Boost.Log

### LINKED ISSUE #532: [ML] Replace log4cxx with Boost.Log

droberts195:
This change replaces log4cxx with Boost.Log.

The most recent release of log4cxx is still version 0.10.0
from 2008, and there is no indication of when a new
version will be released.

log4cxx depends on the Apache Portable Runtime (APR),
and this in turn depends on the expat XML library.  So
by replacing log4cxx with Boost.Log we get to remove
five libraries from our third party dependency chain.

The intention with this change is to preserve the old logging
functionality as much as possible.  In particular the JSON
log messages that are sent to the Elasticsearch JVM through
the logging named pipe remain close enough to the log4cxx
JSON documents that no changes to the Java code are
needed.

Closes #532


lib/core/CJsonLogLayout.cc
```diff
@@ -5,54 +5,50 @@
...
-const int64_t PID(static_cast<int64_t>(ml::core::CProcess::instance().id()));
-const std::string THREAD_NAME("thread");
-const std::string MESSAGE_NAME("message");
-const std::string NDC_NAME("ndc");
-const std::string CLASS_NAME("class");
-const std::string METHOD_NAME("method");
-const std::string FILE_NAME("file");
-const std::string LINE_NAME("line");
-const std::string PROPERTIES_NAME("properties");
+const int64_t PID{static_cast<std::int64_t>(ml::core::CProcess::instance().id())};
```
tveasey:
```suggestion
const std::int64_t PID{static_cast<std::int64_t>(ml::core::CProcess::instance().id())};
```

lib/core/CJsonLogLayout.cc
```diff
@@ -63,132 +59,103 @@ bool CJsonLogLayout::locationInfo() const {
...
     writer.StartObject();
 
     writer.String(LOGGER_NAME);
-    LOG4CXX_ENCODE_CHAR(logger, event->getLoggerName());
-    writer.String(logger);
+    writer.String(LOGGER);
 
     writer.String(TIMESTAMP_NAME);
-    writer.Int64(event->getTimeStamp() / 1000);
+    const boost::posix_time::ptime& timeStamp{
```
tveasey: nit: the uses the uses of extract in this file it seems like a good candidates for auto since the extract already defines the values' types
```suggestion
    const auto& timeStamp =
```

lib/core/CJsonLogLayout.cc
```diff
@@ -63,132 +59,103 @@ bool CJsonLogLayout::locationInfo() const {
...
-        LOG4CXX_ENCODE_CHAR(ndc, logNdc);
-        writer.String(ndc);
-    }
-
     if (m_LocationInfo) {
-        const spi::LocationInfo& locInfo = event->getLocationInformation();
 
-        const std::string& className = locInfo.getClassName();
-        if (!className.empty()) {
+        std::string functionName{boost::log::extract<std::string>(
```
tveasey: No need to copy?

lib/core/CJsonLogLayout.cc
```diff
@@ -63,132 +59,103 @@ bool CJsonLogLayout::locationInfo() const {
...
+
+        std::size_t argsStartPos{functionName.find('(')};
+        if (argsStartPos != std::string::npos) {
+            functionName.erase(argsStartPos);
+        }
+        std::size_t returnTypeSeparatorPos{functionName.rfind(' ')};
+        if (returnTypeSeparatorPos != std::string::npos) {
+            functionName.erase(0, returnTypeSeparatorPos + 1);
+        }
+        std::size_t lastScopeOperatorPos{functionName.rfind("::")};
```
tveasey: Maybe factor out into cc local method or `CJsonLogLayout` helper for readability.

lib/core/CLogger.cc
```diff
@@ -6,61 +6,121 @@
...
       m_OrigStderrFd(-1), m_FatalErrorHandler(defaultFatalErrorHandler) {
     CCrashHandler::installCrashHandler();
+    // These formatter factories are not needed for programmatic configuration
+    // of the format, but without them formats defined in settings files don't
+    // work correctly.
+    boost::log::register_simple_formatter_factory<ELevel, char>("Severity");
+    boost::log::register_simple_filter_factory<ELevel, char>("Severity");
+    boost::log::register_formatter_factory(
+        "TimeStamp", boost::make_shared<CTimeStampFormatterFactory>());
+    auto loggingCorePtr{boost::log::core::get()};
```
tveasey: With auto I always use copy initialisation to avoid having to think about the `auto-deduced-a-std::initializer_list` problem. Maybe stick to that convention in this file?
```suggestion
    auto loggingCorePtr = boost::log::core::get();
```

lib/core/CLogger.cc
```diff
@@ -6,61 +6,121 @@
...
+        boost::log::attributes::current_thread_id());
     this->reset();
 }
 
 CLogger::~CLogger() {
-    log4cxx::LogManager::shutdown();
-    m_Logger = nullptr;
+    auto loggingCorePtr{boost::log::core::get()};
+    loggingCorePtr->flush();
+    loggingCorePtr->remove_all_sinks();
```
tveasey: Feels like we could factor out a cc local `resetSink(shared_ptr< sinks::sink >)` function which if it is supplied `nullptr` just calls `flush` and `remove_all_sinks` on `boost::log::core::get()`.

include/core/CLogger.h
```diff
@@ -11,24 +11,23 @@
...
-#include <log4cxx/level.h>
-#include <log4cxx/logger.h>
+#include <boost/log/sources/severity_logger.hpp>
 
 #include <functional>
-#include <memory>
-#include <stdio.h>
+#include <iosfwd>
+
+#include <stdio.h> // fileno() is not C++ so need the C header
```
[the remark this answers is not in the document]
droberts195: I tried it and it didn't work, hence the comment `// fileno() is not C++ so need the C header`.  `fileno` is actually not even in the C standard, which I guess is why it hasn't been pulled across in the C++ wrappers around C headers.  It's POSIX, but they've put it in a C header.

include/core/LogMacros.h
```diff
@@ -76,16 +105,19 @@
...
-    LOG4CXX_FATAL(ml::core::CLogger::instance().logger(), "" message);         \
+#define LOG_ABORT(message)                                                                   \
+    BOOST_LOG_STREAM_SEV(ml::core::CLogger::instance().logger(), ml::core::CLogger::E_Fatal) \
+    LOG_LOCATION_INFO                                                                        \
+    message;                                                                                 \
     ml::core::CLogger::fatal()
 
 // Log at a level specified at runtime as a string, for example
-// LOG_AT_LEVEL("WARN", << "Stay away from here " << username)
+// LOG_AT_LEVEL(ml::core::CLogger::E_Warn, << "Stay away from here " << username)
```
edsavage: nit: this comment requires updating, now the level is specified as an enum, not a string

lib/core/CJsonLogLayout.cc
edsavage: suggestion: pull out the functionName processing into a separate function - it's probably worthy of a unit test too

droberts195: Bits of the string are erased.  This is to try to preserve the functionality log4cxx had to split the compiler's `__PRETTY_FUNCTION__` or `__FUNCSIG__` into separate class and a method values.

include/core/CLogger.h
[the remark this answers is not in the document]
tveasey: Ok great, now I get the comment!

lib/core/CJsonLogLayout.cc
tveasey: Right, yes I missed that actual modifications: was thinking you were just searching it.

3rd_party/3rd_party.sh
```diff
@@ -292,27 +201,5 @@ case `uname` in
         fi
         ;;
 
-    SunOS)
```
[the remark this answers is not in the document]
droberts195: There was a previous PR in the old repo to remove Solaris support but it missed this bit.  This is just removing a bit of dead code since I'm refactoring the file so much anyway.  It's not like this PR is sneaking in removal of support for a platform.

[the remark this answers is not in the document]
edsavage: Yeah, I get that :-) Just shedding a tear for the passing of an old friend...

PR #658, opened 2019-09-11
[ML] Refactoring of the persistence for regression

valeriy42:
This PR introduces following changes:
- Persister is writing `{"index"...}` and `\0` for every persistence state to comply with requirements for bulk operations
- `JobId` is passed over JSON Specification instead of the command line
- Constants for `ML_STATE` and `REGRESSION_TRAIN_STATE_TYPE` are moved to a dedicated header file

### DIFF
--- a/include/api/ElasticsearchStateIndex.h
+++ b/include/api/ElasticsearchStateIndex.h
@@ -14,6 +14,10 @@ namespace api {
 static const std::string ML_STATE_INDEX(".ml-state");
 static const std::string MODEL_STATE_TYPE("model_state");
 static const std::string REGRESSION_TRAIN_STATE_TYPE("predictive_model_train_state");
+
+std::string getRegressionStateId(const std::string& jobId) {
+    return jobId + '_' + REGRESSION_TRAIN_STATE_TYPE;
+}
 }
 }
 
--- a/include/api/ElasticsearchStateIndex.h
+++ b/include/api/ElasticsearchStateIndex.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#ifndef INCLUDED_ml_api_ElasticsearchStateIndex_h
+#define INCLUDED_ml_api_ElasticsearchStateIndex_h
+
+#include <string>
+
+namespace ml {
+namespace api {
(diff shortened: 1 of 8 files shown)


include/api/ElasticsearchStateIndex.h
```diff
@@ -0,0 +1,20 @@
...
+ */
+#ifndef INCLUDED_ml_api_ElasticsearchStateIndex_h
+#define INCLUDED_ml_api_ElasticsearchStateIndex_h
+
+#include <string>
+
+namespace ml {
+namespace api {
+//! Elasticsearch index for state
+static const std::string ML_STATE_INDEX(".ml-state");
```
dimitris-athanasiou: We should delete this. The c++ should not need to know which index we're persisting into.

lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -197,10 +199,18 @@ void CDataFrameAnalysisRunner::setToFinished() {
...
 CDataFrameAnalysisRunner::TStatePersister CDataFrameAnalysisRunner::statePersister() {
     return [this](std::function<void(core::CStatePersistInserter&)> persistFunction) -> void {
-        auto persistStream = m_Spec.persistStream();
-        if (persistStream != nullptr) {
-            core::CJsonStatePersistInserter inserter{*persistStream};
-            persistFunction(inserter);
+        auto persister = m_Spec.persister();
+        if (persister != nullptr) {
+            auto persistStream = persister->addStreamed(
+                ML_STATE_INDEX, m_Spec.jobId() + '_' + REGRESSION_TRAIN_STATE_TYPE);
```
dimitris-athanasiou: Let's just pass an empty string here instead of the index. The `CSingleStreamDataAdder` does not make use of the argument.

droberts195: I'd rather do this in a followup PR (which I can do at some point) that also changes other places where we reference the `.ml-state` index in the C++ code.  This method doesn't reference the `CSingleStreamDataAdder` class, so assuming it's the underlying implementation is breaching encapsulation and will trip up anyone who tries to use this code with a different type of data adder for test/debugging purposes.

dimitris-athanasiou: Sounds good.

include/api/ElasticsearchStateIndex.h
dimitris-athanasiou: Ignore this in the light of Dave's comment to do this as a refactoring later.

PR #660, opened 2019-09-11
[ML] Restore regression state from named pipe

valeriy42:
This PR implements restoring the Analyzer from a data in named pipe. It also introduces the usage of `CStateCompressor` and `CStateDecompressor` for compression and decompression of the state.

As a minor refactoring, memory estimation callback was moved from a lambda function into a dedicated method for symmetry with other callback functions.


include/api/CDataFrameAnalysisRunner.h
```diff
@@ -148,6 +148,9 @@ class API_EXPORT CDataFrameAnalysisRunner {
     //! \return Callback function for writing state using given persist inserter
     TStatePersister statePersister();
 
+protected:
```
dimitris-athanasiou: Do we need this? There is already a protected `spec()` function that returns it. If we do need it for some reason then should we remove the function?

valeriy42: Done

include/api/CDataFrameBoostedTreeRunner.h
```diff
@@ -61,6 +64,9 @@ class API_EXPORT CDataFrameBoostedTreeRunner final : public CDataFrameAnalysisRu
     TBoostedTreeFactoryUPtr m_BoostedTreeFactory;
     TBoostedTreeUPtr m_BoostedTree;
     std::atomic<std::int64_t> m_Memory;
+
+    bool restoreBoostedTree(core::CDataFrame& frame,
+                            CDataFrameAnalysisSpecification::TDataSearcherUPtr& restoreSearcher);
```
tveasey: nit: as per coding standards, can we keep this in the private member functions section.

include/api/ElasticsearchStateIndex.h
```diff
@@ -15,7 +15,7 @@ static const std::string ML_STATE_INDEX(".ml-state");
 static const std::string MODEL_STATE_TYPE("model_state");
 static const std::string REGRESSION_TRAIN_STATE_TYPE("predictive_model_train_state");
 
-std::string getRegressionStateId(const std::string& jobId) {
+static std::string getRegressionStateId(const std::string& jobId) {
```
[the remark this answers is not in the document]
tveasey: As with static constants seems better not to duplicate into every cc it's included into.

valeriy42:
retest

lib/api/unittest/CDataFrameAnalyzerTest.cc
```diff
@@ -1191,3 +1189,54 @@ void CDataFrameAnalyzerTest::testRunBoostedTreeTrainingWithStateRecoverySubrouti
...
+    CTestDataSearcher dataSearcher(persistedStates.back());
+    auto decompressor{std::make_unique<core::CStateDecompressor>(dataSearcher)};
+    decompressor->setStateRestoreSearch(api::ML_STATE_INDEX,
+                                        api::getRegressionStateId("testJob"));
+    auto stream{decompressor->search(1, 1)};
+    return maths::CBoostedTreeFactory::constructFromString(*stream, *frame);
+}
+
+std::unique_ptr<core::CDataFrame>
+CDataFrameAnalyzerTest::passDataToAnalyzer(const TDoubleVec& weights,
```
tveasey: My feeling is if we're going to factor this out we should have a utility which allows you define the target relationship as a callback and use it in all test code which adds data for regression, i.e. something like `setupRegressionData`. As it is there are some unit tests which generate and pass data to the analyser and one that doesn't.

include/api/CDataFrameBoostedTreeRunner.h
```diff
@@ -52,6 +52,7 @@ class API_EXPORT CDataFrameBoostedTreeRunner final : public CDataFrameAnalysisRu
                                                std::size_t totalNumberRows,
                                                std::size_t partitionNumberRows,
                                                std::size_t numberColumns) const override;
+    std::function<void(int64_t)> memoryEstimator();
```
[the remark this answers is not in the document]
valeriy42: done

include/api/ElasticsearchStateIndex.h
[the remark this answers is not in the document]
valeriy42: Done.

lib/api/CDataFrameBoostedTreeRunner.cc
```diff
@@ -142,6 +131,19 @@ CDataFrameBoostedTreeRunner::CDataFrameBoostedTreeRunner(const CDataFrameAnalysi
     }
 }
 
+std::function<void(int64_t)> CDataFrameBoostedTreeRunner::memoryEstimator() {
```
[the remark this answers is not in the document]
valeriy42: done

include/api/CDataFrameBoostedTreeRunner.h
valeriy42: Done. Unfortunately, CLion misplaces the extracted methods sometimes.

lib/api/unittest/CDataFrameAnalyzerTest.cc
valeriy42: Let's discuss your idea offline to avoid misunderstandings.

valeriy42: Done

include/api/ElasticsearchStateIndex.h
```diff
@@ -11,13 +11,11 @@
 namespace ml {
 namespace api {
 //! Elasticsearch index for state
-static const std::string ML_STATE_INDEX(".ml-state");
-static const std::string MODEL_STATE_TYPE("model_state");
-static const std::string REGRESSION_TRAIN_STATE_TYPE("predictive_model_train_state");
+extern const std::string ML_STATE_INDEX;
```
tveasey: Windows link errors are caused by these needing `API_EXPORT`.
```suggestion
API_EXPORT
extern const std::string ML_STATE_INDEX;
```

include/api/ElasticsearchStateIndex.h
```diff
@@ -11,13 +11,11 @@
...
+extern const std::string MODEL_STATE_TYPE;
+extern const std::string REGRESSION_TRAIN_STATE_TYPE;
```
tveasey: As above

include/api/ElasticsearchStateIndex.h
```diff
@@ -11,13 +11,11 @@
...
 
-std::string getRegressionStateId(const std::string& jobId) {
-    return jobId + '_' + REGRESSION_TRAIN_STATE_TYPE;
-}
+std::string getRegressionStateId(const std::string& jobId);
```
tveasey:
```suggestion
API_EXPORT
std::string getRegressionStateId(const std::string& jobId);
```

valeriy42:
This PR implements restoring the Analyzer from a data in named pipe. It also introduces the usage of CStateCompressor and CStateDecompressor for compression and decompression of the state.

As a minor refactoring, memory estimation callback was moved from a lambda function into a dedicated method for symmetry with other callback functions.

PR #673, opened 2019-09-16
Added unit test to CXMeansOnline1dTest.cc

edsavage:
Plugged a `TODO` in the clustering unit tests

### DIFF
--- a/lib/maths/unittest/CXMeansOnline1dTest.cc
+++ b/lib/maths/unittest/CXMeansOnline1dTest.cc
@@ -181,7 +181,7 @@ void CXMeansOnline1dTest::testMixtureOfGaussians() {
 
     // Test 1:
     //   * Cluster 1 = N(7, 1),     100 points
-    //   * Cluster 2 = N(15, 2.25), 200 points
+    //   * Cluster 2 = N(16, 2.25), 200 points
     //   * Cluster 3 = N(35, 2.25), 150 points
     maths::CXMeansOnline1d::TSizeDoublePr2Vec dummy;
     {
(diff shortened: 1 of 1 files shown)


lib/maths/unittest/CXMeansOnline1dTest.cc
```diff
@@ -699,7 +699,71 @@ void CXMeansOnline1dTest::testAdaption() {
...
+                std::sqrt(maths::CBasicStatistics::variance(expectedClusters[j])),
+                clusters[j].spread(),
+                0.04 * std::max(std::sqrt(maths::CBasicStatistics::variance(expectedClusters[j])),
+                                clusters[j].spread()));
+        }
+    };
+
+    addSamplesToClusterer(mode1.size(), {mode1});
+    addSamplesToClusterer(mode1.size() + mode2.size(), {mode2a, mode2b});
+    addSamplesToClusterer(samples.size(), {mode1});
```
tveasey:
```suggestion
    addSamplesToClusterer(samples.size(), {mode3});
```

lib/maths/unittest/CXMeansOnline1dTest.cc
```diff
@@ -699,7 +699,71 @@ void CXMeansOnline1dTest::testAdaption() {
...
+    test::CRandomNumbers rng;
+
+    TDoubleVec mode1;
+    rng.generateNormalSamples(3000.0, 10000.0, 100u, mode1);
+    TDoubleVec mode2a;
+    rng.generateNormalSamples(3000.0, 10000.0, 100u, mode2a);
+    TDoubleVec mode2b;
+    rng.generateNormalSamples(4000.0, 90000.0, 100u, mode2b);
+    TDoubleVec mode3;
+    rng.generateNormalSamples(3000.0, 10000.0, 1000u, mode3);
```
tveasey: Would it make sense to use different values for `mode3`? This tests the third case that the data shifts to a new level. You may need to add slightly more data in this case before the mode for N(3000.0, 10000.0) is removed: i.e. would be 20% of data after just 1000 new points.

PR #677, opened 2019-09-17
[ML] Restore regression state from named pipe

valeriy42:
Same as #660 with fixed error in calling default callback functions.


dimitris-athanasiou:
I will test the java side using this branch to check everything works fine end to end and report back here.

dimitris-athanasiou:
I confirm I could run a job without errors and the state index contained the model state after the job was done.

PR #680, opened 2019-09-17
Revert "Revert "[ML] Replace log4cxx with Boost.Log (#655)""

valeriy42:
Putting #655  back in code


droberts195:
I actually reverted the revert directly on the master branch, since there are absolutely no differences to the original set of changes that was temporarily reverted.  So there's probably no need to merge this PR.  We might as well leave it open until the CI is finished though.  Then if the CI fails because of some clash with some other recent change I can push the fix to this PR.  If the CI succeeds just close without merging.

droberts195:
CI passed, so closing 🎉

PR #681, opened 2019-09-19
[ML] Extract encoding mapping

valeriy42:


include/maths/CDataFrameCategoryEncoder.h
```diff
@@ -85,12 +86,21 @@ class CMakeDataFrameCategoryEncoder;
...
+    };
     using TBoolVec = std::vector<bool>;
     using TDoubleVec = std::vector<double>;
     using TDoubleVecVec = std::vector<TDoubleVec>;
     using TSizeVec = std::vector<std::size_t>;
     using TSizeVecVec = std::vector<TSizeVec>;
     using TRowRef = core::CDataFrame::TRowRef;
+    using TDoubleEncodingPr = std::pair<double, EEncoding>;
+    using TDoubleEncodingPrVector = std::vector<TDoubleEncodingPr>;
+    using TSizeDoubleEncodingPrVectorMap = std::unordered_map<std::size_t, TDoubleEncodingPrVector>;
```
tveasey: I think if you go down this route I'd be inclined to have this as a map from `encodedRowIndex` to `(EEncoding, TDoubleVec)`. This might as well be a vector since there are no holes. So you'd have a `TEncodingDoubleVecPrVec m_EncodingMap`. You could then do `std::tie(encodingType, encodingValues) = m_EncodingMap[encodedRowIndex]` with `encodingType` being one of `E_OneHot`, `E_Frequency`, `E_TargetMean` and the vector storing the values to use for that category, i.e. `encodingValues[category]` would contain the target mean or frequency.

tveasey: Alternatively, you could actually store the functions to do the mapping in this vector, i.e. `std::vector<std::function<double(std::size_t)>>`. In this case, notionally one-hot would `[m_Category](std::size_t category) { return category == m_Category ? 1.0 : 0.0; }`, target mean would be `[m_Means, m_Fallback](std::size_t category) { return category < m_Means.size() ? m_Means[category] : m_Fallback; }`, etc.

valeriy42:
The functionality is implemented in #689

PR #682, opened 2019-09-20
[ML] Persisting time series decomposition state with readable tags

edsavage:
As the first stage in extracting time series decomposition state for analysis as part of the `model_extractor` project, meaningful 'readable' tags have been added throughout the relevant persistence routines.

Also took the opportunity to attend to several code style warnings flagged up by `clang-tidy`

### DIFF
--- a/lib/maths/CSeasonalTime.cc
+++ b/lib/maths/CSeasonalTime.cc
@@ -25,8 +25,8 @@ namespace ml {
 namespace maths {
 namespace {
 // DO NOT change the existing tags if new sub-classes are added.
-const std::string DIURNAL_TIME_TAG("a");
-const std::string ARBITRARY_PERIOD_TIME_TAG("b");
+const core::TPersistenceTag DIURNAL_TIME_TAG{"a", "diurnal-time"};
+const core::TPersistenceTag ARBITRARY_PERIOD_TIME_TAG{"b", "arbitrary_period_time"};
 }
 
 //////// CSeasonalTime ////////
--- a/lib/maths/CSeasonalTime.cc
+++ b/lib/maths/CSeasonalTime.cc
@@ -25,7 +25,7 @@ namespace ml {
 namespace maths {
 namespace {
 // DO NOT change the existing tags if new sub-classes are added.
-const core::TPersistenceTag DIURNAL_TIME_TAG{"a", "diurnal-time"};
+const core::TPersistenceTag DIURNAL_TIME_TAG{"a", "diurnal_time"};
 const core::TPersistenceTag ARBITRARY_PERIOD_TIME_TAG{"b", "arbitrary_period_time"};
 }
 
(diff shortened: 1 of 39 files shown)


lib/maths/CSeasonalTime.cc
```diff
@@ -25,8 +25,8 @@ namespace ml {
 namespace maths {
 namespace {
 // DO NOT change the existing tags if new sub-classes are added.
-const std::string DIURNAL_TIME_TAG("a");
-const std::string ARBITRARY_PERIOD_TIME_TAG("b");
+const core::TPersistenceTag DIURNAL_TIME_TAG{"a", "diurnal-time"};
```
droberts195: Underscore would be better than hyphen in a field name that might be rendered in JSON.

lib/model/CAnomalyDetectorModelConfig.cc
```diff
@@ -623,7 +623,7 @@ CAnomalyDetectorModelConfig::factory(int identifier,
     }
     result->scheduledEvents(m_ScheduledEvents);
 
-    return result;
+    return std::move(result);
```
droberts195: It's best not to move return values.

If you _don't_ specify `move` then the compiler will choose what it thinks is best from:

1. Copy elision
2. RVO
3. Move

If you specify `move` then you're forcing option 3, which might not be the most efficient in all circumstances.

See also https://stackoverflow.com/a/17473874

edsavage: For context, this change was in response to the compiler warning ```CAnomalyDetectorModelConfig.cc:626:12: warning: prior to the resolution of a defect report against ISO C++11, local variable 'result' would have been copied despite being returned by name, due to its not matching the function return type ('shared_ptr<const ml::model::CModelFactory>' vs 'shared_ptr<ml::model::CModelFactory>') [-Wreturn-std-move-in-c++11]
    return result;
           ^~~~~~
CAnomalyDetectorModelConfig.cc:626:12: note: call 'std::move' explicitly to avoid copying on older compilers
    return result;
           ^~~~~~
           std::move(result)
1 warning generated.```

which is just a minor annoyance.. I'll revert the change.

droberts195: In https://reviews.llvm.org/rG321f24ec42f65403134031bbe99ad3c468b01efe it says:

> Diagnose cases of "return x;" or "throw x;" which in this version of Clang *do* produce moves, but which prior to Clang 3.9 / GCC 5.1 produced copies instead. This is useful in codebases which care about portability to those older compilers.

We are currently using Clang 3.9 / GCC 7.3 in `master` and `7.x`, so should be OK.  Realistically we'll never downgrade for the more recent branches.  The branch that could benefit is `6.8` where we're using Clang 3.8 / GCC 6.2, but not sure it's worth the effort of a PR for just one place.

PR #683, opened 2019-09-20
[ML] Remote obsolete reference to restoreSearcherSupplier in Analyzer

valeriy42:
This PR fixes the way `restoreSearcherSupplier` is passed from main to the business logic.

### DIFF
--- a/include/api/CDataFrameAnalyzer.h
+++ b/include/api/CDataFrameAnalyzer.h
@@ -51,8 +51,7 @@ public:
 
 public:
     CDataFrameAnalyzer(TDataFrameAnalysisSpecificationUPtr analysisSpecification,
-                       TJsonOutputStreamWrapperUPtrSupplier resultsStreamSupplier,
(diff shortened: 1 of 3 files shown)


include/api/CDataFrameAnalyzer.h
```diff
@@ -112,7 +111,6 @@ class API_EXPORT CDataFrameAnalyzer {
     TStrVec m_FieldNames;
     TTemporaryDirectoryPtr m_DataFrameDirectory;
     TJsonOutputStreamWrapperUPtrSupplier m_ResultsStreamSupplier;
-    TDataSearcherUPtrSupplier m_DataSearcher;
```
tveasey: Can we also remove the `TDataSearcherUPtrSupplier` and `#include <core/CDataSearcher.h>`?

valeriy42: Done

PR #695, opened 2019-09-25
[ML] Fix problems found by static code analysis

edsavage:
Attended to a handful of warnings thrown up by `make analyze` on the library code.

Nothing too serious was thrown up - mainly uninitialised / unused variables and a few missing checks for `nullptr`

### DIFF
--- a/include/core/CMemory.h
+++ b/include/core/CMemory.h
@@ -294,10 +294,10 @@ public:
     //! Overload for std::shared_ptr.
     template<typename T>
     static std::size_t dynamicSize(const std::shared_ptr<T>& t) {
-        if (t == nullptr) {
-            return 0;
-        }
-        long uc = t.use_count();
+        // The check for nullptr here may seem unnecessary but there are situations
+        // where an unset shared_ptr can have a use_count greater than 0, see
+        // https://stackoverflow.com/questions/48885252/c-sharedptr-use-count-for-nullptr/48885643
+        long uc{t == nullptr ? 0 : t.use_count()};
         if (uc == 0) {
             return 0;
         }
@@ -700,10 +700,13 @@ public:
     static void dynamicSize(const char* name,
                             const std::shared_ptr<T>& t,
                             CMemoryUsage::TMemoryUsagePtr mem) {
-        if (t == nullptr || t.use_count() == 0) {
+        // The check for nullptr here may seem unnecessary but there are situations
+        // where an unset shared_ptr can have a use_count greater than 0, see
+        // https://stackoverflow.com/questions/48885252/c-sharedptr-use-count-for-nullptr/48885643
+        long uc{t == nullptr ? 0 : t.use_count()};
+        if (uc == 0) {
             return;
         }
-        long uc = t.use_count();
         // If the pointer is shared by multiple users, each one
         // might count it, so divide by the number of users.
         // However, if only 1 user has it, do a full debug.
--- a/include/core/CMemory.h
+++ b/include/core/CMemory.h
@@ -298,6 +298,9 @@ public:
             return 0;
         }
         long uc = t.use_count();
+        if (uc == 0) {
+            return 0;
+        }
         // Note we add on sizeof(long) here to account for the memory
         // used by the shared reference count. Also, round up.
         return (sizeof(long) + staticSize(*t) + dynamicSize(*t) + std::size_t(uc - 1)) / uc;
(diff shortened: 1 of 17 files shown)


include/core/CMemory.h
```diff
@@ -697,26 +700,27 @@ class CORE_EXPORT CMemoryDebug : private CNonInstantiatable {
...
-                // the memory used by the shared reference count.
-                // Also, round up.
-                mem->addItem(ss.str(), (sizeof(long) + CMemory::staticSize(*t) +
-                                        CMemory::dynamicSize(*t) + std::size_t(uc - 1)) /
-                                           uc);
-            }
+        if (t == nullptr || t.use_count() == 0) {
+            return;
+        }
+        long uc = t.use_count();
```
droberts195: Since `use_count` is 0 when the pointer is null, maybe we could _just_ check this? Same in the non-debug version of the method.

Also, this code is reading `use_count` twice and the value could change between the two calls. So I think it’s best to cache its value in `uc` as the first step.

lib/model/CEventRateModel.cc
```diff
@@ -32,15 +31,13 @@
 #include <model/CResourceMonitor.h>
 #include <model/FrequencyPredicates.h>
 
-#include <boost/iterator/counting_iterator.hpp>
-
 #include <algorithm>
+#include <boost/iterator/counting_iterator.hpp>
+#include <stdint.h>
```
droberts195: We usually split the headers into groups:

- Boost headers
- blank line
- C++ headers
- blank line
- C headers

(We've also been gradually switching from C headers to their C++ equivalents, i.e. `<stdint.h>` to `<cstdint>`, but that will also mean adding `std::` in front of all the sized integer types throughout the file, so if that's a lot of extra changes feel free to leave it to a future PR.)

lib/model/CEventRateModel.cc
```diff
@@ -584,6 +581,11 @@ void CEventRateModel::fill(model_t::EFeature feature,
                            bool interim,
                            CProbabilityAndInfluenceCalculator::SParams& params) const {
     const TFeatureData* data{this->featureData(feature, pid, bucketTime)};
+    if (data == nullptr) {
+        LOG_ERROR(<< "Feature data unexpectedly null");
```
droberts195: The other logging for null feature data is at the `TRACE` level.  I think that's OK, on the assumption that the `featureData` method itself always logs at the `ERROR` level before returning `nullptr`.  So I think this should be `TRACE` too.  (In fact, the places in the code that already did null checks on the return value of `featureData` don't log at all - they just continue/return.  `TRACE` is equivalent for production code as we compile it out.)

lib/model/CEventRatePopulationModel.cc
```diff
@@ -466,6 +466,10 @@ void CEventRatePopulationModel::sample(core_t::TTime startTime,
                     .trendWeights(attribute.second.s_Weights)
                     .priorWeights(attribute.second.s_Weights);
                 maths::CModel* model{this->model(feature, cid)};
+                if (model == nullptr) {
+                    LOG_TRACE(<< "Model unexpectedly null");
```
droberts195: In contrast to the `featureData` methods, the `model` methods _can_ returning `nullptr` without logging anything.  That makes me wonder if we should log at the `ERROR` level in these places (6 in this PR) and include `feature` and `cid` in the error messages where `model` might return `nullptr` are genuinely unexpected?  What do you think @tveasey?

tveasey: I think it should never happen. The only thing I wonder about is whether we don't create a model for say hitting memory constraints, but we still have a value for it in the bucket: I think these also get removed, but will double check.

lib/model/CMetricPopulationModel.cc
```diff
@@ -919,6 +927,10 @@ void CMetricPopulationModel::fill(model_t::EFeature feature,
     std::size_t dimension{model_t::dimension(feature)};
     auto data = find(this->featureData(feature, bucketTime), pid, cid);
     const maths::CModel* model{this->model(feature, cid)};
+    if (model == nullptr) {
+        LOG_TRACE(<< "Model unexpectedly null");
+        return;
+    }
```
tveasey: This isn't sufficient. I don't think this should ever happen, but assuming it does this gets dereferenced at on this [line](https://github.com/elastic/ml-cpp/blob/master/lib/model/CMetricPopulationModel.cc#L593). My inclination would be to have these methods return false if the model is missing and take action in the calling code as well.

lib/model/CEventRatePopulationModel.cc
```diff
@@ -1003,6 +1011,10 @@ void CEventRatePopulationModel::fill(model_t::EFeature feature,
                                      CProbabilityAndInfluenceCalculator::SParams& params) const {
     auto data = find(this->featureData(feature, bucketTime), pid, cid);
     const maths::CModel* model{this->model(feature, cid)};
+    if (model == nullptr) {
+        LOG_TRACE(<< "Model unexpectedly null");
+        return;
```
tveasey: Similar considerations apply.

lib/model/CMetricModel.cc
```diff
@@ -536,6 +536,10 @@ void CMetricModel::fill(model_t::EFeature feature,
 
     std::size_t dimension{model_t::dimension(feature)};
     const TFeatureData* data{this->featureData(feature, pid, bucketTime)};
+    if (data == nullptr) {
+        LOG_TRACE(<< "data unexpectedly null");
+        return;
+    }
```
tveasey: Model being null is also unchecked below.

tveasey: Also, as with other cases I think if this fails we need to short circuit in calling code.

lib/model/CEventRateModel.cc
edsavage: Ah yes, missed that. Those includes were added by CLion doing its thing

include/core/CMemory.h
edsavage: Weirdly I've just encountered a case where `use_count` returns 1 for a `nullptr`

```
Exception = EXC_BAD_ACCESS (code=1, address=0x0)
t = {const std::__1::shared_ptr<ml::maths::CPrior> &} nullptr
 __ptr_ = {std::__1::shared_ptr<ml::maths::CPrior>::element_type * | 0x0} NULL
uc = {long} 1
```

I _think_ this may be related to https://stackoverflow.com/questions/48885252/c-sharedptr-use-count-for-nullptr/48885643. So an explicit check for `nullptr` is still needed to trap this case

droberts195: Interesting, thanks for finding that link.  It would still be good to only read `use_count` once, but, like you say, we'll have to keep the `nullptr` check.

PR #698, opened 2019-09-26
[ML] Depth based regularisation for boosted tree training

tveasey:
This implements a smooth depth based regularisation strategy. In particular, we penalise nodes with an exponential cost after the depth exceeds some specified target.

This fits in better with are general hyperparameter optimisation strategy than using a hard cutoff and should in addition lead to better tradeoffs. We use Bayesian optimisation to additionally optimise the parameters of the depth penalty function: these are the desired maximum depth and the tolerance we are prepared to accept in the actual maximum depth.

I made two small refactors: exposing the tree node so I can unit test depth control and factoring out duplicate code to populate the data frame in the boosted tree unit tests.


lib/maths/CBoostedTreeFactory.cc
```diff
@@ -299,20 +317,58 @@ void CBoostedTreeFactory::initializeUnsetRegularizationHyperparameters(core::CDa
...
+
+            double logInitialAlpha{std::log(gainPerNode)};
+            auto applyAlphaStep = [logInitialAlpha](CBoostedTreeImpl& tree,
+                                                    double stepSize, std::size_t step) {
+                tree.m_Regularization.alpha(
+                    std::exp(logInitialAlpha + static_cast<double>(step) * stepSize));
+            };
+            m_LogAlphaSearchInterval =
+                TVector{std::log(gainPerNode)} +
+                this->testLossNewtonLineSearch(frame, allTrainingRowsMask, applyAlphaStep,
```
valeriy42: If you are changing the code here anyway, could you please remove "Newton" from the name of the function. During our last conversation, I confused it with a Newton optimization, since there is a step size and steps, :see_no_evil:  but we are dealing with quadratic approximation here and minimization happens elsewhere.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -115,9 +152,10 @@ void CBoostedTreeImpl::CLeafNodeStatistics::addRowDerivatives(const CEncodedData
 CBoostedTreeImpl::CLeafNodeStatistics::SSplitStatistics
 CBoostedTreeImpl::CLeafNodeStatistics::computeBestSplitStatistics() const {
 
-    // We have two possible regularisation terms we'll use:
+    // We have three possible regularization terms we'll use:
     //   1. Tree size: gamma * "node count"
     //   2. Sum square weights: lambda * sum{"leaf weight" ^ 2)}
+    //   3. Tree depth: alpha * sum{exp(("depth" / "target depth" - 1.0) / "tolerance")}
```
valeriy42: Let's discuss the regularization function offline.

lib/maths/CBoostedTreeFactory.cc
tveasey: I think Newton is ok in the sense it is using a quadratic approximation to find the minimum. I agree it isn't conventional iterated Newton's method, but I still think it captures the gist of what we're doing here. I could switch back to something like `lineSearchForMinimumQuadraticTestLoss` if you prefer.

tveasey:
This models an additive regularisation term to the overall objective which is `\alpha * sum_i{ \exp((d_i/t - 1) / \tau) }` where, `d_i` is the depth of the i'th leaf, `t` is the target depth and `\tau` is the tolerance. In this sense, it behaves exactly like the tree size regularisation as far as the split strategy used in XGBoost is concerned: one can get the change in the regularised objective for a candidate split by adding on the difference in this penalty for the two child nodes and for the node one is considering splitting.

I don't think the exact form of this function is critical except I think it should have the following properties: it should be everywhere positive, decrease rapidly for depth < target depth, increase rapidly for depth > target depth and have a fixed value (say one) at the target depth. However, a key consideration is what should the slope of this function be at the target depth; this determines how much a tree's depth will exceed the target if the loss can be reduced significantly by making further splits. It also, importantly, starts to slightly penalise nodes based on their depth sooner.

Testing on a variety of data sets showed that this tolerance parameter, `\tau`, is important to the eventual quality of the model and I didn't find a single good value.

valeriy42:
Alright, thank you for the detailed explanation.

lib/maths/CBoostedTreeFactory.cc
valeriy42: At this point, I would opt for `testLossLineSearch` for simplification.

tveasey: Done in [e732fe7](https://github.com/elastic/ml-cpp/pull/698/commits/e732fe770fb9d4532853c147c655b98a248dec0d)

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -31,6 +31,9 @@ const std::size_t MAX_REGULARIZER_INDEX{2};
 const std::size_t INITIAL_REGULARIZER_SEARCH_ITERATIONS{8};
 const double MIN_REGULARIZER_SCALE{0.1};
 const double MAX_REGULARIZER_SCALE{10.0};
+const double MIN_MAX_DEPTH{2.0};
```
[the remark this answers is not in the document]
tveasey: See [fe65bbd](https://github.com/elastic/ml-cpp/pull/698/commits/fe65bbd66bc663e764e10d7ef9637f5f610a60e0)

include/maths/CBoostedTreeFactory.h
```diff
@@ -58,10 +58,16 @@ class MATHS_EXPORT CBoostedTreeFactory final {
     CBoostedTreeFactory& minimumFrequencyToOneHotEncode(double frequency);
     //! Set the number of folds to use for estimating the generalisation error.
     CBoostedTreeFactory& numberFolds(std::size_t numberFolds);
-    //! Set the lambda regularisation parameter.
-    CBoostedTreeFactory& lambda(double lambda);
-    //! Set the gamma regularisation parameter.
+    //! Set the alpha regularisation parameter: sum of tree depth factors.
+    CBoostedTreeFactory& alpha(double alpha);
```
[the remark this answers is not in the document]
tveasey: See [21a03c7](https://github.com/elastic/ml-cpp/pull/698/commits/21a03c73e760deb05deab5ffd1fb528a7822a5ce).

lib/maths/CDataFrameCategoryEncoder.cc
```diff
@@ -271,15 +271,9 @@ std::uint64_t CDataFrameCategoryEncoder::checksum(std::uint64_t seed) const {
 }
 
 void CDataFrameCategoryEncoder::acceptPersistInserter(core::CStatePersistInserter& inserter) const {
-    auto persistEncodingVector = [this](core::CStatePersistInserter& inserter) {
-        for (const auto& encoding : m_Encodings) {
-            auto acceptPersistInserter = [&encoding](core::CStatePersistInserter& inserter) {
```
tveasey: The nested parameter names were generating compiler warnings for me. I decided to just factor out one level of lambda as a private member function, which also matches what we do for restore.

PR #703, opened 2019-09-30
[ML] Fix relative location to load resources from

droberts195:
Fixes a bug introduced in #593


droberts195:
For the backport of #593 I will incorporate this fix into the backport PRs.  Hence this PR is only needed on master.

PR #720, opened 2019-10-07
[ML] Switch to Boost.Test

### LINKED ISSUE #156: [ML] Switch unit testing framework from cppunit to boost

droberts195:
Switches our unit test framework from cppunit to Boost.Test.

Closes #156


droberts195:
After the first commit the only directory where `make test` works is `lib/ver`.  But that proves the switch is possible.  Subsequent commits will migrate the other tests.

include/test/CBoostTestJUnitOutput.h
```diff
@@ -0,0 +1,44 @@
...
+//! IMPLEMENTATION DECISIONS:\n
+//! Always logs to the same output file, junit_results.xml,
+//! which makes it clear which file a CI system needs to
+//! pick up to render the test results.
+//!
+TEST_EXPORT
+class CBoostTestJUnitOutput {
+public:
+    CBoostTestJUnitOutput() = delete;
+    CBoostTestJUnitOutput(const CBoostTestJUnitOutput&) = delete;
```
edsavage: Should the assignment operator also be `delete`d?

PR #721, opened 2019-10-08
[ML] Improvements to sparse count modelling

### LINKED ISSUE #696: [ML] Improve modeling of periodic, sparse data
Prospective customer use case has once-daily sampled data of a program executed at a certain time. 

image

Tried to model with 4h bucket span in order to get somewhat timely alerting if a gap exists (a monitored program does not run), but the modeling isn't adequate. Anomalies are not raised at the time of the gap of execution:

image

The modeling and the anomaly results are better with a 1d bucket_span:

image

But, the downside is an increased "delay to alert" in this case.

tveasey:
This simplifies sparse count and sum modelling and migrates to always updating the time series model, but using a weight which decreases in proportion to the number of empty buckets. This means we simply smoothly transition to modelling non-empty buckets for sparse data.

I've also removed the correction to the probability which accounts for the fraction of non-empty buckets: we have the rare function anyway if this is the primary concern. Finally, I changed periodicity testing so that it approximates the old behaviour, i.e. it tends to ignoring empty buckets as their proportion increases.

Closes #696.

### DIFF
--- a/lib/maths/CPeriodicityHypothesisTests.cc
+++ b/lib/maths/CPeriodicityHypothesisTests.cc
@@ -344,7 +344,7 @@ void reweightOutliers(const std::vector<T>& trend,
         CBasicStatistics::COrderStatisticsHeap<TDoubleSizePr, std::greater<TDoubleSizePr>>;
 
     std::size_t period{trend.size()};
-    std::size_t numberOutliers{static_cast<std::size_t>([&] {
+    std::size_t numberOutliers{static_cast<std::size_t>([&period, &values] {
         std::size_t count(std::count_if(
             values.begin(), values.end(), [](const TFloatMeanAccumulator& value) {
                 return CBasicStatistics::count(value) > 0.0;
@@ -1838,7 +1838,7 @@ bool CPeriodicityHypothesisTests::testPeriodWithScaling(const TTimeTimePr2Vec& w
     }
 
     // Compute the degrees of freedom given the alternative hypothesis.
-    double b{[&] {
+    double b{[&windows, &period_, &values, this] {
         TDoubleVec repeats(calculateRepeats(windows, period_, m_BucketLength, values));
         return static_cast<double>(
             std::count_if(repeats.begin(), repeats.end(),
@@ -2056,7 +2056,7 @@ bool CPeriodicityHypothesisTests::testPartition(const TTimeTimePr2Vec& partition
     //   3) The significance of the variance reduction, and
     //   4) The amount of variance reduction.
 
-    auto calculateMeanRepeats = [&](const TTimeTimePr2Vec& w, core_t::TTime p) {
+    auto calculateMeanRepeats = [&values, this](const TTimeTimePr2Vec& w, core_t::TTime p) {
         TMeanAccumulator result;
         result.add(calculateRepeats(w, p, m_BucketLength, values));
         return CBasicStatistics::mean(result);
@@ -2140,7 +2140,7 @@ bool CPeriodicityHypothesisTests::testVariance(const TTimeTimePr2Vec& window,
     LOG_TRACE(<< "  autocorrelation          = " << R);
     LOG_TRACE(<< "  autocorrelationThreshold = " << stats.s_AutocorrelationThreshold);
 
-    meanRepeats = [&] {
+    meanRepeats = [&window, &period_, &buckets, this] {
         TMeanAccumulator result;
         result.add(calculateRepeats(window, period_, m_BucketLength, buckets));
         return CBasicStatistics::mean(result);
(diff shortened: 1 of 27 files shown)


lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1839,10 +1838,12 @@ bool CPeriodicityHypothesisTests::testPeriodWithScaling(const TTimeTimePr2Vec& w
     }
 
     // Compute the degrees of freedom given the alternative hypothesis.
-    TSizeVec repeats(calculateRepeats(windows, period_, m_BucketLength, values));
-    double b{static_cast<double>(
-        std::count_if(repeats.begin(), repeats.end(),
-                      [](std::size_t repeat) { return repeat > 0; }))};
+    double b{[&] {
```
edsavage: Can we avoid using the default capture mode here?

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -2056,13 +2057,9 @@ bool CPeriodicityHypothesisTests::testPartition(const TTimeTimePr2Vec& partition
...
 
     auto calculateMeanRepeats = [&](const TTimeTimePr2Vec& w, core_t::TTime p) {
-        TSizeVec repeats{calculateRepeats(w, p, m_BucketLength, values)};
-        return CBasicStatistics::mean(
-            std::accumulate(repeats.begin(), repeats.end(), TMeanAccumulator{},
-                            [](TMeanAccumulator mean, std::size_t r) {
-                                mean.add(static_cast<double>(r));
-                                return mean;
-                            }));
+        TMeanAccumulator result;
```
edsavage: ... and here (I realise it's not part of this change)

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -2143,13 +2140,11 @@ bool CPeriodicityHypothesisTests::testVariance(const TTimeTimePr2Vec& window,
...
     LOG_TRACE(<< "  autocorrelationThreshold = " << stats.s_AutocorrelationThreshold);
 
-    TSizeVec repeats{calculateRepeats(window, period_, m_BucketLength, buckets)};
-    meanRepeats = CBasicStatistics::mean(
-        std::accumulate(repeats.begin(), repeats.end(), TMeanAccumulator{},
-                        [](TMeanAccumulator mean, std::size_t repeat) {
-                            mean.add(static_cast<double>(repeat));
-                            return mean;
-                        }));
+    meanRepeats = [&] {
```
edsavage: ditto

lib/maths/CTimeSeriesModel.cc
```diff
@@ -2619,18 +2549,13 @@ bool CMultivariateTimeSeriesModel::probability(const CModelProbabilityParams& pa
     };
     using TJointProbability2Vec = core::CSmallVector<SJointProbability, 2>;
 
-    auto updateJointProbabilities = [&](maths_t::EProbabilityCalculation calculation,
-                                        double value_, const TDouble10Vec2Vec& pl,
+    auto updateJointProbabilities = [&](const TDouble10Vec2Vec& pl,
```
edsavage: ditto

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -146,43 +146,33 @@ double ratio(double numerator, double denominator, double zeroDividedByZero) {
...
 
 // Functions to compute influence based on different criteria
+
 //! \brief Computes the value of summed statistics on the set difference.
 class CValueDifference {
 public:
     //! Features.
     bool operator()(const TDouble2Vec& v,
-                    double n,
+                    double /*n*/,
```
edsavage: Is it possible to remove the commented out parameters entirely?

lib/maths/CPeriodicityHypothesisTests.cc
tveasey: Done in [5d3c3a3](https://github.com/elastic/ml-cpp/pull/721/commits/5d3c3a3b79319afb4949e5ad10666d9adf1036d3).

lib/maths/CPeriodicityHypothesisTests.cc
tveasey: Done in [5d3c3a3](https://github.com/elastic/ml-cpp/pull/721/commits/5d3c3a3b79319afb4949e5ad10666d9adf1036d3).

lib/maths/CTimeSeriesModel.cc
tveasey: Done in [5d3c3a3](https://github.com/elastic/ml-cpp/pull/721/commits/5d3c3a3b79319afb4949e5ad10666d9adf1036d3).

PR #726, opened 2019-10-10
[ML] Add state format version for resilience

valeriy42:
This PR adds the version number of the current release (7.5) to the format of the training state using for resilience. In the case of the version mismatch, the deserialization of the training state will fail and training will start from scratch.

I also took an opportunity to refactor the test for persistence state error handling to have a more granular control of the reasons for failure.

### DIFF
--- a/lib/maths/unittest/CBoostedTreeTest.cc
+++ b/lib/maths/unittest/CBoostedTreeTest.cc
@@ -184,6 +184,13 @@ void readFileToStream(const std::string& filename, std::stringstream& stream) {
     std::string str((std::istreambuf_iterator<char>(file)),
                     std::istreambuf_iterator<char>());
     stream << str;
+    stream.flush();
+}
+
+void clearFile(const std::string& filename) {
+    std::ofstream file;
+    file.open(filename, std::ofstream::out | std::ofstream::trunc);
+    file.close();
 }
 }
 
@@ -1256,6 +1263,12 @@ void CBoostedTreeTest::testRestoreErrorHandling() {
     };
     core::CLogger::CScopeSetFatalErrorHandler scope{errorHandler};
 
+    const std::string logFile{"test.log"};
+
+    // log at level ERROR only
+    CPPUNIT_ASSERT(ml::core::CLogger::instance().reconfigureFromFile(
+        "testfiles/testLogErrors.boost.log.ini"));
+
     std::size_t cols{3};
     std::size_t capacity{50};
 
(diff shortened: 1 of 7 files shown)


lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1252,80 +1262,70 @@ void CBoostedTreeTest::testRestoreErrorHandling() {
...
-           "\"hyperparam_feature_bag_fraction\":\"0\",\"hyperparam_feature_sample_probabilities\":\"\"},"
-           "\"eta_override\":\"false;0\",\"feature_bag_fraction_override\":\"false;0\",\"gamma_override\":\"false;0\","
-           "\"lambda_override\":\"false;0\",\"maximum_number_trees_override\":\"true;2\",\"loss\":\"mse\"}";
+    readFileToStream("testfiles/error_bayesian_optimisation_state.json",
+                     errorInBayesianOptimisationState);
     errorInBayesianOptimisationState.flush();
 
     bool throwsExceptions{false};
+    std::stringstream buffer;
+    std::streambuf* old = std::cerr.rdbuf(buffer.rdbuf());
```
droberts195: I don't think we should intercept logs like this:

1. It's making the assumption that default logging goes through `std::cerr`, which is a breach of encapsulation.  For example, I was wondering if I should change it to go through `std::clog` so that it's buffered.
2. It's setting the buffer to a local variable which means that if the corresponding `std::cerr.rdbuf(old)` doesn't get run for some reason then the test suite will core dump in the next test.  And when that happens it will be much more work to figure out what went wrong than looking at a nicely formatted list of test failures in Jenkins.

There's an example of how we've previously asserted on which errors are logged in `CPoissonMeanConjugateTest::testSampleMarginalLikelihoodInSupportBounds`.  You could even use the same logging config file in your test as it's also in the `maths` unit tests.

If you don't like this approach then after 7.5 feature freeze I'd be happy for you to open another PR to add a method to `CLogger` to tell it to log to a supplied stream instead of the current logging location and then change the unit tests that use the current approach over to that.  We could have a `catch (...)` in the tests that use that to ensure the logger is reset before the test exits for any reason.  But please use the current approach for this PR and then do that change (if you think it's worthwhile) separately.

valeriy42: Thank you for pointing this out, @droberts195. I adjusted the test. Could you please take another look.

PR #732, opened 2019-10-11
[ML] Fix potential problem for boosted tree training initialisation

tveasey:
Numeric errors mean that it's possible the sum curvature for a candidate split is identically zero while the gradient is epsilon. This can cause the node gain to appear infinite (when there is no weight regularisation) which in turns causes problems initialising the region we search for optimal hyperparameter values.

Since none of our loss functions should have curvature identically equal to zero, we can safely force the gradient to be zero if detect that the curvature is zero.

\cc @przemekwitek.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -110,13 +110,18 @@ CBoostedTreeImpl::CLeafNodeStatistics::CLeafNodeStatistics(std::size_t id,
         m_Gradients[i].resize(numberSplits);
         m_Curvatures[i].resize(numberSplits);
         for (std::size_t j = 0; j < numberSplits; ++j) {
-            m_Gradients[i][j] = parent.m_Gradients[i][j] - sibling.m_Gradients[i][j];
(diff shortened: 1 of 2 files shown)


tveasey:
No problem! Thanks for raising and helping investigate @przemekwitek.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -110,13 +110,18 @@ CBoostedTreeImpl::CLeafNodeStatistics::CLeafNodeStatistics(std::size_t id,
         m_Gradients[i].resize(numberSplits);
         m_Curvatures[i].resize(numberSplits);
         for (std::size_t j = 0; j < numberSplits; ++j) {
-            m_Gradients[i][j] = parent.m_Gradients[i][j] - sibling.m_Gradients[i][j];
             m_Curvatures[i][j] = parent.m_Curvatures[i][j] - sibling.m_Curvatures[i][j];
+            m_Gradients[i][j] = m_Curvatures[i][j] == 0.0
+                                    ? 0.0
+                                    : parent.m_Gradients[i][j] -
+                                          sibling.m_Gradients[i][j];
```
valeriy42: Could you please add the description of the PR as a comment here for posterity?

PR #736, opened 2019-10-14
[ML] Improve logistic regression numerics

tveasey:
Properly handle numeric errors in the calculation of logistic loss related functions.


tveasey:
Thanks!

PR #738, opened 2019-10-14
[ML] Generate inference definition for classification

valeriy42:
This PR adds the ability to generate inference definitions for classification. The class hierarchy of the `CBoostedTreeRegressionBuilder` was refactored to reflect the new class hierarchy for `CDataFrameRegressionRunner` and `CDataFrameClassificationRunner`.

* `CBoostedTreeRegressionInferenceModelBuilder.{h.cc}` -> `CBoostedTreeInferenceModelBuilder.{h.cc}`

### DIFF
--- a/include/api/CBoostedTreeInferenceModelBuilder.h
+++ b/include/api/CBoostedTreeInferenceModelBuilder.h
@@ -19,8 +19,7 @@ namespace ml {
 namespace api {
 
 //! \brief Builds a a serialisable trained model object by visiting a maths::CBoostedTree object.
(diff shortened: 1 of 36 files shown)


benwtrent:
@valeriy42 will we need a different sort of output aggregator for the logistic results?

valeriy42:
> @valeriy42 will we need a different sort of output aggregator for the logistic results?

No, it's still summing up over all trees without any specific weights (all weights are 1.0).

valeriy42:
This PR is replaced by #743

PR #740, opened 2019-10-14
[ML] Snapshot boosted tree training state after final train

tveasey:
We should be snapshotting the training state after the final train on the full data set. This incidentally means we can also re-enable the save/restore unit test case we'd disabled when all parameters are supplied by the user.

### DIFF
--- a/lib/api/unittest/CDataFrameAnalyzerTest.cc
+++ b/lib/api/unittest/CDataFrameAnalyzerTest.cc
@@ -995,7 +995,7 @@ void CDataFrameAnalyzerTest::testRunBoostedTreeRegressionTrainingWithStateRecove
         };
 
         finalIteration = params.numberUnset() * numberRoundsPerHyperparameter;
-        finalIteration = std::max(finalIteration, std::size_t{1}) - 1;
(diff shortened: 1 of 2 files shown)


lib/api/unittest/CDataFrameAnalyzerTest.cc
```diff
@@ -995,8 +994,11 @@ void CDataFrameAnalyzerTest::testRunBoostedTreeRegressionTrainingWithStateRecove
                                   params.s_FeatureBagFraction, &persisterSupplier);
         };
 
-        finalIteration = params.numberUnset() * numberRoundsPerHyperparameter - 1;
-        rng.generateUniformSamples(0, finalIteration - 1, 3, intermediateIterations);
+        finalIteration = params.numberUnset() * numberRoundsPerHyperparameter;
+        finalIteration = std::max(finalIteration, std::size_t{1}) - 1;
```
valeriy42:
```suggestion
        finalIteration = std::max(params.numberUnset() * numberRoundsPerHyperparameter - 1, std::size_t{0});
```

valeriy42: nit: I would have only one-step assignment for `finalIteration` since the logic is not too complicated.

PR #743, opened 2019-10-15
[ML] Generate inference definition for classification NG

valeriy42:
This PR adds the ability to generate inference definitions for classification. The class hierarchy of the `CBoostedTreeRegressionBuilder` was refactored to reflect the new class hierarchy for `CDataFrameRegressionRunner` and `CDataFrameClassificationRunner`.

*    `CBoostedTreeRegressionInferenceModelBuilder.{h.cc}` -> `CBoostedTreeInferenceModelBuilder.{h.cc}`

This PR overhauls #738


include/api/CInferenceModelDefinition.h
```diff
@@ -57,7 +56,7 @@ class API_EXPORT CWeightedMode : public CAggregateOutput {
     //! Construct with a weight vector of \p size with all entries equal to \p weight.
     CWeightedMode(std::size_t size, double weight);
     void addToDocument(rapidjson::Value& parentObject, TRapidJsonWriter& writer) const override;
-    const std::string& stringType() override;
+    const std::string& stringType() const final;
```
tveasey: This makes me wonder whether this class should actually be final, in which case we could just change this back to override.

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -0,0 +1,193 @@
...
+                                       splitFeature, leftChild, rightChild, gain);
+}
+
+CBoostedTreeInferenceModelBuilder::CBoostedTreeInferenceModelBuilder(
+    TStringVec fieldNames,
+    std::size_t dependentVariableColumnIndex,
+    const TStringSizeUMapVec& categoryNameMap) {
+    // filter filed names containing only "."
+    fieldNames.erase(std::remove(fieldNames.begin(), fieldNames.end(), "."),
+                     fieldNames.end());
```
tveasey: The field names I store on the data frame do not include the special control fields. So I think this is redundant.

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -0,0 +1,193 @@
...
+    // filter dependent variable field name
+    assert(dependentVariableColumnIndex < fieldNames.size());
+    fieldNames.erase(fieldNames.begin() +
+                     static_cast<std::ptrdiff_t>(dependentVariableColumnIndex));
+    m_FieldNames = fieldNames;
```
tveasey:
```suggestion
    m_FieldNames = std::move(fieldNames);
```

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -0,0 +1,193 @@
...
+CInferenceModelDefinition& CBoostedTreeInferenceModelBuilder::definition() {
+    return m_Definition;
+}
+
+void CRegressionInferenceModelBuilder::setTargetType() {
+    this->definition().trainedModel()->targetType(CTrainedModel::ETargetType::E_Regression);
+}
+
+CRegressionInferenceModelBuilder::CRegressionInferenceModelBuilder(
+    const CBoostedTreeInferenceModelBuilder::TStringVec& fieldNames,
```
tveasey: Since the type is visible via inheritance
```suggestion
    const TStringVec& fieldNames,
```
and three other cases below

lib/api/CInferenceModelDefinition.cc
```diff
@@ -170,6 +171,10 @@ CTrainedModel::ETargetType CEnsemble::targetType() const {
     return this->CTrainedModel::targetType();
 }
 
+const CTrainedModel::TStringVec& CEnsemble::featureNames() const {
+    return CTrainedModel::featureNames();
```
tveasey: not: consistency
```suggestion
    return this->CTrainedModel::featureNames();
```

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -0,0 +1,193 @@
...
+    this->categoryNameMap(categoryNameMap);
+    m_Definition.fieldNames(fieldNames);
+    m_Definition.trainedModel(std::make_unique<CEnsemble>());
+    m_Definition.typeString(INFERENCE_MODEL);
+}
+
+CBoostedTreeInferenceModelBuilder::TStringDoubleUMap
+CBoostedTreeInferenceModelBuilder::encodingMap(std::size_t inputColumnIndex,
+                                               const TDoubleVec& map_) {
+    TStringDoubleUMap map;
```
tveasey: Question: do we ever actually need to lookup in this map or is it just for outputting as name value pairs in the JSON? If it is the latter, maybe we should switch the type `std::vector<std::pair<std::string, double>>`: it would be more space and memory efficient.

include/api/CInferenceModelDefinition.h
tveasey: In general, I wonder whether more of the classes related to the inference model schema should be marked as final.

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -0,0 +1,193 @@
...
+
+CBoostedTreeInferenceModelBuilder::CBoostedTreeInferenceModelBuilder(
+    TStringVec fieldNames,
+    std::size_t dependentVariableColumnIndex,
+    const TStringSizeUMapVec& categoryNameMap) {
+    // filter filed names containing only "."
+    fieldNames.erase(std::remove(fieldNames.begin(), fieldNames.end(), "."),
+                     fieldNames.end());
+    // filter dependent variable field name
+    assert(dependentVariableColumnIndex < fieldNames.size());
```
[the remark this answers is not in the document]
valeriy42: Done

lib/api/CBoostedTreeInferenceModelBuilder.cc
valeriy42: It's not to filter control fields, but the additional columns in the data frame after the target column:
```    TStrVec fieldNames{"numeric_col", "categorical_col", "target_col", ".", "."};```

File: lib/api/CBoostedTreeInferenceModelBuilder.cc (hunk 3 above)
Comment (valeriy42) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: Done :+1:

File: lib/api/CBoostedTreeInferenceModelBuilder.cc (hunk 4 above)
Comment (valeriy42) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: Get fixed after merge.

File: lib/api/CInferenceModelDefinition.cc (hunk 5 above)
Comment (valeriy42) [reply lib/api/CInferenceModelDefinition.cc]: done

File: lib/api/CBoostedTreeInferenceModelBuilder.cc (hunk 6 above)
Comment (valeriy42) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: I changed the type to `TStrVecVec`.

**Comment by valeriy42** [comment]
Thank you @tveasey for the review. The special fields are not getting filtered automatically, so I need to keep this in code.

**Comment by valeriy42** [comment]
retest

**Comment by valeriy42** [comment]
retest

File: lib/api/CBoostedTreeInferenceModelBuilder.cc (hunk 2 above)
Comment (tveasey) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: Sorry these are what I was referring to: one is the field we use to pass control messages from the Java. I explicitly don't copy them to the data frame see [this](https://github.com/elastic/ml-cpp/blob/master/lib/api/CDataFrameAnalyzer.cc#L250) line.

Comment (tveasey) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: So my question would be are you sure this is still the case, i.e. I changed this recently. If you're somehow getting them still we should probably investigate.

Comment (tveasey) [reply lib/api/CBoostedTreeInferenceModelBuilder.cc]: Talked about this offline and it seems like they might still be there. This needs investigation, but shouldn't hold this PR up.

File: include/api/CBoostedTreeInferenceModelBuilder.h [hunk 8]
```diff
@@ -66,7 +70,30 @@ class API_EXPORT CBoostedTreeRegressionInferenceModelBuilder
     TStrVec m_FieldNames;
     TStrVec m_FeatureNames;
 };
+
+class API_EXPORT CRegressionInferenceModelBuilder : public CBoostedTreeInferenceModelBuilder {
```
tveasey: maybe
```suggestion
class API_EXPORT CRegressionInferenceModelBuilder final : public CBoostedTreeInferenceModelBuilder {
```

include/api/CBoostedTreeInferenceModelBuilder.h
```diff
@@ -66,7 +70,30 @@ class API_EXPORT CBoostedTreeRegressionInferenceModelBuilder
...
+    CRegressionInferenceModelBuilder(TStrVec fieldNames,
+                                     std::size_t dependentVariableColumnIndex,
+                                     const TStrVecVec& categoryNames);
+
+protected:
+    void setTargetType() override;
+    void setAggregateOutput(CEnsemble* ensemble) const override;
+};
+
+class API_EXPORT CClassificationInferenceModelBuilder : public CBoostedTreeInferenceModelBuilder {
```
tveasey:
```suggestion
class API_EXPORT CClassificationInferenceModelBuilder final : public CBoostedTreeInferenceModelBuilder {
```

benwtrent:
FWIW, I am running into issues still. It seems that `field` `feature_name` are out of sync with the contained value maps. 

```
 {
                "frequency_encoding" : {
                  "field" : "dayOfWeek",
                  "feature_name" : "dayOfWeek_frequency",
                  "frequency_map" : {
                    "Rain" : 0.14402581095941017,
                    "Heavy Fog" : 0.14402581095941017,
                    "Thunder & Lightning" : 0.14402581095941017,
                    "Cloudy" : 0.14402581095941017,
                    "Damaging Wind" : 0.14402581095941017,
                    "Hail" : 0.14402581095941017,
                    "Sunny" : 0.14402581095941017,
                    "Clear" : 0.14402581095941017
                  }
                }
              },
```
```
{
                "target_mean_encoding" : {
                  "field" : "OriginWeather",
                  "feature_name" : "OriginWeather_targetmean",
                  "target_map" : {
                    "DE" : 40.89416058394161,
                    "NO" : 35.125000000000014,
                    "PR" : 73.12500000000004,
                    "RU" : 46.20906801007566,
                    "FI" : 48.266129032258114,
                    "JP" : 46.46416910122013,
                    "DK" : 43.88059701492539,
...
```

PR #747, opened 2019-10-16
[ML] Fix VS linking problem

valeriy42:
Since VS was complaining about multiple exported instances of `CRapidJsonLineWriter`, I changed the code in order to use `CRapidJsonConcurrentLineWriter` as an instantiation of `CRapidJsonLineWriter`.


droberts195:
retest

lib/api/unittest/CBoostedTreeInferenceModelBuilderTest.cc
```diff
@@ -256,7 +256,7 @@ void CBoostedTreeInferenceModelBuilderTest::testIntegrationClassification() {
 }
 
 void CBoostedTreeInferenceModelBuilderTest::testJsonSchema() {
-    std::size_t numberExamples = 1000;
+    std::size_t numberExamples = 200;
```
przemekwitek: Why is this change in this PR? Is it somehow related?

valeriy42: No you a right, it's unrelated, I will move it to another PR.

include/core/CRapidJsonLineWriter.h
```diff
@@ -29,7 +29,7 @@ template<typename OUTPUT_STREAM,
          typename TARGET_ENCODING = rapidjson::UTF8<>,
          typename STACK_ALLOCATOR = rapidjson::CrtAllocator,
          unsigned WRITE_FLAGS = rapidjson::kWriteDefaultFlags>
-class CORE_EXPORT CRapidJsonLineWriter
```
droberts195: Please can you also remove the header from line 12.

valeriy42: Done

valeriy42: Done

valeriy42:
retest

valeriy42:
retest

PR #765, opened 2019-10-22
[ML] Tidy ups for data frame analysis runner

tveasey:
1. Moves common memory monitoring stuff up to base class to reduce duplication,
2. Moves `CParameters` out of `CDataFrameAnalysisConfigReader` so we can forward declare,
3. Renames `CDataFrameBoostedTreeRunner` hierarchy to allow for other model types,
4. Migrates to singleton `CDataFrameAnalysisConfigReader` for training regression and classification models.


lib/api/CDataFrameAnalysisRunner.cc
```diff
@@ -182,6 +182,20 @@ CDataFrameAnalysisRunner::TProgressRecorder CDataFrameAnalysisRunner::progressRe
...
+CDataFrameAnalysisRunner::TMemoryMonitor
+CDataFrameAnalysisRunner::memoryMonitor(counter_t::ECounterTypes counter) {
+    return [counter, this](std::int64_t delta) {
+        std::int64_t memory{m_Memory.fetch_add(delta)};
+        if (memory >= 0) {
+            core::CProgramCounters::counter(counter).max(memory);
+        } else {
+            // Something has gone wrong with memory estimation. Trap this case
+            // to avoid underflowing the peak memory usage statistic.
+            LOG_DEBUG(<< "Memory estimate " << memory << " is negative!");
```
droberts195: I know you've only moved it and not introduced it in this PR, but this is going to be silent in our QA testing and in production.  Should it be a warning?  (Even if it should, feel free to change in a future PR since this one is green.)

tveasey: [ec9a652](https://github.com/elastic/ml-cpp/pull/765/commits/ec9a6520bd58c7687017688ee04b04046b152988)

PR #769, opened 2019-10-23
[ML] Identity encode the target variable

tveasey:
We should always just pass back the target variable unmodified in the encoded feature vector.

### DIFF
--- a/include/maths/CDataFrameCategoryEncoder.h
+++ b/include/maths/CDataFrameCategoryEncoder.h
@@ -84,14 +84,17 @@ enum EEncoding {
 //! \brief Performs encoding of the categorical columns in a data frame.
 //!
 //! DESCRIPTION:\n
-//! We select appropriate encodings for categorical values based on their cardinality,
-//! the information they carry about the regression target variable and so on. We use
-//! a mixture of one-hot encoding for the categories which carry the most information
-//! (because this is lossless), mean target encoding for categories where we have a
-//! reasonable sample set and a catch all indicator feature for all rare categories.
+//! We consider:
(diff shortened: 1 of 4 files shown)


include/maths/CDataFrameCategoryEncoder.h
```diff
@@ -84,14 +84,17 @@ enum EEncoding {
...
-//! the information they carry about the regression target variable and so on. We use
-//! a mixture of one-hot encoding for the categories which carry the most information
-//! (because this is lossless), mean target encoding for categories where we have a
-//! reasonable sample set and a catch all indicator feature for all rare categories.
+//! We consider:
+//!   -# One-hot encoding for the categories which carry the most information (because
+//!      this is lossless)
+//!   -# Mean target encoding for categories where we have a reasonable sample in the
+//!      training data and
+//!   -# Frequency encoding
```
valeriy42: If I understand correctly:
* Every category gets encoded using frequency encoding
* Categories with a reasonable sample size get additionally target mean encoding, other categories would have a fallback value in the corresponding features
* Selected important categories get additionally (to frequency and target mean) encoded using one-hot encoding

If this is correct, then maybe it is better to describe this that way to avoid misunderstanding.

tveasey: Not exactly. We will only choose mean target and frequency encoding if their MICe with the target variable is greater than a cutoff: see `discardNuisanceFeatures`. Importantly, if we decide we need to prune available features to n, we select the these n by looking at MICe with the target and average MICe with the features chosen so far: see `selectFeatures`.

PR #773, opened 2019-10-23
[ML] Refactor CLogger to use custom stream

valeriy42:
Following the [discussion](https://github.com/elastic/ml-cpp/pull/726#discussion_r333475371) this PR adds a way to use custom stream in `CLogger`.

### DIFF
--- a/lib/core/CLogger.cc
+++ b/lib/core/CLogger.cc
@@ -276,7 +276,7 @@ bool CLogger::reconfigure(const std::string& pipeName, const std::string& proper
     return this->reconfigureLogToNamedPipe(pipeName);
 }
 
-bool CLogger::reconfigure(boost::shared_ptr<std::ostream>& streamPtr) {
+bool CLogger::reconfigure(boost::shared_ptr<std::ostream> streamPtr) {
     if (streamPtr->good()) {
         auto newSink{boost::make_shared<TTextOStreamSynchronousSink>()};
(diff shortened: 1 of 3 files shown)


lib/core/CLogger.cc
```diff
@@ -277,6 +276,16 @@ bool CLogger::reconfigure(const std::string& pipeName, const std::string& proper
     return this->reconfigureLogToNamedPipe(pipeName);
 }
 
+bool CLogger::reconfigure(boost::shared_ptr<std::ostream>& streamPtr) {
```
droberts195: If this took its argument by value, would it avoid the need to do the `boost::static_pointer_cast` in the calling code?  I think it should because `boost::shared_ptr<T>` has a constructor that takes a `boost::shared_ptr<Y>` where `Y` is convertible to `T`.

If taking the argument by value does allow the `boost::static_pointer_cast` to be avoided then I think it's worth it because the pattern that is set by this PR is likely to be duplicated in several other tests in the future, and it's nice if it's cleaner.

I know this means an extra increment and decrement of an atomic counter so is very slightly less efficient, but it's only for test code and it's not like we're doing that millions of times.

valeriy42:
Thank you @droberts195 for the review. I agree, the costs of copying the pointer and increasing/decreasing the counters may be justified by improved simplicity.

PR #784, opened 2019-10-25
[ML] Stratified cross validation for regression

tveasey:
This implements stratified fractional cross validation for regression, i.e. independently and uniformly random sample from each decile of the target variable. It also better controls the proportion of samples per fold (to be nearest integer to "total training examples" / "number folds") for classification.


lib/maths/CDataFrameUtils.cc
```diff
@@ -57,6 +61,162 @@ bool doReduce(std::pair<std::vector<READER>, bool> readResults,
...
+    auto copyBucketRowCounts = [](TDoubleVec counts_, TDoubleVec& counts) {
+        counts = std::move(counts_);
+    };
+    auto reduceBucketRowCounts = [](TDoubleVec counts_, TDoubleVec& counts) {
+        for (std::size_t i = 0; i < counts.size(); ++i) {
+            counts[i] += counts_[i];
+        }
+    };
+
+    TDoubleVec weights;
```
valeriy42: The variable name `weights` is confusing here. Can you maybe rename it to something more self-explanatory, i.e. `decentileCounts`?

tveasey: Ok good suggestion. I renamed in [this](https://github.com/elastic/ml-cpp/pull/784/commits/7c23e983bf26fe663ffc982ae2adf4c18a7949da) commit. (Note I don't use decile because I passed in the number of buckets to optionally disable.)

lib/maths/unittest/CDataFrameUtilsTest.cc
```diff
@@ -534,51 +534,66 @@ BOOST_AUTO_TEST_CASE(testStratifiedCrossValidationRowMasks) {
...
+        }
+
+        for (std::size_t i = 1; i < numberBins; ++i) {
+            TMeanVarAccumulator testTargetDecileMoments;
+            for (std::size_t fold = 0; fold < numberFolds[0]; ++fold) {
+                testTargetDecileMoments.add(targetDecile[fold][i]);
+            }
+            LOG_DEBUG(<< "variance in test set target percentile = "
+                      << maths::CBasicStatistics::variance(testTargetDecileMoments));
+            BOOST_REQUIRE(maths::CBasicStatistics::variance(testTargetDecileMoments) < 0.02);
```
droberts195: It's worth noting that `BOOST_TEST_REQUIRE` here would print the values either side of the `<` in the event of failure, whereas `BOOST_REQUIRE` won't.

It seems that `BOOST_REQUIRE(condition)` is the equivalent of `BOOST_TEST_REQUIRE((condition))`.  I wasn't aware of this until now, and `BOOST_REQUIRE` is the cleaner solution for complex conditions where diagnostics cannot be printed.  But for this line the improved diagnostics _could_ be printed by `BOOST_TEST_REQUIRE`.

droberts195: I changed this in #788

tveasey: Thanks

PR #790, opened 2019-10-29
[ML] Stop cross-validation early if the test loss isn't promising

tveasey:
This targets runtime by stopping cross-validation early if the test loss from the folds for which we've already trained suggest it is unlikely that the parameter values under test will be good. We account for the variation in test loss across the folds by fitting an OLS to the test losses from previous rounds of training to predict the missing folds' test errors. This gives us both a predicted loss and its variance for each missing fold. We use these to decide if there is any point proceeding with training and to correct the test loss we supply to Bayesian Optimisation.

This is somewhat experimental. On the data sets I've tested on we are able to stop training early on many rounds for little or no degradation in QoR. However, it'll need more thorough evaluation than I've performed so far. It is useful to get it in so it goes through the QA suite and also so it's available for others to test internally.

### DIFF
--- a/include/maths/CBoostedTreeFactory.h
+++ b/include/maths/CBoostedTreeFactory.h
@@ -57,8 +57,10 @@ public:
     CBoostedTreeFactory& minimumFrequencyToOneHotEncode(double frequency);
     //! Set the number of folds to use for estimating the generalisation error.
     CBoostedTreeFactory& numberFolds(std::size_t numberFolds);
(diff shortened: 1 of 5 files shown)


tveasey:
On second thoughts, this is going to part of a bigger piece of work targeting performance improvements for 7.6. There are a few things to consider and this is likely going to require some iterations, so I'm going to create a feature branch for this work. I'll open pull requests against this branch.

PR #793, opened 2019-10-29
[ML] Add classification labels for classification definition

### LINKED ISSUE #785: [ML] Inference writer not writing classification labels
When the inference model is saved into the ES inference index, the classification labels are not being written. This is even though the data is labeled with a `yes|no` type of string value. 

Withing the `ensemble` object, the java side is expecting the field `classification_labels` to be included so that the appropriate label can be written when inferencing against the model.

valeriy42:
This PR adds information about `classification_labels` to the inference definition for classifiers.

Closes #785

### DIFF
--- a/lib/api/CBoostedTreeInferenceModelBuilder.cc
+++ b/lib/api/CBoostedTreeInferenceModelBuilder.cc
@@ -169,7 +169,7 @@ void CClassificationInferenceModelBuilder::setAggregateOutput(CEnsemble* ensembl
 }
 
 CInferenceModelDefinition&& CClassificationInferenceModelBuilder::build() {
-    auto definition{this->CBoostedTreeInferenceModelBuilder::build()};
+    auto&& definition{this->CBoostedTreeInferenceModelBuilder::build()};
     const auto& trainedModel{definition.trainedModel()};
     trainedModel->classificationLabels(m_ClassificationLabels);
     return std::move(definition);
--- a/lib/api/CBoostedTreeInferenceModelBuilder.cc
+++ b/lib/api/CBoostedTreeInferenceModelBuilder.cc
@@ -169,8 +169,8 @@ void CClassificationInferenceModelBuilder::setAggregateOutput(CEnsemble* ensembl
 }
 
 CInferenceModelDefinition&& CClassificationInferenceModelBuilder::build() {
-    auto&& definition{this->CBoostedTreeInferenceModelBuilder::build()};
-    const auto& trainedModel{definition.trainedModel()};
+    auto&& definition = this->CBoostedTreeInferenceModelBuilder::build();
+    const auto& trainedModel = definition.trainedModel();
     trainedModel->classificationLabels(m_ClassificationLabels);
     return std::move(definition);
 }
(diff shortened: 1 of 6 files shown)


lib/api/unittest/CBoostedTreeInferenceModelBuilderTest.cc
```diff
@@ -215,10 +216,15 @@ BOOST_AUTO_TEST_CASE(testIntegrationClassification) {
     BOOST_REQUIRE_EQUAL(expectedSize, trainedModel->size());
     BOOST_TEST_REQUIRE("logistic_regression" ==
                        trainedModel->aggregateOutput()->stringType());
+    const auto& classificationLabels{trainedModel->classificationLabels()};
+    BOOST_REQUIRE(classificationLabels.is_initialized() == true);
```
droberts195: I think it's setting a bad example that people will accidentally copy to use `BOOST_REQUIRE` with an assertion that has a simple comparison.  In this case it's obvious that if it fails the actual value was `false` instead of `true`, but if people copy the macro to other places where the RHS is not a boolean then we'll lose information for less obvious failures.

Therefore I think this line should be changed to either:

```
    BOOST_TEST_REQUIRE(classificationLabels.is_initialized() == true);
```

or:

```
    BOOST_REQUIRE(classificationLabels.is_initialized());
```

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -160,11 +160,19 @@ CClassificationInferenceModelBuilder::CClassificationInferenceModelBuilder(
...
+    : CBoostedTreeInferenceModelBuilder(fieldNames, dependentVariableColumnIndex, categoryNames),
+      m_ClassificationLabels{categoryNames[dependentVariableColumnIndex]} {
 }
 
 void CClassificationInferenceModelBuilder::setAggregateOutput(CEnsemble* ensemble) const {
     ensemble->aggregateOutput(std::make_unique<CLogisticRegression>(ensemble->size(), 1.0));
 }
+
+CInferenceModelDefinition&& CClassificationInferenceModelBuilder::build() {
+    auto&& definition{this->CBoostedTreeInferenceModelBuilder::build()};
```
tveasey: nit: I prefer using assignment with auto since this will often deduce `std::initializer_list` which is rarely wanted
```suggestion
    auto&& definition = this->CBoostedTreeInferenceModelBuilder::build();
```

lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -160,11 +160,19 @@ CClassificationInferenceModelBuilder::CClassificationInferenceModelBuilder(
...
+    const auto& trainedModel{definition.trainedModel()};
+    trainedModel->classificationLabels(m_ClassificationLabels);
+    return std::move(definition);
```
tveasey: No need to move the local variable result type.

lib/api/CBoostedTreeInferenceModelBuilder.cc
tveasey: and below

lib/api/unittest/CBoostedTreeInferenceModelBuilderTest.cc
valeriy42: Thank you for spotting this. I changed to the first option.

lib/api/CBoostedTreeInferenceModelBuilder.cc
valeriy42: Done

lib/api/CBoostedTreeInferenceModelBuilder.cc
valeriy42: Unfortunately, this is required, I get a compilation error "cannot bind rvalue to lvalue" otherwise.

PR #799, opened 2019-10-30
Remove duplicated code.

przemekwitek:
Remove duplicated code in `bin/data_frame_analyzer/Main.cc`

### DIFF
--- a/bin/data_frame_analyzer/Main.cc
+++ b/bin/data_frame_analyzer/Main.cc
@@ -184,9 +184,7 @@ int main(int argc, char** argv) {
         std::move(restoreSearcherSupplier));
 
     if (memoryUsageEstimationOnly) {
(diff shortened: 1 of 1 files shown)


przemekwitek:
> There's nothing wrong with this change, but I don't think it will fix any bugs.
> 
> It removes 3 lines of duplicated code, and it's nice to avoid code duplication so I'm happy to merge it.
> 
> But the duplicated code that's removed is just the definition of a lambda, not calling that lambda. So there wasn't any double wrapping of streams before.
> 
> So, like I say, I'm happy for this change to be made but it won't help with [elastic/elasticsearch#47612](https://github.com/elastic/elasticsearch/issues/47612) and therefore the commit message should not reference that issue. The commit title should just be something like "Remove duplicated code".

I've changed the title and description of this PR to reflect that it is only about deduplicating code.

PR #809, opened 2019-11-06
[ML] add geo point formatting string for actual|typical

benwtrent:
Adds new geo_point formatted fields for Java side consumption. These are only written if we have actual|typical values that have length of 2, and are using a `lat_long` function.

This is my first PR against this repo, let me know if I missed something. 

Blocked by https://github.com/elastic/elasticsearch/pull/47050

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -37,6 +37,7 @@
 * Reduce the peak memory used by boosted tree training and fix an overcounting bug
 estimating maximum memory usage. (See {ml-pull}781[#781].)
 * Stratified fractional cross validation for regression. (See {ml-pull}784[#784].)
+* Added `geo_point` supported output for `lat_long` function records. (See {ml-pull}809[#809].)
 
 == {es} version 7.5.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -37,7 +37,7 @@
 * Reduce the peak memory used by boosted tree training and fix an overcounting bug
 estimating maximum memory usage. (See {ml-pull}781[#781].)
 * Stratified fractional cross validation for regression. (See {ml-pull}784[#784].)
-* Added `geo_point` supported output for `lat_long` function records. (See {ml-pull}809[#809].)
+* Added `geo_point` supported output for `lat_long` function records. (See {ml-pull}809[#809], {pull}47050[#47050].)
 
 == {es} version 7.5.0
 
--- a/lib/api/CJsonOutputWriter.cc
+++ b/lib/api/CJsonOutputWriter.cc
@@ -15,6 +15,7 @@
 
 #include <api/CModelSizeStatsJsonWriter.h>
 #include <api/CModelSnapshotJsonWriter.h>
+#include <api/CFieldConfig.h>
 
 #include <algorithm>
 #include <ostream>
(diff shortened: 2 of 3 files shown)


lib/api/unittest/CJsonOutputWriterTest.cc
```diff
@@ -1056,6 +1056,110 @@ BOOST_AUTO_TEST_CASE(testSimpleWrite) {
                         std::string(object2["field_name"].GetString()));
 }
 
+BOOST_AUTO_TEST_CASE(testGeoResultsWrite) {
+    ml::api::CJsonOutputWriter::TStrStrUMap emptyFields;
+
+    std::ostringstream sstream;
+    std::string partitionFieldName("tfn");
+    std::string partitionFieldValue("");
+    std::string overFieldName("pfn");
```
edsavage: Maybe change these string values to `ofv`, `ofn` for consistency?

lib/api/unittest/CJsonOutputWriterTest.cc
```diff
@@ -1056,6 +1056,110 @@ BOOST_AUTO_TEST_CASE(testSimpleWrite) {
...
+    arrayDoc.Parse<rapidjson::kParseDefaultFlags>(sstream.str().c_str());
+
+    BOOST_TEST_REQUIRE(arrayDoc.IsArray());
+    BOOST_REQUIRE_EQUAL(rapidjson::SizeType(3), arrayDoc.Size());
+
+    const rapidjson::Value& object = arrayDoc[rapidjson::SizeType(0)];
+    BOOST_TEST_REQUIRE(object.IsObject());
+
+    BOOST_TEST_REQUIRE(object.HasMember("typical"));
+    //BOOST_REQUIRE_EQUAL(2, object["typical"].GetArray().Size());
```
edsavage: Hmmm.. I'm always in 2 minds about commented out code, maybe delete it if it's not needed?

benwtrent: yeah, I cannot get the tests to do what I want the to do :(. So, very WIP there.

lib/api/CJsonOutputWriter.cc
```diff
@@ -553,6 +557,26 @@ void CJsonOutputWriter::addMetricFields(const CHierarchicalResultsWriter::TResul
...
     m_Writer.addDoubleArrayFieldToObj(TYPICAL, results.s_BaselineMean, *docPtr);
     m_Writer.addDoubleArrayFieldToObj(ACTUAL, results.s_CurrentMean, *docPtr);
+    if (results.s_FunctionName == CFieldConfig::FUNCTION_LAT_LONG) {
+        rapidjson::Value geoResults = m_Writer.makeObject();
+        if (results.s_BaselineMean.size() == 2) {
+            std::ostringstream ptStr;
+            // We don't want scientific notation and geo points only have precision up to 12 digits
+            ptStr << std::fixed;
+            ptStr << std::setprecision(12);
+            ptStr << results.s_BaselineMean[0] << "," << results.s_BaselineMean[1];
```
tveasey: I'd be inclined to factor this out as a lambda (or function), i.e.
```
auto geoPointToString = [](const auto& point) -> std::string {
    std::ostringstream result;
    return (result << std::fixed << std::setprecision(12) << point[0] << "," << point[1]).str();
};
```

edsavage:
Looks good to go @benwtrent

docs/CHANGELOG.asciidoc
```diff
@@ -37,6 +37,7 @@
 * Reduce the peak memory used by boosted tree training and fix an overcounting bug
 estimating maximum memory usage. (See {ml-pull}781[#781].)
 * Stratified fractional cross validation for regression. (See {ml-pull}784[#784].)
+* Added `geo_point` supported output for `lat_long` function records. (See {ml-pull}809[#809].)
```
droberts195: Because of the title of elastic/elasticsearch#47050, as things stand this release note will appear in addition to "Add new geo_results.(actual|typical) fields for lat_lon results".  As a user I would wonder whether these two things were the same or different enhancements, so it's probably best to have just one of them.

Since this changelog is more flexible and can reference two PRs, it's probably best to mark elastic/elasticsearch#47050 as a non-issue so that it doesn't get auto-generated into the release notes, but then mention it in this one too, using `(See {ml-pull}809[#809], {pull}47050[#47050].)`.

PR #814, opened 2019-11-11
[ML] Handle upgrade state to 7.5 to account for sparse count modelling changes

tveasey:
This is a bug, but marked as a non-issue since 7.5 is not yet released. (Note that the bounds are expected to change because they used to be scaled by the non-empty bucket frequency.)

Without this fix
Screenshot 2019-11-11 at 16 28 28

With this fix
Screenshot 2019-11-11 at 16 28 07

### DIFF
--- a/lib/model/CIndividualModel.cc
+++ b/lib/model/CIndividualModel.cc
@@ -35,6 +35,7 @@ namespace model {
 
 namespace {
 
+using TDouble2Vec = core::CSmallVector<double, 2>;
 using TStrCRef = std::reference_wrapper<const std::string>;
 using TStrCRefUInt64Map = std::map<TStrCRef, uint64_t, maths::COrderings::SLess>;
 using TStrCRefStrCRefPr = std::pair<TStrCRef, TStrCRef>;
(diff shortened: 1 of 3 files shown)


lib/model/CIndividualModel.cc
```diff
@@ -378,13 +382,35 @@ bool CIndividualModel::doAcceptRestoreTraverser(core::CStateRestoreTraverser& tr
                         std::cref(this->params()), std::placeholders::_1)))
         RESTORE(MEMORY_ESTIMATOR_TAG,
                 core::CPersistUtils::restore(MEMORY_ESTIMATOR_TAG, m_MemoryEstimator, traverser))
+        RESTORE_BUILT_IN(UPGRADING_PRE_7_5_STATE, upgradingPre7p5State)
     } while (traverser.next());
 
+    const double DEFAULT_CUTOFF_TO_MODEL_EMPTY_BUCKETS{0.2};
```
droberts195: For the benefit of anyone looking at this in the future, this is a local reinstatement of https://github.com/elastic/ml-cpp/pull/721/files#diff-d253406bdc862c2c89a29e7560f828a8L70

PR #818, opened 2019-11-12
Emit prediction_probability field alongside prediction field in ml results.

przemekwitek:
Even though the user can request `top_classes` (containing class names and probabilities) to be emitted, it seems reasonable to also include the top class' prediction probability as a field alongside the prediction field.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
@@ -36,6 +36,7 @@ const std::string BALANCED_CLASS_LOSS{"balanced_class_loss"};
 
 // Output
 const std::string IS_TRAINING_FIELD_NAME{"is_training"};
+const std::string PREDICTION_PROBABILITY_FIELD_NAME{"prediction_probability"};
 const std::string TOP_CLASSES_FIELD_NAME{"top_classes"};
 const std::string CLASS_NAME_FIELD_NAME{"class_name"};
 const std::string CLASS_PROBABILITY_FIELD_NAME{"class_probability"};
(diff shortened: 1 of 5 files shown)


lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -111,6 +112,8 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
     writer.StartObject();
     writer.Key(this->predictionFieldName());
     writer.String(categoryValues[predictedCategoryId]);
+    writer.Key(PREDICTION_PROBABILITY_FIELD_NAME);
+    writer.Double(probabilityOfCategory[0]);
```
dimitris-athanasiou: What do we call `probabilityOfCategory1` above? It seems a bit confusing.

przemekwitek: Ooops, at this point the entries are not sorted yet. Let me fix that and add unit test.

przemekwitek: I've added unit test. In order to do that I needed to split `writeOneRow` method. I don't quite like this but I couldn't find an easy way to mock out the boosted tree.
@tveasey: Any suggestion for improvement here?

As for the naming, I kept the name `probabilityOfCategory1` as I couldn't come up with a better one. I added inline comments so hopefully the meaning of this variable is now clear.

dimitris-athanasiou:
You will need to add an entry in the changelog though.

przemekwitek:
> You will need to add an entry in the changelog though.

Done.

przemekwitek:
run elasticsearch-ci

PR #821, opened 2019-11-13
[ML] Fix restore from checkpoint damaging seasonality modelling for anomaly detection

### LINKED ISSUE #820: [ML] Weekend periodicity lost in first week after reopening anomaly detection job
Reproduced in 6.8.4 with the `event_rate_gen` dataset

The job was run on 5 weeks of data (2015-01-05 -> 2015-02-09), and weekend periodicity has been established. After restarting the job and the rest of the dataset has been processed, it can be noticed that the first week processed after reopening has lost weekend periodicity.
image

## Steps to reproduce
1. Create a job with model plot enabled
2. Run a datafeed on the first 5 weeks of the data, and let it gracefully close
3. Check that the model plot shows that weekend periodicity is being modelled
4. Restart the datafeed for the remainder of the dataset
5. Notice that the first week processed after reopening the job has lost the weekend periodicity

tveasey:
We weren't persisting and restoring all the state we should have been. This should only affect behaviour after a restore.

Fixes #820.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -63,7 +63,7 @@ errors to be logged during boosted tree training. (See {ml-pull}732[#732].)
 
 === Bug Fixes
 
-* Restore from checkpoint could damage seasonality modelling. For example, it could
+* Restore from checkpoint could damage seasonality modeling. For example, it could
 cause seasonal components to be overwritten in error. (See {ml-pull}821[#821].)
 
 == {es} version 7.4.1
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -61,6 +61,11 @@ are unexpectedly missing. (See {ml-pull}721[#721].)
 * Trap numeric errors causing bad hyperparameter search initialisation and repeated
 errors to be logged during boosted tree training. (See {ml-pull}732[#732].)
 
+=== Bug Fixes
+
+* Restore from checkpoint could damage seasonality modelling. For example, it could
+cause seasonal components to be overwritten in error. (See {ml-pull}821[#821].)
+
 == {es} version 7.4.1
 
 === Enhancements
--- a/lib/maths/CSeasonalTime.cc
+++ b/lib/maths/CSeasonalTime.cc
@@ -11,6 +11,7 @@
 #include <core/CPersistUtils.h>
 #include <core/CStatePersistInserter.h>
 #include <core/CStateRestoreTraverser.h>
+#include <core/CStringUtils.h>
 #include <core/Constants.h>
 
 #include <maths/CChecksum.h>
@@ -143,7 +144,7 @@ CDiurnalTime* CDiurnalTime::clone() const {
 }
 
 bool CDiurnalTime::fromString(std::string value) {
-    std::size_t delimiter{value.find(DELIMITER)};
+    std::size_t delimiter{value.rfind(DELIMITER)};
     if (delimiter != std::string::npos) {
         double precedence{0.0};
         if (core::CStringUtils::stringToType(value.substr(delimiter + 1), precedence) == false) {
@@ -223,7 +224,7 @@ CGeneralPeriodTime* CGeneralPeriodTime::clone() const {
 }
 
 bool CGeneralPeriodTime::fromString(std::string value) {
-    std::size_t delimiter{value.find(DELIMITER)};
+    std::size_t delimiter{value.rfind(DELIMITER)};
     if (delimiter != std::string::npos) {
         double precedence{0.0};
         if (core::CStringUtils::stringToType(value.substr(delimiter + 1), precedence) == false) {
--- a/lib/maths/CSeasonalTime.cc
+++ b/lib/maths/CSeasonalTime.cc
@@ -6,6 +6,7 @@
 
 #include <maths/CSeasonalTime.h>
 
+#include <core/CIEEE754.h>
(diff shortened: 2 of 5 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -61,6 +61,11 @@ are unexpectedly missing. (See {ml-pull}721[#721].)
 * Trap numeric errors causing bad hyperparameter search initialisation and repeated
 errors to be logged during boosted tree training. (See {ml-pull}732[#732].)
 
+=== Bug Fixes
+
+* Restore from checkpoint could damage seasonality modelling. For example, it could
```
droberts195: For release notes we have to use the US spelling of modeling with one 'l'.

lib/maths/CSeasonalTime.cc
```diff
@@ -27,6 +27,7 @@ namespace {
 // DO NOT change the existing tags if new sub-classes are added.
 const core::TPersistenceTag DIURNAL_TIME_TAG{"a", "diurnal_time"};
 const core::TPersistenceTag ARBITRARY_PERIOD_TIME_TAG{"b", "arbitrary_period_time"};
+const std::string DELIMITER(",");
```
droberts195: There's a subtle dependency between this and the `DELIMITER` and `PAIR_DELIMITER` in `core::CPersistUtils`.  This one has to be different to those two, since restoration is searching for it starting at the beginning of a string whose beginning portion contains `core::CPersistUtils::DELIMITER` and `core::CPersistUtils::PAIR_DELIMITER`.  So this could be commented.  Alternatively `rfind` could be used instead of `find` when deconstructing the persisted string.

lib/maths/CSeasonalTime.cc
```diff
@@ -133,7 +142,17 @@ CDiurnalTime* CDiurnalTime::clone() const {
     return new CDiurnalTime(*this);
 }
 
-bool CDiurnalTime::fromString(const std::string& value) {
+bool CDiurnalTime::fromString(std::string value) {
+    std::size_t delimiter{value.find(DELIMITER)};
```
droberts195: `rfind` might be more robust to future changes given how the string is being concatenated in the persistence method.

lib/maths/CSeasonalTime.cc
```diff
@@ -198,21 +222,34 @@ CGeneralPeriodTime* CGeneralPeriodTime::clone() const {
     return new CGeneralPeriodTime(*this);
 }
 
-bool CGeneralPeriodTime::fromString(const std::string& value) {
+bool CGeneralPeriodTime::fromString(std::string value) {
+    std::size_t delimiter{value.find(DELIMITER)};
```
droberts195: `rfind` might be more robust to future changes given how the string is being concatenated in the persistence method.

lib/maths/unittest/Makefile
```diff
@@ -88,6 +88,7 @@ SRCS=\
 	CSamplingTest.cc \
 	CSeasonalComponentTest.cc \
 	CSeasonalComponentAdaptiveBucketingTest.cc \
+    CSeasonalTimeTest.cc \
```
droberts195: nit: the indenting is inconsistent

lib/maths/CSeasonalTime.cc
```diff
@@ -15,10 +16,9 @@
 #include <maths/CChecksum.h>
 #include <maths/CIntegerTools.h>
 
-#include <boost/array.hpp>
```
droberts195: It looks like this file had already been changed to `std::array` on master without changing the header, so this is fine on master.  But note that on 6.8 the code is still using `boost::array`, so the header change shouldn't be included in the 6.8 backport.

lib/maths/CSeasonalTime.cc
tveasey: I like the rfind idea. I thought about defining both here to make this explicit, but decided this unit test would fail if anyone changed the delimiter used by CPersistUtils. (On that front, this would also invalidate all our saved state so I don't see it ever happening, but better to be safe.)

lib/maths/CSeasonalTime.cc
tveasey: Yes, I need to rewrite the tests as well on 6.8 so the back port is non-trivial.

docs/CHANGELOG.asciidoc
tveasey: See [2e9d47c](https://github.com/elastic/ml-cpp/pull/821/commits/2e9d47cd0955a5a8c22509e8c239e50438b3c5e0)

lib/maths/CSeasonalTime.cc
tveasey: See [2e9d47c](https://github.com/elastic/ml-cpp/pull/821/commits/2e9d47cd0955a5a8c22509e8c239e50438b3c5e0)

lib/maths/unittest/Makefile
tveasey: See [2e9d47c](https://github.com/elastic/ml-cpp/pull/821/commits/2e9d47cd0955a5a8c22509e8c239e50438b3c5e0)

lib/maths/CSeasonalTime.cc
tveasey: See [2e9d47c](https://github.com/elastic/ml-cpp/pull/821/commits/2e9d47cd0955a5a8c22509e8c239e50438b3c5e0)

tveasey:
retest

PR #825, opened 2019-11-13
[7.5][ML] Fix restore from checkpoint damaging seasonality modelling for anomaly detection

tveasey:
Backport #821.


tveasey:
This branch doesn't have the boost test changes.

PR #828, opened 2019-11-15
[ML] Remove CI parallelism hack

droberts195:
Previously the ML Linux CI workers had 16 CPUs and 14.4GB RAM.
The Jenkins setup has just been changed so that in future they'll
have 16 CPUs and 24GB RAM.

This increase of memory per CPU from 0.9GB to 1.5GB will hopefully
mean that we can utilise all the CPUs on our Linux CI workers
without one of the gcc processes running out of memory.

### DIFF
--- a/dev-tools/docker/docker_entrypoint.sh
+++ b/dev-tools/docker/docker_entrypoint.sh
@@ -23,7 +23,7 @@ cd "$MY_DIR/../.."
 # Note: no need to clean due to the .dockerignore file
 
 # Build the code
-make -j`grep -c '^processor.*[1-6]$' /proc/cpuinfo` # TODO remove the processor hack
+make -j`grep -c '^processor' /proc/cpuinfo` # TODO remove the processor hack
 
 # Strip the binaries
 dev-tools/strip_binaries.sh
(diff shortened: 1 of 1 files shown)


dev-tools/docker/docker_entrypoint.sh
```diff
@@ -23,7 +23,7 @@ cd "$MY_DIR/../.."
 # Note: no need to clean due to the .dockerignore file
 
 # Build the code
-make -j`grep -c '^processor.*[1-6]$' /proc/cpuinfo` # TODO remove the processor hack
+make -j`grep -c '^processor' /proc/cpuinfo` # TODO remove the processor hack
```
edsavage: Probably best to remove the comment at end of line as well

dimitris-athanasiou: Is the comment for removing the processor hack obsolete? Or that's different?

edsavage: snap!

droberts195:
This change knocked 5 minutes off the build time: from 1 hour 19 minutes down to 1 hour 14 minutes.

PR #827, opened 2019-11-14
Upgrade to Eigen 3.3.7

edsavage:
Added Eigen as a git sub-tree under the 3rd_party directory.

Minor modifications were required to the build system and also to the
code to accommodate these changes.

PR #829, opened 2019-11-15
[7.6][ML]Upgrade to Eigen 3.3.7 (#827)

edsavage:
Relocated Eigen as a git sub-tree under the 3rd_party directory.

Minor modifications were required to the build system and also to the
code to accommodate the upgraded Eigen library version and its changed location.

Backports #827


edsavage:
I fully expect that some of the threshold values in `CBoostedTreeTest` will require tweaking and also note that #811 has not yet been back ported, which may prove troublesome.

PR #837, opened 2019-11-19
[ML]Documented developing, building and running the `ml-cpp` project

### LINKED ISSUE #813: [README] Apple Developer tools link in mac os env setup is broken
The macos.md build setup instruction mention this link https://developer.apple.com/technologies/xcode.html, which appears to lead to a Page Not Found.
https://github.com/elastic/ml-cpp/tree/master/build-setup

edsavage:
Closes #813

### DIFF
--- a/build-setup/macos.md
+++ b/build-setup/macos.md
@@ -44,7 +44,7 @@ The above environment variables only need to be set when building tools on macOS
 
 ### Xcode
 
-The first major piece of development software to install is Apple's development environment, Xcode, which can be downloaded from <http://developer.apple.com/technologies/xcode.html> . You will need to register as a developer with Apple. Alternatively, you can get the latest version of Xcode from the App Store.
+The first major piece of development software to install is Apple's development environment, Xcode, which can be downloaded from <https://developer.apple.com/download/> . You will need to register as a developer with Apple. Alternatively, you can get the latest version of Xcode from the App Store.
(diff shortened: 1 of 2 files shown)


edsavage:
I had a thought while labelling this PR. Would it be possible to rig up something withe Jenkins CI so that the build doesn't run when a PR is labelled as "docs" (or similar)?

build-setup/macos.md
```diff
@@ -53,6 +53,7 @@ The first major piece of development software to install is Apple's development
 - If you are using Mojave, you must install Xcode 10.3.x
```
droberts195: We should mention Catalina now, and we also know that Xcode 11 is fine and works on Mojave.  So please change this to:

- If you are using Mojave or Catalina, you must install Xcode 11.2.x

build-setup/macos.md
```diff
@@ -62,6 +63,14 @@ xcode-select --install
 
 at the command prompt.
 
+If (and only if) you are using Mojave, you must install the SDK headers from `/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg`. You can do this by running:
```
droberts195: This isn't true any more.  I fixed the underlying problem in #669 (and removed something very similar to what you've just added from this file).

PR #840, opened 2019-11-21
[ML] Remove unary_function/binary_function and old-style binders

droberts195:
C++11 deprecated the unary_function and binary_function base
classes, together with old-style binders like bind1st and not1.

These base classes and functions are completed removed in C++17,
so they need removing from the codebase before we can upgrade.
However, since they are unnecessary in C++11 and above it is
best to remove them in advance of any (master-only) C++17 work
so that the master and 7.x branches can stay more closely
aligned.

### DIFF
--- a/include/api/COutputHandler.h
+++ b/include/api/COutputHandler.h
@@ -15,7 +15,6 @@
 
 #include <boost/unordered_map.hpp>
 
-#include <functional>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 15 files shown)


droberts195:
Interestingly this seemingly no-op change has had a beneficial effect on memory usage on Windows.

The reason is that Visual Studio 2017 didn't apply the empty base class optimisation in the case of multiple inheritance - see https://developercommunity.visualstudio.com/content/problem/69605/vs2017-c-empty-base-optimization-does-not-work-wit.html

The explanation of why is in https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/

I have opened a dummy PR, #842, to use the secret `/d1reportClassLayoutChanges` flag to check for places where we might be able to benefit from adjusting base classes or adding `__declspec(empty_bases)`.

PR #843, opened 2019-11-22
[ML] Adjust Eigen linear algebra wrapper constructors

droberts195:
This change was driven by problems compiling the previous
code in Visual Studio 2019 in C++17 mode.  However, the
changes should work with other compilers, so it is best
to separate them out of the Visual Studio 2019 upgrade
PR.

There are basically two changes in this PR, but both
relate to the same area of the code:

1. Remove redundant copy constructor definitions from the
   Eigen linear algebra wrappers.  Most classes had two
   copy constructors, one taking a const argument and the
   other non-const.  But since the non-const copy
   constructor just forwarded to the const one there was
   no need to have two.  This then allows the single const
   copy constructor to be specified using = default.
   (Visual Studio 2019 warns if both const and non-const
   copy constructors are defined, so this silences those
   warnings, and also = default reduces boilerplate code.)
2. A workaround for a compilation problem when CKdTree is
   instantiated using a POINT type that is a
   CAnnotatedVector\<CMemoryMappedDenseVector\>\>.  The
   problem is that las::zero() returns a CDenseVector when
   passed a CMemoryMappedDenseVector.  This then cannot be
   converted back to CMemoryMappedDenseVector by Visual Studio
   2019 in C++17 mode unless there is a specific constructor to
   do it.  (It's unclear whether the difference is C++17 or
   some bug in Visual Studio 2019.)  This must have been
   working previously via some intermediate temporary object,
   and this highlighted another problem: the
   CMemoryMappedDenseVector is basically just a reference
   to memory owned by some other object, so if that other
   object is a temporary that's gone out of scope then
   there's a risk of invalid memory reads.  So this change
   makes sure the intermediate temporary object remains
   in scope until the POINT that it's transferred to is
   finished with.

### DIFF
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -240,7 +240,12 @@ public:
         const POINT* nearest{nullptr};
         if (m_Nodes.size() > 0) {
             auto inf = std::numeric_limits<TCoordinatePrecise>::max();
-            POINT distancesToHyperplanes{las::zero(point)};
+            // Sometimes the return type of las::zero() is not POINT. In this case
+            // we must convert it to POINT, but also must be careful that the POINT
+            // doesn't then end up referencing data owned by a temporary that has
+            // been destroyed.
+            auto temp{las::zero(point)};
+            POINT distancesToHyperplanes{std::move(temp)};
             return this->nearestNeighbour(point, m_Nodes[0], distancesToHyperplanes,
                                           0 /*split coordinate*/, nearest, inf);
         }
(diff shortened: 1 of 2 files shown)


edsavage:
LGTM, FWIW

include/maths/CKdTree.h
```diff
@@ -240,7 +240,12 @@ class CKdTree {
         const POINT* nearest{nullptr};
         if (m_Nodes.size() > 0) {
             auto inf = std::numeric_limits<TCoordinatePrecise>::max();
-            POINT distancesToHyperplanes{las::zero(point)};
+            // Sometimes the return type of las::zero() is not POINT. In this case
+            // we must convert it to POINT, but also must be careful that the POINT
+            // doesn't then end up referencing data owned by a temporary that has
+            // been destroyed.
+            auto temp{las::zero(point)};
```
tveasey: Less to think about
```suggestion
            auto temp = las::zero(point);
```

include/maths/CKdTree.h
```diff
@@ -260,7 +265,12 @@ class CKdTree {
             // with infinite distances so we get the correct value for the furthest
             // nearest neighbour at the start of the branch and bound search.
             COrderings::SLess less;
-            POINT distancesToHyperplanes{las::zero(point)};
+            // Sometimes the return type of las::zero() is not POINT. In this case
+            // we must convert it to POINT, but also must be careful that the POINT
+            // doesn't then end up referencing data owned by a temporary that has
+            // been destroyed.
+            auto temp{las::zero(point)};
```
tveasey: As before
```suggestion
            auto temp = las::zero(point);
```

PR #855, opened 2019-11-28
[ML] Add unit test checks for new hyperparameters in supervised learning

### LINKED ISSUE #849: [ML] Trees can have leaves without any data points
It seams that the splitting algorithm continues to split a node even if there are no data points in one of the branches. This leads to creation of degenerated large trees, which are only limited by the maximum tree size. This also breaks algorithms like TreeSHAP, which assume that every leaf contains some data.

### LINKED ISSUE #851: [ML] Add unit test checks for new hyperparameters in supervised learning
For the tests with all hyperparameters set (e.g. `m_RegularizationOverride.*`), we need to check that those would set the actual hyperparameter values (e.g. `m_Regularization.*`). I just noticed that setting `softTreeDepthLimit` and `softTreeDepthTolerance` was missing and the unit test  `testRunBoostedTreeRegressionTrainingWithParams` didn't fail.

valeriy42:
This PR adds check that user-specified hyperparameters are used as final hyperparameters by the boosted trees. Implementation of the unit test required factoring out classes `CBoostedTreeHyperparameters` and `CBoostedTreeRegularization` from `CBoostedTreeImpl`.

Also I added the setting of some user-specified hyperparameters in `CBoostedTreeFactory.cc` that were missing before.

Fixes #851 
Fixes #849

### DIFF
--- a/include/maths/CBoostedTreeHyperparameters.h
+++ b/include/maths/CBoostedTreeHyperparameters.h
@@ -174,7 +174,7 @@ const std::string CBoostedTreeRegularization<T>::REGULARIZATION_SOFT_TREE_DEPTH_
     "regularization_soft_tree_depth_tolerance"};
 
 //! \brief The algorithm parameters we'll directly optimise to improve test error.
-class CBoostedTreeHyperparameters {
+class MATHS_EXPORT CBoostedTreeHyperparameters {
 public:
     using TRegularization = CBoostedTreeRegularization<double>;
 
--- a/include/maths/CBoostedTreeHyperparameters.h
+++ b/include/maths/CBoostedTreeHyperparameters.h
@@ -1,0 +1,206 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#ifndef INCLUDED_ml_maths_CBoostedTreeHyperparameters_h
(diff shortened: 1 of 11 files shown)


include/maths/CBoostedTreeHyperparameters.h
```diff
@@ -0,0 +1,242 @@
...
+    "regularization_leaf_weight_penalty_multiplier"};
+template<typename T>
+const std::string CBoostedTreeRegularization<T>::REGULARIZATION_SOFT_TREE_DEPTH_LIMIT_TAG{
+    "regularization_soft_tree_depth_limit"};
+template<typename T>
+const std::string CBoostedTreeRegularization<T>::REGULARIZATION_SOFT_TREE_DEPTH_TOLERANCE_TAG{
+    "regularization_soft_tree_depth_tolerance"};
+
+//! \brief The algorithm parameters we'll directly optimise to improve test error.
+class CBoostedTreeHyperparameters {
```
tveasey: I don't think Windows will like this:
```suggestion
class MATHS_EXPORT CBoostedTreeHyperparameters {
```

include/maths/CBoostedTreeImpl.h
```diff
@@ -120,11 +123,13 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     //! Visit this tree trainer implementation.
     void accept(CBoostedTree::CVisitor& visitor);
 
+    //! \return Class containing best hyperparameters.
```
tveasey: nit: strictly this is an object of the class but how about
```suggestion
    //! \return The best hyperparameters for validation error found so far.
```

PR #856, opened 2019-11-28
[7.6][ML] Stochastic derivatives for boosted tree training

tveasey:
Backport #811.


droberts195:
@tveasey now you're backporting this it may be appropriate to change some of the test thresholds that had to be adjusted in the second commit to the backport of #827 to account for the fact that these two PRs were backported to 7.x in the opposite order to their additions to master.  See https://github.com/elastic/ml-cpp/pull/829/commits/75a2a1a8ed760289da76da856df6fef6bfc65e2c

For the future it would be best if those test thresholds are the same on master and 7.x if they can be.

PR #857, opened 2019-11-28
[ML] Compute SHAP values for supervised learning

valeriy42:
This PR introduces the computation of SHAP (SHapley Additive exPlanation) values for feature importance. Refer [Consistent Individualized Feature Attribution for Tree Ensembles](https://arxiv.org/abs/1802.03888) by Lundberg et al. for details to the original algorithm.


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+        HANDLE_FATAL(<< "Internal error: no model available for prediction. "
+                     << "Please report this problem.");
+        return;
+    }
+    bool successful;
+    auto treeFeatureImportance =
+        std::make_unique<maths::CTreeShapFeatureImportance>(m_BestForest);
+    TDoubleVecVec shapValues;
+    TDoubleVec shapTotal;
+    size_t numberInputFields = predictionColumn(frame.numberColumns()) - 1;
```
tveasey: Should this not be `m_Encoder->numberEncodedColumns() - 1`, i.e. the number of regressors? Also for consistency:
```suggestion
    std::size_t numberInputFields = predictionColumn(frame.numberColumns()) - 1;
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+                                         core::CDataFrame& frame,
+                                         const TProgressCallback&) {
+    if (m_BestForestTestLoss == INF) {
+        HANDLE_FATAL(<< "Internal error: no model available for prediction. "
+                     << "Please report this problem.");
+        return;
+    }
+    bool successful;
+    auto treeFeatureImportance =
+        std::make_unique<maths::CTreeShapFeatureImportance>(m_BestForest);
```
```suggestion
        std::make_unique<CTreeShapFeatureImportance>(m_BestForest);
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+    TDoubleVecVec shapValues;
+    TDoubleVec shapTotal;
+    size_t numberInputFields = predictionColumn(frame.numberColumns()) - 1;
+    topShapValues = (topShapValues < numberInputFields) ? topShapValues : numberInputFields;
+    std::tie(shapValues, shapTotal) =
+        treeFeatureImportance->shap(frame, *m_Encoder, numberInputFields);
+
+    // get indices of the top elements
+    std::vector<std::size_t> indices(shapTotal.size());
```
tveasey: I think we have the typedef
```suggestion
    TSizeVec indices(shapTotal.size());
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+    std::iota(indices.begin(), indices.end(), 0);
+    std::nth_element(indices.begin(), indices.end(), indices.begin() + topShapValues - 1,
+                     [&shapTotal](std::size_t a, std::size_t b) {
+                         return shapTotal[a] > shapTotal[b];
+                     });
+
+    // write top columns to the result
+    std::size_t offset{frame.numberColumns()};
+    frame.resizeColumns(m_NumberThreads, frame.numberColumns() + topShapValues);
```
tveasey: Although these aren't computed until the end, my inclination is to only resize the data frame once and that this should therefore be reflected in the `numberExtraColumnsForTrain`. The resize operation is expensive and doing it twice means we have to copy the data again.

lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -296,6 +305,12 @@ const std::string CDataFrameTrainBoostedTreeRunner::FEATURE_BAG_FRACTION{"featur
 const std::string CDataFrameTrainBoostedTreeRunner::NUMBER_FOLDS{"number_folds"};
 const std::string CDataFrameTrainBoostedTreeRunner::NUMBER_ROUNDS_PER_HYPERPARAMETER{"number_rounds_per_hyperparameter"};
 const std::string CDataFrameTrainBoostedTreeRunner::BAYESIAN_OPTIMISATION_RESTARTS{"bayesian_optimisation_restarts"};
+const std::string CDataFrameTrainBoostedTreeRunner::TOP_SHAP_VALUES{"top_shap_values"};
+const std::string CDataFrameTrainBoostedTreeRunner::SHAP_PREFIX{"shap_"};
```
tveasey: Maybe
```suggestion
const std::string CDataFrameTrainBoostedTreeRunner::SHAP_PREFIX{"shap."};
```
This way they translated into nested fields and this matches outlier detection feature influence.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+    TStrVec columnNames(frame.columnNames());
+    for (int i = 0; i < topShapValues; ++i) {
+        columnNames[offset + i] = SHAP_PREFIX + frame.columnNames()[indices[i]];
```
tveasey: I can't see where this is defined, but we'll have a circular library dependency problem if this uses the version in `api` library. I'd be inclined to pass this in.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1446,54 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+    }
+    frame.columnNames(columnNames);
+
+    std::tie(std::ignore, successful) = frame.writeColumns(
+        m_NumberThreads, 0, frame.numberRows(), [&](TRowItr beginRows, TRowItr endRows) {
+            for (auto row = beginRows; row != endRows; ++row) {
+                for (int i = 0; i < topShapValues; ++i) {
```
tveasey: nit: can we make loops over `std::size_t` in this function.

lib/maths/CDataFrameCategoryEncoder.cc
```diff
@@ -231,8 +231,12 @@ std::size_t CEncodedDataFrameRowRef::numberColumns() const {
     return m_Encoder->numberEncodedColumns();
 }
 
-CDataFrameCategoryEncoder::CDataFrameCategoryEncoder(CMakeDataFrameCategoryEncoder builder) {
-    m_Encodings = builder.makeEncodings();
+CDataFrameCategoryEncoder::CDataFrameCategoryEncoder(CMakeDataFrameCategoryEncoder builder)
+    : CDataFrameCategoryEncoder(&builder) {
+}
+
+CDataFrameCategoryEncoder::CDataFrameCategoryEncoder(CMakeDataFrameCategoryEncoder* builder) {
```
tveasey: This seems odd to me: if this is passed in by pointer then it could be null, but this can't support it being null. What is forcing us to create this new constructor?

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+                    }
+                    for (int j = 0; j < phi.size(); ++j) {
+                        phi[j] /= m_Trees.size();
+                        phiSum[j] += std::fabs(phi[j]);
+                    }
+
+                    phiVec.emplace_back(std::move(phi));
+                }
+            },
+            std::make_pair(TDoubleVecVec(), TDoubleVec(numberFeatures, 0))));
```
tveasey: We should `reserve` this vector. However, I'm a bit uncomfortable about writing all these values to a vector of vectors and then writing a subset of them back to the data frame. This particularly breaks matters if the frame is stored off heap for size constraints. I feel like a better strategy is to resize the data frame first and write all the values directly into the data frame storage. When we come to write results we can simply write out the top n.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+}
+
+CTreeShapFeatureImportance::TDoubleVec
+CTreeShapFeatureImportance::samplesPerNode(const TTree& tree,
+                                           const core::CDataFrame& frame,
+                                           const CDataFrameCategoryEncoder& encoder) const {
+    auto result = frame.readRows(
+        1, 0, frame.numberRows(),
+        core::bindRetrievableState(
+            [&](std::vector<double>& state, TRowItr beginRows, TRowItr endRows) {
```
tveasey:
```suggestion
            [&](TDoubleVec& state, TRowItr beginRows, TRowItr endRows) {
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+    maxDepthVec.reserve(m_Trees.size());
+    for (auto& tree : m_Trees) {
+        auto samplesPerNode = this->samplesPerNode(tree, frame, encoder);
+        std::size_t maxDepth = this->updateNodeValues(tree, 0, samplesPerNode, 0);
+        maxDepthVec.push_back(maxDepth);
+        m_SamplesPerNode.emplace_back(std::move(samplesPerNode));
+    }
+
+    auto result = frame.readRows(
+        1, 0, frame.numberRows(),
```
tveasey: I think this should be thread-able no? Basically, you get a copy of state for each thread so as long as you can combine and this isn't mutating any other state you should be ok.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+
+ml::maths::CTreeShapFeatureImportance::CTreeShapFeatureImportance(TTreeVec trees)
+    : m_Trees{std::move(trees)}, m_SamplesPerNode() {
+    m_SamplesPerNode.reserve(m_Trees.size());
+}
+
+std::size_t CTreeShapFeatureImportance::updateNodeValues(TTree& tree,
+                                                         std::size_t nodeIndex,
+                                                         const TDoubleVec& samplesPerNode,
+                                                         std::size_t depth) {
```
tveasey: const or better static?

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+
+void CTreeShapFeatureImportance::shapRecursive(const TTree& tree,
+                                               const TDoubleVec& samplesPerNode,
+                                               const CEncodedDataFrameRowRef& encodedRow,
+                                               CTreeShapFeatureImportance::TDoubleVec& phi,
+                                               SPath splitPath,
+                                               std::size_t nodeIndex,
+                                               double parentFractionZero,
+                                               double parentFractionOne,
+                                               int parentFeatureIndex) {
```
tveasey: const or better static?

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+                            incomingFractionOne, splitFeature);
+        this->shapRecursive(tree, samplesPerNode, encodedRow, phi, splitPath, coldIndex,
+                            incomingFractionZero * coldFractionZero, 0.0, splitFeature);
+    }
+}
+
+void CTreeShapFeatureImportance::extendPath(SPath& path,
+                                            double fractionZero,
+                                            double fractionOne,
+                                            int featureIndex) {
```
tveasey: const or better static?

lib/api/CDataFrameTrainBoostedTreeRegressionRunner.cc
```diff
@@ -73,6 +75,18 @@ void CDataFrameTrainBoostedTreeRegressionRunner::writeOneRow(
     writer.Double(row[columnHoldingPrediction]);
     writer.Key(IS_TRAINING_FIELD_NAME);
     writer.Bool(maths::CDataFrameUtils::isMissing(row[columnHoldingDependentVariable]) == false);
+    if (this->topShapValues() > 0) {
+        for (int i = 0; i < row.numberColumns(); ++i) {
```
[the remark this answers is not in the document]
valeriy42: Fixed in 6f81bdad

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -593,7 +593,7 @@ BOOST_AUTO_TEST_CASE(testRunBoostedTreeRegressionTrainingWithStateRecovery) {
                 numberRoundsPerHyperparameter, 12, {}, params.s_Alpha,
                 params.s_Lambda, params.s_Gamma, params.s_SoftTreeDepthLimit,
                 params.s_SoftTreeDepthTolerance, params.s_Eta, params.s_MaximumNumberTrees,
-                params.s_FeatureBagFraction, &persisterSupplier);
+                params.s_FeatureBagFraction, 0, &persisterSupplier);
```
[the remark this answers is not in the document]
valeriy42: fixed

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,248 @@
...
+    for (std::size_t i = 1; i < state.size(); ++i) {
+        for (std::size_t nodeIndex = 0; nodeIndex < totalSamplesPerNode.size(); ++nodeIndex) {
+            totalSamplesPerNode[nodeIndex] += state[i].s_FunctionState[nodeIndex];
+        }
+    }
+
+    return totalSamplesPerNode;
+}
+
+ml::maths::CTreeShapFeatureImportance::CTreeShapFeatureImportance(TTreeVec trees)
```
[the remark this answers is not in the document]
valeriy42: fixed

lib/maths/CBoostedTreeImpl.cc
valeriy42: Fixed both :+1:

lib/maths/CBoostedTreeImpl.cc
valeriy42: fixed

lib/maths/CBoostedTreeImpl.cc
valeriy42: Fixed

lib/api/CDataFrameTrainBoostedTreeRunner.cc
valeriy42: Fixed, also I moved the constant into `CDataFrameRegressionModel`

lib/maths/CBoostedTreeImpl.cc
valeriy42: Following your suggestion, I removed the need for the loops here altogether.

lib/maths/CBoostedTreeImpl.cc
valeriy42: That is why I moved the definition to `CDataFrameRegressionModel` which is a part of `maths`

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Fixed.

lib/maths/CDataFrameCategoryEncoder.cc
valeriy42: Hm, I need this to override the logic of `CMakeDataFrameCategoryEncoder` for unit testing. To this end, inherit from `CMakeDataFrameCategoryEncoder`, but if I pass it by reference to the encoder, I loose polymorphic behavior. Do you have any better idea?

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: I fixed it by writing directly to the data frame.

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: fixed

lib/maths/CBoostedTreeImpl.cc
valeriy42: As discussed, this change would have trade-offs which are not necessarily advantageous.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -1017,5 +1017,10 @@ void CBoostedTreeFactory::noopRecordProgress(double) {
 
 void CBoostedTreeFactory::noopRecordMemoryUsage(std::int64_t) {
 }
+
+CBoostedTreeFactory& CBoostedTreeFactory::shapValues(bool shapValues) {
+    m_TreeImpl->m_ShapValues = shapValues;
```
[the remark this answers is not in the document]
valeriy42: Fixed in 006698f5

lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -157,6 +159,9 @@ CDataFrameTrainBoostedTreeRunner::CDataFrameTrainBoostedTreeRunner(
     if (bayesianOptimisationRestarts > 0) {
         m_BoostedTreeFactory->bayesianOptimisationRestarts(bayesianOptimisationRestarts);
     }
+    if (topShapValues > 0) {
+        m_TopShapValues = topShapValues;
```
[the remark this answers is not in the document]
valeriy42: Fixed in 87dfec17

valeriy42:
Thank you @tveasey for your helpful reviews. I adjusted the code as you suggested. Now I am writing the SHAP values directly into the data frame instead of an extra array. I changed the memory estimation accordingly in the factory class, please let me know if this suffice our I need to adjust the impl class as well.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -817,4 +817,128 @@ BOOST_AUTO_TEST_CASE(testCategoricalFieldsEmptyAsMissing) {
...
+
+        auto frame = setupLinearRegressionData(fieldNames, fieldValues,
+                                               analyzer, weights, values);
+        analyzer.handleRecord(fieldNames, {"", "", "", "", "", "", "$"});
+
+        rapidjson::Document results;
+        rapidjson::ParseResult ok(results.Parse(output.str()));
+        BOOST_TEST_REQUIRE(static_cast<bool>(ok) == true);
+
+        double c1_sum, c2_sum, c3_sum, c4_sum;
```
tveasey: nit: can we stick to coding standards, i.e. c1Sum, etc.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -817,4 +817,128 @@ BOOST_AUTO_TEST_CASE(testCategoricalFieldsEmptyAsMissing) {
...
+                c1_sum += std::fabs(
+                    result["row_results"]["results"]["ml"]["shap.c1"].GetDouble());
+                c2_sum += std::fabs(
+                    result["row_results"]["results"]["ml"]["shap.c2"].GetDouble());
+                c3_sum += std::fabs(
+                    result["row_results"]["results"]["ml"]["shap.c3"].GetDouble());
+                c4_sum += std::fabs(
+                    result["row_results"]["results"]["ml"]["shap.c4"].GetDouble());
+            }
+        }
```
tveasey: I think it might be useful to state the linear function we're fitting here to explain these inequalities, i.e. `50 c_1 + 150 c_2 + 50 c_3 - 50 c_4`. This leads to a question: why do we expect `c1_sum` > `c3_sum`, since the feature values come from the same distribution? Isn't this just an artefact of sampling, in which case could we maybe make w_3 < w_1 < w_2.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -817,4 +817,128 @@ BOOST_AUTO_TEST_CASE(testCategoricalFieldsEmptyAsMissing) {
...
+
+        TStrVec fieldNames{"c1", "c2", "c3", "c4", "c5", ".", "."};
+        TStrVec fieldValues{"", "", "", "", "", "0", ""};
+        test::CRandomNumbers rng;
+        TDoubleVec weights{50, 150, 50, -50};
+
+        TDoubleVec values;
+        rng.generateUniformSamples(-10.0, 10.0, weights.size() * rows, values);
+
+        // make last column categorical
```
tveasey: Isn't this making the first column categorical?

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
tveasey: Ok, reading the code above more carefully, I think this comes from the fact that mean(|c1 feature values|) > mean(|c3 feature values|), etc. So I think the test is good as is, but this needs more explanation.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -817,4 +817,128 @@ BOOST_AUTO_TEST_CASE(testCategoricalFieldsEmptyAsMissing) {
...
+
+        auto frame = setupLinearRegressionData(fieldNames, fieldValues,
+                                               analyzer, weights, values);
+        analyzer.handleRecord(fieldNames, {"", "", "", "", "", "", "$"});
+
+        rapidjson::Document results;
+        rapidjson::ParseResult ok(results.Parse(output.str()));
+        BOOST_TEST_REQUIRE(static_cast<bool>(ok) == true);
+
+        double c1_sum, c2_sum, c3_sum, c4_sum;
```
tveasey: Unused

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -817,4 +817,128 @@ BOOST_AUTO_TEST_CASE(testCategoricalFieldsEmptyAsMissing) {
...
+            }
+        }
+        BOOST_TEST_REQUIRE(c2_sum > c1_sum);
+        BOOST_TEST_REQUIRE(c1_sum > c3_sum);
+        BOOST_TEST_REQUIRE(c1_sum > c4_sum);
+        BOOST_REQUIRE_CLOSE(c3_sum, c4_sum, 80); // c3 and c4 within 80% of each other
+    }
+
+    // test with 0 shap values
+    {
```
tveasey: Could we maybe factor out boilerplate code to set up the input data as a lambda?

lib/maths/unittest/CTreeShapFeatureImportanceTest.cc
```diff
@@ -0,0 +1,189 @@
...
+    std::size_t offset{frame->numberColumns()};
+    treeFeatureImportance->shap(*frame, *encoder, frame->numberColumns(), offset);
+    frame->readRows(1, [&](TRowItr beginRows, TRowItr endRows) {
+        for (auto row = beginRows; row != endRows; ++row) {
+            for (std::size_t col = 0; col < 2; ++col) {
+                BOOST_TEST_REQUIRE((*row)[offset + col] == expectedPhi[row->index()][col]);
+            }
+        }
+    });
+}
```
tveasey: I think a nice test would be to test invariant "sum of shap values" = "prediction - mean" for some randomised input.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1445,5 +1451,50 @@ const double CBoostedTreeImpl::INF{std::numeric_limits<double>::max()};
...
+        TSizeVec indices(shapTotal.size());
+        std::iota(indices.begin(), indices.end(), offset);
+        std::nth_element(indices.begin(), indices.end(),
+                         indices.begin() + topShapValues - 1,
+                         [&shapTotal](std::size_t a, std::size_t b) {
+                             return shapTotal[a] > shapTotal[b];
+                         });
+        m_TopShapIndices = TSizeVec(topShapValues);
+        std::copy(indices.begin(), indices.begin() + topShapValues,
+                  m_TopShapIndices.get().begin());
```
tveasey: I'm reading this as we output top overall shap values for each feature for each example. It seems to me like it would be more useful to return separate result: feature shap values, and then top (local) shap values for each example. If we return indices for all shap values then the writer can simply iterate over them and select the top n. You'd then introduce fields called `shap.<feature foo>`, so different results would have different fields set. WDYT?

include/maths/CBoostedTree.h
```diff
@@ -391,6 +399,8 @@ class MATHS_EXPORT CBoostedTree final : public CDataFrameRegressionModel {
     //! Get the column containing the model's prediction for the dependent variable.
     std::size_t columnHoldingPrediction(std::size_t numberColumns) const override;
 
+    const TOptionalSizeVec& columnsHoldingShapValues() const override;
```
tveasey: How about just returning an empty vector if there aren't anything rather than making this optional? Seems cleaner.

include/maths/CDataFrameRegressionModel.h
```diff
@@ -53,6 +62,14 @@ class MATHS_EXPORT CDataFrameRegressionModel {
     //! Get the column containing the model's prediction for the dependent variable.
     virtual std::size_t columnHoldingPrediction(std::size_t numberColumns) const = 0;
 
+    //! Get the range of column indices with SHAP values
+    //!
+    //! \note The boost::irange can be sustituted by ranges::views::iota once this becomes available in C++20
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+    using TIntVec = std::vector<int>;
+    using TDoubleVec = std::vector<double>;
+    using TDoubleVecVec = std::vector<TDoubleVec>;
+
+    struct SPath {
+        SPath(std::size_t length)
+            : s_FractionOnes(length), s_FractionZeros(length),
+              s_FeatureIndex(length, -1), s_Scale(length), s_NextIndex(0),
+              s_MaxLength(length) {}
+        ~SPath() = default;
```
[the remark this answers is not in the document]
valeriy42: Good catch. I included it while hunting a memory leak, but it has become obsolete since.

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+class MATHS_EXPORT CTreeShapFeatureImportance {
+public:
+    using TTree = std::vector<CBoostedTreeNode>;
+    using TTreeVec = std::vector<TTree>;
+    using TIntVec = std::vector<int>;
+    using TDoubleVec = std::vector<double>;
+    using TDoubleVecVec = std::vector<TDoubleVec>;
+
+    struct SPath {
+        SPath(std::size_t length)
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+
+public:
+    explicit CTreeShapFeatureImportance(TTreeVec trees, std::size_t threads = 1);
+
+    //! Compute SHAP values for the data in \p frame using the specified \p encoder.
+    //!
+    //!\param[in] numberFeatures number of features. If set to -1, it's assument that the number of feature is equal to the
+    //! number of columns in the \p frame.
+    //!\return pair shap values for all points, sum of magnitues of shap values for every feature.
+    CTreeShapFeatureImportance::TDoubleVec shap(core::CDataFrame& frame,
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+                              double parentFractionZero,
+                              double parentFractionOne,
+                              int parentFeatureIndex,
+                              std::size_t offset,
+                              core::CDataFrame::TRowItr& row);
+    static void extendPath(CTreeShapFeatureImportance::SPath& path,
+                           double fractionZero,
+                           double fractionOne,
+                           int featureIndex);
+    static double sumUnwoundPath(const CTreeShapFeatureImportance::SPath& path, int pathIndex);
```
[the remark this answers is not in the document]
valeriy42: fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+    static void unwindPath(CTreeShapFeatureImportance::SPath& path, int pathIndex);
```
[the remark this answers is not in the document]
valeriy42: fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+        std::size_t s_NextIndex;
+        std::size_t s_MaxLength;
+    };
+
+public:
+    explicit CTreeShapFeatureImportance(TTreeVec trees, std::size_t threads = 1);
+
+    //! Compute SHAP values for the data in \p frame using the specified \p encoder.
+    //!
+    //!\param[in] numberFeatures number of features. If set to -1, it's assument that the number of feature is equal to the
```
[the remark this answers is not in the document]
valeriy42: fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+    //! number of columns in the \p frame.
+    //!\return pair shap values for all points, sum of magnitues of shap values for every feature.
```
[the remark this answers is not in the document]
valeriy42: fixed

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+                                                std::size_t offset);
+
+    static CTreeShapFeatureImportance::TDoubleVec
+    samplesPerNode(const TTree& tree,
+                   const core::CDataFrame& frame,
+                   const CDataFrameCategoryEncoder& encoder,
+                   std::size_t numThreads);
+
+    //! Recursively computes inner node values as weighted average of the children (leaf) values
+    //! Returns maximum depth the the tree
```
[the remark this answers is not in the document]
valeriy42: Fixed

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: fixed

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: fixed

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: Good catch. The comment was outdated.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+
+using TRowItr = core::CDataFrame::TRowItr;
+
+CTreeShapFeatureImportance::TDoubleVec
+CTreeShapFeatureImportance::shap(core::CDataFrame& frame,
+                                 const CDataFrameCategoryEncoder& encoder,
+                                 std::size_t numberFeatures,
+                                 std::size_t offset) {
+    numberFeatures = (numberFeatures != -1) ? numberFeatures : frame.numberColumns();
+    std::vector<std::size_t> maxDepthVec;
```
tveasey:
```suggestion
    TSizeVec maxDepthVec;
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+    return phiSum;
+}
+
+CTreeShapFeatureImportance::TDoubleVec
+CTreeShapFeatureImportance::samplesPerNode(const TTree& tree,
+                                           const core::CDataFrame& frame,
+                                           const CDataFrameCategoryEncoder& encoder,
+                                           std::size_t numThreads) {
+    auto result = frame.readRows(
+        numThreads, 0, frame.numberRows(),
```
tveasey: Also, defaults to range to `[0, frame.numberRows())` if not explicitly provided
```suggestion
       m_NumberThreads,
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+CTreeShapFeatureImportance::shap(core::CDataFrame& frame,
+                                 const CDataFrameCategoryEncoder& encoder,
+                                 std::size_t numberFeatures,
+                                 std::size_t offset) {
+    numberFeatures = (numberFeatures != -1) ? numberFeatures : frame.numberColumns();
+    std::vector<std::size_t> maxDepthVec;
+    maxDepthVec.reserve(m_Trees.size());
+    for (auto& tree : m_Trees) {
+        auto samplesPerNode =
+            CTreeShapFeatureImportance::samplesPerNode(tree, frame, encoder, 0);
```
tveasey: This is passing in 0 for number of threads! But I think the method can just use the member variable in any case and we can drop this parameter.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+                                           const CDataFrameCategoryEncoder& encoder,
+                                           std::size_t numThreads) {
+    auto result = frame.readRows(
+        numThreads, 0, frame.numberRows(),
+        core::bindRetrievableState(
+            [&](TDoubleVec& state, TRowItr beginRows, TRowItr endRows) {
+                for (auto row = beginRows; row != endRows; ++row) {
+                    auto encodedRow{encoder.encode(*row)};
+                    auto node{tree[0]};
+                    state[0] += 1;
```
tveasey: nit
```suggestion
                    state[0] += 1.0;
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+                                           const core::CDataFrame& frame,
+                                           const CDataFrameCategoryEncoder& encoder,
+                                           std::size_t numThreads) {
+    auto result = frame.readRows(
+        numThreads, 0, frame.numberRows(),
+        core::bindRetrievableState(
+            [&](TDoubleVec& state, TRowItr beginRows, TRowItr endRows) {
+                for (auto row = beginRows; row != endRows; ++row) {
+                    auto encodedRow{encoder.encode(*row)};
+                    auto node{tree[0]};
```
tveasey: I think we should make this type `const CBoostedTreeNode*` otherwise you are copying the entire node as you traverse the tree for every single row (and knock on changes below).
```suggestion
                    const CBoostedTreeNode* node{&tree[0]};
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+
+CTreeShapFeatureImportance::TDoubleVec
+CTreeShapFeatureImportance::samplesPerNode(const TTree& tree,
+                                           const core::CDataFrame& frame,
+                                           const CDataFrameCategoryEncoder& encoder,
+                                           std::size_t numThreads) {
+    auto result = frame.readRows(
+        numThreads, 0, frame.numberRows(),
+        core::bindRetrievableState(
+            [&](TDoubleVec& state, TRowItr beginRows, TRowItr endRows) {
```
tveasey: How about a more descriptive name?
```suggestion
            [&](TDoubleVec& samplesPerNode, TRowItr beginRows, TRowItr endRows) {
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+        tree, node.leftChildIndex(), samplesPerNode, depth + 1);
+    std::size_t depthRight = CTreeShapFeatureImportance::updateNodeValues(
+        tree, node.rightChildIndex(), samplesPerNode, depth + 1);
+
+    double leftWeight = samplesPerNode[node.leftChildIndex()];
+    double rightWeight = samplesPerNode[node.rightChildIndex()];
+    double averageValue = (leftWeight * tree[node.leftChildIndex()].value() +
+                           rightWeight * tree[node.rightChildIndex()].value()) /
+                          (leftWeight + rightWeight);
+    node.value(averageValue);
```
tveasey: 👍 I like the idea of repurposing the internal node values for this! It is useful information to surface in its own right.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+std::size_t CTreeShapFeatureImportance::updateNodeValues(TTree& tree,
+                                                         std::size_t nodeIndex,
+                                                         const TDoubleVec& samplesPerNode,
+                                                         std::size_t depth) {
+    auto& node{tree[nodeIndex]};
+    if (node.isLeaf()) {
+        return 0;
+    }
+
+    std::size_t depthLeft = CTreeShapFeatureImportance::updateNodeValues(
```
tveasey: nit: can we stick to braces initialisation.

include/maths/CBoostedTree.h
valeriy42: Indeed. However, this now becomes irrelevant as I let the result writer to pick top SHAP values for every row.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+        auto samplesPerNode =
+            CTreeShapFeatureImportance::samplesPerNode(tree, frame, encoder, 0);
+        std::size_t maxDepth =
+            CTreeShapFeatureImportance::updateNodeValues(tree, 0, samplesPerNode, 0);
+        maxDepthVec.push_back(maxDepth);
+        m_SamplesPerNode.emplace_back(std::move(samplesPerNode));
+    }
+
+    auto result = frame.writeColumns(
+        m_NumberThreads, 0, frame.numberRows(),
```
tveasey: Not needed since this is default behaviour.
```suggestion
        m_NumberThreads,
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+        core::bindRetrievableState(
+            [&](auto& state, TRowItr beginRows, TRowItr endRows) {
```
tveasey:
```suggestion
            [&](TDoubleVec& phiSum, TRowItr beginRows, TRowItr endRows) {
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+                for (auto row = beginRows; row != endRows; ++row) {
+                    auto encodedRow{encoder.encode(*row)};
+                    auto& phiSum = state;
```
tveasey: We don't really need reference to argument
```suggestion
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+                    for (int i = 0; i < m_Trees.size(); ++i) {
+                        //                        phi[frame.numberColumns()] += m_Trees[i][0].value();
+                        SPath path(maxDepthVec[i] + 1);
+                        CTreeShapFeatureImportance::shapRecursive(
+                            m_Trees[i], m_SamplesPerNode[i], encoder,
+                            encodedRow, path, 0, 1.0, 1.0, -1, offset, row);
+                        i += 0;
```
tveasey:
```suggestion
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -0,0 +1,258 @@
...
+                    }
+                    for (int j = 0; j < numberFeatures; ++j) {
+                        row->writeColumn(offset + j,
+                                         (*row)[offset + j] / m_Trees.size());
```
tveasey: Can we pass in the normalisation factor and write once inside `shapRecursive`? I found this line confusing until I saw you were first writing them inside `shapRecursive`.

lib/maths/CBoostedTreeImpl.cc
valeriy42: I adjusted the code to output top shap values per row.

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: done

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: done

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Fixed

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: changed here and below.

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: done

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: whoopsy

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Done

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -0,0 +1,138 @@
...
+    //!
+    //!\param[in] numberFeatures number of features. If set to -1, it's assument that the number of feature is equal to the
+    //! number of columns in the \p frame.
+    //!\return pair shap values for all points, sum of magnitues of shap values for every feature.
+    CTreeShapFeatureImportance::TDoubleVec shap(core::CDataFrame& frame,
+                                                const CDataFrameCategoryEncoder& encoder,
+                                                std::size_t numberFeatures,
+                                                std::size_t offset);
+
+    static CTreeShapFeatureImportance::TDoubleVec
```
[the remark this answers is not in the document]
valeriy42: fixed

lib/api/CDataFrameTrainBoostedTreeRegressionRunner.cc
```diff
@@ -18,6 +18,8 @@
 #include <api/CDataFrameAnalysisSpecification.h>
 #include <api/ElasticsearchStateIndex.h>
 
+#include <numeric>
```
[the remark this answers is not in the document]
valeriy42: done

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: I created a fixture to deal with the boiler plate code

lib/maths/unittest/CTreeShapFeatureImportanceTest.cc
valeriy42: Done.

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Good catch!

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Fixed

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: Doc adjusted.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -153,6 +153,20 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
         }
         writer.EndArray();
     }
+
+    if (this->topShapValues() > 0) {
+        auto shapColumnsRange{this->boostedTree().columnsHoldingShapValues()};
```
tveasey: At least in theory `nth_element` needs to guard against the case `topShapValues() > shapColumnsRange.size()` unless that happens elsewhere? (Also, we probably don't need to actually copy the size vector; we could have a begin and end or hold the vector and return by reference, at which point it is more efficient to have a heap of the top values.) We have a class which makes all this easy in the `maths` library; I think using this works out as clean in code. You'd need a factory method to take care of the type of the comparison operator, but something like should work:
```
template<typename LESS>
auto makeLargestShapAccumulator(std::size_t n, LESS less) {
    return maths::CBasicStatistics::COrderStatisticsHeap<std::size_t, LESS>{n, std::size_t{}, less);
}
...
    auto largestShapValues = makeLargestShapAccumulator(this->topShapValues(), [&row](std::size_t lhs, std::size_t rhs) {
                             return std::fabs(row[lhs]) > std::fabs(row[rhs]);
                         });
    largestShapValues.add(this->boostedTree().columnsHoldingShapValues());
    largestShapValues.sort();
    for (auto i : largestShapValues) {
       ...
    }
```

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -1017,5 +1022,11 @@ void CBoostedTreeFactory::noopRecordProgress(double) {
 
 void CBoostedTreeFactory::noopRecordMemoryUsage(std::int64_t) {
 }
+
+CBoostedTreeFactory& CBoostedTreeFactory::topShapValues(std::size_t topShapValues) {
+    m_TopShapValues = topShapValues;
```
tveasey: Do we need this variable on the factory as well: seems redundant?

lib/maths/CDataFrameCategoryEncoder.cc
```diff
@@ -231,8 +231,12 @@ std::size_t CEncodedDataFrameRowRef::numberColumns() const {
     return m_Encoder->numberEncodedColumns();
 }
 
-CDataFrameCategoryEncoder::CDataFrameCategoryEncoder(CMakeDataFrameCategoryEncoder builder) {
-    m_Encodings = builder.makeEncodings();
+CDataFrameCategoryEncoder::CDataFrameCategoryEncoder(CMakeDataFrameCategoryEncoder& parameters) {
```
tveasey: nit: for consistency can we call this `builder`.

lib/maths/CBoostedTreeFactory.cc
valeriy42: I need it to estimate memory usage with resized data frame.

lib/maths/CDataFrameCategoryEncoder.cc
valeriy42: done

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
valeriy42: Done. Great idea!

valeriy42:
This PR introduces the computation of SHAP (SHapley Additive exPlanation) values for feature importance. Refer "Consistent Individualized Feature Attribution for Tree Ensembles" by Lundberg et al. for details to the original algorithm.

PR #859, opened 2019-11-28
[ML] Adding memory usage methods for categorization classes

droberts195:
This is a step towards adding memory instrumentation to
categorization.  On its own this PR has no externally
visible effect, but by splitting out these boilerplate
changes the PR that eventually adds memory instrumentation
to categorization will be smaller.

### DIFF
--- a/include/core/CMemory.h
+++ b/include/core/CMemory.h
@@ -996,7 +996,7 @@ public:
         componentName += "_list";
 
         std::size_t listSize = (memory_detail::EXTRA_NODES + t.size()) *
-                               (sizeof(T) + 4 * sizeof(std::size_t));
+                               (sizeof(T) + 2 * sizeof(std::size_t));
 
         CMemoryUsage::SMemoryUsage usage(componentName, listSize);
         CMemoryUsage::TMemoryUsagePtr ptr = mem->addChild();
--- a/include/core/CMemory.h
+++ b/include/core/CMemory.h
@@ -6,6 +6,7 @@
 #ifndef INCLUDED_ml_core_CMemory_h
 #define INCLUDED_ml_core_CMemory_h
 
+#include <core/BoostMultiIndex.h>
 #include <core/CLogger.h>
 #include <core/CMemoryUsage.h>
 #include <core/CNonInstantiatable.h>
@@ -14,6 +15,7 @@
 #include <boost/any.hpp>
 #include <boost/circular_buffer_fwd.hpp>
 #include <boost/container/container_fwd.hpp>
+#include <boost/mpl/range_c.hpp>
 #include <boost/optional/optional_fwd.hpp>
 #include <boost/shared_array.hpp>
 #include <boost/type_traits/is_pointer.hpp>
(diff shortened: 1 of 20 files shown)


include/core/CMemory.h
```diff
@@ -1009,6 +1035,37 @@ class CORE_EXPORT CMemoryDebug : private CNonInstantiatable {
...
+        // can be of a different type.  To accurately determine the overhead
+        // would require some serious template metaprogramming to interpret the
+        // "typename I" template argument, and it's just not worth it given the
+        // infrequent and relatively simple usage (generally just two indices
+        // in our current codebase).  Therefore there's an approximation here
+        // that the overhead is 2 pointers per entry per index.
+        using TMultiIndex = boost::multi_index::multi_index_container<T, I, A>;
+        constexpr std::size_t indexCount{
+            boost::mpl::size<typename TMultiIndex::index_type_list>::value};
+        std::string componentName(name);
```
edsavage: could `componentName` be `const std::string &`?

include/core/CMemory.h
```diff
@@ -1009,6 +1035,37 @@ class CORE_EXPORT CMemoryDebug : private CNonInstantiatable {
...
+
+        std::size_t items = t.size();
+        CMemoryUsage::SMemoryUsage usage(
+            componentName + "::" + typeid(T).name(),
+            items * (sizeof(T) + 2 * indexCount * sizeof(std::size_t)));
+        CMemoryUsage::TMemoryUsagePtr ptr = mem->addChild();
+        ptr->setName(usage);
+
+        componentName += "_item";
+        for (auto i = t.begin(); i != t.end(); ++i) {
```
edsavage: `const auto& i` ?

include/core/CMemory.h
```diff
@@ -491,6 +493,30 @@ class CORE_EXPORT CMemory : private CNonInstantiatable {
...
+        // "typename I" template argument, and it's just not worth it given the
+        // infrequent and relatively simple usage (generally just two indices
+        // in our current codebase).  Therefore there's an approximation here
+        // that the overhead is 2 pointers per entry per index.
+        using TMultiIndex = boost::multi_index::multi_index_container<T, I, A>;
+        constexpr std::size_t indexCount{
+            boost::mpl::size<typename TMultiIndex::index_type_list>::value};
+        std::size_t mem = 0;
+        if (!memory_detail::SDynamicSizeAlwaysZero<T>::value()) {
+            for (auto i = t.begin(); i != t.end(); ++i) {
```
edsavage: `const auto& i`?

lib/core/CompressUtils.cc
```diff
@@ -117,6 +118,37 @@ bool CCompressUtil::prepareToReturnData(bool finish) {
     return true;
 }
 
+void CCompressUtil::debugMemoryUsage(core::CMemoryUsage::TMemoryUsagePtr mem) const {
+    mem->setName("CCompressUtil");
+    core::CMemoryDebug::dynamicSize("m_FullResult", m_FullResult, mem);
+    if (m_ZlibStrm.state != nullptr) {
+        mem->addItem("m_ZlibStrm", 5952); // See below for where 5952 came from
```
edsavage: Could we maybe also use a named constant for the `5952` value?

include/core/CMemory.h
droberts195: It gets modified 8 lines below.

There would be other ways to do what this method does, but I copied the structure from the adjacent `boost::circular_buffer` method.  I think there's some benefit in keeping the same high level structure for these similar methods.

include/core/CMemory.h
droberts195: The idiomatic way to iterate with iterators is `auto` rather than `const auto&`.  All the equivalent loops in the file are done like this one.  The `auto` variable is the iterator, not the container element.

But it did make me wonder why all these loops are using iterators rather than `for (const auto& i : t) {`.  I tried changing them and got this warning:

```
/Users/dave/ml-cpp/include/core/CMemory.h:334:30: warning: loop variable 'i' is always a copy because the range of type 'const std::vector<bool>' does not return a reference [-Wrange-loop-analysis]
            for (const auto& i : t) {
```

So I think it's because _sometimes_ the `vector` method will have to deal with `vector<bool>`, so that needs the explicit iterator version to avoid the warning, and all the others are done the same way for consistency.

Obviously I could change the loops _except_ the `vector` one, but I do think with this file that there's a benefit in keeping as much as possible consistent between the methods that deal with different types of container.  There are already subtle differences just dealing with the container structure differences.  I'm not sure it's good to introduce a difference in the way looping is done too.

edsavage: Weird.. Thanks for looking in to it!

PR #861, opened 2019-11-29
Blacklist a number of prediction field names.

przemekwitek:
This PR prevents field name clashes in ml results docs on JSON level.

Relates https://github.com/elastic/elasticsearch/issues/48808

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
@@ -68,6 +68,13 @@ CDataFrameTrainBoostedTreeClassifierRunner::CDataFrameTrainBoostedTreeClassifier
                   this->dependentVariableFieldName()) == categoricalFieldNames.end()) {
         HANDLE_FATAL(<< "Input error: trying to perform classification with numeric target.");
     }
+    const std::set<std::string> predictionFieldNameBlacklist{
+        IS_TRAINING_FIELD_NAME, PREDICTION_PROBABILITY_FIELD_NAME, TOP_CLASSES_FIELD_NAME};
+    if (predictionFieldNameBlacklist.count(this->predictionFieldName())) {
+        HANDLE_FATAL(<< "Input error: prediction_field_name must not be equal to any of "
+                     << core::CContainerPrinter::print(predictionFieldNameBlacklist)
+                     << ".");
+    }
 }
 
 CDataFrameTrainBoostedTreeClassifierRunner::CDataFrameTrainBoostedTreeClassifierRunner(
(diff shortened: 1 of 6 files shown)


lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -68,6 +68,13 @@ CDataFrameTrainBoostedTreeClassifierRunner::CDataFrameTrainBoostedTreeClassifier
                   this->dependentVariableFieldName()) == categoricalFieldNames.end()) {
         HANDLE_FATAL(<< "Input error: trying to perform classification with numeric target.");
     }
+    const std::set<std::string> predictionFieldNameBlacklist{
+        IS_TRAINING_FIELD_NAME, PREDICTION_PROBABILITY_FIELD_NAME, TOP_CLASSES_FIELD_NAME};
+    if (predictionFieldNameBlacklist.count(this->predictionFieldName())) {
```
edsavage: I'd prefer an explicit comparison here, e.g `if (predictionFieldNameBlacklist.count(this->predictionFieldName()) > 0)`
Looking through our code there's no consistency but the bulk of such comparisons are explicit.
Incidentally there's nothing about this in the `STYLEGUIDE` either, maybe there should be?

lib/api/CDataFrameTrainBoostedTreeRegressionRunner.cc
```diff
@@ -52,6 +52,12 @@ CDataFrameTrainBoostedTreeRegressionRunner::CDataFrameTrainBoostedTreeRegression
                   this->dependentVariableFieldName()) != categoricalFieldNames.end()) {
         HANDLE_FATAL(<< "Input error: trying to perform regression with categorical target.");
     }
+    const std::set<std::string> predictionFieldNameBlacklist{IS_TRAINING_FIELD_NAME};
+    if (predictionFieldNameBlacklist.count(this->predictionFieldName())) {
```
edsavage: Same comment as above applies here...

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
droberts195: We used to have this in the old Word doc coding standards - section 5.27: "if statements SHOULD be fully qualified".  It got lost when that was converted to markdown.

edsavage: Created #862 to add this to the `STYLEGUIDE`

przemekwitek: > I'd prefer an explicit comparison here, e.g `if (predictionFieldNameBlacklist.count(this->predictionFieldName()) > 0)`

Done.

lib/api/CDataFrameTrainBoostedTreeRegressionRunner.cc
przemekwitek: Done.

PR #863, opened 2019-11-29
[ML] Assorted runtime optimisations for classification and regression

tveasey:
The principal change is to bucket the candidate split points so we only check a subset of the splits when searching for the derivatives to update when computing split statistics. There are also some smaller optimisations which profiling showed up as worthwhile. The impact is largest for the case we have many metric valued features. These are pure runtime optimisations, i.e. the results are unaffected.

### DIFF
--- a/lib/core/CDataFrame.cc
+++ b/lib/core/CDataFrame.cc
@@ -188,7 +188,7 @@ CDataFrame::TRowFuncVecBoolPr CDataFrame::readRows(std::size_t numberThreads,
 
     return numberThreads > 1
                ? this->parallelApplyToAllRows(numberThreads, beginRows, endRows,
-                                              reader, rowMask, false)
+                                              std::move(reader), rowMask, false)
                : this->sequentialApplyToAllRows(beginRows, endRows, reader, rowMask, false);
 }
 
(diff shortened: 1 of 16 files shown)


lib/core/CDataFrame.cc
```diff
@@ -410,7 +408,7 @@ CDataFrame::TRowFuncVecBoolPr
 CDataFrame::parallelApplyToAllRows(std::size_t numberThreads,
                                    std::size_t beginRows,
                                    std::size_t endRows,
-                                   TRowFunc func,
+                                   TRowFunc& func,
```
droberts195: It seems dodgy to pass this by lvalue reference here, as when it's passed on again it's passed on with `std::move(func)`.  Should it be an rvalue reference here too?

Same for `sequentialApplyToAllRows`.

tveasey: This can be an rvalue reference, but it not in `sequentialApplyToAllRows` because I don't move from there and pass down by reference to another function in a loop.

(It ought to be equivalent passing by lvalue reference because the upshot is it'll still get moved into `SCallableWithBoundState`, but perhaps passing by rvalue reference makes the intention clearer.)

PR #864, opened 2019-11-29
[ML] Log best hyperparameters and runtime summary

valeriy42:
This PR changes the logging level for the best hyperparameters from TRACE to INFO. It also adds tracking runtime for the hyperparameter optimization steps and logs the total number of steps, mean runtime per iteration and its stdev.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -1167,18 +1167,18 @@ void CBoostedTreeImpl::captureBestHyperparameters(const TMeanVarAccumulator& los
                 std::sqrt(CBasicStatistics::variance(lossMoments))};
     if (loss < m_BestForestTestLoss) {
         m_BestForestTestLoss = loss;
-        m_BestHyperparameters = SHyperparameters{m_Regularization, m_DownsampleFactor,
-                                                 m_Eta, m_EtaGrowthRatePerTree,
-                                                 m_FeatureBagFraction};
+        m_BestHyperparameters = CBoostedTreeHyperparameters{
+            m_Regularization, m_DownsampleFactor, m_Eta, m_EtaGrowthRatePerTree,
+            m_FeatureBagFraction};
     }
 }
 
 void CBoostedTreeImpl::restoreBestHyperparameters() {
-    m_Regularization = m_BestHyperparameters.s_Regularization;
-    m_DownsampleFactor = m_BestHyperparameters.s_DownsampleFactor;
-    m_Eta = m_BestHyperparameters.s_Eta;
-    m_EtaGrowthRatePerTree = m_BestHyperparameters.s_EtaGrowthRatePerTree;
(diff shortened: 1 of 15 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -485,6 +493,12 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
         m_BestForest = this->trainForest(frame, this->allTrainingRowsMask(), recordMemoryUsage);
         this->recordState(recordTrainStateCallback);
 
+        timeAccumulator.add(stopWatch.stop());
+
+        LOG_INFO(<< "Training finished after " << m_CurrentRound << " iterations. Time per iteration in ms mean: "
+                 << timeAccumulator.s_Moments[0]
```
```suggestion
                 << CBasicStatistics::mean(timeAccumulator)
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -485,6 +493,12 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
...
+                 << " std. dev:  " << std::sqrt(timeAccumulator.s_Moments[1]));
```
tveasey: This is the unbiased variance estimate and what we'd normally use
```suggestion
                 << " std. dev:  " << std::sqrt(CBasicStatistics::variance(timeAccumulator)));
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -444,6 +445,11 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
         LOG_TRACE(<< "Test loss = " << m_BestForestTestLoss);
 
     } else if (m_CurrentRound < m_NumberRounds || m_BestForest.empty()) {
+        using TUIntMeanVarAccumulator =
+            CBasicStatistics::SSampleMeanVar<std::uint64_t>::TAccumulator;
```
tveasey: This doesn't like integers. We could do something to manage integer division better, but it is probably better to stick to double and cast when adding the iteration time.

lib/maths/CBoostedTreeImpl.cc
valeriy42: Oh nice! I learned something new today :D

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -444,6 +445,11 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
         LOG_TRACE(<< "Test loss = " << m_BestForestTestLoss);
 
     } else if (m_CurrentRound < m_NumberRounds || m_BestForest.empty()) {
+        using TUIntMeanVarAccumulator =
```
tveasey: nit: this should be renamed to something like TFloatMeanVarAccumulator. (I'd also be inclined to just go with double: there is no real reason to save memory or runtime here.)

PR #867, opened 2019-11-29
[ML] Upgrade to clang 6.0/Apple clang 10.1 for macOS compilation

droberts195:
We will build the macOS version 8.x of the ML C++ using
Apple clang 10.1 on High Sierra or cross-compiling using
clang 6.0 on Ubuntu 18.04.  This will eventually facilitate
use of C++17 (after separate PRs to upgrade the Windows and
Linux compilers have also been merged).

PR #834, opened 2019-11-19
[ML] Upgrade to Visual Studio 2019 for Windows compilation

droberts195:
We will build version 8.x of the ML C++ using Visual Studio 2019
rather than Visual Studio 2017.  This will eventually facilitate
use of C++17 (after separate PRs to upgrade the Mac and Linux
compilers have also been merged).

Since this change involves rebuilding all dependencies, the
libxml2 dependency is also upgraded from version 2.9.4 to
version 2.9.7.

Closes #1015

PR #871, opened 2019-12-02
[ML] Rename anonymous namespace apply function to avoid clash

droberts195:
In C++17 there is a std::apply function that gets chosen in
preference to the apply function in the anonymous namespace
in the metric and event rate bucket gatherers.

This change renames our function so the compiler does not
have to resolve an ambiguity when it's called.


droberts195:
Exactly the same clash was observed on Windows in #834 and on macOS in #867, so this must be a C++17 thing rather than a quirk of one of the compilers.

Even though C++17 changes are only essential in master, it doesn't hurt to backport to 7.x too, which will reduce the chance of future backport clashes.

Issue #845, opened 2019-11-22
Valgrind error in CTimeSeriesDecompositionDetail


droberts195:
`valgrind` detected an error while running the `CTimeSeriesDecompositionTest/testMixedSmoothAndSpikeyDataProblemCase` unit test:

```
2019-11-22 12:32:54,860835 UTC [1558] DEBUG CTestObserver.cc@23 +------------------------------------------------------------------------+
2019-11-22 12:32:54,861307 UTC [1558] DEBUG CTestObserver.cc@24 |  CTimeSeriesDecompositionTest/testMixedSmoothAndSpikeyDataProblemCase  |
2019-11-22 12:32:54,861749 UTC [1558] DEBUG CTestObserver.cc@25 +------------------------------------------------------------------------+
2019-11-22 12:32:55,294146 UTC [1558] DEBUG CTimeSeriesDecompositionTest.cc@1173 timeseries = [(1400490000, 3), (1400491800, 1), (1400493600, 14), (1400495400, 3), (1400497200, 31), (1400499000, 127), (1400500800, 480), (1400502600, 1121), (1400504400, 2240), (1400506200, 3044)] ...
2019-11-22 12:32:55,340594 UTC [1558] DEBUG CTimeSeriesDecompositionDetail.cc@1290 Detected trend at 1400506200
==1558== Conditional jump or move depends on uninitialised value(s)
==1558==    at 0x72867B8: __muldc3 (libgcc2.c:1991)
==1558==    by 0x6096094: ml::maths::CSignal::fft(std::vector<std::complex<double>, std::allocator<std::complex<double> > >&) (complex:1328)
==1558==    by 0x6096778: ml::maths::CSignal::autocorrelations(std::vector<ml::maths::CBasicStatistics::SSampleCentralMoments<ml::core::CFloatStorage, 1u>, std::allocator<ml::maths::CBasicStatistics::SSampleCentralMoments<ml::core::CFloatStorage, 1u> > > const&, std::vector<double, std::allocator<double> >&) (CSignal.cc:211)
[...]

PR #876, opened 2019-12-03
[ML] Adds "make memcheck" target to build system.

edsavage:
`make memcheck` executes the test suites under the `Valgrind` memory checker tool.

This is the first stage in setting up a periodic Jenkins CI build to run `Valgrind` over the code regularly.


mk/toplevel.mk
```diff
@@ -122,6 +122,32 @@ ifdef TOP_DIR_MKF_LAST
 	@ $(MAKE) test -f $(TOP_DIR_MKF_LAST)
 endif
 
+ifdef TOP_DIR_MKF_FIRST
+FIRST_TEST_CMD=$(MAKE) memcheck -f $(TOP_DIR_MKF_FIRST)
```
droberts195: This can't be called `FIRST_TEST_CMD` - it needs a different name so it doesn't corrupt the `test` target.

lib/core/unittest/valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+   fun:_ZN2ml5maths12_GLOBAL__N_132mostSignificantPeriodicComponentENSt3__16vectorINS0_16CBasicStatistics21SSampleCentralMomentsINS_4core13CFloatStorageELj1EEENS2_9allocatorIS8_EEEE
+   fun:_ZN2ml5maths14testForPeriodsERKNS0_33CPeriodicityHypothesisTestsConfigEllRKNSt3__16vectorINS0_16CBasicStatistics21SSampleCentralMomentsINS_4core13CFloatStorageELj1EEENS4_9allocatorISA_EEEE
+   fun:_ZN2ml5maths30CTimeSeriesDecompositionDetail16CPeriodicityTest4testERKNS1_9SAddValueE
+}
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Cond
+   fun:_ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE11__recommendEm
+   fun:_ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE6__initIPcEENS_9enable_ifIXsr21__is_forward_iteratorIT_EE5valueEvE4typeES9_S9_
+   fun:_ZNSt3__112basi
```
droberts195: Didn't we manage to avoid this problem in #852?

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+   fun:_ZNK5boost7phoenix9evaluator4implIRNS0_5actorINS_5proto7exprns_10basic_exprINS4_6tagns_3tag10shift_leftENS4_7argsns_5list2INS3_INS6_IS9_NSB_INS3_INS_3log11v2_mt_posix11expressions3aux25attribute_output_terminalINS3_INS6_IS9_NSB_INS3_INSG_INS3_INS6_IS9_NSB_INS3_INSG_INS3_INS6_IS9_NSB_INS3_INS6_IS9_NSB_INS3_INS6_IS9_NSB_INS3_INSG_INS3_INS6_INS8_8terminalENSA_4termINS0_8argumentILi2EEEEELl0EEEEENS_10posix_time5ptimeENSD_16fallback_to_noneENSD_3aux14light_functionIFvRNSD_24basic_formatting_ostreamIcNSt3__111char_traitsIcEENSU_9allocatorIcEEEERKSP_EEEEEEENS3_INS6_ISH_NSI_IA7_cEELl0EEEEEEELl2EEEEENS3_INS6_ISH_NSI_IiEELl0EEEEEEELl2EEEEENS3_INS6_ISH_NSI_IA3_cEELl0EEEEEEELl2EEEEEN2ml4core7CLogger6ELevelESQ_NSD_10to_log_funIvEEEEEENS3_INS6_ISH_NSI_IcEELl0EEEEEEELl2EEEEENSU_12basic_stringIcSW_SY_EESQ_S1W_EEEES21_EELl2EEEEEiSQ_S1W_EEEES21_EELl2EEEEENSE_15attribute_actorIS26_SQ_NSE_3tag8smessageES3_EEEELl2EEEEERKNS0_7vector2IRNS0_7vector3IPKS2N_RKNSD_11record_viewERNSF_10stream_refISZ_EEEERKNS0_15default_actionsEEENS4_6envns_9empty_envEEclES2O_S36_RKS38_
+}
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Value8
+   fun:_platform_memmove$VARIANT$Haswell
```
droberts195: This is a machine-specific function that's been chosen based on the machine the suppression file was generated on.  It's probably best to replace it with `...` (see https://wiki.wxwidgets.org/Valgrind_Suppression_File_Howto#Wildcards for syntax).

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+   fun:__vfprintf
+   fun:__v2printf
+   fun:_vsnprintf
+   fun:vsnprintf_l
+   fun:snprintf_l
+   fun:_ZNKSt3__17num_putIcNS_19ostreambuf_iteratorIcNS_11char_traitsIcEEEEE6do_putES4_RNS_8ios_baseEcd
+   fun:_ZNSt3__113basic_ostreamIcNS_11char_traitsIcEEElsEd
+   fun:_ZNSt3__1lsIdcNS_11char_traitsIcEEEERNS_13basic_ostreamIT0_T1_EES7_RKNS_7complexIT_EE
+   fun:_ZN2ml4core14printer_detail12CLeafPrinterIN5boost17integral_constantIbLb0EEEE6print_INSt3__17complexIdEEEENS8_12basic_stringIcNS8_11char_traitsIcEENS8_9allocatorIcEEEERKT_S5_
+   fun:_ZN2ml4core14printer_detail12CLeafPrinterIN5boost17integral_constantIbLb0EEEE5printINSt3__17complexIdEEEENS8_12basic_stringIcNS8_11char_traitsIcEENS8_9allocatorIcEEEERKT_
```
droberts195: It's probably possible to reduce the number of suppressions by chopping some of the functions out of the stacks.  For example, it seems that on macOS Valgrind cannot properly trace the memory through the C++ runtime library and C runtime library when printing a number to an `ostream`.  So if you start the suppression here at `_ZNKSt3__17num_putIcNS_19ostreambuf_iteratorIcNS_11char_traitsIcEEEEE6do_putES4_RNS_8ios_baseEcd` and drop everything beneath it then this single suppression can probably cover several others, and they can be deleted.

Another advantage of this will be that similar suppressions won't have to be added each time somebody adds a new test that prints a number.

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+}
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Cond
+   fun:_ZNSt3__115basic_stringbufIcNS_11char_traitsIcEENS_9allocatorIcEEE8overflowEi
+   fun:_ZNSt3__115basic_streambufIcNS_11char_traitsIcEEE6xsputnEPKcl
+   fun:_ZNSt3__115basic_streambufIcNS_11char_traitsIcEEE5sputnEPKcl
+   fun:_ZNSt3__116__pad_and_outputIcNS_11char_traitsIcEEEENS_19ostreambuf_iteratorIT_T0_EES6_PKS4_S8_S8_RNS_8ios_baseES4_
+   fun:_ZNSt3__124__put_character_sequenceIcNS_11char_traitsIcEEEERNS_13basic_ostreamIT_T0_EES7_PKS4_m
+   fun:_ZNSt3__1lsIcNS_11char_traitsIcEENS_9allocatorIcEEEERNS_13basic_ostreamIT_T0_EES9_RKNS_12basic_stringIS6_S7_T1_EE
```
droberts195: Similarly this stack trace could end at `_ZNSt3__1lsIcNS_11char_traitsIcEENS_9allocatorIcEEEERNS_13basic_ostreamIT_T0_EES9_RKNS_12basic_stringIS6_S7_T1_EE` and cover many similar ones.

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Value8
+   fun:_ZNSt3__115basic_streambufIcNS_11char_traitsIcEEE5sputcEc
+   fun:_ZNSt3__115basic_stringbufIcNS_11char_traitsIcEENS_9allocatorIcEEE8overflowEi
+   fun:_ZNSt3__115basic_streambufIcNS_11char_traitsIcEEE6xsputnEPKcl
+   fun:_ZNSt3__115basic_streambufIcNS_11char_traitsIcEEE5sputnEPKcl
+   fun:_ZNSt3__116__pad_and_outputIcNS_11char_traitsIcEEEENS_19ostreambuf_iteratorIT_T0_EES6_PKS4_S8_S8_RNS_8ios_baseES4_
+   fun:_ZNSt3__124__put_character_sequenceIcNS_11char_traitsIcEEEERNS_13basic_ostreamIT_T0_EES7_PKS4_m
+   fun:_ZNSt3__1lsIcNS_11char_traitsIcEENS_9allocatorIcEEEERNS_13basic_ostreamIT_T0_EES9_RKNS_12basic_stringIS6_S7_T1_EE
```
droberts195: Probably a good rule of thumb would be to end each stack trace at the line containing `basic_ostream`, then delete the resulting duplicates.

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+}
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Cond
+   fun:_ZNK16ImageLoaderMachO9getRPathsERKN11ImageLoader11LinkContextERNSt3__16vectorIPKcNS4_9allocatorIS7_EEEE
+   fun:_ZN11ImageLoader22recursiveLoadLibrariesERKNS_11LinkContextEbRKNS_10RPathChainEPKc
+   fun:_ZN11ImageLoader4linkERKNS_11LinkContextEbbbRKNS_10RPathChainEPKc
+   fun:_ZN4dyld4linkEP11ImageLoaderbbRKNS0_10RPathChainEj
+   fun:_ZN4dyld5_mainEPK12macho_headermiPPKcS5_S5_Pm
+   fun:_ZN13dyldbootstrap5startEPK12macho_headeriPPKclS2_Pm
```
droberts195: `_ZN4dyld4linkEP11ImageLoaderbbRKNS0_10RPathChainEj` runs before `main`, so any problems are out of our control and all the stack trace that include this can be replaced with just one like:

```
  ...
  fun:_ZN4dyld4linkEP11ImageLoaderbbRKNS0_10RPathChainEj
```

valgrind.supp
```diff
@@ -0,0 +1,965 @@
...
+   fun:_ZN5boost3log11v2_mt_posix10attributes20attribute_value_implINS_10posix_time5ptimeEE8dispatchERNS1_15type_dispatcherE
+   fun:_ZNK5boost3log11v2_mt_posix15attribute_value8dispatchERNS1_15type_dispatcherE
+}
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Cond
+   fun:__lo0bits_D2A
+   fun:__d2b_D2A
+   fun:__dtoa
+   fun:__vfprintf
```
droberts195: There are many suppressions containing a stack with `__vfprintf` and then different functions called from it.  Replacing those with `...` would reduce suppressions.

So this one could be:

```
  ...
  fun:__v2printf
  fun:_vsnprintf
  fun:vsnprintf_l
  fun:snprintf_l
  fun:_ZNKSt3__17num_putIcNS_19ostreambuf_iteratorIcNS_11char_traitsIcEEEEE6do_putES4_RNS_8ios_baseEcd
  fun:_ZNSt3__113basic_ostreamIcNS_11char_traitsIcEEElsEd
  fun:_ZNSt3__1lsIdcNS_11char_traitsIcEEEERNS_13basic_ostreamIT0_T1_EES7_RKNS_7complexIT_EE
```

Then hopefully that means even more suppressions can be consolidated.

lib/core/unittest/.gitignore
```diff
@@ -1,2 +1,2 @@
 # Ignore temporary test files
-test.log
+test.log
```
droberts195: There are no changes to this file except removing the newline from the end of the last line.  Please change your editor settings not to do this.  *nix best practice is to have a newline at the end of the last line of the file.  There's an option in the CLion settings for this.

.gitignore
```diff
@@ -62,5 +62,4 @@ junit_results.xml
 cppunit_results.xml
 boost_test_results.xml
 **/.objs/*.*
-!**/.objs/.gitignore
-
+!**/.objs/.gitignore
```
droberts195: It's fine to remove the completely blank line, but please add back the newline at the end of the last non-blank line.  Same for `.dockerignore`.

edsavage:
> The gist of most of my comments is that it would be good to consolidate the suppressions. It will be very fragile having such specific suppressions - any time anyone changes code a suppression will probably need to be added or changed. There are definite patterns, so it must be possible to consolidate using shorter traces that only cover library functions and not ours, and using wildcards (`...`) for details inside the deepest library functions.
> 
> Also, most (all?) of the current suppressions are for macOS. Do we need any for Linux? The Jenkins job will run on Linux.

I'll re-run the `make memcheck` job on Linux with `--gen-suppressions=all` (Once the current run has finished... still going!)

lib/core/unittest/valgrind.supp
edsavage: nope, #852 was targeted specific Valgrind errors thrown up by `CSignal`. I suppressed this error because it appears to be out of our control. I'll review that decision.

droberts195: The unmangled function name in this one is `std::__1::complex<double>& std::__1::complex<double>::operator*=<double>(std::__1::complex<double> const&` (from `llvm-cxxfilt`).  That is different to all the ones related to `ostream`s, and it was `operator*=` for `complex` that was showing up in #845.

edsavage: Ah, ok. Thanks!

droberts195:
We seem to be having more success with `-fsanitize` than with `valgrind`, so I don't think we'll merge this now.

PR #900, opened 2019-12-12
[ML] Cleaning-up SHAP estimation

### LINKED ISSUE #886: [ML] Avoid explicit vector generation in `CBoostedTree::columnsHoldingShapValues()`
At the moment we explicitly generate the complete vector of indices. However, it would be sufficient to use first and last index only to feed the accumulator in `CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow` and `CDataFrameTrainBoostedTreeRegressionRunner::writeOneRow`. We should probably replace `std::vector<std::size_t>` by `boost::integer_range`.

valeriy42:
This PR adds a number of minor adjustments to the original change #857 for SHAP estimation
* ~~Enhancement note added~~
* SHAP columns are returned as an integer range instead of a vector to prevent instantiation
* `splitPath` is passed by reference in the recursive call to avoid creating copies
* "NoNormalization" is removed from the test names, since we don't do any normalization (may become relevant in the future, once we have multiclass classification)

Closes #886


lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
```diff
@@ -139,50 +139,50 @@ struct SFixture {
...
+    double s_Lambda{1.0};
+    double s_Gamma{10.0};
+    double s_SoftTreeDepthLimit{5.0};
+    double s_SoftTreeDepthTolerance{0.1};
+    double s_Eta{0.9};
+    std::size_t s_MaximumNumberTrees{1};
+    double s_FeatureBagFraction{1.0};
+
+    int s_Rows{200};
+    std::stringstream s_Output;
```
tveasey: 👍

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1561,10 +1561,8 @@ void CBoostedTreeImpl::computeShapValues(core::CDataFrame& frame, const TProgres
     }
 }
 
-CBoostedTreeImpl::TSizeVec CBoostedTreeImpl::columnsHoldingShapValues() const {
-    TSizeVec result(m_LastShapColumnIndex - m_FirstShapColumnIndex + 1);
-    std::iota(result.begin(), result.end(), m_FirstShapColumnIndex);
-    return result;
+CBoostedTreeImpl::TSizeRange CBoostedTreeImpl::columnsHoldingShapValues() const {
+    return boost::irange<std::size_t>(m_FirstShapColumnIndex, m_LastShapColumnIndex + 1);
```
tveasey: nit: seems cleaner to use typedef in case we change the type again
```suggestion
    return TSizeRange{m_FirstShapColumnIndex, m_LastShapColumnIndex + 1};
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -157,25 +158,34 @@ void CTreeShapFeatureImportance::shapRecursive(const TTree& tree,
...
         double incomingFractionOne{1.0};
-        int splitFeature = static_cast<int>(tree[nodeIndex].splitFeature());
-        auto it = std::find(splitPath.s_FeatureIndex.begin(),
-                            splitPath.s_FeatureIndex.end(), splitFeature);
-        if (it != splitPath.s_FeatureIndex.end()) {
+        int splitFeature{static_cast<int>(tree[nodeIndex].splitFeature())};
+        auto featureIndexEnd{(splitPath->s_FeatureIndex.begin() + splitPath->nextIndex())};
+        auto it = std::find(splitPath->s_FeatureIndex.begin(), featureIndexEnd, splitFeature);
+        if (it != featureIndexEnd) {
+            // since we path splitPath by reference, we need ot backup the object before unwinding it
```
tveasey: typo:
```suggestion
            // Since we pass splitPath by reference, we need to backup the object before unwinding it.
```

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -112,23 +113,23 @@ void CTreeShapFeatureImportance::shapRecursive(const TTree& tree,
...
+                                               std::unique_ptr<SPath>& splitPath,
                                                std::size_t nodeIndex,
                                                double parentFractionZero,
                                                double parentFractionOne,
                                                int parentFeatureIndex,
                                                std::size_t offset,
                                                core::CDataFrame::TRowItr& row) const {
-
-    CTreeShapFeatureImportance::extendPath(splitPath, parentFractionZero,
+    std::unique_ptr<SPath> backupPath;
```
tveasey: My inclination would be to make this optional type, but pointer is ok. In any case, don't think stops us making splitPath a reference.

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -157,25 +158,34 @@ void CTreeShapFeatureImportance::shapRecursive(const TTree& tree,
...
                             hotIndex, incomingFractionZero * hotFractionZero,
                             incomingFractionOne, splitFeature, offset, row);
+        this->unwindPath(*splitPath, nextIndex);
         this->shapRecursive(tree, samplesPerNode, encoder, encodedRow, splitPath,
                             coldIndex, incomingFractionZero * coldFractionZero,
                             0.0, splitFeature, offset, row);
+        this->unwindPath(*splitPath, nextIndex);
+        if (backupPath) {
+            // now we swap to restore the data before unwinding
+            backupPath.swap(splitPath);
```
tveasey: If splitPath is by reference then this would just be:
```
splitPath = std::move(backupPath);
```

valeriy42: Done

include/maths/CTreeShapFeatureImportance.h
```diff
@@ -124,7 +129,7 @@ class MATHS_EXPORT CTreeShapFeatureImportance {
                        const TDoubleVec& samplesPerNode,
                        const CDataFrameCategoryEncoder& encoder,
                        const CEncodedDataFrameRowRef& encodedRow,
-                       SPath splitPath,
+                       std::unique_ptr<SPath>& splitPath,
```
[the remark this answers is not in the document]
valeriy42: Done

lib/maths/CTreeShapFeatureImportance.cc
valeriy42: Done

valeriy42:
retest

PR #906, opened 2019-12-16
[ML] Instrument Data Frame Analysis Job to report state information to Java backend

### LINKED ISSUE #899: [ML] Instrument Data Frame Analysis Job to report state information to Java backend
In order to provide users more informative insights into the current job state, we should establish a communication from the running data frame analysis job to the Java back-end that goes beyond the simple progress indicator. This information should be provided on iteration-by-iteration basis.

As discussed in the team, we should start with providing information about 
+ complete set of analysis parameters (inclusive inferred once)
+ peak memory usage
+ intermediate quality of results

valeriy42:
In this PR I introduce a new class `CDataFrameAnalysisState`, which is responsible for centralized collection of statistics related to the data frame analytics jobs, i.e. progress, memory usage, etc. Hence, the related variables are also moved out of `CDataFrameAnalysisRunner` class. 

Right now we have two callback functions related to the job state: `progressRecorder` and `memoryUsageRecorder`. However, since we can expect that every new kind of statistics (quality of results, time, parameters) would require another update callback function in `maths` classes, I pass the reference to the `State` object. To avoid dependency from `maths` to `api`, I introduce a state  interface class owned by `maths`.

Related to  #976


lib/maths/COutliers.cc
```diff
@@ -377,8 +379,10 @@ CEnsemble<POINT>::computeOutlierScores(const std::vector<POINT>& points) const {
     TScorerVec scores(points.size());
     m_RecordMemoryUsage(core::CMemory::dynamicSize(scores));
 
+    std::size_t step{0ul};
     for (const auto& model : m_Models) {
         model.addOutlierScores(points, scores, m_RecordMemoryUsage);
+        state.nextStep(step++);
```
valeriy42: IMHO, this constitutes a single step of the outlier detection job. WDYT @tveasey ?

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <cstdint>
+
+namespace ml {
+namespace api {
+
+//! \brief Responsible for collecting data frame analysis job statistics, i.e. memory usage,
+//! progress, parameters, quality of results. The class also implements the functionality to
+//! write the state at different iteration into the results pipe.
+class API_EXPORT CDataFrameAnalysisState : public maths::CDataFrameAnalysisStateInterface {
```
tveasey: I think I'd call this something different. For me "state" would be more suggestive of it somehow being the state needed to perform the analysis. How about `CDataFrameAnalysisStatus` or `CDataFrameAnalysisRunStatistics`?

include/api/CDataFrameOutliersRunner.h
```diff
@@ -32,6 +32,11 @@ class API_EXPORT CDataFrameOutliersRunner final : public CDataFrameAnalysisRunne
     CDataFrameOutliersRunner(const CDataFrameAnalysisSpecification& spec,
                              const CDataFrameAnalysisParameters& parameters);
 
+    //! \return Reference to the analysis state.
+    const CDataFrameAnalysisState& state() const override;
```
tveasey: As above, I'd consider some different naming for these things, i.e. maybe `status`.

include/core/CProgramCounters.h
```diff
@@ -118,7 +118,7 @@ enum ECounterTypes {
     //! The estimated peak memory usage for training a predictive model
     E_DFTPMEstimatedPeakMemoryUsage = 24,
 
-    //! The peak memory usage of outlier detection in bytes
+    //! The peak memory usage for training a predictive model in bytes
```
tveasey: 👍

include/maths/CDataFrameAnalysisStateInterface.h
```diff
@@ -0,0 +1,54 @@
...
+    //! \note The caller should try to ensure that the sum of the values added
+    //! at the end of the analysis is equal to one.
+    //! \note This is converted to an integer - so we can atomically add - by
+    //! scaling by 1024. Therefore, this shouldn't be called with values less
+    //! than 0.001. In fact, it is unlikely that such high resolution is needed
+    //! and typically this would be called significantly less frequently.
+    virtual void updateProgress(double /*fractionalProgress*/){};
+    //! Trigger the next step of the job. This will initiate writing the job state
+    //! to the results pipe.
+    virtual void nextStep(uint32_t /*step*/){};
```
tveasey:
```suggestion
    virtual void nextStep(std::uint32_t /*step*/){};
```

include/maths/CDataFrameAnalysisStateInterface.h
```diff
@@ -0,0 +1,54 @@
...
+    //! Factory for the updateProgress() callback function object.
+    TProgressCallback progressCallback() {
+        return [&](double fractionalProgress) {
```
tveasey: nit:
```suggestion
        return [this](double fractionalProgress) {
```

include/maths/CDataFrameAnalysisStateInterface.h
```diff
@@ -0,0 +1,54 @@
...
+            this->updateProgress(fractionalProgress);
+        };
+    };
+    //! Factory for the updateMemoryUsage() callback function object.
+    TMemoryUsageCallback memoryUsageCallback() {
+        return [&](std::int64_t delta) { this->updateMemoryUsage(delta); };
```
tveasey: nit:
```suggestion
        return [this](std::int64_t delta) { this->updateMemoryUsage(delta); };
```

include/maths/CDataFrameAnalysisStateInterface.h
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#ifndef INCLUDED_ml_maths_CDataFrameAnalysisInterface_h
```
tveasey: As before, I'd prefer a different name than State, but also:
```suggestion
#ifndef INCLUDED_ml_maths_CDataFrameAnalysisStateInterface_h
```

include/maths/CDataFrameAnalysisStateInterface.h
```diff
@@ -0,0 +1,54 @@
...
+#include <maths/ImportExport.h>
+
+#include <cstdint>
+
+namespace ml {
+namespace maths {
+
+//! \brief Interface class for collecting data frame analysis job statistics owned
+//! by the maths module.
+class MATHS_EXPORT CDataFrameAnalysisStateInterface {
```
tveasey: I'm wondering what the rationale is for having `CDataFrameAnalysisStateInterface` and `CDataFrameAnalysisState`, i.e. I'm not sure the extra layer adds anything and `progressCallback` etc could be pushed down. If it is to stub out functionality for the tests. I'd declare a version in the test library which derives from `CDataFrameAnalysisState` which does this.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -126,19 +126,20 @@ void CDataFrameAnalyzer::run() {
...
-
     // We create the writer in run so that when it is finished destructors
     // get called and the wrapped stream does its job to close the array.
 
     // TODO Revisit this can probably be core::CRapidJsonLineWriter.
     auto outStream = m_ResultsStreamSupplier();
     core::CRapidJsonConcurrentLineWriter outputWriter{*outStream};
 
+    CDataFrameAnalysisRunner* analysis{m_AnalysisSpecification->run(*m_DataFrame)};
+    analysis->state().writer(&outputWriter);
```
tveasey: This has to come after the check it is non-null!

lib/maths/COutliers.cc
tveasey: I agree. This seems like a natural place to me as well.

include/api/CDataFrameAnalysisRunner.h
```diff
@@ -169,27 +167,11 @@ class API_EXPORT CDataFrameAnalysisRunner {
...
-
 private:
     const CDataFrameAnalysisSpecification& m_Spec;
 
     std::size_t m_NumberPartitions = 0;
     std::size_t m_MaximumNumberRowsPerPartition = 0;
-
-    std::atomic_bool m_Finished;
-    std::atomic_size_t m_FractionalProgress;
-    std::atomic<std::int64_t> m_Memory;
```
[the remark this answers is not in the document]
valeriy42: fixed

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <core/CProgramCounters.h>
+#include <core/CRapidJsonConcurrentLineWriter.h>
+
+#include <maths/CDataFrameAnalysisStateInterface.h>
+
+#include <api/CDataFrameAnalysisState.h>
+#include <api/ImportExport.h>
+
+#include <cstdint>
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+
+namespace ml {
+namespace api {
+
+//! \brief Responsible for collecting data frame analysis job statistics, i.e. memory usage,
+//! progress, parameters, quality of results. The class also implements the functionality to
+//! write the state at different iteration into the results pipe.
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+        double s_Progress;
+        std::int64_t s_Memory;
+    };
+
+private:
+    void writeState(uint32_t step, CDataFrameAnalysisState::SInternalState&& state);
+
+private:
+    SInternalState m_InternalState;
+    core::CConcurrentQueue<SInternalState, 10> m_StateQueue;
```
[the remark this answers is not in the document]
valeriy42: Comment added that elements can be dropped if the queue is full.

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+    std::atomic_bool m_Finished;
+    std::atomic_size_t m_FractionalProgress;
+    std::atomic<std::int64_t> m_Memory;
+    core::CRapidJsonConcurrentLineWriter* m_Writer;
+};
+
+class API_EXPORT CDataFrameOutliersState : public CDataFrameAnalysisState {
```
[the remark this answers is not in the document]
valeriy42: Done

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+protected:
+    counter_t::ECounterTypes memoryCounterType() override;
```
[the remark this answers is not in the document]
valeriy42: This doesn't make difference, since the class is made final. I would leave it `protected` for consistency.

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+};
+
+class API_EXPORT CDataFrameTrainBoostedTreeState : public CDataFrameAnalysisState {
```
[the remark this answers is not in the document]
valeriy42: Done

include/api/CDataFrameAnalysisState.h
valeriy42: Rename to `Instrumentation`

include/maths/CDataFrameAnalysisStateInterface.h
valeriy42: Fixed

include/maths/CDataFrameAnalysisStateInterface.h
valeriy42: Fixed

include/maths/CDataFrameAnalysisStateInterface.h
valeriy42: Done

include/maths/CDataFrameAnalysisStateInterface.h
valeriy42: As discussed, I want the `maths` to own their own reference to avoid dependency to `api`

lib/api/CDataFrameAnalyzer.cc
valeriy42: Good catch! Fixed now.

include/maths/CDataFrameAnalysisStateInterface.h
valeriy42: Done

include/api/CDataFrameAnalysisState.h
```diff
@@ -0,0 +1,104 @@
...
+
+    //! Reset variables related to the job progress.
+    void resetProgress();
+
+    //! Set pointer to the writer object.
+    void writer(core::CRapidJsonConcurrentLineWriter* writer);
+
+    //! Trigger the next step of the job. This will initiate writing the job state
+    //! to the results pipe.
+    void nextStep(uint32_t step) override;
```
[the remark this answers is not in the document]
valeriy42: Fixed

include/maths/CDataFrameAnalysisInstrumentationInterface.h
```diff
@@ -0,0 +1,54 @@
...
+//! by the maths module.
+class MATHS_EXPORT CDataFrameAnalysisInstrumentationInterface {
+public:
+    using TProgressCallback = std::function<void(double)>;
+    using TMemoryUsageCallback = std::function<void(std::int64_t)>;
+
+public:
+    virtual ~CDataFrameAnalysisInstrumentationInterface() = default;
+    //! Adds \p delta to the memory usage statistics.
+    virtual void updateMemoryUsage(std::int64_t /*delta*/) = 0;
```
tveasey: I wouldn't now comment these out since they'll then show up named in doxygen.

lib/api/CDataFrameAnalyzer.cc
```diff
@@ -126,19 +126,20 @@ void CDataFrameAnalyzer::run() {
...
     // get called and the wrapped stream does its job to close the array.
 
     // TODO Revisit this can probably be core::CRapidJsonLineWriter.
     auto outStream = m_ResultsStreamSupplier();
     core::CRapidJsonConcurrentLineWriter outputWriter{*outStream};
 
+    CDataFrameAnalysisRunner* analysis{m_AnalysisSpecification->run(*m_DataFrame)};
+    if (analysis == nullptr) {
+        return;
+    }
```
tveasey: Can't we move this back to where it was before, i.e. no need to create the stream if it isn't going to be used.

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -0,0 +1,119 @@
...
+    m_Writer = writer;
+}
+
+void CDataFrameAnalysisInstrumentation::nextStep(std::uint32_t step) {
+    m_StateQueue.tryPush(SInternalState(*this));
+    if (m_Writer != nullptr) {
+        while (m_StateQueue.size() > 0) {
+            this->writeState(step, m_StateQueue.pop());
+        }
+    }
```
tveasey: Ok, I'm now wondering about the use of the queue here at all. These operations all happen on the same thread. Currently, there is no protection that the writer isn't written to by multiple things at once. Although in practice `nextStep` is called by single threaded code in the actual runners we also write progress from a different thread in a loop in `CDataFrameAnalyzer`. However, this uses a concurrent line writer so each JSON object is written to a single line and it manages writes from different threads. So I think we can just remove the queue altogether. Also, [this](https://github.com/elastic/ml-cpp/blob/master/lib/api/CDataFrameAnalyzer.cc#L138) todo can then be removed since it does need to support concurrent writes.

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -0,0 +1,115 @@
...
+//! \brief Instrumentation class for collecting data frame analysis job statistics.
+//!
+//! DESCRIPTION:\n
+//! Responsible for collecting data frame analysis job statistics, i.e. memory usage,
+//! progress, parameters, quality of results. The class also implements the functionality to
+//! write the state at different iteration into the results pipe.
+//!
+//! The implementation is based on a queue that can contain max. 10 element. This is needed since  the
+//! analysis starts asynchronously before the result writer (consumer) is created. The time difference
+//! between the two events is negligible, nevertheless, if the queue is full, new elements will be dropped.
```
tveasey: This comment should probably change if we no longer have a queue.

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -0,0 +1,115 @@
...
+        void writeMemory(uint32_t step, core::CRapidJsonConcurrentLineWriter& writer);
+        double s_Progress;
+        std::int64_t s_Memory;
+    };
+
+private:
+    void writeState(uint32_t step, CDataFrameAnalysisInstrumentation::SInternalState&& state);
+
+private:
+    SInternalState m_InternalState;
```
tveasey: Unused?

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -0,0 +1,115 @@
...
+    std::int64_t memory() const;
+
+protected:
+    virtual counter_t::ECounterTypes memoryCounterType() = 0;
+
+private:
+    struct SInternalState {
+        SInternalState() = default;
+        SInternalState(const CDataFrameAnalysisInstrumentation& state);
+        void writeProgress(uint32_t step, core::CRapidJsonConcurrentLineWriter& writer);
```
tveasey:
```suggestion
        void writeProgress(std::uint32_t step, core::CRapidJsonConcurrentLineWriter& writer);
```

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -0,0 +1,115 @@
...
+        void writeMemory(uint32_t step, core::CRapidJsonConcurrentLineWriter& writer);
```
tveasey:
```suggestion
        void writeMemory(std::uint32_t step, core::CRapidJsonConcurrentLineWriter& writer);
```

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -0,0 +1,115 @@
...
+namespace api {
+
+//! \brief Instrumentation class for collecting data frame analysis job statistics.
+//!
+//! DESCRIPTION:\n
+//! Responsible for collecting data frame analysis job statistics, i.e. memory usage,
+//! progress, parameters, quality of results. The class also implements the functionality to
+//! write the state at different iteration into the results pipe.
+//!
+//! The implementation is based on a queue that can contain max. 10 element. This is needed since  the
```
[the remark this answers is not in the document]
valeriy42: Fixed

lib/api/CDataFrameAnalyzer.cc
valeriy42: Code changed as discussed.

include/api/CDataFrameAnalysisInstrumentation.h
valeriy42: Removed

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: I changed the code as discussed

include/api/CDataFrameAnalysisInstrumentation.h
valeriy42: Fixed

include/api/CDataFrameAnalysisInstrumentation.h
valeriy42: Become obsolete

include/api/CDataFrameAnalysisInstrumentation.h
valeriy42: Done

include/maths/CDataFrameAnalysisInstrumentationInterface.h
valeriy42: Thank you. fixed.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -466,9 +469,13 @@ CBoostedTreeImpl::~CBoostedTreeImpl() = default;
 CBoostedTreeImpl& CBoostedTreeImpl::operator=(CBoostedTreeImpl&&) = default;
 
 void CBoostedTreeImpl::train(core::CDataFrame& frame,
-                             const TProgressCallback& recordProgress,
-                             const TMemoryUsageCallback& recordMemoryUsage,
                              const TTrainingStateCallback& recordTrainStateCallback) {
+    if (m_Instrumentation == nullptr) {
+        HANDLE_FATAL(<< "Internal error: analysis state was not initialize. Please report this problem.");
```
tveasey: This is causing unit test failures. I'm not sure we should enforce this: it doesn't feel fatal that the process isn't monitored: it's up to the caller. I think I'd rather just set `recordProgress` and `recordMemoryUsage` to stubs if this is null.

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void CDataFrameAnalysisInstrumentation::writeMemory(std::uint32_t step) {
+    if (m_Writer != nullptr) {
+        m_Writer->StartObject();
+        m_Writer->Key(STEP_TAG);
+        m_Writer->Uint(step);
+        m_Writer->Key(PEAK_MEMORY_USAGE_TAG);
+        m_Writer->Uint64(m_Memory.load());
```
droberts195: `m_Memory` is signed, so should be written using `Int64()` rather than `Uint64`.

valeriy42: Good catch. Thank you!

lib/maths/CBoostedTreeImpl.cc
valeriy42: Thank you for catching this., I solved the issue.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -477,11 +477,14 @@ CBoostedTreeImpl& CBoostedTreeImpl::operator=(CBoostedTreeImpl&&) = default;
 
 void CBoostedTreeImpl::train(core::CDataFrame& frame,
                              const TTrainingStateCallback& recordTrainStateCallback) {
-    auto recordProgress{std::function<void(double)>()};
-    auto recordMemoryUsage{std::function<void(std::int64_t)>()};
+    std::function<void(double)> recordProgress;
+    std::function<void(std::int64_t)> recordMemoryUsage;
```
tveasey: nit: can we use typedefs for these: we already have them and if we change the type in the future it saves having to fix this code.

valeriy42: Fixed in ec7324ed

PR #913, opened 2019-12-19
[ML] Rename result shap prefix

valeriy42:
This PR renames the prefix for the SHAP value fields form `ml.shap.<field_name>` to `ml.feature_importance.<field_name>`.


lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
```diff
@@ -33,6 +34,8 @@ using TDoubleVec = std::vector<double>;
 using TStrVec = std::vector<std::string>;
 using TMeanVarAccumulator = ml::maths::CBasicStatistics::SSampleMeanVar<double>::TAccumulator;
 
+const std::string SHAP_PREFIX{"feature_importance."};
```
droberts195: Is this used?  It looks like you've used the identical constant from `maths::CDataFrameRegressionModel` instead.

valeriy42: Good catch. Indeed, it's not used. Fixed now.

PR #914, opened 2019-12-19
[ML] Add CHANGELOG entry for SHAP feature

valeriy42:
I add missing doc entry about the new SHAP computation feature.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -37,6 +37,11 @@
 
 == {es} version 7.6.0
 
+=== New Features
+
+* Add computation of tree SHAP feature importance values (SHapley Additive
+exPlanation) for classification and regression. (See {ml-pull}857[#857].)
+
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -37,6 +37,11 @@
 
 == {es} version 7.6.0
 
+=== New Features
+
+* Add computation of tree SHAP feature importance values (SHapley Additive
+exPlanation) for classification and regression. (See {ml-pull}857[#857].)
```
droberts195: I think it would be better to stress the user facing functionality, feature importance, and de-emphasise the implementation detail.

Maybe something more like this:

* Add feature importance values to classification and regression results (using tree SHapley Additive
 exPlanation, or SHAP). (See {ml-pull}857[#857].)

valeriy42: I adjusted the text as you suggested.

PR #921, opened 2020-01-08
[ML] Copy categorization data structures before background persist

### LINKED ISSUE #919: [ML] Background persistence of categorization state isn't thread safe
The problem is in https://github.com/elastic/ml-cpp/blob/9979b8af0971b454061d97ba91acc9a3234e4ebe/lib/model/CBaseTokenListDataCategorizer.cc#L407-L410

It uses reference wrappers instead of copying the data structures that are going to be persisted in the background.  The main thread could be changing the structures while they're being persisted.  It was broken in #508.

Unfortunately over the years the comment saying not to use reference wrappers got detached from the actual copying https://github.com/elastic/ml-cpp/blob/0bc5fc80225572ecd8a4711d463b4aacfccc769d/lib/api/CFieldDataCategorizer.cc#L425-L427

droberts195:
All data structures to be persisted need to be copied before a
background persist is started, otherwise inconsistent state could
be persisted as the foreground thread updates the data structures
(or worse, the program could crash).

Fixes #919

### DIFF
--- a/lib/api/CFieldDataCategorizer.cc
+++ b/lib/api/CFieldDataCategorizer.cc
@@ -333,7 +333,7 @@ bool CFieldDataCategorizer::persistState(core::CDataAdder& persister,
 
     LOG_DEBUG(<< "Persist categorizer state");
 
-    return this->doPersistState(m_DataCategorizer->makePersistFunc(),
+    return this->doPersistState(m_DataCategorizer->makeForegroundPersistFunc(),
                                 m_ExamplesCollector, persister);
 }
 
@@ -425,8 +425,8 @@ bool CFieldDataCategorizer::periodicPersistStateInBackground() {
                       // Do NOT add std::ref wrappers
                       // around these arguments - they
                       // MUST be copied for thread safety
-                      m_DataCategorizer->makePersistFunc(), m_ExamplesCollector,
-                      std::placeholders::_1)) == false) {
+                      m_DataCategorizer->makeBackgroundPersistFunc(),
+                      m_ExamplesCollector, std::placeholders::_1)) == false) {
         LOG_ERROR(<< "Failed to add categorizer background persistence function");
         return false;
     }
--- a/lib/api/unittest/CPersistenceManagerTest.cc
+++ b/lib/api/unittest/CPersistenceManagerTest.cc
@@ -294,6 +294,80 @@ BOOST_FIXTURE_TEST_CASE(testDetectorBackgroundPersistStaticsConsistency, CTestFi
     this->foregroundBackgroundCompAnomalyDetectionAfterStaticsUpdate("testfiles/new_mlfields_over.conf");
 }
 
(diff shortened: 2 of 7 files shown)


lib/api/CFieldDataCategorizer.cc
```diff
@@ -425,8 +425,8 @@ bool CFieldDataCategorizer::periodicPersistStateInBackground() {
                       // Do NOT add std::ref wrappers
                       // around these arguments - they
                       // MUST be copied for thread safety
-                      m_DataCategorizer->makePersistFunc(), m_ExamplesCollector,
-                      std::placeholders::_1)) == false) {
+                      m_DataCategorizer->makeBackgroundPersistFunc(),
```
droberts195: Testing note: if `makeBackgroundPersistFunc` is changed to `makeForegroundPersistFunc` in this line then the newly added unit test fails a very high proportion of the time.  This proves that the fix was necessary and the fact that the test always passes with `makeBackgroundPersistFunc` suggests the fix works.

droberts195:
retest

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -294,6 +294,80 @@ BOOST_FIXTURE_TEST_CASE(testDetectorBackgroundPersistStaticsConsistency, CTestFi
     this->foregroundBackgroundCompAnomalyDetectionAfterStaticsUpdate("testfiles/new_mlfields_over.conf");
 }
 
+BOOST_FIXTURE_TEST_CASE(testBackgroundPersistCategorizationConsistency, CTestFixture) {
+
+    static const std::string JOB_ID{"job"};
+    static const std::string FIRST_INPUT{
+        "{ \"message\": \"the quick brown fox jumped over the lazy dog\" }\n"};
```
edsavage: Come on Dave! Where's the originality? The inspiration! ;-) 

We could always pick something from http://clagnut.com/blog/2380/#Perfect_pangrams_in_English_.2826_letters.29 :-D

lib/model/CBaseTokenListDataCategorizer.cc
```diff
@@ -403,13 +403,22 @@ void CBaseTokenListDataCategorizer::acceptPersistInserter(const TTokenMIndex& to
...
         static_cast<void (*)(const TTokenMIndex&, const TTokenListCategoryVec&, core::CStatePersistInserter&)>(
             &CBaseTokenListDataCategorizer::acceptPersistInserter),
         std::cref(m_TokenIdLookup), std::cref(m_Categories), std::placeholders::_1);
 }
 
+CDataCategorizer::TPersistFunc CBaseTokenListDataCategorizer::makeBackgroundPersistFunc() const {
+    return std::bind(
+        static_cast<void (*)(const TTokenMIndex&, const TTokenListCategoryVec&, core::CStatePersistInserter&)>(
+            &CBaseTokenListDataCategorizer::acceptPersistInserter),
+        // Do NOT add std::ref wrappers around these arguments - they MUST be
```
edsavage: 👍

PR #927, opened 2020-01-09
[ML] Include categorizer size in model size stats

### LINKED ISSUE #724: Extend memory instrumentation to categorization
Following elastic/elasticsearch#47516 there is much less slack in the "process overhead" added to the expected size of every ML C++ process.  This will mean we could underestimate memory requirements quite badly in cases where features are used that are not instrumented at all, such at categorization.  (Previously the relatively small memory usage of categorization would have easily been absorbed into the very generous 100MB process overhead.)

To prevent problems where categorization is in use the memory instrumentation of the `autodetect` process should be extended to categorization.

This is actually not as trivial as it sounds because the memory instrumentation code is currently in the `model` library but categorization is in the `api` library.  So as part of this work the resource monitor class needs to be moved to the `api` library.  This will also allow the possibility of instrumenting the outermost parts of anomaly detection as well as categorization.

Instrumenting the memory usage of categorization is also a realistic prerequisite for allowing per-partition categorization, because a high cardinality partition field combined with per-partition categorization could lead to categorization using a lot of memory.  Therefore it makes sense to do the instrumentation first and only then consider adding per-partition categorization.

droberts195:
This change adds the size of the categorizer used in categorization
jobs to the model_bytes field reported in model_size_stats, making
it a more accurate reflection of actual memory usage.

As with anomaly detectors, only the model library classes are
instrumented, which means that the small amount of additional api
library memory usage is not included.

Closes #724


lib/api/unittest/CSingleStreamDataAdderTest.cc
```diff
@@ -73,42 +73,42 @@ void detectorPersistHelper(const std::string& configFileName,
 
     std::string origSnapshotId;
     std::size_t numOrigDocs(0);
-    ml::api::CAnomalyJob origJob(
-        JOB_ID, limits, fieldConfig, modelConfig, wrappedOutputStream,
-        std::bind(&reportPersistComplete, std::placeholders::_1,
-                  std::ref(origSnapshotId), std::ref(numOrigDocs)),
-        nullptr, -1, "time", timeFormat);
```
droberts195: It looks like there are masses of changes in this file but most are whitespace differences caused by me moving braces to adjust object lifetimes.  I did this after changing the resource monitor to reduce memory consumption on unregistering a component.  This test was relying on the old (unintuitive in my opinion) behaviour.

Best to review this file with "ignore whitespace differences" enabled!

edsavage: 👍

lib/api/CFieldDataCategorizer.cc
```diff
@@ -159,12 +162,19 @@ int CFieldDataCategorizer::computeCategory(const TStrStrUMap& dataRowFields) {
         return -1;
     }
 
-    bool exampleAdded = m_ExamplesCollector.add(categoryId, fieldValue);
-    bool searchTermsChanged = this->createReverseSearch(categoryId);
+    bool exampleAdded{m_DataCategorizer->addExample(categoryId, fieldValue)};
```
edsavage: Just realised that while we have a comment in `STYLEGUIDE.md` for (non-)boolean accessor functions:

```Boolean accessor functions MUST be named as isComplete NOT complete```

we don't seem to have anything similar for mutator functions. Should there be? If so what?

Comment (edsavage) [inline lib/api/CFieldDataCategorizer.cc]:
```suggestion
    bool addExample{m_DataCategorizer->addExample(categoryId, fieldValue)};
```

lib/api/CFieldDataCategorizer.cc
```diff
@@ -298,12 +308,7 @@ bool CFieldDataCategorizer::acceptRestoreTraverser(core::CStateRestoreTraverser&
     }
 
     if (traverser.name() == EXAMPLES_COLLECTOR_TAG) {
-        if (traverser.traverseSubLevel(std::bind(
-                &model::CCategoryExamplesCollector::acceptRestoreTraverser,
-                std::ref(m_ExamplesCollector), std::placeholders::_1)) == false ||
-            traverser.haveBadState()) {
-            LOG_ERROR(<< "Cannot restore categorizer, unexpected element: "
-                      << traverser.value());
+        if (m_DataCategorizer->restoreExamplesCollector(traverser) == false) {
```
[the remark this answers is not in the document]
edsavage: 👍  💯

lib/api/CFieldDataCategorizer.cc
```diff
@@ -410,6 +415,9 @@ void CFieldDataCategorizer::acceptPersistInserter(
 bool CFieldDataCategorizer::periodicPersistStateInBackground() {
     LOG_DEBUG(<< "Periodic persist categorizer state");
 
+    // Make sure model size stats are up to date
```
edsavage:
```suggestion
    // Make sure that the model size stats are up to date
```

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -162,8 +162,8 @@ class CTestFixture {
         BOOST_REQUIRE_EQUAL(size_t(1), ml::core::CStringUtils::replaceFirst(
                                            foregroundSnapshotId2, "snap", foregroundState2));
 
-        // Replace the zero byte separators so the expected/actual strings don't get
-        // truncated by CppUnit if the test fails
+        // Replace the zero byte separators to avoid '\0's in the output if the
```
edsavage: 👍

lib/model/CAnomalyDetector.cc
```diff
@@ -173,11 +171,6 @@ void CAnomalyDetector::zeroModelsToTime(core_t::TTime time) {
 
 bool CAnomalyDetector::acceptRestoreTraverser(const std::string& partitionFieldValue,
                                               core::CStateRestoreTraverser& traverser) {
-    // As the model pointer will change during restore, we unregister
-    // the detector from the resource monitor. We can register it
-    // again at the end of restore.
-    m_Limits.resourceMonitor().unRegisterComponent(*this);
```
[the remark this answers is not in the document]
edsavage: 👍

lib/model/CBaseTokenListDataCategorizer.cc
```diff
@@ -31,19 +31,18 @@ const std::string TOKEN_TAG("a");
 const std::string TOKEN_CATEGORY_COUNT_TAG("b");
 const std::string CATEGORY_TAG("c");
 
-const std::string TIME_ATTRIBUTE("time");
-
 const std::string EMPTY_STRING;
 }
 
 CBaseTokenListDataCategorizer::CBaseTokenListDataCategorizer(
+    CLimits& limits,
```
edsavage:
```suggestion
CTokenListDataCategorizerBase::CTokenListDataCategorizerBase(
    CLimits& limits,
```

lib/model/CCategoryExamplesCollector.cc
```diff
@@ -143,6 +144,18 @@ void CCategoryExamplesCollector::clear() {
     m_ExamplesByCategory.clear();
 }
 
+void CCategoryExamplesCollector::debugMemoryUsage(core::CMemoryUsage::TMemoryUsagePtr mem) const {
+    mem->setName("CCategoryExamplesCollector");
+    core::CMemoryDebug::dynamicSize("m_ExamplesByCategory", m_ExamplesByCategory, mem);
+}
+
+//! Get the memory used by this examples collector.
```
edsavage:
```suggestion
//! Return the memory used by this examples collector.
```

lib/model/CResourceMonitor.cc
```diff
@@ -45,20 +44,22 @@ void CResourceMonitor::memoryUsageReporter(const TMemoryUsageReporterFunc& repor
...
+    LOG_TRACE(<< "Registering component: " << &resource);
+    m_Resources.emplace(&resource, std::size_t(0));
 }
 
-void CResourceMonitor::unRegisterComponent(CAnomalyDetector& detector) {
-    auto itr = m_Detectors.find(&detector);
-    if (itr == m_Detectors.end()) {
-        LOG_ERROR(<< "Inconsistency - component has not been registered: " << &detector);
+void CResourceMonitor::unRegisterComponent(CMonitoredResource& resource) {
+    auto itr = m_Resources.find(&resource);
```
edsavage:
```suggestion
LOG_TRACE(<< "Unregistering component: " << &resource);
    auto itr = m_Resources.find(&resource);
```

lib/model/CResourceMonitor.cc
```diff
@@ -45,20 +44,22 @@ void CResourceMonitor::memoryUsageReporter(const TMemoryUsageReporterFunc& repor
...
+    if (itr == m_Resources.end()) {
+        LOG_ERROR(<< "Inconsistency - component has not been registered: " << &resource);
```
edsavage:
```suggestion
        LOG_ERROR(<< "Inconsistency. - Cannot unregister a component that has not yet been registered: " << &resource);
```

lib/model/CResourceMonitor.cc
```diff
@@ -45,20 +44,22 @@ void CResourceMonitor::memoryUsageReporter(const TMemoryUsageReporterFunc& repor
...
         return;
     }
 
-    LOG_TRACE(<< "Unregistering component: " << &detector);
-    m_Detectors.erase(itr);
+    LOG_TRACE(<< "Unregistering component: " << &resource);
```
edsavage:
```suggestion
```

lib/model/CResourceMonitor.cc
```diff
@@ -45,20 +45,22 @@ void CResourceMonitor::memoryUsageReporter(const TMemoryUsageReporterFunc& repor
...
+    m_CurrentMonitoredResourceMemory -= itr->second;
```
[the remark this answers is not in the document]
edsavage: I bet :-/

lib/model/CResourceMonitor.cc
```diff
@@ -205,16 +213,16 @@ std::size_t CResourceMonitor::allocationLimit() const {
...
 }
 
-void CResourceMonitor::memUsage(CAnomalyDetector* detector) {
-    auto itr = m_Detectors.find(detector);
-    if (itr == m_Detectors.end()) {
-        LOG_ERROR(<< "Inconsistency - component has not been registered: " << detector);
+void CResourceMonitor::memUsage(CMonitoredResource* resource) {
+    auto itr = m_Resources.find(resource);
+    if (itr == m_Resources.end()) {
+        LOG_ERROR(<< "Inconsistency - component has not been registered: " << resource);
```
edsavage:
```suggestion
        LOG_ERROR(<< "Inconsistency - component has not yet been registered: " << resource);
```

lib/model/CResourceMonitor.cc
```diff
@@ -205,16 +213,16 @@ std::size_t CResourceMonitor::allocationLimit() const {
...
         return;
     }
     std::size_t modelPreviousUsage = itr->second;
     std::size_t modelCurrentUsage = core::CMemory::dynamicSize(itr->first);
     itr->second = modelCurrentUsage;
-    m_CurrentAnomalyDetectorMemory += (modelCurrentUsage - modelPreviousUsage);
+    m_CurrentMonitoredResourceMemory += (modelCurrentUsage - modelPreviousUsage);
```
edsavage:
```suggestion
    m_MonitoredResourceCurrentMemory += (modelCurrentUsage - modelPreviousUsage);
```

lib/model/CResourceMonitor.cc
```diff
@@ -232,8 +240,8 @@ bool CResourceMonitor::needToSendReport() {
     }
 
     if (!m_AllocationFailures.empty()) {
-        core_t::TTime lastestAllocationError = (--m_AllocationFailures.end())->first;
-        if (lastestAllocationError > m_LastAllocationFailureReport) {
+        core_t::TTime latestAllocationError = (--m_AllocationFailures.end())->first;
```
edsavage:
```suggestion
        core_t::TTime latestAllocationError{--m_AllocationFailures.end())->first}
```

lib/model/CResourceMonitor.cc
```diff
@@ -346,7 +351,7 @@ std::size_t CResourceMonitor::lowLimit() const {
 }
 
 std::size_t CResourceMonitor::totalMemory() const {
-    return m_CurrentAnomalyDetectorMemory + m_ExtraMemory +
+    return m_CurrentMonitoredResourceMemory + m_ExtraMemory +
```
edsavage:
```suggestion
    return m_MonitoredResourceCurrentMemory + m_ExtraMemory +
```

lib/model/unittest/CResourceMonitorTest.cc
```diff
@@ -21,6 +23,22 @@ BOOST_AUTO_TEST_SUITE(CResourceMonitorTest)
 using namespace ml;
 using namespace model;
 
+namespace {
+using TTokenListDataCategorizerKeepsFields =
+    model::CTokenListDataCategorizer<true,  // Warping
```
edsavage: named constants would be better here I think...

lib/model/unittest/CResourceMonitorTest.cc
```diff
@@ -325,7 +366,7 @@ BOOST_FIXTURE_TEST_CASE(testMonitor, CTestFixture) {
         std::size_t origTotalMemory = mon.totalMemory();
 
         // Go up to 10 bytes, triggering a need
-        mon.m_CurrentAnomalyDetectorMemory = 10;
+        mon.m_CurrentMonitoredResourceMemory = 10;
```
edsavage:
```suggestion
        mon.m_MonitoredResourceCurrentMemory = 10;
```

and elsewhere below...

include/api/CFieldDataCategorizer.h
```diff
@@ -78,7 +76,7 @@ class API_EXPORT CFieldDataCategorizer : public CDataProcessor {
     //! Construct without persistence capability
     CFieldDataCategorizer(const std::string& jobId,
                           const CFieldConfig& config,
-                          const model::CLimits& limits,
+                          model::CLimits& limits,
```
edsavage: I thought there was something in `STYLEGUIDE.md` about parameter ordering based on type but I can't find it.. Maybe it was in the Prelert guide? I'll check later.

include/model/CResourceMonitor.h
```diff
@@ -53,8 +52,8 @@ class MODEL_EXPORT CResourceMonitor {
     };
 
 public:
-    using TDetectorPtrSizePr = std::pair<CAnomalyDetector*, std::size_t>;
-    using TDetectorPtrSizeUMap = boost::unordered_map<CAnomalyDetector*, std::size_t>;
+    using TResourcePtrSizePr = std::pair<CMonitoredResource*, std::size_t>;
```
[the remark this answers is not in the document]
droberts195: When I used CLion's refactoring functionality on this line it revealed that this particular typedef isn't used anywhere, so I completely removed it - it must be dead code left over from years ago.  But I renamed the one on the line below as per this suggestion.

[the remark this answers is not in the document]
droberts195: When I used CLion's refactoring functionality on this line it revealed that this particular typedef isn't used anywhere, so I completely removed it - it must be dead code left over from years ago.  But I renamed the one on the line below as per this suggestion.

lib/model/CResourceMonitor.cc
droberts195: Reformatting like this would make the message inconsistent with all the other similar ones - `grep Inconsistency lib/*/*.cc` and you'll see.  Also, these are basically soft asserts, i.e. things that should never happen but if they did we decided not to abort the program but let it carry on.  So hopefully nobody will ever see such a message logged and the exact wording is not particularly important.

include/api/CFieldDataCategorizer.h
droberts195: Yes:

> Function parameters MUST be ordered as [in] [in,out] [out]

Nobody else takes any notice of that any more so there are masses of violations of it which is why I didn't bother here, but since you noticed I'll change this one.

droberts195: Actually, come to think of it this one is flukily already in the mandated order, because `limits` is effectively now an `[in,out]` variable.

lib/model/unittest/CResourceMonitorTest.cc
droberts195: Actually since in this test it doesn't matter exactly how the categorizer is configured I switched it to use the default template arguments in this file.  Then this whole unnamed namespace goes.

include/model/CMonitoredResource.h
```diff
@@ -0,0 +1,68 @@
...
+//! DESCRIPTION:\n
+//! Resources that the CResourceMonitor class monitors must
+//! derive from this interface.
+//!
+//! IMPLEMENTATION DECISIONS:\n
+//! Defaults to the assumption that monitored resources do not support
+//! pruning.  Methods related to pruning have default implementations
+//! consistent with this assumption.
+//!
+class MODEL_EXPORT CMonitoredResource {
```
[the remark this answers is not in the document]
droberts195: Historically we've only used `Base` or `Interface` in abstract base class names where they were introduced a long time after an original single concrete implementation, that concrete implementation had already taken the obvious name, and it was considered confusing to change the name of the original class.  Appending `Base` or `Interface` to abstract base class names is not something that's commonly done in the codebase.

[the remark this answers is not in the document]
edsavage: Discussed internally and decided against this suggestion

lib/api/CFieldDataCategorizer.cc
droberts195: In the old coding standards Word document the section titled "Non-boolean accessor functions MUST be named as 'clientId' NOT 'getClientId'" had a detailed description of "Accessor functions MUST NOT be prefixed with get or set."

I think people are still adhering to setter methods not being prefixed with set.

But this is not a setter method, it's adding an item to a list.  So I think it's better that it does contain a verb to indicate that it's not a setter method.  If it was just `example` then people might assume it was a setter method setting a single example.

lib/model/CCategoryExamplesCollector.cc
droberts195: This doxygen comment shouldn't have been in the `.cc` file at all.  I have removed it.

droberts195:
I think I addressed most of the review points in 49c102c03df6cd77a95b47e82ab9be7eed8b55ac or replies where I didn't make a change.  The review had multiple conflicting suggestions for the same comment in a couple of places, so obviously I couldn't exactly match both suggestions, but I did change those comments.

droberts195:
retest

Issue #922, opened 2020-01-08
Valgrind memcheck errors and warnings


edsavage:
A number of errors and warnings have been picked up by running all the unit tests (apart from those for `seccomp`) under `valgrind --tool=memcheck` .

Log files attached (or soon will be!)

edsavage:
Valgrind log files..


[marple.zip](https://github.com/elastic/ml-cpp/files/4047000/marple.zip)

edsavage:
I think the better way forward with this is to create multiple GitHub entries, one for each `ml-cpp` library (apart from `seccomp`, `seccomp` doesn't play well with `memcheck`. I'll leave this ticket open and use it to track the sub tickets (when created).

droberts195:
Looks like all the real problems are in the `maths` library except for one memory leak in `CTokenListDataCategorizerTest::(anonymous namespace)::checkMemoryUsageInstrumentation` in the `model` library.  I will have a look at the memory leak since it's a test I recently wrote.

(It looks like all the `Leak_PossiblyLost` problems are due to one-off allocations in system libraries that we cannot control, so I am ignoring those.)

tveasey:
Can you post the commit this was created from Ed.

tveasey:
Note that the "Uninitialised value was created by a heap allocation" is probably being generated by not initialising values in `CBoostedTreeHyperparameters` in the default constructor. This is harmless given the usage, but we should zero initialise them.

edsavage:
Will do

droberts195:
All the issues that were found have now been fixed - there was one more in #1445 - so I think this can be closed now.

PR #940, opened 2020-01-16
[ML] Fix memory leak in tokenization tests

droberts195:
This was introduced by #859.  A raw pointer with a confusing
typedef name was assumed to be a smart pointer when it wasn't.

Also checked for the same problem in other places, and didn't
find any, but corrected a couple of local typedef names.  (The
wider problem that CMemoryUsage::TMemoryUsagePtr aliases a
raw pointer will have to wait for a bigger PR.)

Relates #922

### DIFF
--- a/lib/maths/unittest/CNaiveBayesTest.cc
+++ b/lib/maths/unittest/CNaiveBayesTest.cc
@@ -262,8 +262,8 @@ BOOST_AUTO_TEST_CASE(testPropagationByTime) {
 BOOST_AUTO_TEST_CASE(testMemoryUsage) {
     // Check invariants.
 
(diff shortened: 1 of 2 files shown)


droberts195:
> we should migrate `TMemoryUsagePtr` to be a smart pointer, I don't think it's good design being a bare pointer

I agree.

When it was first added it was a `shared_ptr`, but this was passed by value through deep call stacks, leading to lots of atomic operations on the reference count.  I suggested changing this to passing const references to the `shared_ptr` through these deep call stacks to avoid the repeated atomic increments and decrements, but instead it was changed to raw pointers.

I'll open an issue.

PR #948, opened 2020-01-17
[ML] Line search for best learn rate

tveasey:
This switches to performing an initial line search over the learn rate and reducing the range we search in the main optimisation loop. This does cost us some run time, but we generally get small wins in both the QoR and the stability of QoR as borne out by the sample results below.

Because we can't estimate the runtime upfront I've extended the progress monitor to allow us to reset the expected total number of steps. This will advance progress if it needs to; we have no mechanism for reducing progress if the expected number steps increase. In this case, we simply get stuck at the current progress. I use a margin in the initial estimates to reduce the chance of this so we get smoother progress monitoring.

Sample regression results:

| data set | master mean R^2 | master sd R^2 | patch mean R^2 | patch var R^2 |
| --- | --- | --- | --- | --- |
| boston | 0.85 | 0.011 | 0.92 | 0.006 |
| elevators | 0.88 | 0.028 | 0.89 | 0.002 |
| stock | 0.98 | 0.004 | 0.99 | 0.003 |
| houses | 0.68 | 0.026 | 0.66 | 0.016 |

Sample classification results:

| data set | master mean acc | master sd acc | patch mean acc | patch var acc |
| --- | --- | --- | --- | --- |
| bank | 0.87 | 0.003 | 0.88 | 0.006 |
| census | 0.84 | 0.004 | 0.84 | 0.003 |
| clicks | 0.66 | 0.016 | 0.67 | 0.009 |
| qsar | 0.84 | 0.007 | 0.85 | 0.008 |

### DIFF
--- a/lib/core/CLoopProgress.cc
+++ b/lib/core/CLoopProgress.cc
@@ -17,22 +17,22 @@
 namespace ml {
 namespace core {
 namespace {
-const std::string LOOP_SIZE_TAG{"loop_size_tag"};
+const std::string LOOP_RANGE_TAG{"loop_size_tag"};
 const std::string PROGRESS_STEPS_TAG{"progress_steps_tag"};
 const std::string CURRENT_STEP_PROGRESS_TAG{"current_step_progress_tag"};
 const std::string LOOP_POS_TAG{"loop_pos_tag"};
 }
 
 CLoopProgress::CLoopProgress()
-    : m_Size{std::numeric_limits<std::size_t>::max()}, m_Steps{1},
+    : m_Range{std::numeric_limits<std::size_t>::max()}, m_Steps{1},
       m_StepProgress{1.0}, m_RecordProgress{noop} {
 }
 
-CLoopProgress::CLoopProgress(std::size_t size,
+CLoopProgress::CLoopProgress(std::size_t range,
                              const TProgressCallback& recordProgress,
                              double scale,
                              std::size_t steps)
-    : m_Size{size}, m_Steps{std::min(size, steps)},
+    : m_Range{range}, m_Steps{std::min(range, steps)},
       m_StepProgress{scale / static_cast<double>(m_Steps)}, m_RecordProgress{recordProgress} {
 }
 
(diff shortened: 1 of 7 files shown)


lib/maths/CBoostedTreeFactory.cc
```diff
@@ -73,7 +73,7 @@ std::size_t computeMaximumNumberTrees(double eta) {
     return static_cast<std::size_t>(3.0 / eta / MIN_DOWNSAMPLE_FACTOR_SCALE + 0.5);
 }
 
-std::size_t scaleMaximumNumberTrees(std::size_t maximumNumberTrees) {
+std::size_t scaleMaximumNumberTreesForMainLoop(std::size_t maximumNumberTrees) {
```
valeriy42: nit: Does `ForMainLoop` adds important information to the function name?

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -708,6 +709,52 @@ void CBoostedTreeFactory::initializeUnsetDownsampleFactor(core::CDataFrame& fram
...
+            this->testLossLineSearch(frame, applyEta, logMinEta, logMaxEta,
+                                     -mainLoopSearchInterval / 2.0,
+                                     mainLoopSearchInterval / 2.0)
+                .value_or(fallback);
+        m_LogEtaSearchInterval = min(m_LogEtaSearchInterval, TVector{0.0});
+        LOG_TRACE(<< "log eta search interval = ["
+                  << m_LogEtaSearchInterval.toDelimited() << "]");
+        m_TreeImpl->m_Eta = std::exp(m_LogEtaSearchInterval(BEST_REGULARIZER_INDEX));
+        m_TreeImpl->m_EtaGrowthRatePerTree = 1.0 + m_TreeImpl->m_Eta / 2.0;
+        m_TreeImpl->m_MaximumNumberTrees = computeMaximumNumberTrees(m_TreeImpl->m_Eta);
```
valeriy42: Can you use `applyEta` here?

lib/core/unittest/CLoopProgressTest.cc
```diff
@@ -142,6 +143,49 @@ BOOST_AUTO_TEST_CASE(testScaled) {
     }
 }
 
+BOOST_AUTO_TEST_CASE(testIncrementRange) {
+
```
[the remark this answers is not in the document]
tveasey: Definitely. I'll add a comment

lib/maths/CBoostedTreeFactory.cc
tveasey: Well I wanted to somehow capture that this only applies in the main optimisation loop, because we're allowing the downsample factor and this is coupled with the choice of eta parameter. I think this is useful to convey somehow. Maybe I can come up with something better though.

tveasey: (I take your point about other variable naming though)

lib/maths/CBoostedTreeFactory.cc
tveasey: Good suggestion. See [fa55c62](https://github.com/elastic/ml-cpp/pull/948/commits/fa55c62f1d40558d8eebe87151f72c34888ecb1f).

lib/core/unittest/CLoopProgressTest.cc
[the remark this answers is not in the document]
tveasey: See [4392de2](https://github.com/elastic/ml-cpp/pull/948/commits/4392de234ad66223cfbc788a4cce968f48a07305).

lib/core/CLoopProgress.cc
```diff
@@ -42,24 +42,35 @@ void CLoopProgress::progressCallback(const TProgressCallback& recordProgress) {
...
         // Account for the fact that if i is large we may have jumped several steps.
-        std::size_t stride{m_Steps * std::min(m_Pos, m_Size) / m_Size - m_LastProgress};
-
+        std::size_t stride{m_Steps * std::min(m_Pos, m_Range) / m_Range - m_LastProgress};
         m_RecordProgress(static_cast<double>(stride) * m_StepProgress);
         m_LastProgress += stride;
     }
 }
 
+void CLoopProgress::incrementRange(int i) {
```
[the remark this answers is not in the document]
tveasey: See [0587fec](https://github.com/elastic/ml-cpp/pull/948/commits/0587fecf36ae0db1de2621c33e45f0343da7527e).

lib/maths/CBoostedTreeFactory.cc
tveasey: Ok I ended up removing this method altogether. We could probably still do with some rationalisation around the maximum number of trees and how it is used in the factory class, but I'll leave that for a separate change.

tveasey: See [7a13f76](https://github.com/elastic/ml-cpp/pull/948/commits/7a13f76508cd0add593b3017bb28b1427f0d2429)

tveasey:
Thanks for the review @valeriy42. I've now addressed all your comments can you take another look?

Issue #949, opened 2020-01-17
[ML] Poor category reverse search with some token reorderings


droberts195:
To reproduce, feed the following messages through categorization _in this order_:

```
[count_tweets] Killing job
Killing job [count_tweets]
[tweets_by_location] Killing job
Killing job [tweets_by_location]
```

This results in a category definition like this:

```
        {
          "job_id" : "foo",
          "category_id" : 1,
          "terms" : "",
          "regex" : ".*",
          "max_matching_length" : 35,
          "examples" : [
            "Killing job [count_tweets]",
            "Killing job [tweets_by_location]",
            "[count_tweets] Killing job",
            "[tweets_by_location] Killing job"
          ]
        }
```

Note no `terms` and `.*` for `regex`.

droberts195:
There are two problems here.

One is a result of a decision we made when porting our previous product to work with Elasticsearch, and is easy to fix without changing the state format.  We decided to include terms that appear more than once in the message multiple times in the `terms` field of the reverse search.  This is unnecessary for the reverse search to work, but improves human readability.  We didn't do that in the previous product and due to a mistake in the way that decision was implemented some terms are lost if the token order varies between messages.  I will fix this one immediately and backport it to 6.8.  It will make the `terms` correct in the reverse search, but the `regex` with still end up being `.*`.

The other problem is in the choice of ordered token sequence when messages have different token orders.  (This one affected the previous product too - it wasn't introduced in the Elasticsearch migration.)  Since the regex defines a token order we have to pick a set of tokens that have the same order within the messages.  For `count_tweets,Killing,job` and `Killing,job,count_tweets` we are picking `count_tweets` to be the ordered subsequence to be used for the regex.  Clearly this is suboptimal as it would be better to look for the longest common subsequence rather than the common subsequence that starts with the first token of the category definition as it stood prior to the addition of the new message.  But to do this requires a change to the state format, and is thus more complex and risky than the first fix.  To complete the sorry tale of what's happening at the moment, the third message is added because there's sufficient overlap on the highly weighted tokens `Killing` and `job`, but then we find there is no common subsequence at all between `count_tweets` and `tweets_by_location,Killing,job` so no ordered tokens are available to use for a regex.

PR #950, opened 2020-01-17
[ML] Include out-of-order as well as in-order terms in reverse search

droberts195:
When categories include messages with different token orders,
we were not including the tokens whose order varied between
the messages in the terms of the reverse search.  This change
adds these out-of-order tokens that are part of the category
definition back into the reverse search terms.

Fixes the easy part of #949

### DIFF
--- a/lib/api/unittest/CFieldDataCategorizerTest.cc
+++ b/lib/api/unittest/CFieldDataCategorizerTest.cc
@@ -222,6 +222,53 @@ BOOST_AUTO_TEST_CASE(testNodeReverseSearch) {
     BOOST_TEST_REQUIRE(output.find("\"message\"") == std::string::npos);
 }
 
+BOOST_AUTO_TEST_CASE(testJobKilledReverseSearch) {
+    model::CLimits limits;
+    CFieldConfig config;
+    BOOST_TEST_REQUIRE(config.initFromFile("testfiles/new_persist_categorization.conf"));
+
+    std::ostringstream outputStrm;
+    {
+        CNullOutput nullOutput;
+        core::CJsonOutputStreamWrapper wrappedOutputStream(outputStrm);
+        CJsonOutputWriter writer("job", wrappedOutputStream);
+
+        CFieldDataCategorizer categorizer("job", config, limits, nullOutput, writer);
+
+        CFieldDataCategorizer::TStrStrUMap dataRowFields;
+        dataRowFields["message"] = "[count_tweets] Killing job";
+
+        BOOST_TEST_REQUIRE(categorizer.handleRecord(dataRowFields));
+
+        dataRowFields["message"] = "Killing job [count_tweets]";
+
+        BOOST_TEST_REQUIRE(categorizer.handleRecord(dataRowFields));
+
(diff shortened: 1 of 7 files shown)


lib/api/unittest/CFieldDataCategorizerTest.cc
```diff
@@ -222,6 +222,53 @@ BOOST_AUTO_TEST_CASE(testNodeReverseSearch) {
...
+
+    const std::string& output = outputStrm.str();
+    LOG_DEBUG(<< "Output is: " << output);
+
+    // Assert that the reverse search contains all expected tokens when
+    // categorization is run end-to-end (obviously computation of categories and
+    // reverse search creation are tested more thoroughly in the unit tests for
+    // their respective classes, but this test helps to confirm that they work
+    // together)
+    BOOST_TEST_REQUIRE(output.find("\"terms\":\"Killing job\"") != std::string::npos);
```
droberts195: The fix of this PR makes `terms` here `Killing job` where before this PR it would have been empty.

lib/api/unittest/CFieldDataCategorizerTest.cc
```diff
@@ -222,6 +222,53 @@ BOOST_AUTO_TEST_CASE(testNodeReverseSearch) {
...
+    BOOST_TEST_REQUIRE(output.find("\"regex\":\".*\"") != std::string::npos);
```
droberts195: The regex is still poor, `.*`, after this PR.  A followup that will not be backported as far will fix this.

See https://github.com/elastic/ml-cpp/issues/949#issuecomment-575643446 for details.

droberts195:
> I wonder how this will effect existing ML Jobs? Do we adjust the terms when we expand categories?

Yes, we will if some aspect of the category changes as a result of assigning a new message to it.  But it will only make the category definitions stored in the results index better, so I don't think it's a problem.  The category definition stored in the opaque categorizer state that the C++ uses is not changed at all by this fix.  The problem being fixed here is that the way that C++ internal state was reported in a more friendly format was losing some information in this edge case.

The other part of the fix to #949 _will_ involve changing the C++ internal state, and that is why it is too difficult/dangerous to include in a 6.8 patch.

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -297,20 +297,20 @@ bool CTokenListDataCategorizerBase::createReverseSearch(int categoryId,
...
-    for (auto costedCommonUniqueTokenId : costedCommonUniqueTokenIds) {
-        m_ReverseSearchCreator->addCommonUniqueToken(
-            m_TokenIdLookup[costedCommonUniqueTokenId].str(), part1, part2);
-    }
-
-    bool first(true);
-    size_t end(category.outOfOrderCommonTokenIndex());
-    for (size_t index = 0; index < end; ++index) {
-        size_t tokenId(baseTokenIds[index].first);
+    bool first{true};
```
tveasey: I'm not sure about this name. Maybe `firstInOrderToken` would be clearer?

PR #951, opened 2020-01-17
[ML] Fix invalid read in tests

tveasey:
Avoid the need to reserve the vector before calling the `CBoostedTreeNode::split` function. This rejigs this method to avoid reading member variables after calling `resize` on the supplied vector. Shown up by #922.

### DIFF
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -350,7 +350,7 @@ CBoostedTreeNode::TSizeSizePr CBoostedTreeNode::split(std::size_t splitFeature,
     m_Gain = gain;
     m_Curvature = curvature;
     TSizeSizePr result{m_LeftChild.get(), m_RightChild.get()};
-    // Don't access members after calling resize.
+    // Don't access members after calling resize because this object is likely an element of the vector being resized.
     tree.resize(tree.size() + 2);
     return result;
 }
--- a/lib/maths/CBoostedTree.cc
+++ b/lib/maths/CBoostedTree.cc
@@ -349,8 +349,10 @@ CBoostedTreeNode::TSizeSizePr CBoostedTreeNode::split(std::size_t splitFeature,
     m_RightChild = static_cast<TNodeIndex>(tree.size() + 1);
     m_Gain = gain;
     m_Curvature = curvature;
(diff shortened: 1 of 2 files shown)


droberts195:
> We should reserve vector before calling the split `CBoostedTreeNode::split` function

I can see that it causes inefficiency, but why does it cause an invalid read if you don't do this?

tveasey:
The `split` method is called on a member of the vector. That function resizes this vector on [this](https://github.com/elastic/ml-cpp/pull/951/files#diff-4562017aa975b37520b5ea56d78e274cL352) line and if the vector is reallocated, the reads of the member variables returned are from a destructed object.

droberts195:
Yes, but surely the reordering that was done in `CBoostedTree.cc` around `// Don't access members after calling resize` should be enough to fix the invalid read?  The `reserve()` calls would have been an alternative (though less encapsulated, and thus less good) way to fix it.  So in effect this PR contains two fixes for the same problem.

My worry is that if the `reserve()` calls _are_ needed to stop Valgrind reporting a problem then a compiler optimisation is reordering the code around `// Don't access members after calling resize` such that it's being violated.  If this is the case it should really be refactored again such that the compiler is not allowed to do that reordering.

lib/maths/CBoostedTree.cc
```diff
@@ -349,8 +349,10 @@ CBoostedTreeNode::TSizeSizePr CBoostedTreeNode::split(std::size_t splitFeature,
     m_RightChild = static_cast<TNodeIndex>(tree.size() + 1);
     m_Gain = gain;
     m_Curvature = curvature;
+    TSizeSizePr result{m_LeftChild.get(), m_RightChild.get()};
+    // Don't access members after calling resize.
```
droberts195:
```suggestion
    // Don't access members after calling resize because this object is likely an element of the vector being resized.
```

tveasey:
Ok, I take your point. I guess there is no harm in removing the reserve so we'd catch it with valgrind. If the compiler did reorder that statement then it would be a compiler bug, since it clearly has a side effect (on the program correctness), so I think it's unlikely.

tveasey:
I've addressed all your comments now @droberts195. Can you take another look?

tveasey:
retest

PR #966, opened 2020-01-22
[ML] Modernize the categorization code

droberts195:
While working on part 2 of the fix for #949 I noticed that
the categorization code could do with a little modernization,
such as using range based for loops in preference to
iterators where possible.

Additionally there is some dead code left over from the
days of multiple products that used categorization, for
example in our Elasticsearch integration there is never a
need to create a reverse search for a NULL field, so the
method to do that can be removed.

This PR does that modernization and dead code removal
without changing any externally visible functionality.
Then the eventual fix for part 2 of #949 will not be
polluted with lots of irrelevant refactoring.

This PR also adds a couple of low level tests for the
CTokenListCategory class.  One of these highlights the
outstanding problem reported in #949.  That test will
be adjusted when part 2 of #949 is fixed.

### DIFF
--- a/lib/model/unittest/CTokenListCategoryTest.cc
+++ b/lib/model/unittest/CTokenListCategoryTest.cc
@@ -1,0 +1,101 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#include <core/CContainerPrinter.h>
+
+#include <model/CTokenListCategory.h>
+
+#include <boost/test/unit_test.hpp>
+
+#include <algorithm>
+#include <string>
+
+BOOST_AUTO_TEST_SUITE(CTokenListCategoryTest)
+
+BOOST_AUTO_TEST_CASE(testCommonTokensSameOrder) {
+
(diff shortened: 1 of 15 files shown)


edsavage:
👀

lib/model/unittest/CTokenListCategoryTest.cc
```diff
@@ -0,0 +1,101 @@
...
+        {0 /* she */, 1}, {1 /* sells */, 1}, {2 /* seashells */, 1},
+        {3 /* on */, 1},  {4 /* the */, 1},   {5 /* seashore */, 1}};
+    ml::model::CTokenListCategory::TSizeSizeMap baseUniqueTokenIds(
+        baseTokenIds.begin(), baseTokenIds.end());
+
+    ml::model::CTokenListCategory category(false, baseString, baseString.length(), baseTokenIds,
+                                           baseTokenIds.size(), baseUniqueTokenIds);
+
+    std::string newString{"sells seashells on the seashore, she does"};
+    ml::model::CTokenListCategory::TSizeSizePrVec newTokenIds{
```
edsavage: 😄

droberts195: Yoda inspired me

edsavage: :elasticheart:

PR #975, opened 2020-01-28
[ML] Update the inference schema model definition unit test

valeriy42:
Following changes of @benwtrent in the inference model definition, this PR updates the definition file in the corresponding unit test.

### DIFF
--- a/lib/api/CInferenceModelDefinition.cc
+++ b/lib/api/CInferenceModelDefinition.cc
@@ -29,7 +29,6 @@ const std::string JSON_FIELD_TAG{"field"};
 const std::string JSON_FREQUENCY_ENCODING_TAG{"frequency_encoding"};
 const std::string JSON_FREQUENCY_MAP_TAG{"frequency_map"};
 const std::string JSON_HOT_MAP_TAG{"hot_map"};
-const std::string JSON_INPUT_TAG{"input"};
 const std::string JSON_LEAF_VALUE_TAG{"leaf_value"};
 const std::string JSON_LEFT_CHILD_TAG{"left_child"};
 const std::string JSON_LOGISTIC_REGRESSION_TAG{"logistic_regression"};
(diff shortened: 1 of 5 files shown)


lib/api/CInferenceModelDefinition.cc
```diff
@@ -278,11 +278,7 @@ std::string CInferenceModelDefinition::jsonString() {
 }
 
 void CInferenceModelDefinition::addToDocument(rapidjson::Value& parentObject,
-                                              TRapidJsonWriter& writer) const { //input
-    rapidjson::Value inputObject = writer.makeObject();
-    m_Input.addToDocument(inputObject, writer);
-    writer.addMember(JSON_INPUT_TAG, inputObject, parentObject);
```
tveasey: `JSON_INPUT_TAG` is no longer needed I'd guess.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -368,12 +364,8 @@ void CTrainedModel::classificationWeights(const TDoubleVec& classificationWeight
     m_ClassificationWeights = classificationWeights;
 }
 
-void CInferenceModelDefinition::fieldNames(TStringVec&& fieldNames,
-                                           std::size_t dependentVariableColumnIndex) {
-    m_FieldNames = fieldNames;
-    fieldNames.erase(fieldNames.begin() +
-                     static_cast<std::ptrdiff_t>(dependentVariableColumnIndex));
```
tveasey: Question: does it not matter we include control field names here now, i.e. fields `.`?

valeriy42: It does not, since we only return the field names actually used by the models.

tveasey: This is a nice simplification in that case!

lib/api/CInferenceModelDefinition.cc
valeriy42: fixed

PR #979, opened 2020-01-29
[ML] Feature importance unit test for missing data

valeriy42:
This PR tests that feature importance behaves correctly in the face of missing data values.


lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
```diff
@@ -65,6 +65,36 @@ void setupLinearRegressionData(const TStrVec& fieldNames,
...
+        return core::CStringUtils::typeToStringPrecise(result, core::CIEEE754::E_DoublePrecision);
+    };
+
+    for (std::size_t i = 0; i < rows; ++i) {
+        TDoubleVec regressors;
+        rng.generateUniformSamples(0.0, 10.0, cols - 1, regressors);
+
+        fieldValues[0] = target(regressors);
+        for (std::size_t j = 0; j < regressors.size(); ++j) {
+            double regressor = regressors[j] > 9.0 ? core::CDataFrame::valueOfMissing()
```
tveasey: I think if you want to pass missing as a string you need to pass an empty string, i.e. something like:
```
if (regressors[j] <= 9.0) {
    fieldValues[j + 1] = core::CStringUtils::typeToStringPrecise(
                             regressor, core::CIEEE754::E_DoublePrecision);
}
```
(p.s. The approach you use now may work, but it's non-standard.)

lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
```diff
@@ -325,4 +379,38 @@ BOOST_FIXTURE_TEST_CASE(testRunBoostedTreeRegressionFeatureImportanceNoShap, SFi
     }
 }
 
+BOOST_FIXTURE_TEST_CASE(testMissingFeatures, SFixture) {
+    // Test that feature importance behaves correctly when some features are missing:
+    // We randomly 10% of all data in a simple additive model target=c1+c2+c3+c4. Hence,
```
tveasey:
```suggestion
    // We randomly omit 10% of all data in a simple additive model target=c1+c2+c3+c4. Hence,
```

lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
valeriy42: Fixed in 4eb14fdd

valeriy42:
Thank you @tveasey for the review. I adjusted the code as you suggested.

PR #981, opened 2020-01-29
[ML] Interface changes to support multi-parameter loss functions

tveasey:
This anticipates wanting to support loss functions with multiple parameters, i.e. multinomial logistic regression. I'm making these non-functional changes upfront since we're considering adding other loss functions in the 7.7 timeframe and they need to conform to the new interfaces.

### DIFF
--- a/include/api/CDataFrameAnalysisInstrumentation.h
+++ b/include/api/CDataFrameAnalysisInstrumentation.h
@@ -32,8 +32,6 @@ class API_EXPORT CDataFrameAnalysisInstrumentation
 public:
     CDataFrameAnalysisInstrumentation();
 
-    virtual ~CDataFrameAnalysisInstrumentation() = default;
-
     //! Adds \p delta to the memory usage statistics.
     void updateMemoryUsage(std::int64_t delta) override;
 
--- a/include/maths/CBoostedTree.h
+++ b/include/maths/CBoostedTree.h
@@ -8,6 +8,7 @@
 #define INCLUDED_ml_maths_CBoostedTree_h
 
 #include <core/CDataFrame.h>
+#include <core/CSmallVector.h>
 #include <core/CStatePersistInserter.h>
 #include <core/CStateRestoreTraverser.h>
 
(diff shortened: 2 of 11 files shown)


include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -32,8 +32,6 @@ class API_EXPORT CDataFrameAnalysisInstrumentation
 public:
     CDataFrameAnalysisInstrumentation();
 
-    virtual ~CDataFrameAnalysisInstrumentation() = default;
```
valeriy42: We have a pure virtual method in the `protected` area. That is why this virtual destructor is here.

include/maths/CBoostedTree.h
```diff
@@ -113,6 +117,9 @@ namespace boosted_tree {
 
 //! \brief Computes the leaf value which minimizes the loss function.
 class MATHS_EXPORT CArgMinLoss {
+public:
+    using TDouble1Vec = core::CSmallVector<double, 1>;
```
valeriy42: Calling it `TDouble1Vec` implies that the vector has only one element. How would you extend it to the multinomial case, introducing `TDoubleNVec`?

include/api/CDataFrameAnalysisInstrumentation.h
tveasey: You only need to define it a virtual once in the base class (unless you need it to do something non-default). So the necessary cleanup is already ensured by `CDataFrameAnalysisInstrumentation`. This was generating a warning for me, I think because it wasn't marked as override, but I think removing it is a better fix.

include/maths/CBoostedTree.h
tveasey: This is just the convention we use throughout the code base to distinguish from `std::vector`. We use the `N` in `TDouble<N>Vec` to denote the number of values the vector can store before it starts using the heap. For consistency I'd prefer to leave this as is.

include/api/CDataFrameAnalysisInstrumentation.h
tveasey: If you feel strongly I can add back in and mark as override, but this seems cleaner to me. WDYT?

tveasey:
> now generalize to the multi-nomial case

I'm working through all the details of that now. There are a few things we need to do in different places in the code. I'm going to open a meta issue to deal with them. This prepares the way for the loss function returning a gradient vector and Hessian matrix which is just one part. We then need to update the code which caches these values, the code which aggregates them for candidate splits and the code which chooses the next best split to make. I feel like I can make the first steps anticipating having the loss function and then right at the very end create the loss function. We can test each time that I don't change any behaviours and it stops us having one huge PR.

include/api/CDataFrameAnalysisInstrumentation.h
droberts195: There's a whole discussion about this in https://stackoverflow.com/questions/40999791/default-override-of-virtual-destructor

In a large codebase the `static_assert` suggested might make sense, but I am not convinced it's worth it in our small codebase where the people changing the code generally know the class hierarchy in the bit they're changing very well.

One more data point is that in the days before `default` and `override` existed we wouldn't have declared a constructor in this situation.

valeriy42: Alright :+1:

Issue #982, opened 2020-01-30
[ML] Support multinomial logistic regression in data frame analysis


tveasey:
This issue tracks introducing multinomial logistic regression. The main complication of this is that the loss function has multiple parameters and the forest needs to predict all of them. This breaks the assumption that we have a single prediction, gradient and curvature. The task therefore breaks down into some preliminary tasks independent of actually implementing the new loss function.

- [x] Extend `CLoss` and `CArgMinLoss` to support multivalued predictions, gradients and a Hessian matrix
- [x] Rework caching to store multivalued predictions, gradients and the Hessian. This affects how we index values in the data frame, the memory calculation, etc
- [x] Rework the code to compute aggregate derivatives for candidate splits
- [x] Rework the code to choose the best candidate split
- [x] Extend `CBoostedTreeNode` to support multi-valued leaves
- [x] Extend the inference model definition to support multi-valued leaves
- [x] Extend feature importance to support multi-valued leaves
- [x] Write the new loss function
- [x] Wire in to classification api

tveasey:
This is now complete and merged into 7.7.

PR #985, opened 2020-01-31
[ML] Rework caching and reading predictions and derivatives in boosted tree training to anticipate multiple values

tveasey:
This is a refractor in anticipation loss functions with multiple parameters. It is non-functional, but I noticed running the unit tests creating the temporaries (even though they're not allocating any memory) actually costs some run time. It will likely make sense to revisit the type choice when I work on updates to the aggregation code. A step towards #982.

### DIFF
--- a/include/maths/CBoostedTreeImpl.h
+++ b/include/maths/CBoostedTreeImpl.h
@@ -13,6 +13,7 @@
 #include <core/CLogger.h>
 #include <core/CMemory.h>
 #include <core/CPackedBitVector.h>
+#include <core/CSmallVector.h>
 #include <core/CStatePersistInserter.h>
 #include <core/CStateRestoreTraverser.h>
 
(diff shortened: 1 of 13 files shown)


include/maths/CBoostedTreeImpl.h
```diff
@@ -101,14 +102,13 @@ class MATHS_EXPORT CBoostedTreeImpl final {
...
         // We store as follows:
-        //   1. The predicted value for the dependent variable
+        //   1. The predicted values for the dependent variables
         //   2. The gradient of the loss function
-        //   3. The curvature of the loss function
+        //   3. The upper triangle of the hessian of the loss function
         //   4. The example's weight
-        // In the last four rows of the data frame.
-        return 4;
+        return numberLossParameters * (numberLossParameters + 5) / 2 + 1;
```
valeriy42: nit: where does 5 come from? In Line 599 it's just `numberLossParameters * (numberLossParameters + 1) / 2`.

tveasey: This is `prediction + gradient + curvature = numberLossParameters + numberLossParameters + numberLossParameters * (numberLossParameters + 1) / 2 = numberLossParameters * (numberLossParameters + 1 + 4) / 2`

PR #989, opened 2020-02-04
[ML] Add new categorization stats to model_size_stats

droberts195:
This change adds support for the following new model_size_stats
fields:

- categorized_doc_count
- total_category_count
- frequent_category_count
- rare_category_count
- dead_category_count
- categorization_status

Relates elastic/elasticsearch#50749

### DIFF
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -49,6 +49,12 @@ public:
         core_t::TTime s_BucketStartTime;
         std::size_t s_BytesExceeded;
         std::size_t s_BytesMemoryLimit;
+        std::size_t s_CategorizedMessages;
+        std::size_t s_TotalCategories;
+        std::size_t s_FrequentCategories;
+        std::size_t s_RareCategories;
+        std::size_t s_DeadCategories;
+        model_t::ECategorizationStatus s_CategorizationStatus;
     };
 
 public:
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -38,29 +38,30 @@ class CMonitoredResource;
 //! Assess memory used by models and decide on further memory allocations.
 class MODEL_EXPORT CResourceMonitor {
 public:
-    struct MODEL_EXPORT SResults {
-        std::size_t s_Usage;
-        std::size_t s_AdjustedUsage;
-        std::size_t s_ByFields;
-        std::size_t s_PartitionFields;
-        std::size_t s_OverFields;
-        std::size_t s_AllocationFailures;
-        model_t::EMemoryStatus s_MemoryStatus;
-        core_t::TTime s_BucketStartTime;
-        std::size_t s_BytesExceeded;
-        std::size_t s_BytesMemoryLimit;
-        std::size_t s_CategorizedMessages;
-        std::size_t s_TotalCategories;
-        std::size_t s_FrequentCategories;
-        std::size_t s_RareCategories;
-        std::size_t s_DeadCategories;
-        model_t::ECategorizationStatus s_CategorizationStatus;
+    struct MODEL_EXPORT SModelSizeStats {
+        std::size_t s_Usage = 0;
+        std::size_t s_AdjustedUsage = 0;
+        std::size_t s_ByFields = 0;
+        std::size_t s_PartitionFields = 0;
+        std::size_t s_OverFields = 0;
+        std::size_t s_AllocationFailures = 0;
+        model_t::EMemoryStatus s_MemoryStatus = model_t::E_MemoryStatusOk;
+        core_t::TTime s_BucketStartTime = 0;
+        std::size_t s_BytesExceeded = 0;
+        std::size_t s_BytesMemoryLimit = 0;
+        std::size_t s_CategorizedMessages = 0;
+        std::size_t s_TotalCategories = 0;
+        std::size_t s_FrequentCategories = 0;
+        std::size_t s_RareCategories = 0;
+        std::size_t s_DeadCategories = 0;
+        model_t::ECategorizationStatus s_CategorizationStatus = model_t::E_CategorizationStatusOk;
     };
 
 public:
     using TMonitoredResourcePtrSizeUMap =
         boost::unordered_map<CMonitoredResource*, std::size_t>;
-    using TMemoryUsageReporterFunc = std::function<void(const CResourceMonitor::SResults&)>;
+    using TMemoryUsageReporterFunc =
+        std::function<void(const CResourceMonitor::SModelSizeStats&)>;
     using TTimeSizeMap = std::map<core_t::TTime, std::size_t>;
 
     //! The minimum time between prunes
@@ -115,7 +116,7 @@ public:
     void sendMemoryUsageReport(core_t::TTime bucketStartTime);
 
     //! Create a memory usage report
-    SResults createMemoryUsageReport(core_t::TTime bucketStartTime);
+    SModelSizeStats createMemoryUsageReport(core_t::TTime bucketStartTime);
 
     //! We are being told that a class has failed to allocate memory
     //! based on the resource limits, and we will report this to the
--- a/include/model/CTokenListCategory.h
+++ b/include/model/CTokenListCategory.h
@@ -108,15 +108,14 @@ public:
     bool isMissingCommonTokenWeightZero(const PAIR_CONTAINER& uniqueTokenIds) const {
 
         auto testIter = uniqueTokenIds.begin();
-        for (auto commonIter = m_CommonUniqueTokenIds.begin();
-             commonIter != m_CommonUniqueTokenIds.end(); ++commonIter) {
+        for (const auto& commonItem : m_CommonUniqueTokenIds) {
             testIter = std::find_if(testIter, uniqueTokenIds.end(),
-                                    [&commonIter](const auto& testItem) {
-                                        return testItem.first >= commonIter->first;
+                                    [&commonItem](const auto& testItem) {
+                                        return testItem.first >= commonItem.first;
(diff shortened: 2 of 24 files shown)


droberts195:
This PR will fail CI until elastic/elasticsearch#51879.  The Java side change must be merged first, but I wanted to open this now so they can be reviewed together.

include/model/CResourceMonitor.h
```diff
@@ -49,6 +49,12 @@ class MODEL_EXPORT CResourceMonitor {
         core_t::TTime s_BucketStartTime;
         std::size_t s_BytesExceeded;
         std::size_t s_BytesMemoryLimit;
+        std::size_t s_CategorizedMessages;
+        std::size_t s_TotalCategories;
+        std::size_t s_FrequentCategories;
+        std::size_t s_RareCategories;
+        std::size_t s_DeadCategories;
+        model_t::ECategorizationStatus s_CategorizationStatus;
```
tveasey: I know you're only extending, but should this provide defaults? It seems like this object is a recipe for uninitialised variables as it stands.

lib/api/unittest/CModelSnapshotJsonWriterTest.cc
```diff
@@ -110,14 +115,27 @@ BOOST_AUTO_TEST_CASE(testWrite) {
...
-    BOOST_TEST_REQUIRE(modelSizeStats.HasMember("timestamp"));
-    BOOST_REQUIRE_EQUAL(int64_t(1521046309000), modelSizeStats["timestamp"].GetInt64());
-    BOOST_TEST_REQUIRE(modelSizeStats.HasMember("log_time"));
     BOOST_TEST_REQUIRE(modelSizeStats.HasMember("model_bytes_exceeded"));
     BOOST_REQUIRE_EQUAL(int64_t(0), modelSizeStats["model_bytes_exceeded"].GetInt64());
     BOOST_TEST_REQUIRE(modelSizeStats.HasMember("model_bytes_memory_limit"));
     BOOST_REQUIRE_EQUAL(int64_t(50000),
                         modelSizeStats["model_bytes_memory_limit"].GetInt64());
+    BOOST_TEST_REQUIRE(modelSizeStats.HasMember("categorized_doc_count"));
+    BOOST_REQUIRE_EQUAL(int64_t(1000), modelSizeStats["categorized_doc_count"].GetInt64());
```
tveasey: nit: maybe include `cstdint` and `std::int64_t` based on our guidelines.

lib/model/CResourceMonitor.cc
```diff
@@ -272,10 +272,30 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
     res.s_AllocationFailures = 0;
     res.s_MemoryStatus = m_MemoryStatus;
     res.s_BucketStartTime = bucketStartTime;
+    res.s_CategorizedMessages = 0;
+    res.s_TotalCategories = 0;
+    res.s_FrequentCategories = 0;
+    res.s_RareCategories = 0;
+    res.s_DeadCategories = 0;
+    res.s_CategorizationStatus = model_t::E_CategorizationStatusOk;
```
tveasey: I'd make all these defaults in the class definition.

lib/model/CResourceMonitor.cc
```diff
@@ -272,10 +272,30 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
...
     for (const auto& resource : m_Resources) {
         resource.first->updateMemoryResults(res);
     }
     res.s_AllocationFailures += m_AllocationFailures.size();
+    // Categorization status is poor if:
+    // - At least 100 messages have been categorized
+    // and one of the following holds:
+    // - There is only 1 category
+    // - More than 90% of categories have 1 message
```
tveasey:
```suggestion
    // - More than 90% of categories are rare
```

lib/model/CResourceMonitor.cc
```diff
@@ -272,10 +272,30 @@ CResourceMonitor::SResults CResourceMonitor::createMemoryUsageReport(core_t::TTi
...
+    // - The number of categories is greater than 50% of the number of categorized messages
+    // - There are no frequent match categories
+    // - More than 50% of categories are dead
+    if (res.s_CategorizedMessages > 100 &&
+        (res.s_TotalCategories == 1 || 10 * res.s_RareCategories > 9 * res.s_TotalCategories ||
+         2 * res.s_TotalCategories > res.s_CategorizedMessages ||
+         res.s_FrequentCategories == 0 || 2 * res.s_DeadCategories > res.s_TotalCategories)) {
+        res.s_CategorizationStatus = model_t::E_CategorizationStatusPoor;
```
tveasey: This feels like an odd place for this logic: could it perhaps be better encapsulated. Maybe a static member on some class related to categorisation.

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -90,10 +91,11 @@ int CTokenListDataCategorizerBase::computeCategory(bool isDryRun,
         // further checks.  The first condition here ensures that we never say
         // a string with tokens matches the reverse search of a string with no
         // tokens (which the other criteria alone might say matched).
-        bool matchesSearch((baseWeight == 0) == (workWeight == 0) &&
-                           compCategory.maxMatchingStringLen() >= rawStringLen &&
-                           compCategory.isMissingCommonTokenWeightZero(m_WorkTokenUniqueIds) &&
-                           compCategory.containsCommonTokensInOrder(m_WorkTokenIds));
+        bool matchesSearch(
```
tveasey: nit: braces initialization

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -539,6 +528,47 @@ std::size_t CTokenListDataCategorizerBase::memoryUsage() const {
     return mem;
 }
 
+void CTokenListDataCategorizerBase::updateMemoryResults(CResourceMonitor::SResults& results) const {
```
tveasey: I'm not sure about calling this `updateMemoryResults`. This includes other information now. Maybe `updateTelemetry`? I guess there is a slight issue because we've subverted `CResourceMonitor` to do more than monitor resource usage. Perhaps we should think about renaming this in a separate PR.

lib/api/unittest/CModelSnapshotJsonWriterTest.cc
droberts195: This file was also dodgy before my changes because many expected results should have been unsigned rather than signed.  I copied that bug into my new lines, but I'll correct that throughout the file at the same time as adding `std::`.

lib/model/CTokenListDataCategorizerBase.cc
droberts195: CLion makes these renamings pretty trivial so how about `updateModelSizeStats`?  Then naming will be consistent through C++, Java and JavaScript, which might make it easier to trace how these numbers flow through the code.

Many of the new stats _are_ related to resource use - certainly `total_category_count`, `rare_category_count` and `dead_category_count` are indicators of how much memory categorization is using, and whether that usage is suboptimal.  To put the other stats somewhere else would just create complexity.

lib/model/CResourceMonitor.cc
droberts195: Come to think of it I can just put this in `CTokenListDataCategorizerBase::updateModelSizeStats()`.  I was thinking we'd only want to do this calculation once if we start doing per-partition categorization, but since the stats it's based on will accumulate over the partitions the last partition we calculated the status for would use the overall numbers and that's the value that would get reported.  The cost is that we evaluate some 5 simple comparisons per-partition instead of once but that's not going to take long and I agree the encapsulation is much better moving this formula into categorization code.

(Plus of course while we're _not_ doing per-partition categorization the cost concerns are all theoretical anyway as we only have one categorizer object.)

include/model/CTokenListCategory.h
```diff
@@ -99,11 +99,35 @@ class MODEL_EXPORT CTokenListCategory {
...
+    template<typename PAIR_CONTAINER>
+    bool isMissingCommonTokenWeightZero(const PAIR_CONTAINER& uniqueTokenIds) const {
+
+        auto commonIter = m_CommonUniqueTokenIds.begin();
+        auto testIter = uniqueTokenIds.begin();
+        while (commonIter != m_CommonUniqueTokenIds.end() &&
+               testIter != uniqueTokenIds.end()) {
+            if (commonIter->first < testIter->first) {
+                return false;
+            }
```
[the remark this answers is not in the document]
droberts195: I more-or-less did this. The second part of the condition has to be `testIter->first != commonIter->first || testIter->second != commonIter->second` instead of simply `*testIter != *commonIter` because in some cases one is a `std::pair<const std::size_t, std::size_t>` and the other is a `std::pair<std::size_t, std::size_t>`.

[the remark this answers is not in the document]
tveasey: Ok, great. It's a shame comparison doesn't work with const. That feels like an error in the std::pair implementation.

lib/model/CTokenListDataCategorizerBase.cc
tveasey: Ok cool. I definitely think that's better.
> To put the other stats somewhere else would just create complexity.

Btw, I completely agree we don't want these somewhere else. It's more I wonder if we should think about renaming `CResourceMonitor` at some point. I'm happy for that not to happen in this PR though.

droberts195: ++ to not renaming `CResourceMonitor` in this PR.  That would touch a very large number of files unrelated to categorization.

PR #996, opened 2020-02-12
[ML] Account for data frame memory in the peak memory usage estimate for classification and regression

tveasey:
We weren't accounting for the data frame's memory usage when reporting peak memory usage for training. Also, we were missing some state when estimating peak memory usage during the line searches to initialise hyperparameters.

I've also taken the opportunity to more fully migrate instrumentation of training to use the new instrumentation object and provide better automatic stubbing.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -31,6 +31,7 @@ using namespace boosted_tree_detail;
 using TStrVec = CBoostedTreeImpl::TStrVec;
 using TRowItr = core::CDataFrame::TRowItr;
 using TMeanVarAccumulator = CBoostedTreeImpl::TMeanVarAccumulator;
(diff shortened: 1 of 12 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -133,7 +132,7 @@ CBoostedTreeImpl::CBoostedTreeImpl(std::size_t numberThreads,
       m_BestHyperparameters{
           m_Regularization,       m_DownsampleFactor,   m_Eta,
           m_EtaGrowthRatePerTree, m_MaximumNumberTrees, m_FeatureBagFraction},
-      m_Instrumentation{instrumentation} {
+      m_Instrumentation{instrumentation != nullptr ? instrumentation : &INSTRUMENTATION_STUB} {
```
valeriy42: good idea.

PR #1000, opened 2020-02-13
[6.8] Convert PowerShell error objects from stderr back to strings

droberts195:
When running remotely PowerShell converts output to stderr
as error objects.  These then cause the script to fail.

This change adds a step to convert any such error objects
originating from the main Gradle invocation back to strings.

This is required to get the new CI setup to work on Windows.

Backport of #997


droberts195:
retest

PR #1003, opened 2020-02-17
[ML] Improvements to upfront memory estimation for data frame analyses

tveasey:
This change fixes a double counting bug for the memory used by the extra columns for classification and regression model training. It also means they only count towards the data frame memory usage: previously they were wrongly being treated as features for memory estimation purposes. 

Incidentally, it also fixes the memory reported by the counter `E_DFOEstimatedPeakMemoryUsage`, which was missing the extra columns' memory usage.

Finally, it tidies up instrumentation of outlier detection, to more fully use the new instrumentation class, and corrects the memory estimates in `testRunOutlierDetectionPartitioned`.


tveasey:
retest

PR #1011, opened 2020-02-18
[ML] Extend boosted tree node to support multiple values

tveasey:
This leaves open the tasks of updating the inference model definition and tree SHAP for which I've left TODOs.

This is the next step towards #982.

### DIFF
--- a/lib/maths/CTreeShapFeatureImportance.cc
+++ b/lib/maths/CTreeShapFeatureImportance.cc
@@ -43,10 +43,10 @@ CTreeShapFeatureImportance::CTreeShapFeatureImportance(TTreeVec trees, std::size
     : m_Trees{std::move(trees)}, m_NumberThreads{threads} {
 }
 
-size_t CTreeShapFeatureImportance::updateNodeValues(TTree& tree,
-                                                    std::size_t nodeIndex,
-                                                    std::size_t depth) {
-    auto& node{tree[nodeIndex]};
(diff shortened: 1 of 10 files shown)


lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -43,10 +43,10 @@ CTreeShapFeatureImportance::CTreeShapFeatureImportance(TTreeVec trees, std::size
...
 }
 
-size_t CTreeShapFeatureImportance::updateNodeValues(TTree& tree,
-                                                    std::size_t nodeIndex,
-                                                    std::size_t depth) {
-    auto& node{tree[nodeIndex]};
+std::size_t CTreeShapFeatureImportance::updateNodeValues(TTree& tree,
+                                                         std::size_t nodeIndex,
+                                                         std::size_t depth) {
+    auto& node = tree[nodeIndex];
```
valeriy42: Is there a reason to change it to an assigment?

tveasey: This was just reflex :) I can change it back. Basically, I always use assignment with `auto` for the reason that, at least until C++17, using with braces often wraps types in an unnecessary `initializer_list`, i.e. the type of `i` in `auto i{0}` is `std::initializer_list<int>`. This is to some extent a personal preference, but in my experience using assignment deduces the type I'd want in all cases I've preferred to use `auto` and I prefer the code to be as uniform as possible. My internal rule is always use `{}-initialization` except with `auto` for which always use assignment.

PR #1013, opened 2020-02-18
[ML] Upgrade gcc to version 7.5

droberts195:
This change is primarily to upgrade the gcc we use from version 7.3
to version 7.5.

A secondary change is to build a recent version of the binutils
package from source for use in our builds instead of relying on
whatever happened to come with the build machine.

Both changes are driven by the desire to build Tensorflow in the
future.  gcc 7.3 suffers from an internal compiler error when
trying to compile Tensorflow.  The bug is fixed in gcc 7.5.  Also,
the assembler that comes with binutils 2.20 is too old to assemble
the compiler output when building Tensorflow, and version 2.20 is
what CentOS 6 ships with.  Since our build platform for released
product is CentOS 6 we also need to install a newer version of
binutils.


3rd_party/licenses/gcc-runtime-INFO.csv
```diff
@@ -1,2 +1,2 @@
 name,version,revision,url,license,copyright
-GCC Runtime Library,6.2.0,6ac74a62ba7258299cf85fbef9bf45333ddf10c0,https://gcc.gnu.org,GPL-3.0 WITH GCC-exception-3.1,"Copyright (C) 2016 Free Software Foundation, Inc."
+GCC Runtime Library,7.5.0,b2d961e7342b5ba4e57adfa81cb189b738d10901,https://gcc.gnu.org,GPL-3.0 WITH GCC-exception-3.1,"Copyright (C) 2019 Free Software Foundation, Inc."
```
droberts195: If you are eagle-eyed you will notice that the old version was wrong here.  I opened #1014 to fix that.

PR #1018, opened 2020-02-20
[ML] Extend inference model serialisation to support multivalued leaves

tveasey:
Also, some more groundwork for SHAP.

Another step towards #982.


tveasey:
retest

PR #1019, opened 2020-02-21
[ML] Change the way hardening flags are set for libgcc_s.so.1

droberts195:
The gcc build process makes it hard to set extra linker flags for
the libgcc_s.so.1 library.  We had a [hack](https://github.com/elastic/ml-cpp/blob/07bd27fa66ccacdf6bfe333e3e5dab3335344679/dev-tools/docker/linux_image/Dockerfile#L29) in our build steps that
worked for gcc 7.3, but stopped working in gcc 7.5 with the result
that Debian's lintian tool flags the library as being insecure - see
elastic/elasticsearch#52554.  This change alters the hack to simply
add the required flags directly to the linker command that is used
for libgcc_s.so.1.

Relates elastic/elasticsearch#52554

### DIFF
--- a/build-setup/linux.md
+++ b/build-setup/linux.md
@@ -63,7 +63,7 @@ Unlike most automake-based tools, gcc must be built in a directory adjacent to t
 tar zxvf gcc-7.5.0.tar.gz
 cd gcc-7.5.0
 contrib/download_prerequisites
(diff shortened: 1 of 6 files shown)


droberts195:
The problem being fixed was never released, hence this PR is marked `>non-issue`.

PR #1020, opened 2020-02-21
[ML] Change the test temp directory location

droberts195:
When we first added the test temp directory functionality we
used the system temporary directory.  However, it's simpler
to just put it under the "build" directory at the top level
of the repo.  This neatly solves the problem of multiple
users clashing on the same server, or even the same user
testing multiple ml-cpp repos on the same machine at the
same time.

It may also help with a performance problem in Windows CI,
where the Jenkins workspace is excluded from checks by Windows
Defender.  So under Elastic CI, files written to the system
temporary directory get scanned by Windows Defender but files
written to subdirectories of the cloned repo do not.


droberts195:
The idea of changing the temp directory for `cl.exe` did not speed things up.  However, adding `cl.exe` as an excluded process in the Windows Defender settings _did_ help a lot.  This in combination with changing the temporary directory used by the unit tests themselves brought the build time back to what it was before we changed the CI worker image - see https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+windows/1312/.

The rough timings are:

* Without either optimisation: ~71 minutes
* With ML unit test temporary directory moved to workspace: ~61 minutes
* With ML unit test temporary directory moved to workspace and `cl.exe` excluded from Windows Defender: ~54 minutes

I logged into the CI worker manually to add the exclusion in https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+windows/1312/.  Tomorrow I will make enquiries about getting a Windows Defender exclusion for `cl.exe` baked into the CI worker image the next time it is rebuilt.

PR #1022, opened 2020-02-24
[ML] Format memory instrumentation output

valeriy42:
This PR activates memory instrumentation and formats it according to the [specified JSON format](https://github.com/elastic/ml-json-schemas-private/blob/master/schemas/instrumentation/memory_usage.schema.json).


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -64,14 +69,14 @@ void CDataFrameAnalysisInstrumentation::writer(core::CRapidJsonConcurrentLineWri
...
+void CDataFrameAnalysisInstrumentation::nextStep(std::uint32_t step) {
+    this->writeState(step);
 }
 
 void CDataFrameAnalysisInstrumentation::writeState(std::uint32_t step) {
-    this->writeProgress(step);
-    this->writeMemory(step);
+    // this->writeProgress(step);
+    int64_t timestamp = core::CTimeUtils::toEpochMs(core::CTimeUtils::now());
+    this->writeMemory(timestamp);
```
tveasey: nit: use std::int64_t, or how about just `this->writeMemory(core::CTimeUtils::toEpochMs(core::CTimeUtils::now());`

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include <string>
+
+BOOST_AUTO_TEST_SUITE(CDataFrameAnalysisInstrumentationTest)
+
+using namespace ml;
+
+BOOST_AUTO_TEST_CASE(testMemoryState) {
+    std::string jobId("JOB123");
+    std::int64_t memoryUsage = 1000;
```
tveasey: nit: can we use `{}`-initialisation.

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
```diff
@@ -0,0 +1,46 @@
...
+    rapidjson::Document results;
+    rapidjson::ParseResult ok(results.Parse(s_Output.str()));
+    BOOST_TEST_REQUIRE(static_cast<bool>(ok) == true);
+    BOOST_TEST_REQUIRE(results.IsArray() == true);
+
+    const auto& result{results[0]};
+    BOOST_TEST_REQUIRE(result["job_id"].GetString() == jobId);
+    BOOST_TEST_REQUIRE(result["type"].GetString() == "analytics_memory_usage");
+    BOOST_TEST_REQUIRE(result["peak_usage_bytes"].GetInt64() == memoryUsage);
+    BOOST_REQUIRE_SMALL(result["timestamp"].GetInt64() - timestamp, 10l);
```
tveasey: I wonder if this could fail in the CI environment: we've occasionally seen long pauses. How about getting time now before and after writing `s_Output` and checking `result["timestamp"].GetInt64() >= time before` and `result["timestamp"].GetInt64() <= time after`?

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: I will have to write the time stamp into analysis object as well. Thus, I prefer to keep it as a variable.

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
valeriy42: Changed in 16c7147

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
valeriy42: Fixed in 16c7147

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -64,14 +69,14 @@ void CDataFrameAnalysisInstrumentation::writer(core::CRapidJsonConcurrentLineWri
...
-    // TODO reactivate state writing, once the Java backend can accept it
-    //    this->writeState(step);
+void CDataFrameAnalysisInstrumentation::nextStep(std::uint32_t step) {
+    this->writeState(step);
 }
 
 void CDataFrameAnalysisInstrumentation::writeState(std::uint32_t step) {
-    this->writeProgress(step);
-    this->writeMemory(step);
+    // this->writeProgress(step);
```
dimitris-athanasiou: I am a bit confused with this commented out line. Will progress be written out from here instead of `CDataFrameAnalyzer`?

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -89,11 +94,15 @@ void CDataFrameAnalysisInstrumentation::writeProgress(std::uint32_t step) {
     }
 }
 
-void CDataFrameAnalysisInstrumentation::writeMemory(std::uint32_t step) {
+void CDataFrameAnalysisInstrumentation::writeMemory(std::int64_t timestamp) {
     if (m_Writer != nullptr) {
         m_Writer->StartObject();
-        m_Writer->Key(STEP_TAG);
-        m_Writer->Uint(step);
+        m_Writer->Key(TYPE_TAG);
```
dimitris-athanasiou: This doesn't lead to the expected format. Java expects named top level objects, so in this case we should expect:

```
"analysis_memory_usage": {
  "job_id": "foo",
  "timestamp: 123566778,
  "peak_usage_bytes": 42
}
```

valeriy42: I updated the code as we discussed.

dimitris-athanasiou:
This is ready to be merged now from the java side.

dimitris-athanasiou:
run tests

dimitris-athanasiou:
run elasticsearch-ci

dimitris-athanasiou:
retest

PR #1023, opened 2020-02-24
[ML] Rework computing SHAP to avoid using data frame storage

tveasey:
This also finishes up multi-parameter loss function support.

Note I don't save and restore `CTreeShapFeatureImportance` because train is always run and so this will always be reinitialised before any SHAP values are computed. For failover we should almost certainly keep track of SHAP values already computed and written, but it might make more sense for the Java to pass this information. I'm therefore deferring this detail to a later PR.

Finally, I've removed multithreading of the SHAP code, because it only gets passed one row at a time now. We should be able to parallelise over trees, but this will require a small tweak to parallel_for_each. Since we don't currently pass number threads from Java, I'll make this change in a follow on PR.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -242,19 +242,7 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
     }
 
     this->computeProbabilityAtWhichToAssignClassOne(frame);
-
-    // Populate number samples reaching each node.
-    CTreeShapFeatureImportance::computeNumberSamples(m_NumberThreads, frame,
-                                                     *m_Encoder, m_BestForest);
-
-    if (m_NumberTopShapValues > 0) {
-        // Create the SHAP calculator.
-        m_TreeShap = std::make_unique<CTreeShapFeatureImportance>(
-            frame, *m_Encoder, m_BestForest, m_NumberTopShapValues);
-    } else {
-        // Make sure the internal node values are set anyway.
-        CTreeShapFeatureImportance::computeInternalNodeValues(m_BestForest);
-    }
+    this->initializeTreeShap(frame);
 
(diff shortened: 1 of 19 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -414,6 +368,21 @@ void CBoostedTreeImpl::computeProbabilityAtWhichToAssignClassOne(const core::CDa
...
+    CTreeShapFeatureImportance::computeNumberSamples(m_NumberThreads, frame,
+                                                     *m_Encoder, m_BestForest);
+
+    if (m_NumberTopShapValues > 0) {
+        // Create the SHAP calculator.
+        m_TreeShap = std::make_unique<CTreeShapFeatureImportance>(
+            frame, *m_Encoder, m_BestForest, m_NumberTopShapValues);
+    } else {
+        // Set internal node values anyway.
+        CTreeShapFeatureImportance::computeInternalNodeValues(m_BestForest);
```
valeriy42: At the moment, we don't have internal node values as a part of the inference model definition. IMO, we should either communicated them to java as well or omit to compute them if SHAP is not required.

tveasey: Good point! I think it would be quite nice to have them (independent of having feature importance) since they could be used for the basis of interesting tree visualisations. However, there is no point in having them if they aren't exposed. I'll comment this out and leave a comment to revisit.

tveasey: See [24d7c4e](https://github.com/elastic/ml-cpp/pull/1023/commits/24d7c4e3aa866f0c5edc25d971b8319a0cb4e2d4).

tveasey:
Thanks @valeriy42. I followed your suggestion, but left a TODO. Can you take another look.

PR #1031, opened 2020-02-27
[ML] Instrumentation information for boosted tree learning

valeriy42:
This PR enables instrumentation of the supvervised learning jobs. I needed to add `maths::CDataFrameTrainBoostedTreeInstrumentationInterface` to have an interface for storing validation error results, hyperparameters, and timing.

Since I am passing `iteration` directly over the new interface, I also refactored `nextStep` to use an optional string `phase` as an argument. At the moment, we are not tagging phases, but we intend to in the future. Thus, I added a TODO there.

Finally, I extracted the common routines for generating regression and classification data into `CDataFrameAnalyzerTrainingFactory`. It is not used for feature importance testing so far, cause the generation procedure there is a little bit different. I intend to unify the code in a follow-up PR.


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -64,48 +106,226 @@ void CDataFrameAnalysisInstrumentation::writer(core::CRapidJsonConcurrentLineWri
...
+
+void CDataFrameTrainBoostedTreeInstrumentation::writeAnalysisStats(std::int64_t timestamp) {
+    auto* writer{this->writer()};
+    if (writer != nullptr) {
+        writer->StartObject();
+        writer->Key(JOB_ID_TAG);
+        writer->String(this->jobId());
+        writer->Key(TIMESTAMP_TAG);
+        writer->Int64(timestamp);
+        if (m_Type == E_Regression) {
```
tveasey: Maybe migrate to switch statement: if we ever add more types the missing cases would then generate a warning.

include/maths/CDataFrameAnalysisInstrumentationInterface.h
```diff
@@ -50,16 +54,96 @@ class MATHS_EXPORT CDataFrameAnalysisInstrumentationInterface {
...
+              s_TreeSizePenaltyMultiplier{treeSizePenaltyMultiplier},
+              s_LeafWeightPenaltyMultiplier{leafWeightPenaltyMultiplier} {};
+        double s_DepthPenaltyMultiplier = 0.0;
+        double s_SoftTreeDepthLimit = 0.0;
+        double s_SoftTreeDepthTolerance = 0.0;
+        double s_TreeSizePenaltyMultiplier = 0.0;
+        double s_LeafWeightPenaltyMultiplier = 0.0;
+    };
+    struct SHyperparameters {
+        double s_Eta = 0.1;
```
tveasey: I would initialise all the values in this class to something very obviously wrong: if we miss overwriting them that's a bug and we want to be able to pick that up easily when we review the output, e.g.
```suggestion
        double s_Eta = -1.0;
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -39,6 +41,9 @@ namespace {
 // quantiles anyway. So we amortise their compute cost w.r.t. training trees
 // by only refreshing once every MINIMUM_SPLIT_REFRESH_INTERVAL trees we add.
 const double MINIMUM_SPLIT_REFRESH_INTERVAL{3.0};
+const std::string HYPERPARAMETER_OPTIMIZATION_PHASE{"hyperparameter_optimization"};
+const std::string TRAINING_FINAL_TREE_PHASE{"training_final_tree"};
+const std::array<std::string, 1> REGRESSION_LOSSES{CMse::NAME};
```
tveasey: My inclination would be to add an `isRegression()` to the loss function hierarchy rather than rely on this hidden array: it means that when you add a loss function you know you have to define this quantity rather. As it is it would be easy to miss that you need to update this array.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -423,13 +441,15 @@ CBoostedTreeImpl::crossValidateForest(core::CDataFrame& frame) {
...
-        std::tie(forest, loss) = this->trainForest(
+        TDoubleVec lossValues;
+        std::tie(forest, loss, lossValues) = this->trainForest(
             frame, m_TrainingRowMasks[fold], m_TestingRowMasks[fold], m_TrainingProgress);
         LOG_TRACE(<< "fold = " << fold << " forest size = " << forest.size()
                   << " test set loss = " << loss);
         lossMoments.add(loss);
         m_FoldRoundTestLosses[fold][m_CurrentRound] = loss;
         numberTrees.push_back(static_cast<double>(forest.size()));
+        m_Instrumentation->lossValues(std::to_string(fold), std::move(lossValues));
```
tveasey: Is there any reason to convert this to a string here? I can't think of one.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -556,7 +582,7 @@ CBoostedTreeImpl::trainForest(core::CDataFrame& frame,
 
     LOG_TRACE(<< "Trained one forest");
 
-    return {forest, stoppingCondition.bestLoss()};
+    return {forest, stoppingCondition.bestLoss(), losses};
```
```suggestion
    return {forest, stoppingCondition.bestLoss(), std::move(losses)};
```

lib/test/CDataFrameAnalyzerTrainingFactory.cc
```diff
@@ -0,0 +1,116 @@
...
+        frame->parseAndWriteRow(
+            core::CVectorRange<const TStrVec>(fieldValues, 0, weights.size() + 1));
+    }
+
+    frame->finishWritingRows();
+
+    return frame;
+}
+}
+}
```
tveasey: No new line will generate a compiler warning
```suggestion
}

```

lib/maths/CBoostedTreeImpl.cc
valeriy42: Yes, when I am going to write the loss values in rapid json, `fold` becomes the key and it has to be a string. I cannot convert it to string during writing since the string has to exist in memory until the writer finishes writing, otherwise, it just puts some random bytes as the key.

lib/maths/CBoostedTreeImpl.cc
valeriy42: I would like to avoid working on the same files as you at the moment. I will put a TODO and change it in a follow-up PR, ok?

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -27,10 +30,13 @@ namespace api {
 //! progress, parameters, quality of results. The class also implements the functionality to
 //! write the state at different iteration into the results pipe.
 class API_EXPORT CDataFrameAnalysisInstrumentation
-    : public maths::CDataFrameAnalysisInstrumentationInterface {
+    : virtual public maths::CDataFrameAnalysisInstrumentationInterface {
```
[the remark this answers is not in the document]
tveasey: We talked about this offline and feel like it need not hold this PR up at this point. We can refactor this detail in a follow on change if we like.

lib/maths/CBoostedTreeImpl.cc
tveasey: Agreed. (In fact, I need to add an `isClassification` or equivalent for the PR which wires in multiclass, so we can wait for this to arrive.)

valeriy42:
Thank you for your comments @tveasey. I addressed your comments. It would be great if you could review the changes and let me know if it's ok.

include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -99,26 +110,67 @@ class API_EXPORT CDataFrameAnalysisInstrumentation
...
+    //! List of \p lossValues of validation error for the given \p fold.
+    void lossValues(std::string fold, TDoubleVec&& lossValues) override;
+    //! \return Structure contains hyperparameters.
+    SHyperparameters& hyperparameters() override { return m_Hyperparameters; }
+
+protected:
     counter_t::ECounterTypes memoryCounterType() override;
+
+private:
+    using TLossMap = std::unordered_map<std::string, TDoubleVec>;
```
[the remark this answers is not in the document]
valeriy42: Fixed using std::vector

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -114,6 +152,179 @@ counter_t::ECounterTypes CDataFrameTrainBoostedTreeInstrumentation::memoryCounte
...
+    m_ElapsedTime += delta;
+}
+
+void CDataFrameTrainBoostedTreeInstrumentation::lossType(const std::string& lossType) {
+    m_LossType = lossType;
+}
+
+void CDataFrameTrainBoostedTreeInstrumentation::lossValues(std::string fold,
+                                                           TDoubleVec&& lossValues) {
+    m_LossValues.emplace(std::make_pair(fold, lossValues));
```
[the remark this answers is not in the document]
valeriy42: Fixed

valeriy42:
retest

valeriy42:
retest

PR #1032, opened 2020-02-28
[ML] Correct logistic loss function

tveasey:
This was a bug in the computation of the best leaf weight when predictions for all its training rows are identical. In this case, we should find the weight to *add* to the current prediction which minimises the cross-entropy.

### DIFF
--- a/lib/maths/CBoostedTreeLoss.cc
+++ b/lib/maths/CBoostedTreeLoss.cc
@@ -144,6 +144,8 @@ CArgMinLogisticImpl::TDoubleVector CArgMinLogisticImpl::value() const {
     // case we only need one pass over the data and can compute the optimal
     // value from the counts of the two categories.
     if (this->bucketWidth() == 0.0) {
+        // This is the (unique) predicted value for the rows in leaf by the forest
+        // so far (i.e. without the weight for the leaf we're about to add).
         double prediction{(m_PredictionMinMax.min() + m_PredictionMinMax.max()) / 2.0};
         objective = [&](double weight) {
             double logOdds{prediction + weight};
--- a/lib/maths/CBoostedTreeLoss.cc
+++ b/lib/maths/CBoostedTreeLoss.cc
@@ -144,11 +144,13 @@ CArgMinLogisticImpl::TDoubleVector CArgMinLogisticImpl::value() const {
     // case we only need one pass over the data and can compute the optimal
     // value from the counts of the two categories.
     if (this->bucketWidth() == 0.0) {
-        objective = [this](double weight) {
+        double prediction{(m_PredictionMinMax.min() + m_PredictionMinMax.max()) / 2.0};
+        objective = [&](double weight) {
+            double logOdds{prediction + weight};
             double c0{m_CategoryCounts(0)};
             double c1{m_CategoryCounts(1)};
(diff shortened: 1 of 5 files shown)


lib/maths/CBoostedTreeLoss.cc
```diff
@@ -144,11 +144,13 @@ CArgMinLogisticImpl::TDoubleVector CArgMinLogisticImpl::value() const {
     // case we only need one pass over the data and can compute the optimal
     // value from the counts of the two categories.
     if (this->bucketWidth() == 0.0) {
-        objective = [this](double weight) {
+        double prediction{(m_PredictionMinMax.min() + m_PredictionMinMax.max()) / 2.0};
```
```suggestion
        double predictionSofar{(m_PredictionMinMax.min() + m_PredictionMinMax.max()) / 2.0};
```

valeriy42: I would add `Sofar` or something similar to signify that we are talking about the prediction of the forest so far.

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1220,7 +1220,7 @@ BOOST_AUTO_TEST_CASE(testLogisticRegression) {
 
     LOG_DEBUG(<< "mean log relative error = "
               << maths::CBasicStatistics::mean(meanLogRelativeError));
-    BOOST_TEST_REQUIRE(maths::CBasicStatistics::mean(meanLogRelativeError) < 0.5);
+    BOOST_TEST_REQUIRE(maths::CBasicStatistics::mean(meanLogRelativeError) < 0.52);
```
valeriy42: Do you have an intuition why the error goes up here?

tveasey: No! Interestingly though (looking at the debug) the hyperparameter choices become more stable with this change.

My best guess is there are multiple ways the code can "learn" to avoid the bug. For example, it'll potentially choose hyperparameter values which result in deeper trees so that it'll try and take the non-buggy path more often. Later trees added to the forest correct for the bug, since they're more likely to have different values for at least one of the rows so take the non-buggy path.

I've tested on some additional data sets and I haven't seen evidence of any accuracy degradation, but it will be interesting to review the results on our overnights.

lib/maths/CBoostedTreeLoss.cc
tveasey: Because I use `m_PredictionMinMax` I feel like it is more consistent to call this just prediction. I could change the name of this variable, but I think perhaps the best thing to do is just add an explaining comment. See [5a9d840](https://github.com/elastic/ml-cpp/pull/1032/commits/5a9d84009279d5aa635d4f0db21f685219550d0c).

lib/maths/unittest/CBoostedTreeTest.cc
tveasey: So of course there was a subtle undefined behaviour bug with capture by reference. In fact, the result is still slightly worse for one data set in this test, but slightly better for a different one, with the average error being back to almost exactly what it was before.

I do still think the comments above somewhat explain why the previous behaviour wasn't causing a more obvious degradation in accuracy.

tveasey:
retest

tveasey:
retest

PR #1034, opened 2020-03-03
[ML] Distinguish missing and empty categorical values

tveasey:
Currently, it is impossible to distinguish missing categorical fields and fields whose values are the empty string (the exchange format between C++ and Java doesn't support optional strings). This can cause issues for inference, which does understand the difference. Also, since these could be semantically different we would ideally distinguish them from the standpoint of training.

This change introduces a custom string to denote a missing value. By default this is the `\0` character, but is configurable. This will also allow us to clean up the handling of the missing target category for test data passed to classification, but I've left TODOs until the Java has been updated or we'll break integration tests.

I also found myself needing to introduce another parameter to `CDataFrameAnalysisSpecificationFactory`. I've taken the opportunity to introduce setters for the various parameters since the many optional parameters were becoming unsustainable.


droberts195:
It seems that the test of numeric missing field value is actually causing a fatal error that stops the test suite:

```
FATAL CLogger.cc@385 Input error: you can't use a number (0) to denote a missing field value.
```

So that test needs to replace the fatal error handler.

But I’m happy to merge as soon as the CI passes.

tveasey:
It turns out the `\0` is converted to zero by our `stringToType` method. It checks and returns false if the `std::string` is `empty`. It feels to me like this should instead check the `c_str` length. However, this is a slightly risky change and not needed for this PR, so I just added special case handling.

PR #1036, opened 2020-03-04
[ML] Remove special case handling for missing target field for classification

tveasey:
Following on from #1034, we can simplify the handling of missing target field values for classification. Java will supply the missing string value to denote these are missing rather than the empty string.


tveasey:
retest

droberts195:
> Just be careful about the merge timing. Wait until a few minutes after the Java change has successfully completed an intake build, not just until the Java change is merged.

The Java change passed an intake build: https://elasticsearch-ci.elastic.co/job/elastic+elasticsearch+master+git+push/3204/

So this PR is safe to merge now.

PR #1056, opened 2020-03-13
[ML] Wire in multiclass classification to the data frame analyzer command

tveasey:
This wires up multiclass classification so it can be run by the data frame analyzer and also . Marking as a non-issue since the feature was documented in #1037.

I need some more unit testing, so keeping as WIP at the moment, but the code changes are complete. I've also left some TODOs for supporting maximize minimum class recall, which I'll revisit in a later PR.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
@@ -123,8 +123,8 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
     core::CRapidJsonConcurrentLineWriter& writer,
     maths::CTreeShapFeatureImportance* featureImportance) const {
 
-    TDoubleVec probabilities{readClassProbabilities(row)};
-    TDoubleVec scores{readClassScores(row)};
+    auto probabilities = readClassProbabilities(row);
+    auto scores = readClassScores(row);
 
     double actualClassId{row[columnHoldingDependentVariable]};
     std::size_t predictedClassId(std::max_element(scores.begin(), scores.end()) -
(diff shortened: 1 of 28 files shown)


lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -216,14 +204,29 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writePredictedCategoryValue(
     }
 }
 
+CDataFrameTrainBoostedTreeClassifierRunner::TLossFunctionUPtr
+CDataFrameTrainBoostedTreeClassifierRunner::loss(std::size_t numberClasses) {
+    using namespace maths::boosted_tree;
```
droberts195: Is this needed?

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -239,6 +242,7 @@ CDataFrameTrainBoostedTreeClassifierRunner::inferenceModelDefinition(
 }
 
 // clang-format off
+const std::size_t CDataFrameTrainBoostedTreeClassifierRunner::MAX_NUMBER_CLASSES{30};
```
droberts195: It's probably worth adding a comment to say that if the `30` is ever changed it also needs changing in https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/dataframe/analyses/Classification.java#L57

tveasey: This likely will change at least before 7.7. I'll add a comment.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
tveasey: I use `CBinomialLogisticLoss` and `CMultinomialLogisticLoss` unqualified in the next line. I found with auto formatting this made the code a bit more readable.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
tveasey: I linked the named variable: since this comment is unlikely to get changed if the code is refactored or moved, but hopefully we won't change the name. See [1ce554b](https://github.com/elastic/ml-cpp/pull/1056/commits/1ce554bd5baeb18b97fc119335418d63ea3f96e4).

PR #1062, opened 2020-03-16
[ML] Adding num_matches and preferred_to_categories output to categories

benwtrent:
Two new fields added to our category output when they are written back to Java.

* `num_matches` indicates the total number of matches this category has seen
* `preferred_to_categories` indicates that this particular category is matched now instead of the listed categories. 

Java side https://github.com/elastic/elasticsearch/pull/54214

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -34,12 +34,6 @@
 
 * The macOS build platform for the {ml} C++ code is now High Sierra running Xcode 10.1,
   or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {ml-pull}867[#867].)
-== {es} version 7.8.0
-
-=== Enhancements
-
-* Adds new `num_matches` and `preferred_to_categories` fields to category output.
-(See {ml-pull}1062[#1062])
 
 == {es} version 7.8.0
 
@@ -49,6 +43,8 @@
   when CPU is constrained. (See {ml-pull}1109[#1109].)
 * Take `training_percent` into account when estimating memory usage for classification and regression. 
   (See {ml-pull}1111[1111].)
+* Adds new `num_matches` and `preferred_to_categories` fields to category output.
+  (See {ml-pull}1062[#1062])
 
 == {es} version 7.7.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -62,7 +62,8 @@ the build from version 2.20 to 2.34.  (See {ml-pull}1013[#1013].)
 * Remove all memory overheads for computing tree SHAP values. (See {ml-pull}1023[#1023].)
 * Distinguish between empty and missing categorical fields in classification and regression
 model training. (See {ml-pull}1034[#1034].)
-
+* Adds new `num_matches` and `preferred_to_categories` fields to category output.
+(See {ml-pull}1062[#1062])
 === Bug Fixes
 
 * Use largest ordered subset of categorization tokens for category reverse search regex.
--- a/lib/api/CFieldDataCategorizer.cc
+++ b/lib/api/CFieldDataCategorizer.cc
@@ -168,7 +168,9 @@ int CFieldDataCategorizer::computeCategory(const TStrStrUMap& dataRowFields) {
     if (exampleAdded || searchTermsChanged) {
         m_JsonOutputWriter.writeCategoryDefinition(
             categoryId, m_SearchTerms, m_SearchTermsRegex, m_MaxMatchingLength,
-            m_DataCategorizer->examplesCollector().examples(categoryId));
+            m_DataCategorizer->examplesCollector().examples(categoryId), 
+            m_DataCategorizer->numMatches(categoryId),
+            m_DataCategorizer->usurpedCategories(categoryId));
         if (categoryId % 10 == 0) {
             // Even if memory limiting is disabled, force a refresh occasionally
             // so the user has some idea what's going on with memory.
--- a/lib/api/CFieldDataCategorizer.cc
+++ b/lib/api/CFieldDataCategorizer.cc
@@ -89,10 +89,13 @@ bool CFieldDataCategorizer::handleRecord(const TStrStrUMap& dataRowFields) {
     // Non-empty control fields take precedence over everything else
     TStrStrUMapCItr iter = dataRowFields.find(CONTROL_FIELD_NAME);
     if (iter != dataRowFields.end() && !iter->second.empty()) {
+        // Always handle control messages, but signal completion of handling ONLY if we are the last handler
+        // e.g. flush requests are acknowledged here if this is the last handler
+        bool msgHandled = this->handleControlMessage(iter->second, !m_OutputHandler.consumesControlMessages());
         if (m_OutputHandler.consumesControlMessages()) {
             return m_OutputHandler.writeRow(dataRowFields, m_Overrides);
         }
-        return this->handleControlMessage(iter->second);
+        return msgHandled;
     }
 
     m_OutputFieldCategory =
@@ -112,7 +115,7 @@ void CFieldDataCategorizer::finalise() {
 
     // Make sure model size stats are up to date
     m_Limits.resourceMonitor().forceRefresh(*m_DataCategorizer);
-
+    writeOutChangedCategories();
     // Pass on the request in case we're chained
     m_OutputHandler.finalise();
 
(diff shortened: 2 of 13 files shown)


lib/api/unittest/CJsonOutputWriterTest.cc
```diff
@@ -1298,7 +1298,7 @@ BOOST_AUTO_TEST_CASE(testWriteCategoryDefinition) {
         ml::core::CJsonOutputStreamWrapper outputStream(sstream);
         ml::api::CJsonOutputWriter writer("job", outputStream);
 
-        writer.writeCategoryDefinition(categoryId, terms, regex, maxMatchingLength, examples);
+        writer.writeCategoryDefinition(categoryId, terms, regex, maxMatchingLength, examples, std::size_t(0), vector<int>);
```
droberts195:
```suggestion
        writer.writeCategoryDefinition(categoryId, terms, regex, maxMatchingLength, examples, 0, {});
```

I wouldn't bother explicitly saying the `0` is `size_t` in this case as the function being called enforces it.  Although pedantically you're creating an `int` that's got to be widened to a `size_t` in reality the optimizer won't generate code that does that.

And for your empty `vector` you should be able to pass an `initializer_list`, i.e. `{}`.

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -618,6 +618,45 @@ CTokenListDataCategorizerBase::calculateCategorizationStatus(std::size_t categor
...
+    }
+    auto iter = std::find_if(m_CategoriesByCount.begin(), m_CategoriesByCount.end(), 
+                             [categoryId](const TSizeSizePr& pr){
+                                 return pr.second == static_cast<std::size_t>(categoryId);
+                                 });
+    if (iter == m_CategoriesByCount.end()) {
+        LOG_WARN(<< "Could not find category count for category: " << categoryId);
+        return usurped;
+    }
+    iter++;
```
droberts195: It's best to use `++iter` if you don't care what the old value was.  This is because an iterator can be a non-trivial object and post-increment needs to copy that object so that it can return the old version.  (Maybe this can get optimized away as dead code if the `operator++` is inlined, but still it's better style to use pre-increment than post-increment with non-trivial objects if you don't care about the old value.)

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -618,6 +618,45 @@ CTokenListDataCategorizerBase::calculateCategorizationStatus(std::size_t categor
...
+    if (categoryId < 1 || static_cast<std::size_t>(categoryId) > m_Categories.size()) {
+        LOG_ERROR(<< "Programmatic error - invalid ML category: " << categoryId);
+        return usurped;
+    }
+    auto iter = std::find_if(m_CategoriesByCount.begin(), m_CategoriesByCount.end(), 
+                             [categoryId](const TSizeSizePr& pr){
+                                 return pr.second == static_cast<std::size_t>(categoryId);
+                                 });
+    if (iter == m_CategoriesByCount.end()) {
+        LOG_WARN(<< "Could not find category count for category: " << categoryId);
```
droberts195:
```suggestion
        LOG_WARN(<< "Could not find category definition for category: " << categoryId);
```

lib/api/CJsonOutputWriter.cc
```diff
@@ -909,6 +913,15 @@ void CJsonOutputWriter::writeCategoryDefinition(int categoryId,
...
     }
     m_Writer.EndArray();
+    m_Writer.String(NUM_MATCHES);
+    m_Writer.Uint64(numMatches);
+    m_Writer.String(PREFERRED_TO_CATEGORIES);
+    m_Writer.StartArray();
+    for (TIntVecCIter iter = usurpedCategories.begin(); iter != usurpedCategories.end(); ++iter) {
+        int id = *iter;
+        m_Writer.Int(id);
+    }
```
droberts195:
```suggestion
    for (int id : usurpedCategories) {
        m_Writer.Int(id);
    }
```

(There might be other places where we still have the verbose form because for many years the code had to work with old compilers.)

lib/api/CFieldDataCategorizer.cc
```diff
@@ -168,7 +168,9 @@ int CFieldDataCategorizer::computeCategory(const TStrStrUMap& dataRowFields) {
     if (exampleAdded || searchTermsChanged) {
```
droberts195: This condition is the biggest problem with this whole PR.  We only write category definitions when quite a major thing happens to them.  This means the number of matches recorded will be very out-of-date in some cases.

We could go to the opposite extreme and remove this condition altogether.  Then the numbers of matches would always be up-to-date, but it would kill performance as it would mean every single message categorized would lead to a document being indexed in Elasticsearch.  (Most of these documents would overwrite old ones, so the total number of searchable documents would not increase, but the indexing load would be horrific.)

So if we are to record the number of matches for each category then we need a policy on when the numbers will be accurate.  Immediately after a `_flush` or `_close` might be a good one, as that's the policy for when anomalies generated accurately reflect the data posted to a job.  Then we would have implement code to write all categories on `_flush` or `_close`, document that the number of matches is only accurate after a `_flush` or `_close`, and make sure our own UI didn't display numbers of matches that might be highly misleading.

We could possibly also make "number of matches has increased by 10% since the last time this category definition was written" and "this category definition hasn't been written for 3 wall clock hours" extra conditions in this test.  They wouldn't completely solve the consistency problem, but might mean the numbers were good enough for some uses.

The can of worms with number of matches being inconsistent between categories is the reason we haven't included it in the output in the past.  It has been thought about before 😉

benwtrent: i agree, this sort of thing came to mind last night as i was falling asleep :). 

I will dig around and experiment. My gut is we periodically update the category if there is a "large" change in `numMatched` and on `_flush` and `_close`.

docs/CHANGELOG.asciidoc
```diff
@@ -62,7 +62,8 @@ the build from version 2.20 to 2.34.  (See {ml-pull}1013[#1013].)
 * Remove all memory overheads for computing tree SHAP values. (See {ml-pull}1023[#1023].)
 * Distinguish between empty and missing categorical fields in classification and regression
 model training. (See {ml-pull}1034[#1034].)
-
+* Adds new `num_matches` and `preferred_to_categories` fields to category output.
+(See {ml-pull}1062[#1062])
```
droberts195, line 66: Please move this to the 7.8.0 section

lib/api/CFieldDataCategorizer.cc
```diff
@@ -492,7 +501,7 @@ bool CFieldDataCategorizer::handleControlMessage(const std::string& controlMessa
         break;
     case 'f':
         // Flush ID comes after the initial f
-        this->acknowledgeFlush(controlMessage.substr(1));
+        this->acknowledgeFlush(controlMessage.substr(1), lastHandler);
         break;
     default:
         LOG_WARN(<< "Ignoring unknown control message of length "
```
droberts195, line 507: This warning should be inside an `if (lastHandler)` test, because otherwise it will generate spam for all the anomaly detector control messages.

lib/api/CFieldDataCategorizer.cc
```diff
@@ -506,13 +515,46 @@ bool CFieldDataCategorizer::handleControlMessage(const std::string& controlMessa
...
+
+void CFieldDataCategorizer::writeOutChangedCategories() {
+    int numCategories = static_cast<int>(m_DataCategorizer->numCategories());
+    if (numCategories == 0) {
+        return;
+    }
+    std::string searchTerms;
+    std::string searchTermsRegex;
+    std::size_t maxLength;
+    bool wasCached(false);
```
droberts195, line 538: nit: braced initialisation is preferred now, i.e. `bool wasCached{false};`

lib/api/CFieldDataCategorizer.cc
```diff
@@ -506,13 +515,46 @@ bool CFieldDataCategorizer::handleControlMessage(const std::string& controlMessa
...
+    for (int categoryId = 1; categoryId <= numCategories; categoryId++) {
+        if (m_DataCategorizer->categoryChangedAndReset(categoryId)) {
+            searchTerms.clear();
+            searchTermsRegex.clear();
+            maxLength = 0;
```
droberts195, line 543: I don't think you need to reset these 3 variables here.  They're only local variables and `m_DataCategorizer->createReverseSearch` will return `false` if it doesn't assign to them (or if it doesn't then that's the bug that should be fixed instead of working around it here).

PR #1069, opened 2020-03-19
[ML] Speed ups for multiclass classification

tveasey:
These mainly revolve around reducing the number of new and delete calls from creating temporary vectors. I also adjusted the k-means parameters. The cost of computing leaf values is still dominating (compared to binary case) so this needs more work. This is changing unreleased code so marking as non-issue.


tveasey:
retest

PR #1070, opened 2020-03-19
[ML] More speed ups for multiclass classification

tveasey:
This makes a number of changes to minimise copying vectors when performing streaming k-means. This mainly revolves around moving collections into place where possible. I also made some small tidy ups in the related code and enabled `CBoostedTreeTest/testMultinomialLogisticRegression` which is (just) running fast enough. This mainly affects unreleased code so marking as a non-issue.

### DIFF
--- a/include/maths/CKdTree.h
+++ b/include/maths/CKdTree.h
@@ -154,17 +154,21 @@ public:
         POINT s_Point;
     };
     using TNodeVec = std::vector<SNode>;
+    using TNodeVecItr = typename TNodeVec::iterator;
     using TNodeVecCItr = typename TNodeVec::const_iterator;
 
     //! \brief Iterates points in the tree.
+    template<typename MAYBE_CONST_POINT, typename ITR>
     class TPointIterator
-        : public boost::random_access_iterator_helper<TPointIterator, POINT, std::ptrdiff_t, const POINT*, const POINT&> {
+        : public boost::random_access_iterator_helper<TPointIterator<MAYBE_CONST_POINT, ITR>, MAYBE_CONST_POINT> {
     public:
         TPointIterator() = default;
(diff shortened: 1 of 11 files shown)


include/maths/CKdTree.h
```diff
@@ -36,6 +36,18 @@ namespace kdtree_detail {
...
+    using type = T;
+};
+template<typename T>
+struct unwrapped_ref<std::reference_wrapper<T>> {
+    using type = T;
+};
+template<typename T>
+struct unwrapped_ref<std::reference_wrapper<const T>> {
+    using type = T;
+};
```
droberts195: We have some functions like this in `core/UnwrapRef.h`.  Could those be used instead?

include/maths/CKdTree.h
```diff
@@ -512,6 +550,11 @@ class CKdTree {
         return coordinate == m_Dimension ? 0 : coordinate;
     }
 
+    //! Cast away any reference wrapper.
+    inline static const TPoint& unwrap(const POINT& p) {
+        return static_cast<const TPoint&>(p);
+    }
```
droberts195: Could `core::unwrap_ref()` from `core/UnwrapRef.h` do this?

include/maths/CKdTree.h
tveasey: Good point! I forgot about those.

tveasey: See [dbb74db](https://github.com/elastic/ml-cpp/pull/1070/commits/dbb74dbcd91239394e2ae1bada60489a4cd00cef).

include/maths/CKdTree.h
tveasey: See [dbb74db](https://github.com/elastic/ml-cpp/pull/1070/commits/dbb74dbcd91239394e2ae1bada60489a4cd00cef).

PR #1082, opened 2020-03-24
[ML] Rework the approximate loss function for choosing leaf values for multiclass classification

tveasey:
This change switches to using a weighted random sample of the leaf's training examples to determine the partition with which to approximate the log-loss objective for computing optimal leaf values. (Unfortunately, I couldn't get k-means fast enough for this purpose.)

The sample weights are chosen based on the mean distance between each prediction and the others in the random sample, i.e. they repel one another. This generates a more uniform sampling of the points, which results in lower average radius for the partition subsets. The result is a better approximation to the true loss function than using a uniform random sample.

### DIFF
--- a/include/maths/CSampling.h
+++ b/include/maths/CSampling.h
@@ -21,8 +21,10 @@
 #include <boost/random/mersenne_twister.hpp>
 #include <boost/random/uniform_int_distribution.hpp>
 
+#include <algorithm>
 #include <cstddef>
 #include <functional>
+#include <limits>
 #include <vector>
 
 namespace ml {
@@ -76,7 +78,8 @@ public:
 
             double weight{this->weight(x)};
             double p{uniformSample(m_Rng, 0.0, 1.0)};
-            if (p * (m_StreamWeight + weight) < m_SampleWeight + weight) {
+            double exchangeWeight{weight - this->minWeight()};
+            if (p * (m_StreamWeight + exchangeWeight) < m_SampleWeight + exchangeWeight) {
                 std::size_t slot{this->sampleToReplace(weight)};
                 if (slot < m_TargetSampleSize) {
                     m_SampleWeight += weight - this->doSample(slot, x, weight);
@@ -95,6 +98,7 @@ public:
     private:
         virtual double seedWeights() = 0;
         virtual double weight(const T& x) = 0;
+        virtual double minWeight() const = 0;
         virtual std::size_t sampleToReplace(double weight) = 0;
         virtual double doSample(std::size_t slot, const T& x, double weight) = 0;
 
@@ -157,6 +161,7 @@ public:
             return static_cast<double>(this->targetSampleSize());
         }
         double weight(const T&) override { return 1.0; }
+        double minWeight() const override { return 1.0; }
         std::size_t sampleToReplace(double) override {
             return uniformSample(this->rng(), 0, this->targetSampleSize() + 1);
         }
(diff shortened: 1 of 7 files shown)


include/maths/CSampling.h
```diff
@@ -42,6 +45,71 @@ class MATHS_EXPORT CSampling : private core::CNonInstantiatable {
...
+        std::size_t targetSampleSize() const { return m_TargetSampleSize; }
+
+        //! Reset in preparation for resampling.
+        virtual void reset() {
+            m_SampleSize = 0;
+            m_StreamWeight = m_SampleWeight = 0.0;
+        }
+
+        //! Sample the value \p x uniformly at random.
+        void sample(const T& x) {
```
valeriy42: Can you please explain the sampling algorithm or give a reference to the description?

lib/maths/CBoostedTreeLoss.cc
```diff
@@ -361,25 +333,34 @@ CArgMinMultinomialLogisticLossImpl::value() const {
...
     for (std::size_t i = 0; i < NUMBER_RESTARTS; ++i) {
-        for (int j = 0; j < x0.size(); ++j) {
+        for (int j = 0; j < w0.size(); ++j) {
             double alpha{CSampling::uniformSample(m_Rng, 0.0, 1.0)};
-            x0(j) = weightBoundingBox[0](j) +
+            w0(j) = weightBoundingBox[0](j) +
                     alpha * (weightBoundingBox[1](j) - weightBoundingBox[0](j));
         }
-        LOG_TRACE(<< "x0 = " << x0.transpose());
+        LOG_TRACE(<< "x0 = " << w0.transpose());
```
```suggestion
        LOG_TRACE(<< "w0 = " << w0.transpose());
```

lib/maths/unittest/CSamplingTest.cc
```diff
@@ -214,4 +219,104 @@ BOOST_AUTO_TEST_CASE(testMultivariateNormalSample) {
...
+        [&](std::size_t slot, const TVector& value) { samples[slot] = value; }};
+    TDissimilaritySampler dissimilaritySampler{numberSamples};
+
+    test::CRandomNumbers rng;
+
+    TMeanAccumulator percentageSeparationIncrease;
+
+    TVector x{4};
+    TDoubleVec p;
+    TDoubleVec components;
```
valeriy42: Should you reserve the space for the vectors accordingly?

tveasey: `rng.generateLogNormalSamples` calls reserve for me below. However, `p` is a hangover from an earlier version so I'll tidy up.

tveasey: [9c43087](https://github.com/elastic/ml-cpp/pull/1082/commits/9c43087df595c1eb012d76465343487470fac424).

include/maths/CSampling.h
tveasey: I'll write some fuller comments.

Ideally I'd use a full quasi-random sampling scheme, but I wanted something v.fast and streaming. In the end, I adapted the idea behind k-means++ initialisation, but wanted to avoid having to do nearest neighbour search to get the probability with which to sample each point. This was also based on trying to get something in place quickly given closeness of 7.7 (and we maybe can refine).

So what does this gain us?

For interest, here is an example of some testing I did using random logit values for the leaf predictions. (We get similar behaviour for a variety of distributions and often noticeably lower errors by spreading out the samples better.)

Screenshot 2020-03-25 at 18 30 10

The error in approximate loss surface using random samples:
Screenshot 2020-03-25 at 18 29 17

The error in the approximate loss surface using dissimilar samples:
Screenshot 2020-03-25 at 18 30 42

Flattening the error surfaces:
Screenshot 2020-03-25 at 18 41 29

tveasey: I added a fuller description for the new sampling in [34302ba](https://github.com/elastic/ml-cpp/pull/1082/commits/34302ba97a86af44219f12a1129f1286e1168f0d).

PR #1083, opened 2020-03-25
[ML] Avoid temporaries in vector expressions

tveasey:
Generally, Eigen elides all vector temporaries in vector expressions. The one exception is when we assign to a `CDenseVector`. We use placeholder vectors in various places to avoid additional allocations and we're not getting the full benefits from these as a result.

This gets about 30% improvement in multiclass classification runtime.


dimitris-athanasiou:
Removed `enhancement` label as this impacts an unreleased feature.

PR #1086, opened 2020-03-26
[ML] Fix out-of-bounds read in multiclass feature importance unit test

tveasey:
Note also that the default case generates a compiler warning for me because all enums are explicitly handled.

The error is in test code only and also isn't released so I've marked as a non-issue.

### DIFF
--- a/lib/api/CDataFrameAnalysisInstrumentation.cc
+++ b/lib/api/CDataFrameAnalysisInstrumentation.cc
@@ -267,9 +267,6 @@ void CDataFrameTrainBoostedTreeInstrumentation::writeAnalysisStats(std::int64_t 
         case E_Classification:
             writer->Key(CLASSIFICATION_STATS_TAG);
             break;
-        default:
-            LOG_ERROR(<< "Supervised learning type unknown or not set.");
-            return;
         }
         writer->StartObject();
         writer->Key(JOB_ID_TAG);
--- a/lib/api/CDataFrameAnalysisInstrumentation.cc
+++ b/lib/api/CDataFrameAnalysisInstrumentation.cc
@@ -235,6 +235,10 @@ void CDataFrameOutliersInstrumentation::writeParameters(rapidjson::Value& parent
     }
 }
 
+CDataFrameTrainBoostedTreeInstrumentation::CDataFrameTrainBoostedTreeInstrumentation(const std::string& jobId)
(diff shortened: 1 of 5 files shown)


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -267,9 +267,6 @@ void CDataFrameTrainBoostedTreeInstrumentation::writeAnalysisStats(std::int64_t
         case E_Classification:
             writer->Key(CLASSIFICATION_STATS_TAG);
             break;
-        default:
-            LOG_ERROR(<< "Supervised learning type unknown or not set.");
-            return;
```
droberts195: ++ to removing this.

With `clang++` we build with the `-Werror-switch` option so if another enum value was added to the enum definition but not added to the switch then the file wouldn't compile on macOS, so we would catch it in the PR build at the latest.

lib/api/unittest/CDataFrameAnalyzerFeatureImportanceTest.cc
```diff
@@ -142,10 +142,15 @@ void setupMultiClassClassificationData(const TStrVec& fieldNames,
     maths::CPRNG::CXorOShiro128Plus rng;
     std::uniform_real_distribution<double> u01;
     int numberFeatures{static_cast<int>(weights.size())};
-    TDoubleVec w{weights};
     int numberClasses{static_cast<int>(classes.size())};
+    TDoubleVec storage(numberClasses * numberFeatures);
+    for (int i = 0; i < numberClasses; ++i) {
```
droberts195: Should it be `numberFeatures`?  Or if not on this line somewhere.  There is an inconsistency somewhere between lines 146 and 149.

tveasey: Good point. In fact, because we resize the vector (we're okay), the extra weights are just zero, so it's harmless but not intended. I'll correct this. Also Eigen expects the data to be column major and I'd intended to use the same weights for each class so I've switch that as well.

tveasey: Fixed in [240937b](https://github.com/elastic/ml-cpp/pull/1086/commits/240937bd915bf2adb6bd4a69d4ca5b3338093c03).

PR #1089, opened 2020-03-26
[ML] Extend .gitignore to work with VSCode

valeriy42:
This mini PR adds hidden directories of Visual Studio Code and clangd into the `.gitignore` file. No functionality is changed.


valeriy42:
@droberts195 thank you for the comment. `.dockerignore` is updated now as well.

PR #1092, opened 2020-03-27
[ML] Speedup calculating leaf values for multiclass classification

tveasey:
The optimisation objective is convex so there is no need for restarts. Also, since we're getting unique vectors there is no need to stable sort them. Finally, I increased the size of the approximating partition somewhat because the optimisation is significantly cheaper.

This also truncates lambda to be (small and) positive. This has a couple of benefits: 1) it means we always have a unique minimum (otherwise the normalisation constraint means any {w_j' = k  + w_j} for {w_j} a minimum is also a minimum), 2) it naturally limits weights when all labels are equal so we no longer need to truncate them.

### DIFF
--- a/lib/maths/CBoostedTreeLoss.cc
+++ b/lib/maths/CBoostedTreeLoss.cc
@@ -304,7 +304,7 @@ CArgMinMultinomialLogisticLossImpl::value() const {
     // The optimisation objective is convex. To see this note that we can write
     // it as sum_i{ f_ij(w) } + ||w||^2 with f_ij(w) = -[log(softmax_j(z_i + w))].
     // Since the sum of convex functions is convex and ||.|| is clearly convex we
-    // just require the f_ij are convex. This is a standard result and follows from
+    // just require the f_ij to be convex. This is a standard result and follows from
     // the fact that their Hessian is of the form H = diag(p) - p p^t where 1-norm
     // of p is one. Convexity follows if this is positive definite. To verify note
     // that x^t H x = ||p^(1/2) x||^2 ||p^(1/2)||^2 - (p^t x)^2, which is greater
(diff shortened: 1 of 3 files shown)


tveasey:
retest

lib/maths/CBoostedTreeLoss.cc
```diff
@@ -301,75 +301,34 @@ void CArgMinMultinomialLogisticLossImpl::merge(const CArgMinLossImpl& other) {
 CArgMinMultinomialLogisticLossImpl::TDoubleVector
 CArgMinMultinomialLogisticLossImpl::value() const {
 
-    using TMinAccumulator = CBasicStatistics::SMin<double>::TAccumulator;
+    // The optimisation objective is convex. To see this note that we can write
+    // it as sum_i{ f_ij(w) } + ||w||^2 with f_ij(w) = -[log(softmax_j(z_i + w))].
+    // Since the sum of convex functions is convex and ||.|| is clearly convex we
+    // just require the f_ij are convex. This is a standard result and follows from
```
```suggestion
    // just require the f_ij to be convex. This is a standard result and follows from
```

lib/maths/CBoostedTreeLoss.cc
```diff
@@ -301,75 +301,34 @@ void CArgMinMultinomialLogisticLossImpl::merge(const CArgMinLossImpl& other) {
...
+    // the fact that their Hessian is of the form H = diag(p) - p p^t where 1-norm
+    // of p is one. Convexity follows if this is positive definite. To verify note
+    // that x^t H x = ||p^(1/2) x||^2 ||p^(1/2)||^2 - (p^t x)^2, which is greater
```
valeriy42: The first term of the difference is not clear to me. However, I can see how the positive definiteness follows from Cauchy-Schwarz assuming all entries of p are non-negative and sum to 1. Does this assumption hold?

tveasey: Yes, `sum_i{ [p]_i } = 1`, so `sum_i{ p_i x_i^2 } = sum_i{ p_i x_i^2 } sum_i{ p_i }`. This is used to rewrite the first term. In this form, we then simply take `y = p^(1/2) x` and `z = p^(1/2)` for `[p]_i = [p]_i^(1/2)` and apply Cauchy-Schawrz to the pair of vectors `y` and `z`.

PR #1096, opened 2020-03-27
[ML] Instrumentation parameter names should match analysis config where applicable

tveasey:
We should the same field names for instrumentation JSON that we do for the analysis configuration, so reported values can just be pasted (unmodified) into a new job config if desired. This cuts over to use the strings defined in the api runner objects and changes some of them in the process.

I think this should ideally released in 7.7 to save renaming after we've released this functionality so I've marked as non-issue.


dimitris-athanasiou:
I'll get this worked in the java side which has to be merged before this PR is merged.

dimitris-athanasiou:
retest

tveasey:
retest

PR #1102, opened 2020-03-30
[ML] Speed up the lat_long function

tveasey:
Clustering can be the bottleneck for the `lat_long` function.

This reworks the calculation of the distances to the selected points in k-means++ initialisation. Before we were creating a k-d tree for each point we added and looking up nearest neighbours. This is unnecessary since we can simply update the distances directly, i.e. `distance_i = min(distance_i, distance(selected, x_i))`. The other main speedup comes from the fact that in #1037 I reworked online k-means to remove the points buffer. This saves us memory and we can spend this memory by accumulating more points before we re-cluster.

I also made some tweaks to cutdown the number of allocations (principally by moving various variables into place). Finally, I corrected `CKMeansOnline::split` to copy all the parent clustering parameters into each split and made a couple of other small tidy ups.

In total, I get around a 60% improvement in runtime from these changes for `CXMeansOnline`.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -35,6 +35,9 @@
 * The macOS build platform for the {ml} C++ code is now High Sierra running Xcode 10.1,
   or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {ml-pull}867[#867].)
 
+== {es} version 7.8.0
+* Speedup anomaly detection for the lat_long function. (See {ml-pull}1102[#1102].)
+
 == {es} version 7.7.0
 
 === New Features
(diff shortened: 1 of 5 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -35,6 +35,9 @@
 * The macOS build platform for the {ml} C++ code is now High Sierra running Xcode 10.1,
   or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {ml-pull}867[#867].)
 
+== {es} version 7.8.0
+* Speedup anomaly detection for the lat_long function. (See {ml-pull}1102[#1102].)
```
droberts195, line 39: The should be an `=== Enhancements` header in between these two lines.

PR #1109, opened 2020-03-31
[ML] Reduce analysis process priority for CPU scheduling

droberts195:
When many ML jobs are running at the same time and there is
insufficient CPU it is more important to prioritise the JVM
performance so that search and indexing latency remains
acceptable.

This change is an attempt to do this by increasing the "nice"
setting of the analysis processes (not controller) to reduce
their CPU scheduling priority.

This change should have little or no effect on performance on
machines that are not CPU constrained.

The exact numbers may need to be tweaked in a followup PR.

### DIFF
--- a/include/core/CProcessPriority.h
+++ b/include/core/CProcessPriority.h
@@ -26,6 +26,8 @@ namespace core {
 //! IMPLEMENTATION DECISIONS:\n
 //! This is a static class - it's not possible to construct an instance of it.
 //!
+//! On Linux and macOS we reduce the CPU scheduling priority using "nice".
+//!
 //! On Linux we also attempt to increase the OOM killer adjustment for the
 //! process such that it is more likely to be killed than other processes
 //! when the Linux kernel decides that there isn't enough free memory.
--- a/include/core/CProcessPriority.h
+++ b/include/core/CProcessPriority.h
@@ -27,6 +27,9 @@ namespace core {
 //! This is a static class - it's not possible to construct an instance of it.
 //!
 //! On Linux and macOS we reduce the CPU scheduling priority using "nice".
+//! This should be done after connecting named pipes to the JVM, because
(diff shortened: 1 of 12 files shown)


droberts195:
The most recent test failure shows a problem with the ES integration tests.  The ES build system runs multiple test suites in parallel.  When the ML native multi node tests run in parallel to other suites the other suites take almost all the CPU resulting in the native multi node tests timing out.

droberts195:
Actually ignore the previous comment.  That build timed out because of an inconsistency between the Java and C++ (the same one described in https://github.com/elastic/elasticsearch/issues/50313#issuecomment-607143736).

include/core/CProcessPriority.h
```diff
@@ -26,19 +26,21 @@ namespace core {
...
 public:
-    //! Reduce whatever priority measures are deemed appropriate for the
+    //! Reduce priority for memory if deemed appropriate for the
     //! current OS.
-    static void reducePriority();
+    static void reduceMemoryPriority();
+
+    //! Reduce priority for CPU if deemed appropriate for the
+    //! current OS.
+    static void reduceCpuPriority();
```
tveasey, line 43: How about a `static void reducePriority()` method which calls both as well? We generally always call both.

droberts195: The reason I didn't do this is because we want to reduce memory priority before installing system call filtering, but want to reduce CPU priority after connecting the named pipes.

Maybe that's premature optimisation, but the named pipe connection code is quite time-sensitive and you don't get any useful errors if it fails for an obscure reason.  I was worried that it might sporadically fail due to lack of CPU but this would be impossible to prove.

I can add this reasoning to the Doxygen comment of the class for future reference.

tveasey: Ok that makes sense. I think this is worth a comment at the point at which you call the `reduceCpuPriority`.

PR #1113, opened 2020-04-02
[ML] Multiclass maximise minimum recall

tveasey:
This implements maximise minimum class recall for multiclass classification when assigning class labels, which is often a better objective when classes are imbalanced in the training data.

### DIFF
--- a/lib/maths/CDataFrameUtils.cc
+++ b/lib/maths/CDataFrameUtils.cc
@@ -1208,7 +1208,8 @@ CDataFrameUtils::maximizeMinimumRecallForMulticlass(std::size_t numberThreads,
                     }
                 }
             },
-            TDoubleVectorVec(numberClasses + 1, TDoubleVector::Zero(numberClasses)));
+            TDoubleVectorVec(numberClasses + 1,
+                             TDoubleVector{TDoubleVector::Zero(numberClasses)}));
         auto copyObjectiveAndGradient = [](TDoubleVectorVec state, TDoubleVectorVec& result) {
             result = std::move(state);
         };
--- a/lib/maths/CDataFrameUtils.cc
+++ b/lib/maths/CDataFrameUtils.cc
@@ -1094,6 +1094,7 @@ CDataFrameUtils::maximizeMinimumRecallForMulticlass(std::size_t numberThreads,
     sampleMask.extend(false, rowMask.size() - sampleMask.size());
     LOG_TRACE(<< "# row indices = " << rowIndices.size());
 
+    // Compute the count of each class in the sample set.
     auto readClassCounts = core::bindRetrievableState(
         [&](TDoubleVector& counts, TRowItr beginRows, TRowItr endRows) {
             for (auto row = beginRows; row != endRows; ++row) {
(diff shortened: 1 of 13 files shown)


lib/maths/CDataFrameUtils.cc
```diff
@@ -586,13 +589,9 @@ CDataFrameUtils::categoryFrequencies(std::size_t numberThreads,
...
-        if (doReduce(frame.readRows(numberThreads, 0, frame.numberRows(),
-                                    readCategoryCounts, &rowMask),
-                     copyCategoryCounts, reduceCategoryCounts, result) == false) {
-            HANDLE_FATAL(<< "Internal error: failed to calculate category"
-                         << " frequencies. Please report this problem.")
-            return result;
-        }
+        doReduce(frame.readRows(numberThreads, 0, frame.numberRows(),
+                                readCategoryCounts, &rowMask),
+                 copyCategoryCounts, reduceCategoryCounts, result);
```
tveasey, line 594: The check was unnecessary here since readCategoryCounts can't fail.

lib/maths/CDataFrameUtils.cc
```diff
@@ -1049,6 +997,263 @@ CDataFrameUtils::metricMicWithColumnDataFrameOnDisk(const CColumnValue& target,
...
+    //
+    //   min_w{max_j{f_j(w)}} = min_w{max_j{1 - c_j(w) / n_j}}
+    //
+    // We can write f_j(w) as
+    //
+    //    max_j{sum_i{1 - 1{argmax_i(w_i p_i) == j}} / n_j}                     (1)
+    //
+    // where 1{.} denotes the indicator function. (1) has a smooth relaxation given
+    // by f_j(w) = max_j{sum_i{1 - softmax_j(w_i p_i)} / n_j}. Note that this isn't
+    // convex so we use multiple restarts.
```
valeriy42, line 1158: nice! :+1:

lib/maths/CDataFrameUtils.cc
```diff
@@ -1049,6 +997,263 @@ CDataFrameUtils::metricMicWithColumnDataFrameOnDisk(const CColumnValue& target,
...
+        double loss;
+        std::tie(w0, loss) = lbfgs.minimize(objective, objectiveGradient, std::move(w0));
+        LOG_TRACE(<< "weights = " << w0.transpose() << ", loss = " << loss);
+        if (minLoss.add(loss)) {
+            minWeights = std::move(w0);
+        }
+        for (std::size_t j = 0; j < numberClasses; ++j) {
+            interpolate(j) = CSampling::uniformSample(rng, 0.0, 1.0);
+        }
+        w0 = (a + interpolate.cwiseProduct(b - a)).array().exp();
```
valeriy42, line 1246: It seems to me that this should be at the beginning of the for-loop. Otherwise, you try with w0=(1,1,1..,1) at first and this can be outside of your bounds.

tveasey: I actually intended to do this. My thought was we should always include the best solution in the vicinity of doing nothing, i.e. all weights being equal. Since we use use line search with backtracking we are then *guarantying* that we never do worse than not reweighting at all, which is a nice property since the optimisation objective is complex. WDYT?

tveasey: I reworked initialisation slightly and added a comment.

I still think it is worth trying the one vector for reason outlined. I also now bake in the fact we expect weights to be (roughly) a monotonic decreasing function of class recalls. (This isn't guaranteed because it depends how close probabilities are and what the predicted classes are for the error cases.)

I also reduced the number of restarts because trying out with a wider variety of numbers of classes and range of recalls I didn't see evidence we needed as many restarts after this change. See [this](https://github.com/elastic/ml-cpp/pull/1113/commits/21ba4feb399f86ec251d11ee7f5d840e83f170bf) commit.

PR #1114, opened 2020-04-02
[ML] Improve bad input handling in distribution models

tveasey:
In particular, it more reliably stops non-finite inputs polluting the model.

### DIFF
--- a/lib/maths/CLogNormalMeanPrecConjugate.cc
+++ b/lib/maths/CLogNormalMeanPrecConjugate.cc
@@ -578,7 +578,7 @@ public:
         for (std::size_t i = 0u; i < m_Samples.size(); ++i) {
             double residual = m_Samples[i];
             double n = maths_t::countForUpdate(m_Weights[i]);
-            if (residual <= 0.0 || !CMathsFuncs::isFinite(residual) ||
+            if (residual <= x || !CMathsFuncs::isFinite(residual) ||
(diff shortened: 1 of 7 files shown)


lib/maths/CLogNormalMeanPrecConjugate.cc
```diff
@@ -577,10 +577,11 @@ class CLogSampleSquareDeviation : core::CNonCopyable {
         result = 0.0;
         for (std::size_t i = 0u; i < m_Samples.size(); ++i) {
             double residual = m_Samples[i];
-            if (residual <= 0.0) {
+            double n = maths_t::countForUpdate(m_Weights[i]);
+            if (residual <= x || !CMathsFuncs::isFinite(residual) ||
```
droberts195, line 581: Was it intentional to change `<= 0.0` to `<= x`?

tveasey: Yes. This is actually was actually a bug I spotted while working on this. It causes us to discard a small number of sample we shouldn't be.

tveasey: Although, related tests have failed, so let me double check this.

tveasey: I got the new condition wrong as well :(. Corrected.

PR #1117, opened 2020-04-02
[ML] Add getpriority/setpriority to Linux system call whitelist

droberts195:
This is to support #1109.

The risk with setpriority is that it can possibly be used on a
different process.  However, it is extremely unlikely that the
user that Elasticsearch is running as in production will have
been granted permission to call it.  Running as root is banned
and it's an extra admin action to grant the privilege to any
other user and there is no good reason for it to have been
granted to the Elasticsearch user.


droberts195:
retest

PR #1122, opened 2020-04-03
[ML] Fix off-by-one error in usurped categories

### LINKED ISSUE #1121: [CI] SEGV during Kibana snapshot promotion test
The `autodetect` process crashed during https://kibana-ci.elastic.co/job/elasticsearch+snapshots+verify/446/

The error was:

```
   │      Fatal error: 'si_signo 11, si_code: 1, si_errno: 0, address: 0x7f3eac0d55bd, library: /home/robert/github/kibana/.es/es-test-cluster/modules/x-pack-ml/platform/linux-x86_64/bin/../lib/libMlModel.so, base: 0x7f3eabcf2000, normalized address: 0x3e35bd', version: 8.0.0-SNAPSHOT (build a7f071a14bc835)
```

Checking which function that address is in reveals:

```
[dave@linux lib]$ objdump -T libMlModel.so | grep '^00000000003e[23]' | sort
00000000003e2de0 g    DF .text  0000000000000aaa  Base        _ZN2ml5model29CTokenListDataCategorizerBase17usurpedCategoriesEi
00000000003e3890 g    DF .text  000000000000144f  Base        _ZNK2ml5model29CTokenListDataCategorizerBase20updateModelSizeStatsERNS0_16CResourceMonitor15SModelSizeStatsE
[dave@linux lib]$ c++filt _ZN2ml5model29CTokenListDataCategorizerBase17usurpedCategoriesEi
ml::model::CTokenListDataCategorizerBase::usurpedCategories(int)
```

droberts195:
The mistake was that m_CategoriesByCount has indices into
m_Categories, not category IDs (which are one more than
the indices so that they start at 1 rather than 0).

Also fixed a couple of edge cases where reference arguments
were not being set despite a function returning success.

Fixes #1121

### DIFF
--- a/lib/model/CTokenListDataCategorizerBase.cc
+++ b/lib/model/CTokenListDataCategorizerBase.cc
@@ -200,6 +200,9 @@ bool CTokenListDataCategorizerBase::createReverseSearch(int categoryId,
             return false;
         }
 
+        // Ensure all reference arguments are set before returning true
+        wasCached = false;
+        maxMatchingLength = 0;
         return true;
     }
 
@@ -207,8 +210,8 @@ bool CTokenListDataCategorizerBase::createReverseSearch(int categoryId,
     maxMatchingLength = category.maxMatchingStringLen();
 
     // If we can retrieve cached reverse search terms we'll save a lot of time
-    if (category.cachedReverseSearch(part1, part2) == true) {
-        wasCached = true;
+    wasCached = category.cachedReverseSearch(part1, part2);
+    if (wasCached) {
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -200,15 +200,18 @@ bool CTokenListDataCategorizerBase::createReverseSearch(int categoryId,
             return false;
         }
 
+        // Ensure all reference arguments are set before returning true
+        wasCached = false;
+        maxMatchingLength = 0;
```
tveasey, line 205: `wasCached` and `maxMatchingLength` are unset in the earlier return path `if (m_ReverseSearchCreator == nullptr)`.

I guess they're unused if the return value is false, but it strikes me as safer to just initialise them like this at the very start of the function.

droberts195: Yes, the idea is that there are no guarantees on the reference variables if the return value is false.  Still, I'll move them to the beginning so future changes don't accidentally introduce a problem.

droberts195:
retest

Issue #1106, opened 2020-03-31
[ML] Memory estimates way too high for very simple analyses


droberts195:
Steps to reproduce:

1. Import the Iowa liquor sales dataset using the file data visualizer (ping me if you need the file)
2. Create a new regression data frame analytics job to analyze it, setting training percent to 5, dependent variable to `Sale (Dollars)` and excluding every variable _except_ `Bottle Volume (ml)` and `Store Number` from the analysis.  (So effectively we're predicting one number from two others on 5% of the 380000 rows, i.e. 19000 rows.)

<img width="868" alt="Screenshot 2020-03-31 at 13 05 05" src="https://user-images.githubusercontent.com/7405510/78025233-90f8ef80-7351-11ea-9be1-e99320f03961.png">

3. Start the analysis and wait for it to finish.
4. Look at the job details.  The memory limit recommended by the UI was around 1.2**GB**.  The actual memory required was less than 12MB.

<img width="1711" alt="Screenshot 2020-03-31 at 13 09 49" src="https://user-images.githubusercontent.com/7405510/78025368-d3223100-7351-11ea-8f42-1ab1bad6f95d.png">

Part of the problem here is https://github.com/elastic/kibana/issues/60496, because the memory estimate didn't get updated when I added the exclude fields.  However, a considerable part of the problem is in the C++ estimation code.  If I run the estimate in dev console using the final config it's still 25 times bigger than it needs to be:

```
POST _ml/data_frame/analytics/_explain
{
    "source": {
      "index": [
        "iowa"
      ],
      "query": {
        "match_all": {}
      }
    },
    "analysis": {
      "regression": {
        "dependent_variable": "Sale (Dollars)",
        "prediction_field_name": "Sale (Dollars)_prediction",
        "training_percent": 5
      }
    },
    "analyzed_fields": {
      "includes": [],
      "excludes": [
        "Address",
        "Bottles Sold",
        "Category",
        "Category Name",
        "City",
        "County",
        "County Boundaries of Iowa",
        "Iowa ZIP Code Tabulation Areas",
        "Item Description",
        "Item Number",
        "Pack",
        "State Bottle Cost",
        "State Bottle Retail",
        "Store Location",
        "Store Name",
        "Iowa Watersheds (HUC 10)",
        "Iowa Watershed Sub-Basins (HUC 08)",
        "County Number",
        "Invoice/Item Number",
        "US Counties",
        "Zip Code",
        "Vendor Name",
        "Vendor Number",
[...]

PR #1125, opened 2020-04-03
[ML] Eagerly discard node statistics for leaves which we will never split

tveasey:
This is a simple memory optimisation for regression and classification model training.

Since we limit the maximum tree size, we only need retain statistics for the maximum gain "max tree size" - "current tree size" leaves. Since we add one leaf statistic for each node we add to the tree we only ever need at most "maximum tree size" / 2 node statistics in total.

We use an upper bound for the leaf statistics memory estimates so this also indirectly helps improve our memory estimate accuracy and so helps with #1106. For example, this reduced estimates by around 25% testing.

### DIFF
--- a/lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
+++ b/lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
@@ -321,8 +321,8 @@ BOOST_AUTO_TEST_CASE(testRunBoostedTreeRegressionTraining) {
               << "ms");
 
     BOOST_TEST_REQUIRE(core::CProgramCounters::counter(
-                           counter_t::E_DFTPMEstimatedPeakMemoryUsage) < 6000000);
-    BOOST_TEST_REQUIRE(core::CProgramCounters::counter(counter_t::E_DFTPMPeakMemoryUsage) < 1500000);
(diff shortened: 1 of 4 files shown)


lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -321,8 +321,8 @@ BOOST_AUTO_TEST_CASE(testRunBoostedTreeRegressionTraining) {
               << "ms");
 
     BOOST_TEST_REQUIRE(core::CProgramCounters::counter(
-                           counter_t::E_DFTPMEstimatedPeakMemoryUsage) < 6000000);
-    BOOST_TEST_REQUIRE(core::CProgramCounters::counter(counter_t::E_DFTPMPeakMemoryUsage) < 1500000);
+                           counter_t::E_DFTPMEstimatedPeakMemoryUsage) < 4500000);
+    BOOST_TEST_REQUIRE(core::CProgramCounters::counter(counter_t::E_DFTPMPeakMemoryUsage) < 1600000);
```
tveasey, line 325: We're not using more memory; however, we now properly account for the memory used by container for the leaf statistics because we presize it before estimating its to memory usage.

PR #1126, opened 2020-04-06
[ML] Round up data frame analytics memory estimates to next MB

### LINKED ISSUE #1110: [ML] Switch data frame analytics memory estimate from KB to MB
For anomaly detection memory limits/estimates are done in terms of whole megabytes.

For data frame analytics the estimates are done in kilobytes and limits can be as granular as single bytes.

Originally estimates were going to be in tenths of megabytes, because the feeling was that many analyses would use less than a megabyte and rounding up would be too wasteful.  Since tenths of megabytes to not round nicely to byte values we decided to go with kilobytes instead.

However, as #1106 showed, even quite simple analyses are using ~10MB, so rounding up to the next megabyte would not be a major problem.

Rounding estimates to the nearest megabyte would avoid the excessive precision problem noted in elastic/elasticsearch#54506.  It would also improve consistency with anomaly detection.

We would still have to accept limits set in units other than megabytes, as jobs that have such limits will already exist.  However, we could nudge future jobs towards using whole numbers of megabytes for their memory limits by always returning estimates as whole numbers of megabytes.

/cc @peteharverson

droberts195:
Previously data frame analytics memory estimates were
rounded to the nearest kilobyte, but this results in
excessive precision for large analyses.  This changes
the estimates to always be reported in whole megabytes,
rounded up from the low level estimate.

Closes #1110
Closes elastic/elasticsearch#54506

### DIFF
--- a/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
+++ b/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
@@ -189,19 +189,19 @@ BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor0Rows) {
 }
 
 BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor1Row) {
-    testEstimateMemoryUsage(1, "4kB", "4kB", 0);
+    testEstimateMemoryUsage(1, "1mb", "1mb", 0);
 }
 
-BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor10Rows) {
-    testEstimateMemoryUsage(10, "12kB", "10kB", 0);
+BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor10000Rows) {
+    testEstimateMemoryUsage(10000, "5mb", "2mb", 0);
 }
 
-BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor100Rows) {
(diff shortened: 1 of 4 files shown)


droberts195:
retest

droberts195:
retest

droberts195:
retest

lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
```diff
@@ -189,19 +189,19 @@ BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor0Rows) {
...
 
-BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor100Rows) {
-    testEstimateMemoryUsage(100, "57kB", "35kB", 0);
+BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor100000Rows) {
+    testEstimateMemoryUsage(100000, "40mb", "9mb", 0);
 }
 
-BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor1000Rows) {
-    testEstimateMemoryUsage(1000, "403kB", "142kB", 0);
+BOOST_AUTO_TEST_CASE(testEstimateMemoryUsageFor1000000Rows) {
```
przemekwitek, line 203: nit: Test name is off by one `0`

PR #1127, opened 2020-04-07
[ML] Reduce differences in regression and classification results between target platforms

tveasey:
This addresses some of the underlying causes for variation in regression and classification results for different platforms: `std::log` and `std::exp` don't always agree in the last decimal place and our split order predicate wasn't a total order.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -33,173 +33,125 @@
 === Enhancements
 
 * The macOS build platform for the {ml} C++ code is now High Sierra running Xcode 10.1,
-  or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {
-    ml - pull}867[#867].)
+  or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {ml-pull}867[#867].)
 
 == {es} version 7.8.0
 
 === Enhancements
 
-* Speedup anomaly detection for the lat_long function. (See {
-    ml - pull}1102[#1102].)
+* Speedup anomaly detection for the lat_long function. (See {ml-pull}1102[#1102].)
 * Reduce CPU scheduling priority of native analysis processes to favor the ES JVM
-  when CPU is constrained. (See {
-    ml - pull}1109[#1109].)
+  when CPU is constrained. (See {ml-pull}1109[#1109].)
 * Take `training_percent` into account when estimating memory usage for classification and regression. 
-  (See {
-    ml - pull}1111[1111].)
-* Improve robustness of anomaly detection to bad input data. (See {
-    ml - pull}1114[#1114].)
+  (See {ml-pull}1111[1111].)
+* Support maximize minimum recall when assigning class labels for multiclass classification.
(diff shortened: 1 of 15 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -47,6 +47,8 @@
 * Improve robustness of anomaly detection to bad input data. (See {ml-pull}1114[#1114].)
 * Adds new `num_matches` and `preferred_to_categories` fields to category output.
   (See {ml-pull}1062[#1062])
+* Ensure classification and regression results agree for all our target operating systems.
+  (See {ml-pull}1127[1127].)
```
droberts195, line 51: The `1127` inside the square brackets should be preceded by `#`.

include/maths/CTools.h
```diff
@@ -668,6 +668,17 @@ class MATHS_EXPORT CTools : private core::CNonInstantiatable {
     //! Compute \f$x^2\f$.
     static double pow2(double x) { return x * x; }
 
+    //! Compute a value from \p x which will be stable across platforms.
+    static double stable(double x) {
+        return STABLE_EPS * std::floor(x / STABLE_EPS + 0.5);
```
droberts195, line 673: Could this cause problems if `x` is very large?  This is dividing by a number of the order of 1e-14 right?  So if `x` was 1e300, would anything go wrong?

Just to prove there are no nasty edge cases it would be good to add a unit test that exercises this function for these arguments:

* -1e305
* -1.0
* -1e-305
* 0.0
* 1e-305
* 1.0
* 1e305

tveasey: This is a good point. It won't be near overflowing given how it's used, but, in fact, we have some tools for dealing with this correctly already I should have used.

tveasey: Ok I just completely reworked this to drop the last bit in a way which doesn't rely on any standard library functions: using `frexp` and `ldexp` defeats the purpose. This implementation clearly doesn't overflow, so instead I added some tests for bitwise comparison of `stableLog` and `stableExp` for some random inputs. These would fail if they're different on any of our target platforms. See [7803c5c](https://github.com/elastic/ml-cpp/pull/1127/commits/7803c5c31fabb352f764b76b99885ad6f14382a2).

docs/CHANGELOG.asciidoc
tveasey: See [5f6500b](https://github.com/elastic/ml-cpp/pull/1127/commits/5f6500b278ec5c3b95dc83a25c51e9fb71231618).

include/maths/CTools.h
```diff
@@ -22,6 +22,7 @@
 #include <boost/static_assert.hpp>
 
 #include <array>
+#include <cinttypes>
```
droberts195, line 25: Maybe cstdint would be better - see https://stackoverflow.com/a/7597070/3790930

tveasey: See [5d99c3e](https://github.com/elastic/ml-cpp/pull/1127/commits/5d99c3eec3b8d38f4d6ec429346cfebecf62f416).

PR #1132, opened 2020-04-08
[ML] Native compilation and unit test for Linux on aarch64

droberts195:
This PR adds the ability to compile and unit test the code on
the aarch64 architecture.

It is NOT workable aarch64 support.  That would additionally
require the ability to cross compile for aarch64 from the
x86_64 release manager and CI VMs, and also changes to the
Gradle scripts that assemble the final artifacts.

However, the relative ease of getting the code to compile and
run on aarch64 suggests that supporting this platform is a
realistic goal.

The unit test changes highlight that aarch64 floating point
calculations produce slightly different results to the
equivalent x86_64 calculations.

### DIFF
--- a/mk/linux.mk
+++ b/mk/linux.mk
@@ -6,7 +6,8 @@
 
 OS=Linux
 
-CPP_PLATFORM_HOME=$(CPP_DISTRIBUTION_HOME)/platform/linux-x86_64
+HARDWARE_ARCH:=$(shell uname -m)
+CPP_PLATFORM_HOME=$(CPP_DISTRIBUTION_HOME)/platform/linux-$(HARDWARE_ARCH)
 
 CC=gcc
 CXX=g++ -std=gnu++14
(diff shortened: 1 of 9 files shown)


mk/linux.mk
```diff
@@ -43,21 +48,30 @@ RESOURCES_DIR=resources
...
+BOOSTPROGRAMOPTIONSLIBS=-lboost_program_options-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER)
+BOOSTTHREADLIBS=-lboost_thread-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER) -lboost_system-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER)
+BOOSTFILESYSTEMLIBS=-lboost_filesystem-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER) -lboost_system-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER)
+BOOSTDATETIMELIBS=-lboost_date_time-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER)
+BOOSTTESTLIBS=-lboost_unit_test_framework-gcc$(BOOSTGCCVER)-mt-$(BOOSTARCH)-$(BOOSTVER)
 RAPIDJSONINCLUDES=-isystem $(CPP_SRC_HOME)/3rd_party/rapidjson/include
+ifeq ($(HARDWARE_ARCH),x86_64)
 RAPIDJSONCPPFLAGS=-DRAPIDJSON_HAS_STDSTRING -DRAPIDJSON_SSE42
+else
+RAPIDJSONCPPFLAGS=-DRAPIDJSON_HAS_STDSTRING
```
tveasey, line 73: Why not use NEON optimisations?
```suggestion
RAPIDJSONCPPFLAGS=-DRAPIDJSON_HAS_STDSTRING -RAPIDJSON_NEON
```

mk/linux.mk
```diff
@@ -23,9 +24,13 @@ COVERAGE=--coverage
 endif
 endif
 
+ifeq ($(HARDWARE_ARCH),x86_64)
+SIMDFLAGS=-msse4.2 -mfpmath=sse
```
tveasey, line 28: Eigen [supports](http://eigen.tuxfamily.org/index.php?title=FAQ#Which_SIMD_instruction_sets_are_supported_by_Eigen.3F) NEON instruction sets. I think we should aim to build with these.

droberts195: The page you linked says:

> On 64-bit ARM, SIMD is enabled by default, you don't have to do anything extra.

mk/linux.mk
```diff
@@ -23,9 +24,15 @@ COVERAGE=--coverage
 endif
 endif
 
+ifeq ($(HARDWARE_ARCH),aarch64)
+ARCHCFLAGS=-march=armv8-a+crc+crypto
```
droberts195, line 28: It turns out that all aarch64 CPUs have SIMD instructions, so there's no need to tell the compiler it can use them - it just will.  The [Amazon Graviton](https://aws.amazon.com/ec2/graviton/) processors used in EC2 aarch64 instances also support the CRC and Crypto instructions, so I told the compiler it can use them.  These were added to the ARMv8 architecture fairly early on, so should not restrict usage too much.  I suspect any hardware that didn't support these instructions would be unable to run ML well anyway, e.g. a 5 year old Raspberry Pi.

mk/linux.mk
tveasey: Ok interesting. I'd only skimmed the code and I thought all the SIMD code was behind compile time flags. That's good in that case.

PR #1134, opened 2020-04-14
[7.x][ML] Native compilation and unit test for Linux on aarch64

droberts195:
This PR adds the ability to compile and unit test the code on
the aarch64 architecture.

It is NOT workable aarch64 support.  That would additionally
require the ability to cross compile for aarch64 from the
x86_64 release manager and CI VMs, and also changes to the
Gradle scripts that assemble the final artifacts.

However, the relative ease of getting the code to compile and
run on aarch64 suggests that supporting this platform is a
realistic goal.

The unit test changes highlight that aarch64 floating point
calculations produce slightly different results to the
equivalent x86_64 calculations.

Backport of #1132


droberts195:
retest

PR #1137, opened 2020-04-14
[ML] Fix periodic persistence of categorizer state

### LINKED ISSUE #1136: [ML] Categorizer state is not background persisted
The intention is that we persist both anomaly detector state and categorizer state in the background periodically while a job is running (every 3-4 hours by default).

However, since version 7.4.0 the categorizer state has not been persisted in the background.  Instead an error message like this is logged:

```
[2020-03-23T14:01:06,813][ERROR][org.elasticsearch.xpack.ml.process.logging.CppLogMessageHandler] [instance-0000000001] [my_job] [autodetect/15732] [CFieldDataTyper.cc@414] NULL persistence manager
```

The problem was introduced by the changes of #550.  Prior to that the categorizer only needed a pointer to the background persister when it was unchained, i.e. running as part of a dedicated categorizer program.  In #550 it was changed to use a new persistence manager class regardless of whether it was chained or not, but when chained (which is always the case in current shipped product) it was not passed a pointer to the persistence manager.  Hence in production it never does background persistence.  (The unit tests cannot detect this because the bug is in the `main()` function of a production program.  It would need an integration test that uses the complete program to find the problem.)

droberts195:
Periodic persistence of categorizer state has not
worked since version 7.4.0.  This change makes it work
again.

Additionally, the persistence manager pointer argument
to the categorizer and anomaly detector classes is no
longer defaulted.  It can still be null, because it's
a major pain for unit tests to have to pass a real
persistence manager, but at least this must be
explicitly specified now, hopefully prompting anyone
who constructs such an object in production code to
pass a non-null persistence manager.

FIxes #1136

### DIFF
--- a/bin/autodetect/Main.cc
+++ b/bin/autodetect/Main.cc
@@ -254,11 +254,11 @@ int main(int argc, char** argv) {
     }
 
     // The anomaly job knows how to detect anomalies
-    ml::api::CAnomalyJob job(jobId, limits, fieldConfig, modelConfig,
-                             wrappedOutputStream, persistenceManager.get(),
+    ml::api::CAnomalyJob job(jobId, limits, fieldConfig, modelConfig, wrappedOutputStream,
                              std::bind(&ml::api::CModelSnapshotJsonWriter::write,
                                        &modelSnapshotWriter, std::placeholders::_1),
-                             maxQuantileInterval, timeField, timeFormat, maxAnomalyRecords);
+                             persistenceManager.get(), maxQuantileInterval,
+                             timeField, timeFormat, maxAnomalyRecords);
 
     if (!quantilesStateFile.empty()) {
         if (job.initNormalizer(quantilesStateFile) == false) {
--- a/bin/autodetect/Main.cc
+++ b/bin/autodetect/Main.cc
@@ -227,7 +227,7 @@ int main(int argc, char** argv) {
     }
 
     using TPersistenceManagerUPtr = std::unique_ptr<ml::api::CPersistenceManager>;
-    const TPersistenceManagerUPtr periodicPersister{
+    const TPersistenceManagerUPtr persistenceManager{
         [persistInterval, isPersistInForeground, &persister,
          &bucketPersistInterval]() -> TPersistenceManagerUPtr {
             if (persistInterval >= 0 || bucketPersistInterval > 0) {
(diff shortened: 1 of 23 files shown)


bin/autodetect/Main.cc
```diff
@@ -279,19 +279,20 @@ int main(int argc, char** argv) {
 
     // The categorizer knows how to assign categories to records
     ml::api::CFieldDataCategorizer categorizer(jobId, fieldConfig, limits, outputChainer,
-                                               fieldDataCategorizerOutputWriter);
+                                               fieldDataCategorizerOutputWriter,
+                                               persistenceManager.get());
```
droberts195, line 283: This line is the actual fix.  All the other changes relate to trying to make clearer that this extra argument was not optional in production.

An automated CI test to prove this works is tricky as it would need to wait for a background persistence to trigger and complete.  It is probably better that a test for the presence of categorizer state documents be added to the long running QA tests.

include/api/CAnomalyJob.h
```diff
@@ -144,12 +143,12 @@ class API_EXPORT CAnomalyJob : public CDataProcessor {
                 CFieldConfig& fieldConfig,
                 model::CAnomalyDetectorModelConfig& modelConfig,
                 core::CJsonOutputStreamWrapper& outputBuffer,
+                CPersistenceManager* persistenceManager,
```
tveasey, line 146: Definitely +1 to making this non-optional.

Another option, which would make the error impossible in production code, would be make this constructor protected and only expose a version taking the persistence manager by reference. We'd then need a `class CTestAnomalyJob : public api::CAnomalyJob` which sets it to null.

Since it is unlikely that we'll have much churn around this code, I'm happy for you to decide if this is worthwhile.

droberts195: Unfortunately I don't think the production code can pass the persistence manager by reference as there is an obscure option to never persist state:

https://github.com/elastic/elasticsearch/blob/0f2f00a20a6eae5b1fa9931f3b300746eb28e1c6/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/process/autodetect/AutodetectBuilder.java#L218-L219

However, good idea about having a derived class for tests.  I will make _all_ the arguments mandatory in production code and introduce a test class that defaults the same ones that were defaulted previously.  Then I don't need to change the argument order which will make the change cleaner.

droberts195:
retest

bin/autodetect/Main.cc
droberts195: > An automated CI test to prove this works is tricky as it would need to wait for a background persistence to trigger and complete.

This turned out to be less tricky than I thought, because we request a background persist when a datafeed that is doing both lookback and real-time transitions from lookback to real-time.  I have added a test that checks that categorization state is persisted in this situation in elastic/elasticsearch#55243 and pleasingly it fails before this C++ PR is merged but succeeds if run against C++ compiled locally.

PR #1139, opened 2020-04-15
[ML] Normalize hardware architecture from Java

droberts195:
Some versions of Java report amd64 and i386 where most tools
would use x86_64 and x86.  These need to be normalized.

This should have been in #1135.  This PR is master only as for the
backport of #1135 I can incorporate this change before merging.


droberts195:
retest

droberts195:
retest

PR #1141, opened 2020-04-15
[7.7][ML] Fix periodic persistence of categorizer state

droberts195:
Periodic persistence of categorizer state has not
worked since version 7.4.0.  This change makes it work
again.

Additionally, the persistence manager pointer argument
to the categorizer and anomaly detector classes is no
longer defaulted.  It can still be null, because it's
a major pain for unit tests to have to pass a real
persistence manager, but at least this must be
explicitly specified now, hopefully prompting anyone
who constructs such an object in production code to
pass a non-null persistence manager.

Backport of #1137


droberts195:
retest

PR #1142, opened 2020-04-15
[ML] Ensure the performance critical data are 16 byte aligned for data frame analyses

tveasey:
Eigen uses explicit intrinsic instructions for various operations. Currently, we enable SSE 4.2 (although should perhaps consider AVX as well longer term). These benefit considerably from using 16 byte alignment for the memory backing the vector/matrix. Eigen ensures this for memory it allocates itself, but in a number of performance critical pieces of code we use mapped matrices and vectors to cutdown on allocations (see [here](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html) for documentation of this type). It is possible to supply the alignment to the map class if this is known, but before this change we were supplying unaligned here. So, loads/stores were not optimised.

This change extends to support alignments up to 32 bytes and chooses 16 byte alignment by default. There are two areas where I now have to manage alignment as a result: in the data frame and in the accumulators of the loss function gradients. The high-level strategy is as follows:
1. Use the `Eigen::aligned_allocator` to ensure that the start of vector storage are aligned to 32 bytes. (Note that it is cleaner to manage the alignment of Eigen allocations globally and I've selected 32 bytes because this is more future proof.)
2. Insert pads into the vectors so that we maintain the alignment of addresses to the start of the rows and the start of the accumulated loss derivatives.

The pads are calculated by rounding up the capacity of the row to a multiple of the alignment and by rounding up the start position of derivatives as necessary.

I've tested this across of a range of benchmark sets on my i9 laptop. The speed up is fairly stable and around a 15% mean improvement for a range of regression and binary classification tasks. This does not change the results.


lib/core/CDataFrame.cc
```diff
@@ -562,7 +597,11 @@ CDataFrame::sequentialApplyToAllRows(std::size_t beginRows,
         break;
     }
 
-    return {{std::move(func)}, true};
+    TRowFuncVec funcs;
+    funcs.reserve(1);
+    funcs.emplace_back(std::move(func));
+
+    return TRowFuncVecBoolPr{std::move(funcs), true};
```
droberts195, line 604: It would be worth adding a comment about why it's been done like this so that somebody doesn't simplify it in the future.

mk/linux.mk
```diff
@@ -75,7 +75,7 @@ else
 RAPIDJSONCPPFLAGS=-DRAPIDJSON_HAS_STDSTRING -DRAPIDJSON_SSE42
 endif
 EIGENINCLUDES=-isystem $(CPP_SRC_HOME)/3rd_party/eigen
-EIGENCPPFLAGS=-DEIGEN_MPL2_ONLY
+EIGENCPPFLAGS=-DEIGEN_MPL2_ONLY -DEIGEN_MAX_ALIGN_BYTES=32
```
droberts195, line 78: Please merge in latest master and either add this to `mk/linux_cross_compile_linux.mk` too or else add a comment saying it's a deliberate difference from the flags used on x86_64.

tveasey:
> The default malloc on 64 bit Linux and Windows already returns 16 byte aligned memory, so these changes will only help for padding within large arrays, not for the start of the arrays.

Agreed and I haven't tested this yet. However, it is a tiny proportion, but more importantly, the key is to tell Eigen that the memory is aligned, via the new alignment template parameter. If you specify unaligned, the default, it won't exploit the fact is aligned.

mk/linux.mk
tveasey: Done in [c405438](https://github.com/elastic/ml-cpp/pull/1142/commits/c405438d6cdcb4eaa49eec80eebd7e85a3881a8a).

lib/core/CDataFrame.cc
tveasey: I added an explanation in this [06f7e07](https://github.com/elastic/ml-cpp/pull/1142/commits/06f7e07f4b2cbcd6cc0553c381e4f91d116d12e2) commit.

PR #1146, opened 2020-04-16
[ML] Fix macOS sandbox rules

droberts195:
Two rules were missing:

1. We were blocking our own SIGIO signals used to wake up blocking
   file reads.  (This is quite nasty as this functionality is only
   used in edge cases.)
2. We were blocking the setpriority() calls made by nice().

I added the (debug deny) rule so that any blocked calls get logged
to the system console.  This makes it easier to see which rules are
required but missing.

I also changed the code style to match what our current style guide
says.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -63,9 +63,9 @@
 
 === Bug Fixes
 
-* Fixed background persistence of categorizer state. (See {ml-pull}1137[#1137],
+* Fixed background persistence of categorizer state (See {ml-pull}1137[#1137],
   issue: {ml-issue}1136[#1136].)
-* Fixed macOS sandbox rules. (See {ml-pull}1146[#1146].)
+
 
 == {es} version 7.7.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -63,9 +63,9 @@
 
 === Bug Fixes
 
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -63,9 +63,9 @@
 
 === Bug Fixes
 
-* Fixed background persistence of categorizer state (See {ml-pull}1137[#1137],
+* Fixed background persistence of categorizer state. (See {ml-pull}1137[#1137],
   issue: {ml-issue}1136[#1136].)
-
+* Fixed macOS sandbox rules. (See {ml-pull}1146[#1146].)
```
davidkyle, line 68: Nit, a little more detail can't hurt

Update macOS sandbox rules to allow calling `nice` . (See {ml-pull}1146[#1146].)

droberts195: I was just about to say that the other change, to allow the signal used to cancel blocking IO calls, was more important.  If that is disallowed then it could cause the controller to hang indefinitely if the JVM doesn't connect to it (e.g. when ML is disabled) and then the JVM crashes rather than exiting gracefully.  However, then I realised that the controller doesn't install the sandbox at all, so doesn't suffer that problem.

Given that, I don't think this needs backporting to 7.7 after all and doesn't need separately release noting because it should have just been part of the CPU priority reduction change that's already release noted for 7.8.

So I think I'll mark this a `>non-issue` and remove the release note altogether.

droberts195:
retest

PR #1156, opened 2020-04-20
[ML] Fix memory usage estimation for vectors with a custom allocator

tveasey:
#1142 showed up that we weren't accounting for memory used by vector with a custom allocator. This is a pre-existing bug in memory estimation, but the aligned allocator was the first code to expose it. Since this is not released I've marked this as a non-issue.

### DIFF
--- a/include/core/CMemory.h
+++ b/include/core/CMemory.h
@@ -327,8 +327,8 @@ public:
     }
 
     //! Overload for std::vector.
-    template<typename T>
-    static std::size_t dynamicSize(const std::vector<T>& t) {
+    template<typename T, typename A>
+    static std::size_t dynamicSize(const std::vector<T, A>& t) {
         std::size_t mem = 0;
         if (!memory_detail::SDynamicSizeAlwaysZero<T>::value()) {
             for (auto i = t.begin(); i != t.end(); ++i) {
@@ -781,9 +781,9 @@ public:
     }
 
     //! Overload for std::vector.
-    template<typename T>
+    template<typename T, typename A>
     static void dynamicSize(const char* name,
-                            const std::vector<T>& t,
+                            const std::vector<T, A>& t,
                             const CMemoryUsage::TMemoryUsagePtr& mem) {
         std::string componentName(name);
 
--- a/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
@@ -162,9 +162,9 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
     }
 
     if (featureImportance != nullptr) {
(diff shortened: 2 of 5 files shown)


lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -162,9 +162,9 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
     }
 
     if (featureImportance != nullptr) {
+        int numberClasses{static_cast<int>(classValues.size())};
```
droberts195, line 165: Historically we always used `size_t` to iterate container indices.  Why are we switching to `int`?

include/core/CMemory.h
```diff
@@ -327,8 +327,8 @@ class CORE_EXPORT CMemory : private CNonInstantiatable {
     }
 
     //! Overload for std::vector.
-    template<typename T>
-    static std::size_t dynamicSize(const std::vector<T>& t) {
+    template<typename T, typename A>
+    static std::size_t dynamicSize(const std::vector<T, A>& t) {
```
droberts195, line 331: It's funny that this was missed for `vector` but none of the other containers.

lib/core/unittest/CDataFrameTest.cc
```diff
@@ -429,7 +429,7 @@ BOOST_FIXTURE_TEST_CASE(testMemoryUsage, CTestFixture) {
     //   1) 1075 + 4 times the root directory length bytes for on disk, and
     //   2) data size + doc ids size + 900 byte overhead in main memory.
     std::size_t maximumMemory[]{1075 + 4 * rootDirectory.length(),
-                                rows * (cols + 1) * 4 + 900};
+                                rows * (4 * ((cols + 3) / 4) + 1) * 4 + 900};
```
droberts195, line 432: It might be worth adding a comment here to say the rounding up is due to alignment padding.  In a couple of years time it won't be clear to everyone why this formula is this complex.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
tveasey: Eigen's index type is unsigned, so we use `int` when we're iterating over components of an Eigen vector.

include/core/CMemory.h
tveasey: I know! I think all the other template arguments for other containers got added after the initial implementation. For some reason it was just missed from vector.

lib/core/unittest/CDataFrameTest.cc
tveasey: Added in [d0155db](https://github.com/elastic/ml-cpp/pull/1156/commits/d0155db8ceb24d9d46c4e6214df44551051c5add).

PR #1159, opened 2020-04-22
[ML] Report analytics progress as phase_progress

dimitris-athanasiou:
Changes the way we report progress for data frame analytics
from a plain integer `progress_percent` to an object called
`phase_progress` which contains an integer `progress_percent`
and the `phase` name. In this commit we only have one phase
called `analyzing` but this change allows us to introduce new
phases.

### DIFF
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -42,7 +42,7 @@ const std::string PHASE_PROGRESS{"phase_progress"};
 const std::string PHASE{"phase"};
 const std::string PROGRESS_PERCENT{"progress_percent"};
 
-// Temporarily the only progress phase
+// Progress phases
 const std::string ANALYZING{"analyzing"};
 
 // Row result fields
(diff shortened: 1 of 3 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -35,9 +35,16 @@ const char FINISHED_DATA_CONTROL_MESSAGE_FIELD_VALUE{'$'};
...
 const std::string ROW_RESULTS{"row_results"};
-const std::string PROGRESS_PERCENT{"progress_percent"};
 const std::string ANALYZER_STATE{"analyzer_state"};
 
+// Phase progress
+const std::string PHASE_PROGRESS{"phase_progress"};
+const std::string PHASE{"phase"};
+const std::string PROGRESS_PERCENT{"progress_percent"};
+
+// Temporarily the only progress phase
```
tveasey, line 45: How about just...
```suggestion
// Progress phases
```
then the comment won't immediately go out-of-date.

PR #1167, opened 2020-04-24
[ML] Limit categorization memory usage

### LINKED ISSUE #1130: [ML] Categorization should take notice of hard_limit memory status
If a job's memory status goes to `hard_limit` then we stop modelling new entities in anomaly detection.  However, categorization can still create new categories.  If there are many new categories then this can cause a very significant overrun of the configured memory limit.

Some possibilities:

* When a job is in `hard_limit` status no new categories should be created.  The input document that could not be categorized should be discarded as it cannot take part in anomaly detection without a category.  A new statistic in the model size stats should be incremented to record the number of documents discarded for this reason.
* When a job is in `soft_limit` status, we stop recording examples for the category.

droberts195:
Anomaly detection jobs have a model_memory_limit setting.  This is
supposed to restrict the amount of memory the job can use, however,
in the past the limit only applied to anomaly detection and not to
categorization.

This change applies memory limiting to categorization as follows:

- When a job is in hard_limit status no new categories will be
  created.  The input document that could not be categorized is
  discarded as it cannot take part in anomaly detection without a
  category.  The failed_category_count statistic is incremented
  each time this happens.
- When a job is in soft_limit status, we stop recording examples
  for the category.

Fixes #1130

### DIFF
--- a/include/model/CDataCategorizer.h
+++ b/include/model/CDataCategorizer.h
@@ -53,6 +53,16 @@ public:
     using TPersistFunc = std::function<void(core::CStatePersistInserter&)>;
     using TIntVec = std::vector<int>;
 
+    //! A soft categorization failure means downstream components can continue,
+    //! by considering the input record to be in some "uncategorizable"
+    //! category.
+    static const int SOFT_CATEGORIZATION_FAILURE_ERROR;
+
+    //! A hard categorization failure means processing of the input record must
+    //! cease.  (This is generally used to prevent excessive memory
+    //! accumulation.)
+    static const int HARD_CATEGORIZATION_FAILURE_ERROR;
+
 public:
     CDataCategorizer(CLimits& limits, const std::string& fieldName);
 
@@ -142,6 +152,10 @@ public:
     //! Once called, the category is marked as unchanged, until the category
     //! changes again.
     virtual bool categoryChangedAndReset(int categoryId) = 0;
+
+    //! Is it permissable to create new categories?  New categories are
+    //! not permitted when memory use has exceeded the limit.
+    bool areNewCategoriesAllowed();
 
 protected:
     //! Used if no fields are supplied to the computeCategory() method.
--- a/lib/api/CFieldDataCategorizer.cc
+++ b/lib/api/CFieldDataCategorizer.cc
@@ -48,12 +48,9 @@ CFieldDataCategorizer::CFieldDataCategorizer(const std::string& jobId,
                                              COutputHandler& outputHandler,
                                              CJsonOutputWriter& jsonOutputWriter,
                                              CPersistenceManager* persistenceManager)
-    : m_JobId{jobId}, m_Limits{limits}, m_OutputHandler{outputHandler},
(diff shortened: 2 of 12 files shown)


droberts195:
retest

lib/api/CFieldDataCategorizer.cc
```diff
@@ -30,30 +30,30 @@ namespace api {
...
-      m_NumRecordsHandled(0), m_OutputFieldCategory(m_Overrides[MLCATEGORY_NAME]),
-      m_MaxMatchingLength(0), m_JsonOutputWriter(jsonOutputWriter),
-      m_CategorizationFieldName(config.categorizationFieldName()),
-      m_CategorizationFilter(), m_PersistenceManager(persistenceManager) {
+    : m_JobId{jobId}, m_Limits{limits}, m_OutputHandler{outputHandler},
+      m_ExtraFieldNames{1, MLCATEGORY_NAME}, m_WriteFieldNames{true},
+      m_NumRecordsHandled{0}, m_OutputFieldCategory{m_Overrides[MLCATEGORY_NAME]},
+      m_MaxMatchingLength{0}, m_JsonOutputWriter{jsonOutputWriter},
+      m_CategorizationFieldName{config.categorizationFieldName()},
+      m_CategorizationFilter{}, m_PersistenceManager{persistenceManager} {
```
tveasey, line 56: nit: there doesn't seem to be any need to default initialise `m_CategorizationFilter` here. Also, how about initialising builtin types `m_NumRecordsHandled`, etc in the class body?

include/model/CDataCategorizer.h
```diff
@@ -53,6 +53,16 @@ class MODEL_EXPORT CDataCategorizer : public CMonitoredResource {
     using TPersistFunc = std::function<void(core::CStatePersistInserter&)>;
     using TIntVec = std::vector<int>;
 
+    //! A soft categorization failure means downstream components can continue,
+    //! by considering the input record to be in some "uncategorizable"
+    //! category.
+    static const int SOFT_CATEGORIZATION_FAILURE_ERROR;
```
tveasey, line 59: Good idea!

lib/model/CDataCategorizer.cc
```diff
@@ -56,13 +58,23 @@ std::size_t CDataCategorizer::memoryUsage() const {
 }
 
 bool CDataCategorizer::addExample(int categoryId, const std::string& example) {
+    // Don't add examples if we're memory-limited
+    if (m_Limits.resourceMonitor().getMemoryStatus() != model_t::E_MemoryStatusOk) {
```
tveasey, line 62: Is there any reason not to use `this->areNewCategoriesAllowed()` here? I couldn't think of one.

droberts195: `areNewCategoriesAllowed()` returns `false` when the job is in `hard_limit`.  But we stop adding examples as soon as the job goes to `soft_limit`.  It's in the spec for memory limiting in #1130 and the PR description, but I will expand the comment in the code to say this.

PR #1169, opened 2020-04-25
[ML] Get Eigen dynamically instead of committing to repo

droberts195:
Up to now we have had Eigen committed to the repo.  This
change alters the way we obtain Eigen to dynamically cloning
the Eigen repo when necessary.  We do a shallow clone of
just the tagged commit we require to reduce the download
size.  This functionality requires at least Git version 1.8.3, but
that's pretty old so shouldn't be a problem for anyone.

This change also updates the 3rd party information for Eigen
to reflect the fact that it's now hosted on GitLab instead of
BitBucket (see http://eigen.tuxfamily.org/index.php?title=News:Migration_to_GitLab.com_scheduled_on_the_December_4th).


3rd_party/licenses/eigen-INFO.csv
```diff
@@ -1,2 +1,2 @@
 name,version,revision,url,license,copyright
-Eigen,3.3.7,323c052e17310984a40dda9426e0435bed7178fa,http://eigen.tuxfamily.org,MPL-2.0,
+Eigen,3.3.7,21ae2afd4edaa1b69782c67a54182d34efe43f9c,http://eigen.tuxfamily.org,MPL-2.0,
```
droberts195, line 2: The reason the commit hash changed is that Eigen switched from Mercurial to Git last December.  We are still using exactly the same code as before.

droberts195:
Note to reviewers: the first commit just removes all the Eigen files, and is huge.  The changes that are worth reviewing in detail are in the second commit.  So it would be best to just look at the second commit when reviewing.

Issue #1015, opened 2020-02-19
[ML] Upgrade compilers so that we can use C++17 in the 8.x timeframe


droberts195:
Currently in our codebase we are able to use C++11 in the 6.x branches and C++14 in the 7.x branches.

It would be good if we can use C++17 in the 8.x branches (once they are eventually created).  Enabling this will require compiler upgrades on our release build environments for all platforms:

- [x] macOS - requires Xcode 10 or above - #867
- [x] Linux - requires gcc 7 (partial) or 9 (complete) - #1170
- [ ] Windows - requires Visual Studio 2019 - #1352

When the compilers are upgraded we will also take the opportunity to upgrade libxml2 from version 2.9.4 to version 2.9.7 with minimal extra effort.

droberts195:
#834 had to be backed out as the release manager VM had some components missing.  #1352 will be the replacement once the release manager VM is updated.

PR #1170, opened 2020-04-27
[ML] Upgrade to gcc 9.3 for Linux compilation

droberts195:
We will build version 8.x of the ML C++ using gcc 9.3 rather
than gcc 7.5. This will eventually facilitate use of C++17
(after separate PRs to upgrade the Mac and Windows compilers
have also been merged).

This also moves the build platform for Linux builds we ship
from RHEL/CentOS 6 to RHEL/CentOS 7 to match the supported
platforms change for Elasticsearch 8.x.  This permits a
couple of tidyups:

1. We no longer need the Boost.Math workaround to switch off
   its incredibly slow mechanism for avoiding a glibc bug
   that was fixed in glibc 2.14.  (We worked around it in a
   different way.)
2. We can get more seccomp constants from OS headers instead
   of duplicating them.

Since this change involves rebuilding all dependencies, the
libxml2 dependency is also upgraded from version 2.9.4 to
version 2.9.7.

Relates #1015


droberts195:
retest

droberts195:
One of the latest failures is:

```
[Exception] - critical check "[0, 10, 20, 30, 40, 50, 60, 70, 80]" == core::CContainerPrinter::print(task.s_TenPercentProgressPoints) has failed [[0, 10, 20, 30, 40, 50, 60, 70, 80] != [0, 10, 20, 30, 40, 50, 60, 70, 80, 90]]
 == [File] - CBoostedTreeTest.cc
 == [Line] -1437
```

I am thinking that assertion is going to be a thorn in our sides forever unless it can be adjusted to be more deterministic.  We could make it accept any progress level above 70, but if it's accepting such a wide range of possible results there's a question of what it's actually testing.

The other failure is:

```
[Exception] - difference{0.178941} between 0.0{0} and modelBias[i][0]{-0.17894063818585687} exceeds 0.16
 == [File] - CBoostedTreeTest.cc
 == [Line] -355
```

This one may be partially a symptom of the gcc 9.3 upgrade, as the value has never been this far out on gcc 7.5.  However, the constant multiplier that's currently 8.0 has already been increased twice (7.0 -> 7.3 -> 8.0), so this may be a sign of some algorithm that's vulnerable to very slight changes in floating point behaviour either between platforms or between compiler versions.  We could bump it up to 9.0 to get the test passing again, but it would be good for someone who understands what's going on to take a closer look.

droberts195:
retest

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1432,22 +1432,22 @@ BOOST_AUTO_TEST_CASE(testProgressMonitoring) {
...
-                if (task.s_TenPercentProgressPoints.size() != 11 ||
-                    task.s_TenPercentProgressPoints.front() != 0 ||
-                    task.s_TenPercentProgressPoints.back() != 100) {
-                    BOOST_REQUIRE_EQUAL("[0, 10, 20, 30, 40, 50, 60, 70, 80, 90]",
-                                        core::CContainerPrinter::print(task.s_TenPercentProgressPoints));
-                }
+                // Just assert that training started as there is no guarantee on how far it's got
+                BOOST_TEST_REQUIRE(task.s_TenPercentProgressPoints.empty() == false);
+                BOOST_TEST_REQUIRE("[0]" != core::CContainerPrinter::print(
+                                                task.s_TenPercentProgressPoints));
```
tveasey, line 1450: This is interesting. Did you see cases where progress where we got [0, 10]? For final training we know how many trees we plan to use so it is surprising that this is very variable.

droberts195: No, I only ever saw 70 to 100.  But if this is at the mercy of thread scheduling on the test machine then all we can safely say is that it's started.  Asserting 70 to 100 will probably be OK for a few years and then we'll start testing on a machine with different timing characteristics and we'll get a 50 or 60.

tveasey: This test isn't actually affected by threading. The reason it can finish with less than 100% progress is we estimate the number of steps upfront and this estimate can be too large. That said this need not hold this PR up further. I'll maybe revisit these checks in a different change.

droberts195: Oh, sorry, I misunderstood what the problem was.

PR #1177, opened 2020-04-28
[ML] Reduce false positives for the periodic component test for anomaly detection

tveasey:
While investigating changes in our QA suite from #1158, I noticed, particularly for sparse data, the periodicity test generates false positives.

These turn out to be because of how we account for piecewise scaling of the periodic component. We allow piecewise linear scaling of the periodic component in the time window we test. However, we didn't directly impose a limit on the minimum number of repeats we require per scale. This change requires (as with unscaled periodicity testing) that we have at least three repeats *per* scale selected. I also took the opportunity to slightly rework the combination of soft conditions we check to make it clearer what is happening. 

The downside of detecting periodicity when none is present is that the model is generally less robust to anomalous behaviour. Also, the periodic component tends to be unstable. Both these effects tend to lead to poor estimation of anomaly severity: in particular, less significant anomalies are scored more highly than more significant anomalies by chance because of model instability.

### DIFF
--- a/lib/maths/CPeriodicityHypothesisTests.cc
+++ b/lib/maths/CPeriodicityHypothesisTests.cc
@@ -2095,7 +2095,7 @@ bool CPeriodicityHypothesisTests::testPartition(const TTimeTimePr2Vec& partition
         return CBasicStatistics::mean(result);
     };
 
-    CSmoothCondition condition;
+    CSmoothCondition correlationCondition;
     double R{-1.0};
 
     TFloatMeanAccumulatorVec partitionValues;
@@ -2121,9 +2121,10 @@ bool CPeriodicityHypothesisTests::testPartition(const TTimeTimePr2Vec& partition
                                      MINIMUM_REPEATS_TO_TEST_VARIANCE};
         LOG_TRACE(<< "  mean repeats per segment = " << meanRepeatsPerSegment);
 
-        condition = std::max(
-            condition, softGreaterThan(R, stats.s_AutocorrelationThreshold, 0.1) &&
-                           softGreaterThan(meanRepeatsPerSegment, 1.0, 0.2));
+        correlationCondition =
+            std::max(correlationCondition,
+                     softGreaterThan(R, stats.s_AutocorrelationThreshold, 0.1) &&
+                         softGreaterThan(meanRepeatsPerSegment, 1.0, 0.2));
         R = std::max(R, RW);
     }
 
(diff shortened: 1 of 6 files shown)


lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -1231,24 +1263,25 @@ CPeriodicityHypothesisTests::best(const TNestedHypothesesVec& hypotheses) const
...
-                     CTools::logisticFunction(summary.s_ScaleSegments, 0.3, 0.0, -1.0)};
+                                          50.0 - CONFIDENCE_INTERVAL / 2.0)};
+            v = v == summary.s_VarianceThreshold * vmin[0]
+                    ? 1.0
+                    : v / summary.s_VarianceThreshold / vmin[0];
+            double p{(softLessThan(v, 1.0, 0.2) &&
+                      softGreaterThan(summary.s_R, summary.s_AutocorrelationThreshold, 0.1) &&
+                      softGreaterThan(summary.s_DF / DFmin[0], 1.0, 0.2) &&
+                      softLessThan(summary.s_TrendSegments, 0.0, 0.3) &&
+                      softLessThan(summary.s_ScaleSegments, 0.0, 0.3))
```
valeriy42, line 1283: nice :+1:

lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -69,6 +69,38 @@ const std::size_t MINIMUM_REPEATS_TO_TEST_AMPLITUDE{4};
...
+class CSmoothCondition {
+public:
+    explicit CSmoothCondition(double value = 0.0) : m_Value{value} {}
+
+    operator bool() const { return m_Value > 1.0; }
+    bool operator<(const CSmoothCondition& rhs) const {
+        return m_Value < rhs.m_Value;
+    }
+
+    double p() const { return m_Value; }
```
valeriy42, line 83: what does `p` mean? Is this the p-value of a hypothesis test? I would call it `pValue` of `probability` then.

tveasey: It isn't a p-value (in the conventional sense) it is more how close the condition is to passing or not. The code called it p in the past which is why I called it p here, but it isn't a good name. This is really a fuzzy logical operation with AND represented by multiplication (except it there is an implicit threshold for being actually true: a logical operation just holds if all constituent conditions' truth values are 0.5, so the value isn't normalised as it should). Let me come up with something better.

tveasey: I went ahead and renamed some stuff in this area. I think this is better. See [cbf9d37](https://github.com/elastic/ml-cpp/pull/1177/commits/cbf9d3775e8dfd01cc16ed06fd8f6cdbbe4af372).

valeriy42: Indeed, stating it explicitly that you are using fuzzy logic makes is much clearer :+1:

tveasey:
retest

PR #1183, opened 2020-05-02
[ML] Fix restore for boosted tree and improve tests

tveasey:
Various state wasn't getting restored by `CBoostedTreeImpl` and we weren't initialising the example weights. We also weren't properly testing persist and restore:
1. We weren't checking persist after running training in `CBoostedTreeTest:: testPersistRestore` so not all state was present,
2. We weren't checking training produced the same values after restore (because we weren't actually restoring from the saved state) in `CDataFrameAnalyzerTrainingTest ::testOneRunOfBoostedTreeTrainingWithStateRecovery`.

I also took the opportunity to switch over the `std::exp` in the MSLE loss to their stable equivalents since it also improved stability of results after restore. Restoring from training state is currently broken and there isn't a good path to fixing this so I just discard old state.


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1399,21 +1412,11 @@ void CBoostedTreeImpl::acceptPersistInserter(core::CStatePersistInserter& insert
...
     core::CPersistUtils::persist(TRAINING_ROW_MASKS_TAG, m_TrainingRowMasks, inserter);
-    core::CPersistUtils::persist(BEST_FOREST_TAG, m_BestForest, inserter);
-    core::CPersistUtils::persist(BEST_HYPERPARAMETERS_TAG, m_BestHyperparameters, inserter);
-    core::CPersistUtils::persist(ETA_OVERRIDE_TAG, m_EtaOverride, inserter);
-    core::CPersistUtils::persist(FEATURE_BAG_FRACTION_OVERRIDE_TAG,
-                                 m_FeatureBagFractionOverride, inserter);
-    core::CPersistUtils::persist(MAXIMUM_NUMBER_TREES_OVERRIDE_TAG,
-                                 m_MaximumNumberTreesOverride, inserter);
-    inserter.insertValue(LOSS_TAG, m_Loss->name());
-    core::CPersistUtils::persist(NUMBER_TOP_SHAP_VALUES_TAG, m_NumberTopShapValues, inserter);
```
tveasey, line 1412: I've made these fully alphabetic because it is easier to spot missing values that way.

tveasey:
retest

tveasey:
Reapplying to master.

PR #1185, opened 2020-05-04
[ML] Fix override of maximum permitted trees for classification and regression

tveasey:
This was being reset when searching for the best learn rate. We were also missing a unit test that user overrides for hyperparameters are respected.


tveasey:
retest

PR #1193, opened 2020-05-05
[ML] Reset memory status from soft limit to ok below threshold

### LINKED ISSUE #1131: [ML] Discrepancy in resetting of soft_limit status
If an anomaly detection job is approaching its memory limit then we start pruning old models more aggressively and the memory status is set to `soft_limit`.  If the memory usage then reduces such that aggressive pruning is no longer required then we do not immediately reset the memory status to `ok`.  However, if the job is closed and reopened, or if it switches nodes because the node it was running on leaves the cluster then the memory status _is_ reset to `ok` if pruning is no longer required.

This discrepancy should be addressed.  If aggressive pruning is no longer required then the job's memory status should be reset to `ok` without requiring a process restart.

droberts195:
We start pruning models when the memory used goes above the
prune threshold, and set the job's memory status to soft
limit at the same time.  If you close and reopen the job and
memory usage is lower than the prune threshold then we don't
prune and the memory limit is ok.  Prior to this change there
was a discrepancy where if you didn't close and reopen the
job but memory usage dropped below the prune threshold then
we continued to prune and left the memory status in soft
limit.

This change makes the case of not closing and reopening the
job more consistent.  If memory usage drops below the prune
threshold and the prune window expands to the default level
then we stop pruning and set the memory status back to ok.

Closes #1131

### DIFF
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -127,12 +127,18 @@ public:
     //! We are being told that aggressive pruning has taken place
     //! to avoid hitting the resource limit, and we should report this
     //! to the user when we can
-    void acceptPruningResult();
+    void acceptPruningStartResult();
+
+    //! We are being told that aggressive pruning to avoid hitting the
+    //! resource limit is no longer necessary, and we should report this
+    //! to the user when we can
+    void acceptPruningEndResult();
 
     //! Accessor for no limit flag
     bool haveNoLimit() const;
 
     //! Prune models where necessary
+    //! \return Was pruning required?
     bool pruneIfRequired(core_t::TTime endTime);
 
     //! Accounts for any extra memory to the one
@@ -243,7 +249,7 @@ private:
     //! The largest that the prune window can grow to - determined from the models
     std::size_t m_PruneWindowMaximum;
 
-    //! The smallest that the prune window can shrink to - 4 weeks
+    //! The smallest that the prune window can shrink to - determined from the models
     std::size_t m_PruneWindowMinimum;
 
     //! Don't do any sort of memory checking if this is set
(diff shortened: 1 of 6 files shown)


include/model/CResourceMonitor.h
```diff
@@ -127,12 +127,18 @@ class MODEL_EXPORT CResourceMonitor {
     //! We are being told that aggressive pruning has taken place
     //! to avoid hitting the resource limit, and we should report this
     //! to the user when we can
-    void acceptPruningResult();
+    void acceptPruningStartResult();
```
tveasey, line 130: This matched what we had before, but I find these names rather strange. How about:
```suggestion
    void startPruning();
```

lib/model/CResourceMonitor.cc
```diff
@@ -187,11 +187,21 @@ bool CResourceMonitor::pruneIfRequired(core_t::TTime endTime) {
...
+                    // If we increase the prune window to a size that's bigger
+                    // than what we started with then we should stop pruning to
+                    // be consistent with what would happen if the job was
+                    // closed and reopened
+                    LOG_DEBUG(<< "Pruning no longer necessary.");
+                    m_PruneWindow = m_PruneWindowMaximum;
+                    m_HasPruningStarted = false;
+                    this->acceptPruningEndResult();
+                } else {
+                    LOG_TRACE(<< "Expanding window, to " << m_PruneWindow);
```
tveasey, line 203: nit:
```suggestion
                    LOG_TRACE(<< "Expanding window to " << m_PruneWindow);
```

include/model/CResourceMonitor.h
droberts195: `startPruning()` sort of implies that it's going to do more than maybe update the memory status.  So I renamed it but moved the debug and setting of `m_HasPruningStarted` into this method too.  And the equivalent for `endPruning()`.  Does that sound OK to you?  I am just testing it locally now.

tveasey: That sounds much better to me

droberts195:
retest

PR #1194, opened 2020-05-05
[7.8][ML] Add a constant to the prediction which minimises the unregularised loss for classification and regression

tveasey:
Backport #1192.


tveasey:
retest

PR #1197, opened 2020-05-05
[ML] Fix restore for boosted tree and improve tests

tveasey:
Various state wasn't getting restored by `CBoostedTreeImpl` and we weren't initialising the example weights. We also weren't properly testing persist and restore:
1. We weren't checking persist after running training in `CBoostedTreeTest:: testPersistRestore` so not all state was present,
2. We weren't checking training produced the same values after restore (because we weren't actually restoring from the saved state) in `CDataFrameAnalyzerTrainingTest ::testOneRunOfBoostedTreeTrainingWithStateRecovery`.

I also took the opportunity to switch over the `std::exp` in the MSLE loss to their stable equivalents since it also improved stability of results after restore. Restoring from training state is currently broken and there isn't a good path to fixing this so I just discard old state.


lib/maths/CBoostedTreeLoss.cc
```diff
@@ -623,39 +621,38 @@ CArgMinPseudoHuberImpl::TObjective CArgMinPseudoHuberImpl::objective() const {
...
-            }
-        } catch (const std::exception& e) {
-            LOG_ERROR(<< "Error restoring loss function " << lossFunctionName
-                      << " " << e.what());
-            return nullptr;
-        }
-        LOG_ERROR(<< "Error restoring loss function. Unknown loss function type '"
-                  << lossFunctionName << "'.");
-        return nullptr;
-    } while (traverser.next());
```
tveasey, line 649: The while loop was generating a warning for me, since the condition is unreachable I just removed the loop.

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -294,43 +294,50 @@ TLossFunctionUPtr createLossFunction(TLossFunctionType lossFunctionType,
...
     std::size_t trainRows{500};
     std::size_t testRows{200};
     std::size_t cols{3};
     std::size_t capacity{250};
+
     // TODO reactivate test for huber and MSLE
     for (auto lossFunctionType : {TLossFunctionType::E_MseRegression /*, TLossFunctionType::E_MsleRegression,
           TLossFunctionType::E_HuberRegression*/}) {
-        auto generatePiecewiseConstant = [lossFunctionType](test::CRandomNumbers& rng,
-                                                            std::size_t cols) {
```
tveasey, line 307: I was getting variable shadowing warnings here so I just moved before the setting of the variables whose names clash. And same below.

droberts195:
This will need backporting to both 7.x and 7.8 now

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1234,12 +1234,12 @@ void CBoostedTreeImpl::restoreBestHyperparameters() {
     m_EtaGrowthRatePerTree = m_BestHyperparameters.etaGrowthRatePerTree();
     m_MaximumNumberTrees = m_BestHyperparameters.maximumNumberTrees();
     m_FeatureBagFraction = m_BestHyperparameters.featureBagFraction();
-    LOG_INFO(<< "loss* = " << m_BestForestTestLoss
-             << ", regularization* = " << m_Regularization.print()
-             << ", downsample factor* = " << m_DownsampleFactor << ", eta* = " << m_Eta
-             << ", eta growth rate per tree* = " << m_EtaGrowthRatePerTree
-             << ", maximum number trees* = " << m_MaximumNumberTrees
-             << ", feature bag fraction* = " << m_FeatureBagFraction);
+    LOG_TRACE(<< "loss* = " << m_BestForestTestLoss
```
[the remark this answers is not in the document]
tveasey: Yes that was my thinking. We have this in instrumentation anyway.

tveasey:
arm CI is WIP: merging since this is otherwise clean.

PR #1204, opened 2020-05-06
Adjust JDK for running ES integration tests on aarch64

droberts195:


droberts195:
The aarch64 part of the build failed with this:

```
[Exception] - critical check "[0, 10, 20, 30, 40, 50, 60, 70, 80, 90]" == core::CContainerPrinter::print(task.s_TenPercentProgressPoints) has failed [[0, 10, 20, 30, 40, 50, 60, 70, 80, 90] != [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100]]
 == [File] - CBoostedTreeTest.cc
 == [Line] -1419
```

That may well be a sign of different threading characteristics on the different architecture.

droberts195:
retest

droberts195:
Two different failures in the latest attempt:

```
[Exception] - critical check core::CAlignment::maxAlignment(&addresses[i]) == core::CAlignment::E_Aligned32 has failed [16 != 32]
 == [File] - CAlignmentTest.cc
 == [Line] -27

[Exception] - critical check core::CAlignment::isAligned( &addresses[i], core::CAlignment::E_Aligned32) has failed
 == [File] - CAlignmentTest.cc
 == [Line] -49
```

tveasey:
That's more worrying. It could be that the `Eigen`'s aligned allocator doesn't work on aarch64. If this is the case it will be a problem since it also relies on it internally. Although in theory Eigen supports aarch64. We may have to look at ways of passing flags that the memory is unaligned if this is possible for the basic `Matrix` type. It would be good to verify which index i is unaligned.

droberts195:
It looks like https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89357

That is fixed in gcc 9.3 apparently, which we will be upgrading to on master soon (for C++17).  But I hadn't planned to upgrade on 7.x.  There may be some workaround we can use in 7.x.  There is lots of discussion in that bug report that I haven't had time to fully digest.

droberts195:
@tveasey how about this for a plan:

- Merge and backport this PR, as it doesn’t make any problems worse
- Merge #1170 to master
- Monitor whether the alignment problem occurs on 7.x but not master, which would indicate it is the gcc bug fixed in version 9.3, or if it continues to occur on both branches, which means it’s something different

tveasey:
I think this sounds like a good plan.

tveasey:
Btw, if it is alignas not working we can also work around it in the test (so don't need to upgrade 7.x). I can allocate a large array and search for the first aligned address. The fact that other tests aren't failing makes me think that no code in `Eigen` relies on this and we don't use alignas in non-test code.

PR #1211, opened 2020-05-07
[7.x] Adjust JDK for running ES integration tests on aarch64

droberts195:

Backport of #1204


droberts195:
Test failures are aarch64 only and the same ones noted in https://github.com/elastic/ml-cpp/pull/1204#issuecomment-625258104

PR #1212, opened 2020-05-07
[7.8] Adjust JDK for running ES integration tests on aarch64

droberts195:

Backport of #1204


droberts195:
Test failure is aarch64 only and is this:

```
fatal: destination path 'elasticsearch' already exists and is not an empty directory.
```

That needs fixing on master too, so I will deal with it in a new PR.

PR #1213, opened 2020-05-08
Make ES integration tests work on reused Jenkins workers

droberts195:
The code to run ES integration tests on ml-cpp PR builds was
assuming the PR build would be on a completely clean Jenkins
worker.  This is not always the case, and for a reused worker
the old clone of the elasticsearch repo needs to be deleted
before doing the shallow (single commit) clone for the current
build.

PR #1217, opened 2020-05-08
[7.x] Make ES integration tests work on reused Jenkins workers

droberts195:
The code to run ES integration tests on ml-cpp PR builds was
assuming the PR build would be on a completely clean Jenkins
worker.  This is not always the case, and for a reused worker
the old clone of the elasticsearch repo needs to be deleted
before doing the shallow (single commit) clone for the current
build.

Backport of #1213

### DIFF
--- a/dev-tools/run_es_tests.sh
+++ b/dev-tools/run_es_tests.sh
@@ -61,6 +61,7 @@ function pickCloneTarget {
 pickCloneTarget
 
 cd "$1"
+rm -rf elasticsearch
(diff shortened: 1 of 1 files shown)


droberts195:
https://gradle-enterprise.elastic.co/s/j5n67zxsicjyy shows:

```
X-Pack is not supported and Machine Learning is not available for [linux-aarch64]; you can use the other X-Pack features (unsupported) by setting xpack.ml.enabled: false in elasticsearch.yml
```

That check needs removing from the ML Java code.

Interesting that didn't happen on master...

PR #1220, opened 2020-05-09
[ML] Work around gcc alignas bug

droberts195:
Prior to gcc 9.3 there is a bug where alignas sometimes wrongly
caps alignment requests of more than 16 bytes at 16 bytes.

The old style alignment attribute does not suffer this problem so
the workaround is to use that instead.

This change will be reverted when we upgrade to gcc 9.3.

### DIFF
--- a/lib/core/unittest/CAlignmentTest.cc
+++ b/lib/core/unittest/CAlignmentTest.cc
@@ -12,6 +12,14 @@
 #include <array>
 #include <vector>
 
+// TODO: revert this when we upgrade to gcc 9.3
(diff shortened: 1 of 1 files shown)


droberts195:
`>non-issue` as the code that used `alignas` has never been released.

Eigen has always used compiler-specific alignment directives (i.e. `__attribute__((aligned(x)))` with gcc/clang and `__declspec(align(x))` with Visual C++), so any aligned allocations Eigen did internally prior to 7.8 are fine.

PR #1221, opened 2020-05-10
[ML] Only log failure to ignore SIGPIPE once

droberts195:
The warning if something stops SIGPIPE being ignored has
always been deferred as the signal calls are made before
main() has run.  However, the place the warning was deferred
to could result in it being logged many times because the
deferred logging was in a method that could be called many
times.

This change moves that deferred logging to a dedicated
method that can just be called once.

### DIFF
--- a/include/core/CNamedPipeFactory.h
+++ b/include/core/CNamedPipeFactory.h
@@ -90,6 +90,11 @@ public:
     //! Default path for named pipes.
     static std::string defaultPath();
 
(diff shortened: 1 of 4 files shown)


droberts195:
I labelled this `>non-issue` as it's so inconsequential and there's no evidence anyone has ever suffered as a result.  But if people disagree I can relabel it as a `>bug`.

PR #1223, opened 2020-05-10
[7.8][ML] Work around gcc alignas bug

droberts195:
Prior to gcc 9.3 there is a bug where alignas wrongly caps alignment
requests of more than 16 bytes at 16 bytes.

The old style alignment attribute does not suffer this problem so
the workaround is to use that instead.

This change will be reverted when we upgrade to gcc 9.3.

Backport of #1220


droberts195:
retest

Issue #1202, opened 2020-05-06
[ML] Improve adaption to time translations of periodic patterns


tveasey:
We have observed some issues with poor adaption and instability on a data set which exhibits a stable periodic pattern which occasionally shifts in time. This leads to "false positive" anomalies.

The instability stems from poor adaption, so this issue is to cover work to improve model adaption for this case.

tveasey:
Resolved by #1614.

PR #1224, opened 2020-05-11
[ML] Improve adaption to periodic anomalies for longer bucket lengths

tveasey:
We track where anomalies occur for each seasonal component and increase the rate at which we learn them if there is strong evidence they're occurring periodically. However, we were only doing this for shorter bucket lengths, meaning we could take much longer than intended to adapt for long buckets.

This makes a related change to improve the stability of the adaption of seasonal components by including some mean reversion. The vast majority of signals we model tend to have stable periodic patterns and so mean reverting as we age out old data we better match their expected behaviour.

This is a step towards #1202.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -35,6 +35,13 @@
 * The macOS build platform for the {ml} C++ code is now High Sierra running Xcode 10.1,
   or Ubuntu 18.04 running clang 6.0 for cross compilation. (See {ml-pull}867[#867].)
 
+== {es} version 7.9.0
+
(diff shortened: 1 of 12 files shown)


tveasey:
retest

tveasey:
Thanks for the review @droberts195.

> There should be a plan to monitor the effect and make further tweaks depending on what happens across a range of test jobs.

I've been testing this myself across a representative collection of our QA jobs. However, I have a couple of changes I'd like to make which interact. Discussing it further with Wei we decided to group them together and test as a whole. So I'm going to close this PR and reopen against a branch.

PR #1157, opened 2020-04-22
[ML] Fix a bug in adaptive local regression

tveasey:
We were hitting the error when some the time intervals we use to split cyclic components into local regression models had negative length. With this change, I impose a constraint that they have positive length explicitly. I also moved the (rather specific code) which ensures that every interval has a minimum length, so that it receives at least one update per period, out of `CTools`. At the same time I simplified it to make explicitly handling the constraint easier.

I was able to reproduce the error following the instruction in [this](https://github.com/elastic/ml-cpp/issues/142#issuecomment-599684378) comment and verified that this change fixes it. Closes #142.

PR #1225, opened 2020-05-11
[ML] Fix possible cause for "Bad variance scale nan" log errors

tveasey:
Following on from #1114, there are a couple of places where we have code which could in theory generate 0/0. These would be a candidate root cause for #364 along with #1157.

This should generally have no impact on results until we'd actually hit the problem case.

### DIFF
--- a/lib/maths/CCalendarComponentAdaptiveBucketing.cc
+++ b/lib/maths/CCalendarComponentAdaptiveBucketing.cc
@@ -276,7 +276,9 @@ void CCalendarComponentAdaptiveBucketing::refresh(const TFloatVec& oldEndpoints)
                 static_cast<double>(oldCentres[l - 1]));
             largeErrorCount += w * oldLargeErrorCounts[l - 1];
             count += w * w * CBasicStatistics::count(m_Values[l - 1]);
-            double scale{count / CBasicStatistics::count(value)};
+            double scale{count == CBasicStatistics::count(value)
+                             ? 1.0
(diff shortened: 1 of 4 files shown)


lib/maths/CCalendarComponentAdaptiveBucketing.cc
```diff
@@ -276,7 +276,9 @@ void CCalendarComponentAdaptiveBucketing::refresh(const TFloatVec& oldEndpoints)
                 static_cast<double>(oldCentres[l - 1]));
             largeErrorCount += w * oldLargeErrorCounts[l - 1];
             count += w * w * CBasicStatistics::count(m_Values[l - 1]);
-            double scale{count / CBasicStatistics::count(value)};
+            double scale{count == CBasicStatistics::count(value)
```
droberts195, line 279: It's not so obvious here as in the other places that this is defending against `CBasicStatistics::count(value) == 0`, so it would be worth adding a comment like:

```
// Defend against division by zero - if CBasicStatistics::count(value) is zero then count must be too
```

(I am assuming that's the reasoning, but if I was reading this outside of a PR that was about avoiding division by zero it wouldn't immediately occur to me.)

tveasey: I added a comment in [da27f31](https://github.com/elastic/ml-cpp/pull/1225/commits/da27f31fbfea1d9590d7a083e4bfcb955aa17596).

PR #1237, opened 2020-05-13
[ML] Fix timestamp issue for instrumentation

### LINKED ISSUE #1234: [ML] Fix timestamp for DFA instrumentation data
At the moment the timestamp is `now()` in seconds converted to milliseconds. This leads to the stats data being overwritten if an iteration takes less than a second (very small jobs).

valeriy42:
This PR changes the resolution of the timestamps from seconds to milliseconds. This fixes #1234

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -90,6 +90,7 @@
 * Improve the initialization of seasonal components for anomaly detection. (See
   {ml-pull}1201[#1201], issue: {ml-issue}#1178[#1178].)
 * Fix possible root cause for "Bad variance scale nan" log errors. (See {ml-pull}1225[#1225].)
+* Change instrumentation timestamp resolution to millisecionds. (See {ml-pull}1237[#1237].)
 
 == {es} version 7.7.1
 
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -90,6 +90,7 @@
 * Improve the initialization of seasonal components for anomaly detection. (See
   {ml-pull}1201[#1201], issue: {ml-issue}#1178[#1178].)
 * Fix possible root cause for "Bad variance scale nan" log errors. (See {ml-pull}1225[#1225].)
+* Change instrumentation timestamp resolution to millisecionds. (See {ml-pull}1237[#1237].)
```
droberts195, line 93: We don't need to release note this for both 7.7.1 and 7.8.0 because 7.7.1 will be released first.

docs/CHANGELOG.asciidoc
```diff
@@ -99,6 +100,7 @@
   issue: {ml-issue}1136[#1136].)
 * Fix classification job failures when number of classes in configuration differs 
   from the number of classes present in the training data. (See {ml-pull}1144[#1144].)
+* Change instrumentation timestamp resolution to millisecionds. (See {ml-pull}1237[#1237].)
```
droberts195, line 103: typo: millisecionds -> milliseconds

Also, say it's data frame analytics, as the ML release notes for anomaly detection and data frame analytics all go in the same section.

docs/CHANGELOG.asciidoc
valeriy42: Thanks, I updated the changelog.

docs/CHANGELOG.asciidoc
valeriy42: Updated

valeriy42:
retest

PR #1238, opened 2020-05-13
Adding forecast parameter max_model_memory 

benwtrent:
This commit adds new parameter max_model_memory.

This parameter is defaulted to `20mb` (the previous static setting). 
Some larger forecasts can take a very long time to run. But if a user
has enough memory, they may opt to run more of the forecast in memory
without spooling to disk.

elasticsearch change https://github.com/elastic/elasticsearch/pull/57254

### DIFF
--- a/lib/api/CForecastRunner.cc
+++ b/lib/api/CForecastRunner.cc
@@ -35,7 +35,7 @@ const std::string CForecastRunner::ERROR_NO_DATA_PROCESSED(
     "Forecast cannot be executed as job requires data to have been processed and modeled");
 const std::string CForecastRunner::ERROR_NO_CREATE_TIME("Forecast create time must be specified and non zero");
 const std::string CForecastRunner::ERROR_BAD_MEMORY_STATUS("Forecast cannot be executed as model memory status is not OK");
-const std::string CForecastRunner::ERROR_MEMORY_LIMIT("Forecast cannot be executed as forecast memory usage is predicted to exceed 20MB while disk space is exceeded");
+const std::string CForecastRunner::ERROR_BAD_MODEL_MEMORY_LIMIT("Forecast max_model_memory must not exceed 500MB");
 const std::string CForecastRunner::ERROR_MEMORY_LIMIT_DISK(
     "Forecast cannot be executed as forecast memory usage is predicted to exceed 500MB");
 const std::string CForecastRunner::ERROR_MEMORY_LIMIT_DISKSPACE(
@@ -300,10 +300,13 @@ bool CForecastRunner::pushForecastJob(const std::string& controlMessage,
                                       prerequisites.s_IsSupportedFunction;
         totalMemoryUsage += prerequisites.s_MemoryUsageForDetector;
 
-        if (totalMemoryUsage >= MAX_FORECAST_MODEL_MEMORY &&
+        if (totalMemoryUsage >= forecastJob.s_MaxForecastModelMemory &&
             forecastJob.s_TemporaryFolder.empty()) {
             // note: for now MAX_FORECAST_MODEL_MEMORY is a static limit, a user can not change it
-            this->sendErrorMessage(forecastJob, ERROR_MEMORY_LIMIT);
+            this->sendErrorMessage(
+                forecastJob, "Forecast cannot be executed as forecast memory usage is predicted to exceed " +
+                                 std::to_string(forecastJob.s_MaxForecastModelMemory) +
+                                 " while disk space is exceeded");
             return false;
         }
     }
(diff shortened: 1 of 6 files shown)


benwtrent:
jenkins retest this please

benwtrent:
Tests are failing due to:
```
CForecastRunnerTest.cc: In member function 'void CForecastRunnerTest::testValidateProvidedMaxMemoryLimit::test_method()':
CForecastRunnerTest.cc:337:31: error: 'struct ml::api::CForecastRunner::SForecast' is private within this context
  337 |     ml::api::CForecastRunner::SForecast forecastJob;
      |                               ^~~~~~~~~
In file included from /ml-cpp/include/api/CAnomalyJob.h:23,
```
My new tests are identical to the others :/ and I did not change the access level to any methods. Not sure why this is failing :(

tveasey:
The magic you are missing is:
```
friend struct CForecastRunnerTest::testValidateProvidedMaxMemoryLimit;
```
etc. See [this](https://github.com/elastic/ml-cpp/blob/master/include/api/CForecastRunner.h#L294). I'm not 100% sure about all this friending. If it is just for that struct, maybe we should expose the struct. The public interface doesn't use `SForecast` so it isn't like it'll be accidentally abused.

lib/api/CForecastRunner.cc
```diff
@@ -337,7 +340,7 @@ bool CForecastRunner::pushForecastJob(const std::string& controlMessage,
 
     // 2nd loop over the detectors to clone models for forecasting
     bool persistOnDisk = false;
-    if (totalMemoryUsage >= MAX_FORECAST_MODEL_MEMORY) {
+    if (totalMemoryUsage >= forecastJob.s_MaxForecastModelMemory) {
```
droberts195, line 343: If we are letting people use 500MB of memory for forecasting then there needs to be a better link to the overall job model memory.

There is already a check on line 274 that the job memory status is `ok` for forecasting to be permitted.  That means the job is using no more than 60% of its configured model memory limit.

When we were letting people use 20MB it was possible that this could cause the job's overall model memory to exceed the job-wide limit, but this would only be by a maximum of 20MB, so it was considered a reasonable risk to take.  But with this change it would be possible that a job with a 100MB total model memory limit could be told that it was allowed to use 500MB for a forecast.  Obviously it wouldn't use that much in reality - if we assume that the forecast would use no more memory than the rest of the job then it would be 60MB max in practice in this example, but that would still be causing an overrun of the configured limit that was used to decide which nodes were big enough to accommodate the job.

So I am thinking we need an additional cap on `forecastJob.s_MaxForecastModelMemory`.  Given that the soft limit definition is 60% of permitted, maybe `forecastJob.s_MaxForecastModelMemory` should be capped at 40% of the job's overall model memory limit.  This should be done in terms of values obtained from `m_ResourceMonitor` so that future code changes don't break the link between the 60% and the 40%.

Then we need to decide how this capping will be notified.  We could make it an error so that if you ask for a forecast memory limit that's inappropriately high given the job's memory limit then the forecast isn't even attempted.  This could also be checked on the Java side in advance so that it can be reported in the REST API response.  Or alternatively we could cap the limit, attempt the forecast with the capped limit, and just report an informational message about the capping.  That would be nicer in the case where the forecast still worked in reasonable time, but we should think through how we'd troubleshoot a case where a user reported slow forecasting even when they'd asked for a high forecast memory limit.  How would the person doing support know the requested forecast limit was capped.

lib/api/CForecastRunner.cc
```diff
@@ -437,6 +442,11 @@ bool CForecastRunner::parseAndValidateForecastRequest(const std::string& control
     }
 
     // from now we have a forecast ID and can send error messages
+    if (forecastJob.s_MaxForecastModelMemory >= MAX_FORECAST_MODEL_PERSISTANCE_MEMORY) {
```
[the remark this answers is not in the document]
benwtrent: agreed.

lib/api/CForecastRunner.cc
benwtrent: There are two caps.

1. The absolute largest size of the forecast (I think 500 MB is sane as that is the limit of our disk overflow)

2. The limit of forecast allowed in memory. I think making this a maximum of `40%` of the job configured model memory limit is good (or the original 20 MB, whichever is larger).


I am not 100% sure if we should throw an error or possibly have the Java side write out an audit message. I will think a bit more about support, etc.

benwtrent: I think allowing the forecast to continue is good. We should indicate though that it will spool to disk and might take longer than desired. The Java side should write an audit message indicating this.

benwtrent: I am going to have the C++ side throw an error if the requested memory usage is above 40% of the configured model memory max. 

On the Java side, I will adjust the limit so that it will take `std::min((0.40 * job_max_memory), requested_forecast_limit)` and write an audit message + log which limit is used. 

The C++ side will also log a warning whenever a forecast has to be spooled to disk. This is not a "failed" state, but definitely a performance concern and I think warrants a "warning" log message.

droberts195:
retest

PR #1239, opened 2020-05-14
[ML] Fix weights to maximize minimum recall for multiclass classification when the training data is missing classes

tveasey:
This is a follow on from #1231.

We were still running into problems if the Java doesn't sample a class at all in the training data. This points to a problem with the stratified sampling implementation in Java, but we need to make the weight calculation defensive.

Since this is a correction to change #1113, which hasn't been released, I've marked this as a non-issue, but it must go out at the same time in 7.8.

### DIFF
--- a/lib/maths/CDataFrameUtils.cc
+++ b/lib/maths/CDataFrameUtils.cc
@@ -1072,27 +1072,32 @@ CDataFrameUtils::maximizeMinimumRecallForMulticlass(std::size_t numberThreads,
     std::size_t numberSamples{
         static_cast<std::size_t>(std::min(1000.0, rowMask.manhattan()))};
 
-    TStratifiedSamplerPtr sampler;
-    std::tie(sampler, std::ignore) = classifierStratifiedCrossValidationRowSampler(
-        numberThreads, frame, targetColumn, rng, numberSamples, rowMask);
-
-    TSizeVec rowIndices;
-    frame.readRows(1, 0, frame.numberRows(),
-                   [&](TRowItr beginRows, TRowItr endRows) {
-                       for (auto row = beginRows; row != endRows; ++row) {
-                           sampler->sample(*row);
-                       }
-                   },
-                   &rowMask);
-    sampler->finishSampling(rng, rowIndices);
-    std::sort(rowIndices.begin(), rowIndices.end());
     core::CPackedBitVector sampleMask;
-    for (auto row : rowIndices) {
-        sampleMask.extend(false, row - sampleMask.size());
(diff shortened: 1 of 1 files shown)


lib/maths/CDataFrameUtils.cc
```diff
@@ -1072,27 +1072,32 @@ CDataFrameUtils::maximizeMinimumRecallForMulticlass(std::size_t numberThreads,
...
-                       }
-                   },
-                   &rowMask);
-    sampler->finishSampling(rng, rowIndices);
-    std::sort(rowIndices.begin(), rowIndices.end());
     core::CPackedBitVector sampleMask;
-    for (auto row : rowIndices) {
-        sampleMask.extend(false, row - sampleMask.size());
-        sampleMask.extend(true);
+    if (numberSamples < static_cast<std::size_t>(rowMask.manhattan())) {
```
valeriy42, line 1076: Maybe it would be helpfull to add a comment for posterity here why the if-clause is necessary.

lib/maths/CDataFrameUtils.cc
```diff
@@ -1119,6 +1124,11 @@ CDataFrameUtils::maximizeMinimumRecallForMulticlass(std::size_t numberThreads,
              copyClassCountsAndRecalls, reduceClassCountsAndRecalls, classCountsAndRecalls);
     TDoubleVector classCounts{classCountsAndRecalls.topRows(numberClasses)};
     TDoubleVector classRecalls{classCountsAndRecalls.bottomRows(numberClasses)};
+    // If a class count is zero the derivative of the loss functin w.r.t. that
+    // class is not well defined. In practice, the objective is independent of
+    // such a a class so the choice for its count and recall is not important,
```
valeriy42, line 1129:
```suggestion
    // such a class so the choice for its count and recall is not important,
```

lib/maths/CDataFrameUtils.cc
tveasey: See [57ba59e](https://github.com/elastic/ml-cpp/pull/1239/commits/57ba59e7c41b8ee444e45b6108b8091c73d2e9fd).

lib/maths/CDataFrameUtils.cc
tveasey: See [57ba59e](https://github.com/elastic/ml-cpp/pull/1239/commits/57ba59e7c41b8ee444e45b6108b8091c73d2e9fd).

tveasey:
This only failed with a gradle issue on aarch64, so I'm going to merge.

PR #1242, opened 2020-05-14
[7.7][ML] Fix timestamp issue for instrumentation

### LINKED ISSUE #1234: [ML] Fix timestamp for DFA instrumentation data

valeriy42:
This PR changes the resolution of the timestamps from seconds to milliseconds. This fixes #1234


valeriy42:
retest

valeriy42:
retest

PR #1247, opened 2020-05-14
Report model change annotations via result stream

przemekwitek:
This PR makes model change annotations reported via result stream.
It is achieved the following way:
- model classes from `maths` module report model changes via `TModelChangeCallback` functional type
- concrete instantiation of `TModelChangeCallback` creates a `CAnnotation` object and puts it into in-memory collection `m_CurrentBucketStats.s_Annotations`
- annotations from models' `s_Annotations` are collected by `CAnomalyJob` and written out to the stream using `CAnnotationJsonWriter` class

Relates elastic/elasticsearch#55781

### DIFF
--- a/include/maths/CModel.h
+++ b/include/maths/CModel.h
@@ -94,7 +94,6 @@ class MATHS_EXPORT CModelAddSamplesParams {
 public:
     using TDouble2Vec = core::CSmallVector<double, 2>;
     using TDouble2VecWeightsAryVec = std::vector<maths_t::TDouble2VecWeightsAry>;
-    using TModelChangeCallback = std::function<void(core_t::TTime, const std::string&)>;
 
 public:
     //! Set whether or not the data are integer valued.
@@ -123,9 +122,9 @@ public:
     const TDouble2VecWeightsAryVec& priorWeights() const;
 
     //! Set the model change callback.
-    CModelAddSamplesParams& onModelChange(const TModelChangeCallback& addAnnotation);
+    CModelAddSamplesParams& onModelChange(const maths_t::TModelChangeCallback& addAnnotation);
     //! Get the model change callback.
-    const TModelChangeCallback& onModelChange() const;
+    const maths_t::TModelChangeCallback& onModelChange() const;
 
 private:
     //! The data type.
@@ -139,7 +138,8 @@ private:
     //! The prior sample weights.
     const TDouble2VecWeightsAryVec* m_PriorWeights = nullptr;
     //! The model change callback.
-    const TModelChangeCallback* m_ModelChangeCallback = nullptr;
+    maths_t::TModelChangeCallback m_ModelChangeCallback = [](core_t::TTime,
+                                                             const std::string&) {};
 };
 
 //! \brief The extra parameters needed by CModel::probability.
--- a/include/maths/CModel.h
+++ b/include/maths/CModel.h
@@ -94,6 +94,7 @@ class MATHS_EXPORT CModelAddSamplesParams {
 public:
     using TDouble2Vec = core::CSmallVector<double, 2>;
     using TDouble2VecWeightsAryVec = std::vector<maths_t::TDouble2VecWeightsAry>;
+    using TModelChangeCallback = std::function<void(core_t::TTime, const std::string&)>;
 
 public:
     //! Set whether or not the data are integer valued.
@@ -121,6 +122,11 @@ public:
     //! Get the prior sample weights.
     const TDouble2VecWeightsAryVec& priorWeights() const;
 
+    //! Set the model change callback.
+    CModelAddSamplesParams& onModelChange(const TModelChangeCallback& addAnnotation);
+    //! Get the model change callback.
+    const TModelChangeCallback& onModelChange() const;
+
 private:
     //! The data type.
     maths_t::EDataType m_Type = maths_t::E_MixedData;
@@ -132,6 +138,8 @@ private:
     const TDouble2VecWeightsAryVec* m_TrendWeights = nullptr;
     //! The prior sample weights.
     const TDouble2VecWeightsAryVec* m_PriorWeights = nullptr;
+    //! The model change callback.
+    const TModelChangeCallback* m_ModelChangeCallback = nullptr;
 };
 
 //! \brief The extra parameters needed by CModel::probability.
--- a/include/maths/CModel.h
+++ b/include/maths/CModel.h
@@ -121,10 +121,11 @@ public:
     //! Get the prior sample weights.
     const TDouble2VecWeightsAryVec& priorWeights() const;
 
-    //! Set the model change callback.
-    CModelAddSamplesParams& onModelChange(const maths_t::TModelChangeCallback& addAnnotation);
-    //! Get the model change callback.
-    const maths_t::TModelChangeCallback& onModelChange() const;
+    //! Set the model annotation callback.
+    CModelAddSamplesParams&
+    annotationCallback(const maths_t::TModelAnnotationCallback& modelAnnotationCallback);
+    //! Get the model annotation callback.
+    const maths_t::TModelAnnotationCallback& annotationCallback() const;
 
 private:
     //! The data type.
@@ -138,8 +139,8 @@ private:
     //! The prior sample weights.
     const TDouble2VecWeightsAryVec* m_PriorWeights = nullptr;
     //! The model change callback.
-    maths_t::TModelChangeCallback m_ModelChangeCallback = [](core_t::TTime,
-                                                             const std::string&) {};
+    maths_t::TModelAnnotationCallback m_ModelAnnotationCallback =
+        [](core_t::TTime, const std::string&) {};
 };
 
 //! \brief The extra parameters needed by CModel::probability.
--- a/lib/api/CAnnotationJsonWriter.cc
+++ b/lib/api/CAnnotationJsonWriter.cc
@@ -1,0 +1,86 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+#include <api/CAnnotationJsonWriter.h>
+#include <core/CLogger.h>
+#include <core/CTimeUtils.h>
+
+namespace ml {
+namespace api {
+namespace {
+// JSON field names
+const std::string ANNOTATION_RESULT_TYPE("annotation");
+const std::string TIMESTAMP("timestamp");
+const std::string END_TIMESTAMP("end_timestamp");
+const std::string ANNOTATION("annotation");
+const std::string CREATE_TIME("create_time");
+const std::string CREATE_USERNAME("create_username");
+const std::string MODIFIED_TIME("modified_time");
+const std::string MODIFIED_USERNAME("modified_username");
+const std::string TYPE("type");
+const std::string JOB_ID("job_id");
+const std::string DETECTOR_INDEX("detector_index");
+const std::string PARTITION_FIELD_NAME("partition_field_name");
+const std::string PARTITION_FIELD_VALUE("partition_field_value");
+const std::string OVER_FIELD_NAME("over_field_name");
+const std::string OVER_FIELD_VALUE("over_field_value");
+const std::string BY_FIELD_NAME("by_field_name");
+const std::string BY_FIELD_VALUE("by_field_value");
+}
+
+CAnnotationJsonWriter::CAnnotationJsonWriter(core::CJsonOutputStreamWrapper& outStream)
+    : m_Writer(outStream) {
+}
+
+void CAnnotationJsonWriter::writeResult(const std::string& jobId,
+                                        const model::CAnnotation& annotation) {
+
+    rapidjson::Value doc = m_Writer.makeObject();
+    write(jobId, annotation, doc);
+
+    rapidjson::Value wrapper = m_Writer.makeObject();
+    m_Writer.addMember(ANNOTATION_RESULT_TYPE, doc, wrapper);
+    m_Writer.write(wrapper);
+    m_Writer.Flush();
+}
+
+void CAnnotationJsonWriter::write(const std::string& jobId,
+                                  const model::CAnnotation& annotation,
+                                  rapidjson::Value& doc) {
+
+    m_Writer.addStringFieldCopyToObj(JOB_ID, jobId, doc, true);
+    m_Writer.addStringFieldCopyToObj(ANNOTATION, annotation.annotation(), doc);
+    // time is in Java format - milliseconds since the epoch
+    m_Writer.addTimeFieldToObj(TIMESTAMP, annotation.time(), doc);
+    m_Writer.addTimeFieldToObj(END_TIMESTAMP, annotation.time(), doc);
+
+    core_t::TTime currentTime(core::CTimeUtils::now());
+    m_Writer.addTimeFieldToObj(CREATE_TIME, currentTime, doc);
+    m_Writer.addTimeFieldToObj(MODIFIED_TIME, currentTime, doc);
+    m_Writer.addStringFieldCopyToObj(CREATE_USERNAME, "_xpack", doc);
+    m_Writer.addStringFieldCopyToObj(MODIFIED_USERNAME, "_xpack", doc);
+    m_Writer.addStringFieldCopyToObj(TYPE, "annotation", doc);
+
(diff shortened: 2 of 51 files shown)


lib/maths/CModel.cc
```diff
@@ -130,6 +130,17 @@ CModelAddSamplesParams::priorWeights() const {
...
 
+CModelAddSamplesParams&
+CModelAddSamplesParams::onModelChange(const TModelChangeCallback& modelChangeCallback) {
+    m_ModelChangeCallback = &modelChangeCallback;
+    return *this;
+}
+
+const CModelAddSamplesParams::TModelChangeCallback&
+CModelAddSamplesParams::onModelChange() const {
+    return *m_ModelChangeCallback;
```
droberts195, line 141: This line will cause a fatal crash if either:

1. `m_ModelChangeCallback` is `nullptr`
2. `m_ModelChangeCallback` is not `nullptr` but the `TModelChangeCallback` it points at is an empty `std::function`

It would be much safer if:

1. `m_ModelChangeCallback` was held by value rather than by pointer
2. `m_ModelChangeCallback` defaulted to a no-op function if unset

I guess the reason you didn't do this originally is that this would cause construction and destruction of may `std::function` objects since we would have to construct and destruct for every piece of input data.  In that case I think this method should at least be defensive, i.e.:

```
    if (m_ModelChangeCallback != nullptr && *m_ModelChangeCallback != nullptr) {
        return *m_ModelChangeCallback;
    } else {
        LOG_ERROR(<< "m_ModelChangeCallback unexpectedly not set");
        return reference_to_a_static_TModelChangeCallback_object_that_is_a_noop;
    }
```

Although it shouldn't happen, these runtime crashes caused by null pointers or empty functions are incredibly difficult to track down if they happen in production.

lib/model/CEventRateModel.cc
```diff
@@ -322,12 +324,23 @@ void CEventRateModel::sample(core_t::TTime startTime,
                                   priorWeights[0]);
                 maths_t::setWinsorisationWeight(winsorisationWeight, priorWeights[0]);
 
+                const auto createAndAddAnnotation = [&](core_t::TTime t,
+                                                        const std::string& annotation) {
+                    m_CurrentBucketStats.s_Annotations.emplace_back(
+                        t, annotation, gatherer.searchKey().detectorIndex(),
+                        gatherer.searchKey().partitionFieldName(),
+                        gatherer.partitionFieldValue(),
+                        gatherer.searchKey().overFieldName(), "",
```
droberts195, line 333: You could use `EMPTY_STRING` instead of `""`.  It's a static constant in the base class and avoids a call to `strlen` on the `const char*`.

lib/model/CEventRatePopulationModel.cc
```diff
@@ -459,12 +461,26 @@ void CEventRatePopulationModel::sample(core_t::TTime startTime,
...
+
+                const auto createAndAddAnnotation =
+                    [&](core_t::TTime t, const std::string& annotation) {
+                        m_CurrentBucketStats.s_Annotations.emplace_back(
+                            t, annotation, gatherer.searchKey().detectorIndex(),
+                            gatherer.searchKey().partitionFieldName(),
+                            gatherer.partitionFieldValue(),
+                            gatherer.searchKey().overFieldName(),
+                            gatherer.attributeName(cid),
+                            gatherer.searchKey().byFieldName(), "");
```
droberts195, line 473: You could use `EMPTY_STRING` instead of `""`.  It's a static constant in the base class and avoids a call to `strlen` on the `const char*`.

lib/model/CMetricModel.cc
```diff
@@ -305,12 +307,23 @@ void CMetricModel::sample(core_t::TTime startTime,
                                                    priorWeights[i]);
                 }
 
+                const auto createAndAddAnnotation = [&](core_t::TTime t,
+                                                        const std::string& annotation) {
+                    m_CurrentBucketStats.s_Annotations.emplace_back(
+                        t, annotation, gatherer.searchKey().detectorIndex(),
+                        gatherer.searchKey().partitionFieldName(),
+                        gatherer.partitionFieldValue(),
+                        gatherer.searchKey().overFieldName(), "",
```
droberts195, line 316: You could use `EMPTY_STRING` instead of `""`.  It's a static constant in the base class and avoids a call to `strlen` on the `const char*`.

lib/api/CAnnotationJsonWriter.cc
```diff
@@ -0,0 +1,86 @@
...
+ */
+#include <api/CAnnotationJsonWriter.h>
+#include <core/CLogger.h>
+#include <core/CTimeUtils.h>
+
+namespace ml {
+namespace api {
+namespace {
+// JSON field names
+const std::string ANNOTATION_RESULT_TYPE("annotation");
```
tveasey, line 14: nit: can we stick with braces. 
```suggestion
const std::string ANNOTATION_RESULT_TYPE{"annotation"};
```
(I've already had case where this has caught bugs for me by flagging up unexpected type conversion, so I think it is worthwhile.)

lib/api/unittest/CAnnotationJsonWriterTest.cc
```diff
@@ -0,0 +1,71 @@
...
+
+#include <api/CAnnotationJsonWriter.h>
+
+#include <test/BoostTestCloseAbsolute.h>
+
+#include <rapidjson/document.h>
+#include <rapidjson/prettywriter.h>
+
+#include <boost/test/unit_test.hpp>
+
```
tveasey, line 20:
```suggestion

#include <sstream>
```

lib/maths/CModel.cc
tveasey: As above, I prefer the option of this not being null and defaulting to a no-op if it isn't overridden in the class.

lib/maths/CTimeSeriesChangeDetector.cc
```diff
@@ -297,8 +298,8 @@ void CUnivariateTimeSeriesChangeDetector::addSamples(const TTimeDoublePr1Vec& sa
         core_t::TTime time{samples[i].first};
         weight = weights[i];
         maths_t::setWinsorisationWeight(winsorisation::MINIMUM_WEIGHT, weight);
-        m_TrendModel->addPoint(
-            time, CBasicStatistics::mean(m_TrendModel->value(time, 0.0)), weight);
+        m_TrendModel->addPoint(time, CBasicStatistics::mean(m_TrendModel->value(time, 0.0)),
+                               onModelChange, weight);
```
tveasey, line 302: I don't think we need this here. We don't add seasonal components to the decomposition while we're testing for a change. If this was an optional argument, as suggested above, this code could just stay as it is.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -387,16 +388,19 @@ CTimeSeriesDecompositionDetail::SMessage::SMessage(core_t::TTime time, core_t::T
...
+    double value,
+    const TModelChangeCallback& modelChangeCallback,
+    const maths_t::TDoubleWeightsAry& weights,
+    double trend,
+    double seasonal,
+    double calendar,
+    const TPredictor& predictor,
+    const CPeriodicityHypothesisTestsConfig& periodicityTestConfig)
+    : SMessage{time, lastTime}, s_Value{value},
+      s_ModelChangeCallback(modelChangeCallback), s_Weights{weights}, s_Trend{trend},
```
tveasey, line 403:
```suggestion
      s_ModelChangeCallback{modelChangeCallback}, s_Weights{weights}, s_Trend{trend},
```

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -1317,9 +1324,12 @@ void CTimeSeriesDecompositionDetail::CComponents::handle(const SDetectedSeasonal
         const CPeriodicityHypothesisTestsResult& result{message.s_Result};
         const CExpandingWindow& window{message.s_Window};
         const TPredictor& predictor{message.s_Predictor};
+        const TModelChangeCallback& onModelChange{message.s_ModelChangeCallback};
 
         if (this->addSeasonalComponents(result, window, predictor)) {
-            LOG_DEBUG(<< "Detected seasonal components at " << time);
+            std::string annotation = "Detected seasonal components";
+            LOG_DEBUG(<< annotation << " at " << std::to_string(time));
+            onModelChange(time, annotation);
```
tveasey, line 1332: I think this isn't the right place to add the annotation and it would be nice to include information about the periodic component detected. Let's discuss this a bit offline.

lib/maths/CTimeSeriesModel.cc
```diff
@@ -1501,12 +1502,16 @@ CUnivariateTimeSeriesModel::testAndApplyChange(const CModelAddSamplesParams& par
...
-        m_ChangeDetector->addSamples({{time, values[median].second[0]}}, {weights});
+        const TModelChangeCallback& onModelChange = params.onModelChange();
+        m_ChangeDetector->addSamples({{time, values[median].second[0]}},
+                                     {weights}, onModelChange);
         if (m_ChangeDetector->stopTesting()) {
             m_ChangeDetector.reset();
             m_TrendModel->testingForChange(false);
         } else if (auto change = m_ChangeDetector->change()) {
-            LOG_DEBUG(<< "Detected " << change->print() << " at " << time);
+            std::string annotation = "Detected " + change->print();
```
tveasey, line 1512:
```suggestion
            std::string annotation{"Detected " + change->print()};
```

lib/maths/unittest/CMathsMemoryTest.cc
```diff
@@ -24,14 +25,16 @@ BOOST_AUTO_TEST_SUITE(CMathsMemoryTest)
 using namespace ml;
 using namespace maths;
 
+const ml::maths::CModelAddSamplesParams::TModelChangeCallback NOOP;
```
tveasey, line 28: This is a null function object and this test code will throw. But I think it is better to handle defaulting in the function definition.

lib/maths/unittest/CTimeSeriesChangeDetectorTest.cc
```diff
@@ -52,6 +52,7 @@ using TPriorPtr = std::shared_ptr<maths::CPrior>;
 
 core_t::TTime BUCKET_LENGTH{1800};
 const double DECAY_RATE{0.0002};
+const ml::maths::CModelAddSamplesParams::TModelChangeCallback NOOP;
```
tveasey, line 55: As above, this can probably be dropped.

lib/model/CEventRatePopulationModel.cc
```diff
@@ -459,12 +461,26 @@ void CEventRatePopulationModel::sample(core_t::TTime startTime,
 
             for (auto& attribute : attributeValuesAndWeights) {
                 std::size_t cid = attribute.first;
+
+                const auto createAndAddAnnotation =
```
tveasey, line 465: I think we should clear the `m_CurrentBucketStats.s_Annotations` at the start of this function. It'll get called exactly once per bucket and we'll write them out between calls.

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
```diff
@@ -113,6 +114,7 @@ const core_t::TTime HOUR = core::constants::HOUR;
 const core_t::TTime DAY = core::constants::DAY;
 const core_t::TTime WEEK = core::constants::WEEK;
 const core_t::TTime YEAR = core::constants::YEAR;
+const ml::maths::CModelAddSamplesParams::TModelChangeCallback NOOP;
```
tveasey, line 117: It would be good to test, in at least one test, that the annotations are actually created.

lib/api/unittest/CAnnotationJsonWriterTest.cc
```diff
@@ -0,0 +1,73 @@
...
+    doc.Parse<rapidjson::kParseDefaultFlags>(sstream.str());
+    BOOST_TEST_REQUIRE(!doc.HasParseError());
+    const rapidjson::Value& firstElement = doc[0];
+    BOOST_TEST_REQUIRE(firstElement.HasMember("annotation"));
+    const rapidjson::Value& annotation = firstElement["annotation"];
+    BOOST_TEST_REQUIRE(annotation.HasMember("job_id"));
+    BOOST_REQUIRE_EQUAL(std::string("job-id"),
+                        std::string(annotation["job_id"].GetString()));
+    BOOST_TEST_REQUIRE(annotation.HasMember("timestamp"));
+    BOOST_REQUIRE_EQUAL(int64_t(1000), annotation["timestamp"].GetInt64());
```
droberts195, line 47: FYI it's not necessary any more to have the exact same types when asserting equality.  I know if you copy an existing file as an example you'll see this everywhere, but this is because CppUnit was really picky that both types in an equality assertion were the same.  Boost.Test does not mandate this.  So `int64_t(1000)` could simply be `1000` here.

I don't mind if you leave all these as-is for this PR, as we have the unnecessary casting and object construction all over the place in our tests due to the legacy of CppUnit.  But I just wanted to make sure you knew it's not necessary for the future.

lib/model/CMetricModel.cc
```diff
@@ -305,12 +308,23 @@ void CMetricModel::sample(core_t::TTime startTime,
                                                    priorWeights[i]);
                 }
 
+                const auto createAndAddAnnotation = [&](core_t::TTime t,
+                                                        const std::string& annotation) {
+                    m_CurrentBucketStats.s_Annotations.emplace_back(
```
droberts195, line 313: There needs to be a way to enable/disable generation of the annotations, so that users with huge numbers of partition field values don't find themselves running out of memory when periodicity is detected in many time series in the same bucket.

I think it would be best if the flag to enable or disable annotation generation was passed with the `--modelplotconfig` (see https://github.com/elastic/elasticsearch/blob/0f2f00a20a6eae5b1fa9931f3b300746eb28e1c6/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/process/autodetect/AutodetectBuilder.java#L68 on the Java side).  But it should be a new setting in that file so that model plot and annotations can be enabled/disabled independently.

On the C++ a nice way to implement this would be to make these "on change" callbacks no-ops if annotations were disabled.

I know there is more to implement, like event names, so feel free to leave this for a future PR.  I just wanted to make sure the requirement isn't forgotten and the changes of this PR don't make it harder to enable/disable annotations in the future.

przemekwitek:
> The memory of m_CurrentBucketStats.s_Annotations needs to be accounted for. 

Done.

> Am I correct that at the moment it's never cleared?

I've added clearing in the beginning of the `sample` method in all the models.
By Tom's comment: "I think we should clear the m_CurrentBucketStats.s_Annotations at the start of this function. It'll get called exactly once per bucket and we'll write them out between calls."

IIUC this solution is equivalent to the one you outlined (using swap).

include/maths/CModel.h
```diff
@@ -121,6 +121,11 @@ class MATHS_EXPORT CModelAddSamplesParams {
     //! Get the prior sample weights.
     const TDouble2VecWeightsAryVec& priorWeights() const;
 
+    //! Set the model change callback.
+    CModelAddSamplesParams& onModelChange(const maths_t::TModelChangeCallback& addAnnotation);
```
tveasey, line 125: In line with my previous comments, maybe something which captures that this is related to annotations, something like:
```suggestion
    CModelAddSamplesParams& addAnnotationCallback(const maths_t::TModelChangeCallback& addAnnotation);
```

lib/model/CMetricModel.cc
przemekwitek: I've implemented the Java part in https://github.com/elastic/elasticsearch/pull/57539.
IIUC the Java PR can be submitted before C++ as the `CAnomalyDetectorModelConfig::configureModelPlot` method should ignore unknown properties. Please LMK if that is not the case.

I'm still working on the C++ part. Do you think I should separate it into its own PR or can I modify this one?

przemekwitek: Update: I've implemented C++ part in this PR.

lib/maths/CTimeSeriesDecompositionDetail.cc
przemekwitek: Done.

I've moved reporting the annotation to `addSeasonalComponents` method.
I've also added reporting in `addCalendarComponent` method

lib/maths/unittest/CTimeSeriesDecompositionTest.cc
przemekwitek: How should I go about the unit testing in this case? Should I call `addPoint` method with a model annotation callback and verify it got called? 
Could you point me at the test case which I can take as a base?

PR #1262, opened 2020-05-15
[7.8][ML] Fix weights to maximize minimum recall for multiclass classification when the training data is missing classes

tveasey:
Backport #1239.


tveasey:
retest

PR #1266, opened 2020-05-15
[6.8][ML] Add source URL to 3rd party info when we have to say how to get it

droberts195:
Some third party licenses require that we tell people where they can
obtain the third party source code.

To integrate with the new solution Elastic is putting together to
meet this requirement, this PR adds a sourceURL field to the 3rd
party info files.  For licenses that require we tell people where
to get the code from this is set to the appropriate download URL.
For other licenses it is left blank.

Backport of #1258


droberts195:
retest

PR #1267, opened 2020-05-18
[ML] Special case handling for autocorrelation of a constant

tveasey:
This was returning NaNs before. Given our usage, it was harmless (and so I've marked this as a non-issue), but it's dangerous.

I also adjusted this file to match our current coding style guidelines.

### DIFF
--- a/lib/maths/CSignal.cc
+++ b/lib/maths/CSignal.cc
@@ -13,8 +13,8 @@
 #include <boost/math/constants/constants.hpp>
 
 #include <algorithm>
-#include <cinttypes>
 #include <cmath>
+#include <cstdint>
 
 namespace ml {
 namespace maths {
--- a/lib/maths/CSignal.cc
+++ b/lib/maths/CSignal.cc
@@ -13,6 +13,7 @@
 #include <boost/math/constants/constants.hpp>
 
 #include <algorithm>
+#include <cinttypes>
 #include <cmath>
 
 namespace ml {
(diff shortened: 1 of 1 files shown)


lib/maths/CSignal.cc
```diff
@@ -169,20 +170,26 @@ void CSignal::autocorrelations(const TFloatMeanAccumulatorVec& values, TDoubleVe
...
     }
-    double mean = CBasicStatistics::mean(moments);
-    double variance = CBasicStatistics::maximumLikelihoodVariance(moments);
+    double mean{CBasicStatistics::mean(moments)};
+    double variance{CBasicStatistics::maximumLikelihoodVariance(moments)};
+
+    if (variance == 0.0) {
+        // The autocorrelation of a constant is zero.
+        result.resize(n, 0.0);
+        return;
```
tveasey, line 187: This is the fix

lib/maths/CSignal.cc
```diff
@@ -13,6 +13,7 @@
 #include <boost/math/constants/constants.hpp>
 
 #include <algorithm>
+#include <cinttypes>
```
droberts195, line 16: It should probably be `<cstdint>` instead of `<cinttypes>` - the former contains the types and the latter adds `printf` macros, which we aren't using here.

tveasey: Good catch. My version of clangd occasionally seems to add this header for no good reason. See [37658a9](https://github.com/elastic/ml-cpp/pull/1267/commits/37658a903a834b0f03630de9675c396bf3187beb).

PR #1271, opened 2020-05-20
[ML] Improve time series decomposition adaption

tveasey:
This makes a couple of enhancements aimed at #1202.

In particular, we track where anomalies occur for each seasonal component and increase the rate at which we learn from them if there is strong evidence they're occurring periodically. However, we were only doing this for shorter bucket lengths, meaning we could take much longer than intended to adapt for long buckets.

This also reworks mean reversion of seasonal components to be have a continuous parameterisation. Together with how quickly we adjust the position of the local regression model and how quickly we age out old data, this gives us three parameters which control how fast we'll adapt to changes and stability. I did a grid search over these parameters for a range of different QA data sets and chose the values to with the best "calibration", i.e. comparing the bucket predicted distribution to actual error distribution.

I also made a couple of small tidy ups and improved some function naming.

@droberts195 this is largely #1224 modulo some small tweaks and the parameter search. As discussed I'm raising this against a feature branch so we can do a full assessment with a couple of related changes I have planned. Marking as a non-issue since I'll add documentation only for the collection of changes.

### DIFF
--- a/include/maths/CAdaptiveBucketing.h
+++ b/include/maths/CAdaptiveBucketing.h
@@ -264,6 +264,9 @@ private:
     //! Implements split of \p bucket for derived state.
     virtual void split(std::size_t bucket) = 0;
 
+    //! Compute the p-value of the large error count for \p bucket.
+    double bucketLargeErrorCountPValue(double totalLargeErrorCount, std::size_t bucket) const;
+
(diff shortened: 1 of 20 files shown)


tveasey:
retest

tveasey:
retest

PR #1274, opened 2020-05-26
[ML] Don't lose precision when persisting state

tveasey:
Following on from #1197, this reworks persistence to not lose precision.

Since we chunk and compress state when persisting, using sufficient decimal places to not lose precision doesn't cost us much in terms of model size, and persisting and restoring state is not a performance bottleneck. 

This has the nice property that we can test restore from state produces the same results much more stably, so I simplified and extended `testRunBoostedTreeRegressionTrainingWithStateRecovery` accordingly.

I also took the opportunity to tidy up `CPersistUtils` and in particular fix violations of our coding guidelines.

This affects results for both anomaly detection and data frame analyses, but only after restoring from state.

### DIFF
--- a/include/core/CPersistUtils.h
+++ b/include/core/CPersistUtils.h
@@ -14,17 +14,16 @@
 #include <core/CStringUtils.h>
 #include <core/ImportExport.h>
 
-#include <boost/array.hpp>
 #include <boost/iterator/indirect_iterator.hpp>
 #include <boost/optional.hpp>
-#include <boost/type_traits/integral_constant.hpp>
-#include <boost/type_traits/is_arithmetic.hpp>
-#include <boost/type_traits/remove_const.hpp>
 #include <boost/unordered/unordered_map_fwd.hpp>
 #include <boost/unordered/unordered_set_fwd.hpp>
 
 #include <algorithm>
+#include <array>
+#include <cinttypes>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 namespace ml {
@@ -39,7 +38,7 @@ const TPersistenceTag SIZE_TAG("d", "size");
 
 template<typename T>
 struct remove_const {
-    using type = typename boost::remove_const<T>::type;
+    using type = typename std::remove_const<T>::type;
 };
 
 template<typename U, typename V>
@@ -47,8 +46,8 @@ struct remove_const<std::pair<U, V>> {
     using type = std::pair<typename remove_const<U>::type, typename remove_const<V>::type>;
 };
 
-//! Template specialisation utility classes for selecting various
-//! approaches for persisting and restoring objects.
+//! Template specialisation utility classes for selecting various approaches
+//! for persisting and restoring objects.
 class BasicPersist {};
 class ContainerPersist {};
 class MemberPersist {};
@@ -58,22 +57,19 @@ class ContainerRestore {};
 class MemberRestore {};
 class MemberFromDelimited {};
 
-//! Auxiliary type used by has_const_iterator to test for a nested
-//! typedef.
+//! Auxiliary type used by has_const_iterator to test for a nested typedef.
 template<typename T, typename R = void>
 struct enable_if {
     using type = R;
 };
 
-//! Auxiliary type used by has_persist_function to test for a nested
-//! member function.
+//! Auxiliary type used by has_persist_function to test for a nested member function.
 template<typename T, T, typename R = void>
 struct enable_if_is {
     using type = R;
 };
 
-//! \name Class used to select appropriate persist implementation
-//! for containers.
+//! \name Class used to select the appropriate persist implementation for containers.
 //@{
 template<typename T, typename ITR = void>
 struct persist_container_selector {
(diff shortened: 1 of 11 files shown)


include/core/CPersistUtils.h
```diff
@@ -172,93 +161,86 @@ struct reserve_selector<T, typename enable_if_is<void (T::*)(std::size_t), &T::r
...
-    //! using CStringUtils functions.
+    //! \brief Converts a built in type to a string using CStringUtils functions.
     class CORE_EXPORT CBuiltinToString {
     public:
         CBuiltinToString(const char pairDelimiter)
             : m_PairDelimiter(pairDelimiter) {}
 
         std::string operator()(double value) const {
-            return CStringUtils::typeToStringPrecise(value, CIEEE754::E_SinglePrecision);
+            return CStringUtils::typeToStringPrecise(value, CIEEE754::E_DoublePrecision);
```
tveasey, line 212: This is the important change in this file

include/core/CPersistUtils.h
```diff
@@ -661,14 +611,12 @@ class CORE_EXPORT CPersistUtils {
...
 
-        // Reuse this same string to avoid as many allocations
-        // as possible.
+        // Reuse this same string to avoid as many allocations as possible.
         //
-        // The reservation is 15 because for string implementations
-        // using the short string optimisation we don't want to
-        // cause an unnecessary allocation.
+        // Reserve 17 chars because for string implementations using the short
+        // string optimisation we don't want to cause an unnecessary allocation.
```
droberts195, line 617: This comment is outdated now.

At the time it was written some platforms used the short string optimisation and others didn't.  For those that weren't it was beneficial to reserve in case the lengths of tokens went, say, 8, 9, 10, causing two reallocations.  But for strings using the short string optimisation minimum `capacity` is 15 and tokens of lengths 8, 9, 10 in that order wouldn't cause any allocations.

Now we are in a world where all platforms use the short string optimisation.  Reserving 16 or above forces Windows/Linux strings to do a heap allocation.  But 17 is not enough to avoid a second, for example if the token is `1.2345678901234567e-89`.  So I am thinking it's actually best _not_ to reserve anything.  In the case of many short numbers like `0`, `1`, `0.5` we'll avoid all allocations in this method and in the case of really long representations we might as well do the allocation when it's required.

include/core/CPersistUtils.h
```diff
@@ -271,51 +253,50 @@ class CORE_EXPORT CPersistUtils {
...
 
-    //! \brief Utility to convert a string to a built in type
-    //! using CStringUtils functions.
+    //! \brief Converts a string to a built in type using CStringUtils functions.
     class CORE_EXPORT CBuiltinFromString {
     public:
         CBuiltinFromString(const char pairDelimiter)
             : m_PairDelimiter(pairDelimiter) {
-            m_Token.reserve(15);
+            m_Token.reserve(17);
```
droberts195, line 261: Similar to what I said in the other comment, it might be best not to `reserve` here in the modern world.

include/core/CPersistUtils.h
```diff
@@ -14,17 +14,16 @@
...
 #include <boost/optional.hpp>
-#include <boost/type_traits/integral_constant.hpp>
-#include <boost/type_traits/is_arithmetic.hpp>
-#include <boost/type_traits/remove_const.hpp>
 #include <boost/unordered/unordered_map_fwd.hpp>
 #include <boost/unordered/unordered_set_fwd.hpp>
 
 #include <algorithm>
+#include <array>
+#include <cinttypes>
```
droberts195, line 24: `<cstdint>` is more efficient

include/core/CPersistUtils.h
tveasey: Good point. See [fa3ab02](https://github.com/elastic/ml-cpp/pull/1274/commits/fa3ab02793b790aa31efb0c0cd8524a645f0db6a).

include/core/CPersistUtils.h
tveasey: Changed in [fa3ab02](https://github.com/elastic/ml-cpp/pull/1274/commits/fa3ab02793b790aa31efb0c0cd8524a645f0db6a).

include/core/CPersistUtils.h
tveasey: Changed in [fa3ab02](https://github.com/elastic/ml-cpp/pull/1274/commits/fa3ab02793b790aa31efb0c0cd8524a645f0db6a).

tveasey:
Note @droberts195 [6118c5a](https://github.com/elastic/ml-cpp/pull/1274/commits/6118c5ae556def059b31c57776571ed3864cb6bc) addresses some test fallout in a way that is probably worth reviewing quickly. In particular, I was getting test failures because the checksum precision for doubles increased as result of the print changes. This uncouples the implementation from `CStringUtils`. I'm not totally sold on printing the value as a mechanism to achieve this sort of fuzzy checksum, but it was the easiest way to just restore what we had before, so get things passing again. This is something I might try to revisit at some point.

tveasey:
retest

droberts195:
> 6118c5ae556def059b31c57776571ed3864cb6bc addresses some test fallout

Like you say, what you've done keeps the tests passing with the way things were working before, so it's reasonable for this PR.

PR #1277, opened 2020-05-28
[ML] Parallelise the feature importance calculation for classification and regression over trees

tveasey:
We can split the work to compute feature importance up over the trees in the forest since we just sum up the per tree feature importances at the end.

In order to do this, I needed to expose an interface to `parallel_for_each` which takes a list of functions to run on subsets of the iterator range because we use member variables for the algorithm state to avoid reallocating for each training example, which I need to bind to each function upfront.

### DIFF
--- a/include/core/Concurrency.h
+++ b/include/core/Concurrency.h
@@ -233,19 +233,19 @@ void parallel_for_each(std::size_t start,
     for (std::size_t offset = 0, partitions = functions.size();
          offset < partitions; ++offset, ++start) {
 
-        // Note there is one copy of g for each thread so capture by reference
-        // is thread safe provided f is thread safe.
+        // Note there is one f for each thread so capture by reference is thread
+        // safe provided each f is thread safe.
 
         CLoopProgress progress{end - offset, recordProgress,
                                1.0 / static_cast<double>(partitions)};
 
-        auto& g = functions[offset];
+        auto& f = functions[offset];
         tasks.emplace_back(
             async(defaultAsyncExecutor(),
-                  [&g, partitions, progress](std::size_t start_, std::size_t end_) mutable {
+                  [&f, partitions, progress](std::size_t start_, std::size_t end_) mutable {
                       for (std::size_t i = start_; i < end_;
                            i += partitions, progress.increment(partitions)) {
-                          g(i);
+                          f(i);
                       }
                       return true; // So we can check for exceptions via get.
                   },
@@ -348,16 +348,17 @@ void parallel_for_each(ITR start,
 
     for (std::size_t offset = 0, partitions = functions.size();
          offset < partitions; ++offset, ++start) {
-        // Note there is one copy of g for each thread so capture by reference
-        // is thread safe provided f is thread safe.
+
+        // Note there is one f for each thread so capture by reference is thread
+        // safe provided each f is thread safe.
 
         CLoopProgress progress{size - offset, recordProgress,
                                1.0 / static_cast<double>(partitions)};
 
-        auto& g = functions[offset];
+        auto& f = functions[offset];
         tasks.emplace_back(async(
             defaultAsyncExecutor(),
-            [&g, partitions, offset, size, progress](ITR start_) mutable {
+            [&f, partitions, offset, size, progress](ITR start_) mutable {
 
                 std::size_t i{offset};
 
@@ -369,7 +370,7 @@ void parallel_for_each(ITR start,
 
                 for (ITR j = start_; i < size; i += partitions,
                          incrementByPartitions(j), progress.increment(partitions)) {
-                    g(*j);
+                    f(*j);
                 }
                 return true; // So we can check for exceptions via get.
             },
--- a/include/core/Concurrency.h
+++ b/include/core/Concurrency.h
@@ -151,7 +151,7 @@ async(CExecutor& executor, FUNCTION&& f, ARGS&&... args) {
     // Schedule the task to compute the result.
     executor.schedule(std::move(task));
 
-    return std::move(result);
+    return result;
 }
 
 //! Wait for all \p futures to be available.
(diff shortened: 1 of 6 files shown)


lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -44,29 +53,57 @@ void CTreeShapFeatureImportance::shap(const TRowRef& row, TShapWriter writer) {
...
+                                            las::zero((*m_Forest)[0][0].value()));
+            computeTreeShap.push_back([&encodedRow, i, this](const TTree& tree) {
+                this->shapRecursive(tree, encodedRow, 0, 1.0, 1.0, -1,
+                                    CSplitPath{m_PathStorage[i].begin(),
+                                               m_ScaleStorage[i].begin()},
+                                    0, m_PerThreadShapValues[i]);
+            });
+        }
+
+        core::parallel_for_each(m_Forest->begin(), m_Forest->size(), computeTreeShap);
```
valeriy42, line 83: Will this work if you use `begin()` and `end()` instead of `size()`?

lib/maths/CTreeShapFeatureImportance.cc
```diff
@@ -44,29 +53,57 @@ void CTreeShapFeatureImportance::shap(const TRowRef& row, TShapWriter writer) {
...
+
+        m_ReducedShapValues = m_PerThreadShapValues[0];
+        for (std::size_t i = 1; i < m_PerThreadShapValues.size(); ++i) {
```
valeriy42, line 86: Maybe add a note here that it can be replaces with `std::reduce` and parallel execution once we have C++17 support?

lib/maths/unittest/CTreeShapFeatureImportanceTest.cc
```diff
@@ -545,4 +559,38 @@ BOOST_FIXTURE_TEST_CASE(testSingleTreeShapRandomDataFrame, SFixtureSingleTreeRan
...
+                expectedShap = shap;
+            });
+            s_ThreadedForestFeatureImportance->shap(*row, [&](const TSizeVec& indices,
+                                                              const TStrVec& names,
+                                                              const TVectorVec& shap) {
+                BOOST_REQUIRE_EQUAL(core::CContainerPrinter::print(expectedIndices),
+                                    core::CContainerPrinter::print(indices));
+                BOOST_REQUIRE_EQUAL(core::CContainerPrinter::print(expectedNames),
+                                    core::CContainerPrinter::print(names));
+                BOOST_REQUIRE_EQUAL(core::CContainerPrinter::print(expectedShap),
```
valeriy42, line 587: Why is it necessary to compare the string representatives instead of the numerical values?

tveasey: It was just for code brevity, i.e. it allows me to compare all values (to reasonable accuracy) in one line.

lib/maths/CTreeShapFeatureImportance.cc
tveasey: I don't think this can go until the standard introduces executors, because we definitely want these to be executed in a thread pool rather than have to spawn threads. So this will have to stay until we are able to adopt C++23 based on current plans. Given how far away this is, I'm not sure it is worth a comment at the moment.

include/core/Concurrency.h
```diff
@@ -207,6 +207,54 @@ CORE_EXPORT
...
+    for (std::size_t offset = 0, partitions = functions.size();
+         offset < partitions; ++offset, ++start) {
+
+        // Note there is one copy of g for each thread so capture by reference
+        // is thread safe provided f is thread safe.
+
+        CLoopProgress progress{end - offset, recordProgress,
+                               1.0 / static_cast<double>(partitions)};
+
+        auto& g = functions[offset];
```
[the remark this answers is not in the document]
tveasey: Historically, this was called `g` because it was in some code which also had an `f` :).

I'm not sure there is a great name for this, it is basically the callable which is the unit of work and I use brief names for these unit of work functions throughout this file. I'm not sure there is convention around this, but I think for these sorts of tasks it would be fairly common to just use a single letter, as in this is some function `f`.

On balance I'd therefore choose to just switch to calling this `f` rather than swap to some different naming throughout this file, if that's okay.

lib/maths/CTreeShapFeatureImportance.cc
tveasey: This calls [this](https://github.com/elastic/ml-cpp/pull/1277/files#diff-75e42b5a7634c55d254341641a79d6fcR340) overload. I used this signature because it matched how the loop was written which I needed to factor out. It does look slightly odd. I'll consider if I can migrate to a version which takes start and end iterators here.

tveasey: Ok I reviewing this code I realised I was missing error checking and handling of reentry for the new functions I exposed. I ended up moving the versions I had into `concurrency_detail` and calling these from the new overloads. A nice side effect is I could could change the signature to take two iterators. See [ea0317c](https://github.com/elastic/ml-cpp/pull/1277/commits/ea0317c1b8eeedd820530f3e3ed2e14c280bb7a7).

include/core/Concurrency.h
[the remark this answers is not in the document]
tveasey: I did the rename in [cc7ea3d](https://github.com/elastic/ml-cpp/pull/1277/commits/cc7ea3d7a8fc16e2569b805f2ae7d1d2099a3b76).

tveasey:
Thanks for the review @valeriy42! Note I made a somewhat significant refactor in `Concurrency` to tighten up the new functions' error handling. Other than that I think I've commented on or addressed all your comments. Can you take another look and in particular it's worth checking [ea0317c](https://github.com/elastic/ml-cpp/pull/1277/commits/ea0317c1b8eeedd820530f3e3ed2e14c280bb7a7)?

PR #1279, opened 2020-05-28
[7.x] Adding forecast parameter max_model_memory  (#1238)

benwtrent:
This commit adds new parameter max_model_memory.

This parameter is defaulted to `20mb` (the previous static setting).
Some larger forecasts can take a very long time to run. But if a user
has enough memory, they may opt to run more of the forecast in memory
without spooling to disk.

elasticsearch change https://github.com/elastic/elasticsearch/pull/57254

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,13 @@
 
 //=== Regressions
 
+== {es} version 7.9.0
+
+=== Enhancements
+
+* Add support for larger forecasts in memory via max_model_memory setting.
+  (See {ml-pull}1238[#1238] and {pull}57254[#57254].)
+
 == {es} version 7.8.0
 
 === Enhancements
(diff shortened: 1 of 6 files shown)


benwtrent:
retest

droberts195:
The PR build failure is the same problem that happened in the debug build on master:

```
Undefined symbols for architecture x86_64:
  "__ZN2ml3api15CForecastRunner33DEFAULT_MAX_FORECAST_MODEL_MEMORYE", referenced from:
      __ZN2ml3api15CForecastRunner31parseAndValidateForecastRequestERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEERNS1_9SForecastElmRKNS2_8functionIFvRKSB_SA_EEE in CForecastRunner.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It must be that with the older version of clang we're using in 7.x it needs the address of the constant even in an optimised build.

The problem is fixed by #1286, and I pushed the PR branch for that to the main repo rather than my fork so you can easily get the changes into this PR.  I think if you fetch origin and then cherry-pick ed12ea4999a5b92cd2315c31b75a11a48ede4237 into your branch for this PR then the PR CI for this should work on 7.x.

PR #1286, opened 2020-05-29
[ML] Fix debug build

droberts195:
The default value is passed to boost::property_tree::ptree::get() as
a reference, which requires taking the address of it.  The new
DEFAULT_MAX_FORECAST_MODEL_MEMORY constant in CForecastRunner was
defined in the class declaration, which means its address cannot be
taken.

This PR moves the definition of DEFAULT_MAX_FORECAST_MODEL_MEMORY
outside of the CForecastRunner class declaration so that its address
can be taken.

We got away with this in optimised builds with the latest compilers
due to inlining and dead code elimination, but in debug builds we
don't.

(I also added a few std:: prefixes to move towards the latest code
style.)

### DIFF
--- a/include/api/CForecastRunner.h
+++ b/include/api/CForecastRunner.h
@@ -79,6 +79,7 @@ public:
     static const std::size_t DEFAULT_EXPIRY_TIME = 14 * core::constants::DAY;
 
     //! max memory allowed to use for forecast models
+    //! (not defined inline because we need its address)
     static const std::size_t DEFAULT_MAX_FORECAST_MODEL_MEMORY;
 
     //! Note: This value measures the size in memory, not the size of the persistence,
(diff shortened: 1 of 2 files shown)


droberts195:
For 7.x the change of this PR can be incorporated into the backport PR #1279, so this PR won't need backporting.

include/api/CForecastRunner.h
```diff
@@ -74,28 +73,28 @@ class API_EXPORT CForecastRunner final : private core::CNonCopyable {
...
-    static const size_t MAX_FORECAST_JOBS_IN_QUEUE = 3;
+    static const std::size_t MAX_FORECAST_JOBS_IN_QUEUE = 3;
 
     //! default expiry time
-    static const size_t DEFAULT_EXPIRY_TIME = 14 * core::constants::DAY;
+    static const std::size_t DEFAULT_EXPIRY_TIME = 14 * core::constants::DAY;
 
     //! max memory allowed to use for forecast models
-    static const size_t DEFAULT_MAX_FORECAST_MODEL_MEMORY = 20971520ull; // 20MB
+    static const std::size_t DEFAULT_MAX_FORECAST_MODEL_MEMORY;
```
tveasey, line 82: I think it is worth adding a comment that this isn't defined inline because we need its address.

PR #1294, opened 2020-06-02
[ML] Report memory usage during DFA job initialization

### LINKED ISSUE #1282: [ML] Report memory usage early
We are already collecting memory usage in the initialization stage in BoostedTreeFactory, now we should also report it here (by calling `CDataFrameAnalysisInstrumentation::flush()` one way or another).

valeriy42:
This PR adds memory usage report in two new cases:
1) right after data frame is loaded in the `DataFrameAnalyzer`
2) in the initialization routines on `CBoostedTreeFactory` when memory usage is updated

Closes #1282

### DIFF
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -131,7 +131,8 @@ void CDataFrameAnalyzer::run() {
         auto& instrumentation = analysisRunner->instrumentation();
         CDataFrameAnalysisInstrumentation::CScopeSetOutputStream setStream{
             instrumentation, *outStream};
-        instrumentation.updateMemoryUsage(static_cast<std::int64_t>(m_DataFrame->memoryUsage()));
+        instrumentation.updateMemoryUsage(
+            static_cast<std::int64_t>(m_DataFrame->memoryUsage()));
         instrumentation.flush();
 
         analysisRunner->run(*m_DataFrame);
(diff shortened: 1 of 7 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -131,6 +131,9 @@ void CDataFrameAnalyzer::run() {
         auto& instrumentation = analysisRunner->instrumentation();
         CDataFrameAnalysisInstrumentation::CScopeSetOutputStream setStream{
             instrumentation, *outStream};
+        instrumentation.updateMemoryUsage(
+            static_cast<std::int64_t>(m_DataFrame->memoryUsage()));
+        instrumentation.flush();
```
tveasey, line 136: Since we're accounting for the frame's memory usage here we need to update to only consider the delta when resizing later. For example at [this](https://github.com/elastic/ml-cpp/blob/master/lib/maths/CBoostedTreeFactory.cc#L280) line you'd need to do something like:
```
std::size_t numberLossParameters{m_TreeImpl->m_Loss->numberParameters()};
std::size_t lastMemoryUsage{core::CMemory::dynamicSize(frame)};
m_TreeImpl->m_ExtraColumns = frame.resizeColumns(
    m_TreeImpl->m_NumberThreads, extraColumns(numberLossParameters));
m_TreeImpl->m_Instrumentation->updateMemoryUsage(
    core::CMemory::dynamicSize(frame) - lastMemoryUsage);
```
and similar changes in `COutliers.cc`.

valeriy42: Good catch! :+1: I fixed double counting in fa785e2

valeriy42:
@dimitris-athanasiou a test is failing, because now I am reporting back memory before I have a proper "regression_stats". Is it required in Java to have both "memory_usage_stats" and e.g. "regression_stats" in the result? 

If there are no concerns against it, I would relax this requirement and change the schema for both parts to be optional.

valeriy42:
Thank you for the review @tveasey . Could you please take another look?

PR #1297, opened 2020-06-03
[ML] Improve seasonal component test parameterisation

tveasey:
There is a short window and a long window to test for "fast" and "slow" seasonal components, respectively. Currently, these are not particularly well parameterised as a function of the job's bucket length: 
1. The parameterisation isn't tuned for very short or long job bucket lengths. This can degrade both the speed with which seasonality is detected and increase error in the components' estimated periods.
2. The window bucket lengths aren't a multiple of the job's bucket length for all the common cases. This causes beating in the assignment of values to buckets which can degrade detection and lead to poor component initialisation. 
3. There is redundancy: the sets of bucket lengths the two windows use "intersect"; this wastes time by running the same tests for both the short and long windows when the bucket lengths match.
4. The common long periodic component is annual and the tests aren't optimised for detecting this.

To address the first two deficiencies this change switches to using appropriate parameters for each of the common job bucket lengths. These choices ensure that the short and long windows don't have any overlap in the bucket lengths they consider. Finally, it uses better parameterisation for detecting annual periodicity and introduces a small selection bias when the estimated period is nearly one year.

I've marked as a non-issue because this is a feature branch.

### DIFF
--- a/lib/maths/CPeriodicityHypothesisTests.cc
+++ b/lib/maths/CPeriodicityHypothesisTests.cc
@@ -2370,7 +2370,15 @@ CPeriodicityHypothesisTests::CNestedHypotheses::CBuilder::finishedNested() {
 namespace {
 
 //! Apply selection bias for known common periods.
+//!
+//! The estimated repeat can be in error. The larger the error the less effective
+//! the modelling will be, since the pattern will precess over time. We therefore
+//! bias towards selecting known common periods by applying a multiplier > 1 to
+//! their correlation.
 double applySelectionBias(std::size_t periodBuckets, core_t::TTime bucketLength, double correlation) {
+    // Daily and weekly periods are explicitly handled when testing for periodicity
+    // and we add calendar predictive features to deal with cyclic monthly effects.
+    // This therefore only biases for annual periodicity.
     core_t::TTime period{static_cast<core_t::TTime>(periodBuckets) * bucketLength};
     if (std::abs(period - core::constants::YEAR) <= bucketLength) {
         return 1.1 * correlation;
@@ -2378,7 +2386,7 @@ double applySelectionBias(std::size_t periodBuckets, core_t::TTime bucketLength,
     return correlation;
 }
 
-//! Compute the period snapping close periods to the selection bias.
+//! If the period is close snap it to a common period.
 core_t::TTime selectPeriod(std::size_t periodBuckets, core_t::TTime bucketLength) {
     core_t::TTime period{static_cast<core_t::TTime>(periodBuckets) * bucketLength};
     if (std::abs(period - core::constants::YEAR) <= bucketLength) {
--- a/lib/maths/CPeriodicityHypothesisTests.cc
+++ b/lib/maths/CPeriodicityHypothesisTests.cc
@@ -561,7 +561,7 @@ void CPeriodicityHypothesisTestsResult::trend(CTrendHypothesis value) {
     m_Trend = value;
 }
 
-void CPeriodicityHypothesisTestsResult::removeTrend(TFloatMeanAccumulatorVec& values) const {
+void CPeriodicityHypothesisTestsResult::removeTrendFrom(TFloatMeanAccumulatorVec& values) const {
     if (m_Trend.type() == CTrendHypothesis::E_Linear) {
         removeLinearTrend(values);
     } else if (m_Trend.type() == CTrendHypothesis::E_PiecewiseLinear) {
@@ -570,7 +570,7 @@ void CPeriodicityHypothesisTestsResult::removeTrend(TFloatMeanAccumulatorVec& va
     }
 }
 
-void CPeriodicityHypothesisTestsResult::removeDiscontinuities(TFloatMeanAccumulatorVec& values) const {
+void CPeriodicityHypothesisTestsResult::removeDiscontinuitiesFrom(TFloatMeanAccumulatorVec& values) const {
     if (m_Trend.type() == CTrendHypothesis::E_PiecewiseLinear) {
         TSizeVec segmentation(CTimeSeriesSegmentation::piecewiseLinear(values));
         values = CTimeSeriesSegmentation::removePiecewiseLinearDiscontinuities(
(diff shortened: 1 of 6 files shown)


lib/maths/CPeriodicityHypothesisTests.cc
```diff
@@ -2369,6 +2369,24 @@ CPeriodicityHypothesisTests::CNestedHypotheses::CBuilder::finishedNested() {
 
 namespace {
 
+//! Apply selection bias for known common periods.
+double applySelectionBias(std::size_t periodBuckets, core_t::TTime bucketLength, double correlation) {
+    core_t::TTime period{static_cast<core_t::TTime>(periodBuckets) * bucketLength};
+    if (std::abs(period - core::constants::YEAR) <= bucketLength) {
```
droberts195, line 2375: Is it worth adding a comment to say this is being done for years because years vary in length?  Or if that isn't the reason then it's not clear why this is done for year but not more common periods.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -472,13 +476,130 @@ std::size_t CTimeSeriesDecompositionDetail::CMediator::memoryUsage() const {
...
+    static const TTimeVec& testSchedule(int window, core_t::TTime bucketLength) {
+        return windowParameters(window, bucketLength)->s_TestSchedule;
+    }
+
+    static core_t::TTime shortestComponent(int window, core_t::TTime bucketLength) {
+        return windowParameters(window, bucketLength)->s_ShortestComponent;
+    }
+
+private:
+    struct CParameters {
```
droberts195, line 515: nit: `struct` names should begin with `S`

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -270,6 +270,10 @@ const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_4_TAG("6.4");
 
 // Periodicity Test Tags
+// Version 7.9
+const core::TPersistenceTag SHORT_WINDOW_7_9_TAG{"e", "short_window"};
+const core::TPersistenceTag LONG_WINDOW_7_9_TAG{"f", "long_window"};
```
droberts195, line 275: I think it's a bad idea that the long names here are identical to the ones used in `SHORT_WINDOW_6_3_TAG` and `LONG_WINDOW_6_3_TAG`.  It may not cause a problem immediately, but if the model visualizer is taken further it might in the future.

lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -502,14 +623,16 @@ bool CTimeSeriesDecompositionDetail::CPeriodicityTest::acceptRestoreTraverser(
...
-            SHORT_WINDOW_6_3_TAG, m_Windows[E_Short].reset(this->newWindow(E_Short)),
+            SHORT_WINDOW_7_9_TAG, m_Windows[E_Short] = this->newWindow(E_Short),
             m_Windows[E_Short] && traverser.traverseSubLevel(std::bind(
                                       &CExpandingWindow::acceptRestoreTraverser,
                                       m_Windows[E_Short].get(), std::placeholders::_1)),
             /**/)
+        RESTORE_NO_ERROR(LONG_WINDOW_6_3_TAG, m_Windows[E_Long] = this->newWindow(E_Long))
         RESTORE_SETUP_TEARDOWN(
-            LONG_WINDOW_6_3_TAG, m_Windows[E_Long].reset(this->newWindow(E_Long)),
+            LONG_WINDOW_7_9_TAG, m_Windows[E_Long] = this->newWindow(E_Long),
```
droberts195, line 635: Short and long windows from old versions are ignored, but the format hasn't changed.  So is the intention just to reinitialise from scratch when 7.9 is used for the first time because we think the persisted windows are sub-optimal?  Will this have any impact on anomaly detection?  Will every user who upgrades to 7.9 see strange effects in the model plot immediately after upgrading?

It would be good to have comments here to say what the intention is.

tveasey: This is intentional because otherwise we'll just use the old poor parameterisation for existing jobs. Since the window slides (after it has fully expanded) this just means we loose the temporary state we'd accumulated in the window to the current time, so this seemed worthwhile.

lib/maths/CPeriodicityHypothesisTests.cc
tveasey: The reason for this is that we estimate the period, via maximum cyclic autocorrelation, and this estimate can be in error. If the estimated period is near a year it is likely that the period is actually one year. It is advantageous to use a period as close as possible to the true period since this minimises precession.

Other common periods are already handled: daily and weekly are dealt with explicitly when testing for seasonal components. I debated treating average monthly periodicity as special case as well, but we have a variety of predictive calendar features. These are usually more suitable, since the variation from month to month in the monthly period is relatively large.

tveasey: I'll add a comment.

tveasey: See [93cbad3](https://github.com/elastic/ml-cpp/pull/1297/commits/93cbad37cbef56122803d5743d9ef6989eef8fd0).

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: See [93cbad3](https://github.com/elastic/ml-cpp/pull/1297/commits/93cbad37cbef56122803d5743d9ef6989eef8fd0).

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: See [93cbad3](https://github.com/elastic/ml-cpp/pull/1297/commits/93cbad37cbef56122803d5743d9ef6989eef8fd0).

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: See [93cbad3](https://github.com/elastic/ml-cpp/pull/1297/commits/93cbad37cbef56122803d5743d9ef6989eef8fd0).

tveasey:
Thanks for the review @droberts195. I've addressed all your comments. Can you take another look?

PR #1298, opened 2020-06-03
[ML] Memory limit circuit breaker

valeriy42:


docs/CHANGELOG.asciidoc
```diff
@@ -46,6 +46,8 @@
   (See {ml-pull}1238[#1238] and {pull}57254[#57254].)
 * Parallelize the feature importance calculation for classification and regression
   over trees. (See {ml-pull}1277[#1277].)
+* More realistic memory estimation requires lower memory limits than before
```
tveasey, line 49: I think it is worth mentioning classification and regression specifically since other analyses aren't changed:
```suggestion
* More realistic memory estimation for classification and regression means that these analyses will require
  lower memory limits than before.
```

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -64,10 +67,17 @@ const std::string PROGRESS_PERCENT{"progress_percent"};
 
 const std::size_t MAXIMUM_FRACTIONAL_PROGRESS{std::size_t{1}
                                               << ((sizeof(std::size_t) - 2) * 8)};
+
+std::int64_t bytesToMb(std::int64_t bytes) {
+    std::int64_t mb{static_cast<std::int64_t>(std::ceil(static_cast<double>(bytes) / 1048576))};
```
tveasey, line 72: I'm not 100% this wouldn't generate a warning for integer division by 1048576, but feels cleaner to just implement as integer division, i.e. this is equivalent to:
```suggestion
    return (bytes + 1048575) / 1048576);
```

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -153,6 +162,18 @@ void CDataFrameAnalysisInstrumentation::monitorProgress(
             lastProgress = progress;
             writeProgress(lastTask, lastProgress, &writer);
         }
+        LOG_DEBUG(<< "Memory usage " << instrumentation.memory());
```
tveasey, line 165: This will generate quite a lot of noise in unit tests, I'd suggest downgrading to LOG_TRACE.
```suggestion
        LOG_TRACE(<< "Memory usage " << instrumentation.memory());
```

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -153,6 +162,18 @@ void CDataFrameAnalysisInstrumentation::monitorProgress(
...
+        if (instrumentation.memory() > instrumentation.m_MemoryLimit) {
+            LOG_DEBUG(<< "Memory limit triggered");
+            HANDLE_FATAL(<< "Environment error: required memory "
+                         << instrumentation.memory() << " exceeds the memory limit "
+                         << instrumentation.m_MemoryLimit << ". Please increase the limit to at least "
+                         << static_cast<std::int64_t>(
+                                static_cast<double>(instrumentation.m_MemoryLimit) * MEMORY_LIMIT_INCREMENT)
+                         << " and restart.");
+            // TODO clean up this message and output the information in mb instead of bytes
```
tveasey, line 174: I think this is done no?
```suggestion
```

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -153,6 +162,18 @@ void CDataFrameAnalysisInstrumentation::monitorProgress(
             lastProgress = progress;
             writeProgress(lastTask, lastProgress, &writer);
         }
+        LOG_DEBUG(<< "Memory usage " << instrumentation.memory());
+        if (instrumentation.memory() > instrumentation.m_MemoryLimit) {
+            LOG_DEBUG(<< "Memory limit triggered");
```
tveasey, line 167: Don't think this is needed with the other
```suggestion
```

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1306,7 +1306,7 @@ BOOST_AUTO_TEST_CASE(testEstimateMemoryUsedByTrain) {
         BOOST_TEST_REQUIRE(instrumentation.maxMemoryUsage() < estimatedMemory);
     }
 }
-
+// TODO: test for handle fatal (memory-limit-circuit-breaker)
```
tveasey, line 1309: I think it is worth adding this test as part of this change. You just need to override the fatal handler so it doesn't call `std::terminate`, i.e.
```
TStrVec errors;
auto errorHandler = [&errors](std::string error) { errors.push_back(error); };
core::CLogger::CScopeSetFatalErrorHandler scope{errorHandler};
```
See for example `CDataFrameAnalysisSpecificationTest/testCreate`

docs/CHANGELOG.asciidoc
valeriy42: Updated

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: I removed the trace completely. IMO it has limited benefit and produces a lot of noise.

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: Oh, this is a nice trick!

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: I changed it in the new commit 335ec7b7

lib/api/CDataFrameAnalysisInstrumentation.cc
valeriy42: Deleted

lib/maths/unittest/CBoostedTreeTest.cc
valeriy42: Thank you for the tip. I created tests in `CDataFrameAnalyzerTrainingTest` and `CDataFrameAnalyzerOutlierTest` to check for errors.

valeriy42:
Thank you @tveasey for the comments. I addressed your suggestions. It would be great if you could have another pass when you have time.

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -64,10 +69,32 @@ const std::string PROGRESS_PERCENT{"progress_percent"};
...
+    }
+    value /= 1024;
+    stream << value;
+    stream << " mb";
+    return stream.str();
+}
+
+std::string bytesToString(std::int64_t bytes) {
+    return bytesToString(static_cast<double>(bytes));
+}
```
tveasey, line 91: 👍 . (Almost feels like this is worth factoring out and putting somewhere. I'm sure there are other cases where we want to print memory usage. Not sure where this belongs though so something for another change.)

valeriy42:
Retest

PR #1300, opened 2020-06-05
[ML] Add state checkpoints during coarse parameter search for classification and regression

### LINKED ISSUE #1275: [ML] Persist/restore state during the coarse parameter search step of boosted tree training
Currently, we only persist state to allow us to restore from failure (or the user stopping the data frame analysis) after fine tuning of hyperparameter starts.

We could easily store sufficient information to be able to resume part way through coarse parameter search and I've seen this step cost up to 30% of our total training time.

tveasey:
Searching for initial parameters for boosted tree training is an expensive operation and it is possible to save state to allow this to resume with minimal loss of work. This PR adds checkpointing of sufficient state to resume after failure during this step. It also makes some tweaks to `CPersistUtils`. These include moving values into place where possible and adding some convenience functions to persist pointers.

The basic strategy is to introduce an initialisation stage variable to `CBoostedTreeImpl`. This tracks where initialisation reached at each checkpoint, so that it is possible to resume, skipping stages which have already been performed. It is necessary to additionally persist and restore some factory state.

Closes #1275.

### DIFF
--- a/include/core/CPersistUtils.h
+++ b/include/core/CPersistUtils.h
@@ -334,19 +334,20 @@ public:
         mutable std::string m_Token;
     };
 
-    //! Entry method for objects being restored
+    //! Restore \p collection reading state from \p inserter expecting \p tag for elements.
     template<typename T>
     static bool restore(const std::string& tag, T& collection, CStateRestoreTraverser& traverser) {
         return persist_utils_detail::restore(tag, collection, traverser);
     }
 
-    //! Entry method for objects being persisted
+    //! Persist \p collection passing state to \p inserter using \p tag for elements.
     template<typename T>
     static bool
     persist(const std::string& tag, const T& collection, CStatePersistInserter& inserter) {
         return persist_utils_detail::persist(tag, collection, inserter);
     }
 
+    //! Persist \p collection passing state to \p inserter using \p tag for elements.
     template<typename T>
     static bool
     persist(const TPersistenceTag& tag, const T& collection, CStatePersistInserter& inserter) {
@@ -603,7 +604,7 @@ private:
                 LOG_ERROR(<< "Invalid state " << state);
                 return false;
             }
-            *inserter = element;
+            *inserter = std::move(element);
             ++inserter;
             return true;
         }
@@ -617,7 +618,7 @@ private:
                 LOG_ERROR(<< "Invalid element 0 : element " << token << " in " << state);
                 return false;
             }
-            *inserter = element;
+            *inserter = std::move(element);
             ++inserter;
         }
 
@@ -637,7 +638,7 @@ private:
                           << " in " << state);
                 return false;
             }
-            *inserter = element;
+            *inserter = std::move(element);
 
             ++i;
             lastDelimPos = delimPos;
@@ -936,7 +937,7 @@ private:
                         LOG_ERROR(<< "Restoration error at " << traverser.name());
                         return false;
                     }
-                    container.insert(container.end(), value);
+                    container.insert(container.end(), std::move(value));
                 }
             } while (traverser.next());
             return true;
@@ -953,7 +954,7 @@ private:
                         LOG_ERROR(<< "Restoration error at " << traverser.name());
                         return false;
                     }
-                    *(i++) = value;
+                    *(i++) = std::move(value);
                 }
             } while (traverser.next());
             return true;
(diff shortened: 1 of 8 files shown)


include/core/CPersistUtils.h
```diff
@@ -334,24 +334,48 @@ class CORE_EXPORT CPersistUtils {
         mutable std::string m_Token;
     };
 
-    //! Entry method for objects being restored
+    //! Restore \p collection reading state from \p inserter expecting \p tag for elements.
```
valeriy42, line 337:
```suggestion
    //! Restore \p collection reading state from \p traverser expecting \p tag for elements.
```

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -120,10 +135,22 @@ CBoostedTreeFactory::restoreFor(core::CDataFrame& frame, std::size_t dependentVa
...
+    switch (m_TreeImpl->m_InitializationStage) {
+    case CBoostedTreeImpl::E_FullyInitialized:
+        break;
+    case CBoostedTreeImpl::E_NotInitialized:
+    case CBoostedTreeImpl::E_SoftTreeDepthLimitInitialized:
+    case CBoostedTreeImpl::E_DepthPenaltyMultiplierInitialized:
+    case CBoostedTreeImpl::E_TreeSizePenaltyMultiplierInitialized:
+    case CBoostedTreeImpl::E_LeafWeightPenaltyMultiplierInitialized:
+    case CBoostedTreeImpl::E_DownsampleFactorInitialized:
+    case CBoostedTreeImpl::E_EtaInitialized:
```
valeriy42, line 147: Are you actually setting all these stage values somewhere?

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1647,21 +1647,96 @@ BOOST_AUTO_TEST_CASE(testPersistRestore) {
...
+                               .numberFolds(2)
+                               .maximumNumberTrees(2)
+                               .maximumOptimisationRoundsPerHyperparameter(3)
+                               .trainingStateCallback(writeCheckpoint)
+                               .buildFor(*frame, cols - 1);
+        core::CJsonStatePersistInserter inserter(expectedState);
+        boostedTree->acceptPersistInserter(inserter);
+        expectedState.flush();
+    }
+
```
valeriy42, line 1726: Nice test overall, I think you need to check  that `checkpoints` contains the expected number of elements, otherwise mot may be empty and the loop below is useless.

tveasey: Great point!

tveasey:
Thanks @valeriy42 !

>  I've seen only Uninitialized and FullyInitialized. Did I miss it?

It happens in skipCheckpointIfAtOrAfter, i.e. at [this](https://github.com/elastic/ml-cpp/pull/1300/files#diff-192fe6f9f53bb6f2064155bc0f9c083aR1333) line.

lib/maths/CBoostedTreeFactory.cc
tveasey: As below, this happens in `skipCheckpointIfAtOrAfter`.

lib/maths/unittest/CBoostedTreeTest.cc
tveasey: See [ea2299c](https://github.com/elastic/ml-cpp/pull/1300/commits/ea2299cdcea8c00055b68725b31dc153580fafc2).

include/core/CPersistUtils.h
tveasey: See [f6804a0](https://github.com/elastic/ml-cpp/pull/1300/commits/f6804a083088665fb4598869e321a17020d6277b).

lib/maths/unittest/CBoostedTreeTest.cc
tveasey: Note, I decided to just add a test that there is at least one checkpoint so we don't fail on changes to when we checkpoint.

tveasey:
Thanks for the review @valeriy42! I've addressed all your comments. Could you take another look?

tveasey:
retest

tveasey:
Aarch64 CI seems to be unavailable at the moment. I'm going to go ahead and merge this anyway, since it already passed in an earlier version, and this is still unsupported.

PR #1305, opened 2020-06-09
[ML] Tailor the data frame slice size per analysis type

tveasey:
When a data frame is sliced by memory it should have been accounting for the additional columns that each analysis type will require. Additionally, now we support threading analyses, it is useful to set the slice size based on the thread count for boosted tree training, since we parallelise work by splitting it over the data frame slices.

I'm marking this as a non-issue because it is a low level detail which is mainly only visible when threading is enabled for classification and regression. Since this is new in 7.9, I don't think it needs documenting separately.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeRunner.cc
@@ -187,6 +187,19 @@ std::size_t CDataFrameTrainBoostedTreeRunner::numberExtraColumns() const {
     return m_BoostedTreeFactory->numberExtraColumnsForTrain();
 }
 
+std::size_t CDataFrameTrainBoostedTreeRunner::dataFrameSliceCapacity() const {
+    std::size_t sliceCapacity{core::dataFrameDefaultSliceCapacity(
+        this->spec().numberColumns() + this->numberExtraColumns())};
+    std::size_t numberThreads{this->spec().numberThreads()};
+    if (numberThreads > 1) {
+        // Use at least one slice per thread because we parallelize work over slices.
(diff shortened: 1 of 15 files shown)


lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -187,6 +187,25 @@ std::size_t CDataFrameTrainBoostedTreeRunner::numberExtraColumns() const {
...
+        std::size_t numberRows{this->spec().numberRows()};
+
+        // Use at least one slice per thread because we parallelize work over slices.
+        std::size_t capacityForOneSlicePerThread{(numberRows + numberThreads - 1) / numberThreads};
+        sliceCapacity = std::min(sliceCapacity, capacityForOneSlicePerThread);
+
+        // Round the slice size so number threads is a divisor of the number of slices.
+        std::size_t numberSlices{numberRows / sliceCapacity};
+        sliceCapacity = numberRows /
+                        (numberThreads * ((numberSlices + numberThreads / 2) / numberThreads));
```
valeriy42, line 204: Why do you need this, why is it not enough to have `sliceCapacity` as computed in L199?

tveasey: What this is trying to do is to choose a slice size for which `"rows" / "slice size" mod "number threads" = 0` (1). (Note that L199 doesn't guaranty this because it's possible that `sliceCapacity < capacityForOneSlicePerThread`.) We distribute work by assigning each thread `"slices" / "number threads"` slices to process, so this condition ensures we have nearly equal numbers of rows for each slice to process, i.e. it is effectively load balancing.

PR #1312, opened 2020-06-11
[ML] Improve data access pattern computing best splits for classification and regression

tveasey:
The feature derivatives for boosted tree splits are laid out in increasing feature index order, but when we sample a feature bag we don't ensure it is sorted. This means we are getting random access rather than linear scan over the derivatives when computing the best split. This switches to sorting the feature bag after sampling and rejigs the function signatures to avoid allocating a vector for every node.

### DIFF
--- a/lib/maths/CSampling.cc
+++ b/lib/maths/CSampling.cc
@@ -200,7 +200,7 @@ void doCategoricalSampleWithoutReplacement(RNG& rng,
     }
 
     // Construct the transform function.
-    for (std::size_t i = 1; i < probabilities.size(); ++i) {
+    for (std::size_t i = 1; i < m; ++i) {
         probabilities[i] += probabilities[i - 1];
     }
 
@@ -226,9 +226,8 @@ void doCategoricalSampleWithoutReplacement(RNG& rng,
         }
     }
 
-    // The sampled values are at the end of the vector so copy into place.
-    std::copy(result.begin() + m, result.end(), result.begin());
-    result.resize(n);
+    // The sampled values are at the end of the vector.
+    result.erase(result.begin(), result.begin() + m);
 }
 
 //! Implementation of multivariate normal sampling.
--- a/lib/maths/CSampling.cc
+++ b/lib/maths/CSampling.cc
@@ -153,25 +153,25 @@ void doCategoricalSampleWithReplacement(RNG& rng,
         return;
     }
 
-    std::size_t p = probabilities.size();
+    std::size_t m{probabilities.size()};
 
     // Construct the transform function.
-    for (std::size_t i = 1u; i < p; ++i) {
+    for (std::size_t i = 1; i < m; ++i) {
         probabilities[i] += probabilities[i - 1];
     }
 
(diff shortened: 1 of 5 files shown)


tveasey:
retest

tveasey:
Thanks for the review @droberts195! While I was in the area I realised:
1. I could avoid all allocations in categorical sample without replacement.
2. There was a bug in the case more samples than values were asked for: missing return. (This wasn't actually being exercised in the code, but it is clearly worth fixing.)

I also added a unit test. I decided it isn't worth breaking this out into a separate change (it is also essentially related to trying to improve cache performance), but it is worth also having a look at [e548cc6](https://github.com/elastic/ml-cpp/pull/1312/commits/e548cc632002991f8b556fa7439f845de0bf67ca0)

lib/maths/CSampling.cc
```diff
@@ -191,45 +191,44 @@ void doCategoricalSampleWithoutReplacement(RNG& rng,
...
+        for (std::size_t j = s + 1; j < m; ++j) {
+            probabilities[j - 1] = probabilities[j] - ps;
+            std::swap(result[j - 1], result[j]);
         }
-        indices.erase(indices.begin() + s[0]);
     }
+
+    // The sampled values are at the end of the vector so copy into place.
+    std::copy(result.begin() + m, result.end(), result.begin());
+    result.resize(n);
```
droberts195, line 231: Would `result.erase(result.begin(), result.begin() + m);` be less efficient than these two lines?  It seems to me it would be doing basically the same thing but in one line.

lib/maths/CSampling.cc
```diff
@@ -191,45 +191,44 @@ void doCategoricalSampleWithoutReplacement(RNG& rng,
...
+    result.assign(boost::counting_iterator<std::size_t>(0),
+                  boost::counting_iterator<std::size_t>(m));
+
+    if (n >= m) {
+        return;
     }
 
     // Construct the transform function.
-    for (std::size_t i = 1u; i < p; ++i) {
+    for (std::size_t i = 1; i < probabilities.size(); ++i) {
```
droberts195, line 203: In other places you switched from `probabilities.size()` to `m`, but here you've switched the other way.

lib/maths/CSampling.cc
tveasey: It guess it is implementation dependent, but any sane implementation would do what I do with the two lines...

Actually, the standard requires it to be "Linear on the number of elements erased (destructions) plus the number of elements after the last element deleted (moving).". This pretty much forces the hand of the implementation so should be as fast. Seems like it is worth tidying.

tveasey: See [d957489](https://github.com/elastic/ml-cpp/pull/1312/commits/d957489ee19d6eedc001373c3fac354aec037986).

lib/maths/CSampling.cc
tveasey: See [d957489](https://github.com/elastic/ml-cpp/pull/1312/commits/d957489ee19d6eedc001373c3fac354aec037986).

PR #1316, opened 2020-06-15
[ML] Generate memory size info for compressed model definition

valeriy42:
This PR adds the ability to generate  `model_size_info` objects used in Java to estimate the heap size for the inference models. I added a slim subclass `CSizeInfo` to all relevant classes which contains reference to the definition objects and generates Json with the necessary information when called. 

We do a double work for counting nodes and leaves of the tree when computing `numOperations` and generating JSON. I did it to avoid memory overhead.

I am not writing the `model_size_info` in the results yet, I intend to do it in a subsequent PR.  Hence this PR is marked as a non-issue.


include/api/CInferenceModelDefinition.h
```diff
@@ -113,8 +113,21 @@ class API_EXPORT CTrainedModel : public CSerializableToJson {
...
+        virtual std::size_t numOperations() const = 0;
+
+    private:
+        const CTrainedModel& m_TrainedModel;
+    };
+    using TSizeInfoUPtr = std::unique_ptr<CSizeInfo>;
+
 public:
-    ~CTrainedModel() override = default;
+    virtual ~CTrainedModel() override = default;
```
tveasey, line 130: nit: I don't think this is needed since the implicit destructor is virtual if the base class destructor is virtual:
```suggestion
```

include/api/CInferenceModelDefinition.h
```diff
@@ -229,6 +269,23 @@ class API_EXPORT CEnsemble final : public CTrainedModel {
 };
 
 class API_EXPORT CEncoding : public CSerializableToJson {
+public:
+    class CSizeInfo : public CSerializableToJson {
+    public:
+        ~CSizeInfo() override = default;
+
```
tveasey, line 276: As before:
```suggestion
```

include/api/CInferenceModelDefinition.h
```diff
@@ -247,6 +306,21 @@ class API_EXPORT CEncoding : public CSerializableToJson {
...
 class API_EXPORT CFrequencyEncoding final : public CEncoding {
 public:
+    class CSizeInfo final : public CEncoding::CSizeInfo {
+    public:
+        explicit CSizeInfo(const CFrequencyEncoding& encoding);
+        void addToDocument(rapidjson::Value& parentObject, TRapidJsonWriter& writer) const override;
+        const std::string& typeString() const override;
+
+    private:
+        using TSizeVec = std::vector<std::size_t>;
```
tveasey, line 316: nit: how about moving this typedef into protected `CEncoding::CSizeInfo` to save duplicating.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -227,6 +285,50 @@ void CEnsemble::classificationWeights(TDoubleVec classificationWeights) {
...
+        numOperations += model->sizeInfo()->numOperations();
+    }
+    return numOperations;
+}
+
+void CEnsemble::CSizeInfo::addToDocument(rapidjson::Value& parentObject,
+                                         TRapidJsonWriter& writer) const {
+    this->CTrainedModel::CSizeInfo::addToDocument(parentObject, writer);
+    rapidjson::Value featureNameLengthsArray =
+        writer.makeArray(m_Ensemble->featureNames().size());
```
tveasey, line 308: You assign to `rapidjson::Value` in quite a few places rather than using `{}`-initialisation, but there isn't a problem with this right? I'd prefer:
```suggestion
    rapidjson::Value featureNameLengthsArray{
        writer.makeArray(m_Ensemble->featureNames().size())};
```
for consistency.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -408,6 +533,62 @@ void CInferenceModelDefinition::dependentVariableColumnIndex(std::size_t depende
...
+    return std::make_unique<CSizeInfo>(*this);
+}
+
+CInferenceModelDefinition::CSizeInfo::CSizeInfo(const CInferenceModelDefinition& definition)
+    : m_Definition{definition} {
+}
+
+std::string CInferenceModelDefinition::CSizeInfo::jsonString() {
+    std::ostringstream stream;
+    {
```
tveasey, line 546: I think it is worth a comment to say this scope block is to finish writing the object in the `CJsonOutputStreamWrapper` destructor.

tveasey: Although you remove them afterwards so is this necessary: maybe remove this and update `std::string resultString(jsonStr, 1, jsonStr.size() - 2)` to `std::string resultString(jsonStr, 1, jsonStr.size() - 1)`?

lib/api/CInferenceModelDefinition.cc
```diff
@@ -408,6 +533,62 @@ void CInferenceModelDefinition::dependentVariableColumnIndex(std::size_t depende
...
+}
+
+void CInferenceModelDefinition::CSizeInfo::addToDocument(rapidjson::Value& parentObject,
+                                                         TRapidJsonWriter& writer) const {
+    // parse preprocessing
+    TEncodingSizeUPtrVec encodingSizeItems;
+    encodingSizeItems.reserve(m_Definition.preprocessors().size());
+    for (const auto& preprocessor : m_Definition.preprocessors()) {
+        encodingSizeItems.push_back(preprocessor->sizeInfo());
+    }
```
tveasey, line 567: Is there any reason to copy these into a vector? It strikes me you could change:
```
for (const auto& encoding : encodingSizeItems) {
```
to 
```
for (const auto& preprocessor : m_Definition.preprocessors()) {
    auto encodingSizeInfo = preprocessor->sizeInfo();
```

lib/api/CInferenceModelDefinition.cc
```diff
@@ -446,6 +627,37 @@ const CTargetMeanEncoding::TStringDoubleUMap& CTargetMeanEncoding::targetMap() c
     return m_TargetMap;
 }
 
+CTargetMeanEncoding::CSizeInfo::CSizeInfo(const CTargetMeanEncoding& encoding)
+    : CSizeInfo(&encoding) {
+}
+
+CTargetMeanEncoding::CSizeInfo::CSizeInfo(const CTargetMeanEncoding* encoding)
+    : CEncoding::CSizeInfo::CSizeInfo(encoding), m_Encoding{*encoding} {
+}
```
tveasey, line 636: why not?
```suggestion
CTargetMeanEncoding::CSizeInfo::CSizeInfo(const CTargetMeanEncoding& encoding)
    : CEncoding::CSizeInfo::CSizeInfo{&encoding}, m_Encoding{encoding} {
}
```

lib/api/CInferenceModelDefinition.cc
```diff
@@ -491,10 +719,45 @@ const CFrequencyEncoding::TStringDoubleUMap& CFrequencyEncoding::frequencyMap()
     return m_FrequencyMap;
 }
 
+CFrequencyEncoding::CSizeInfo::CSizeInfo(const CFrequencyEncoding& encoding)
+    : CSizeInfo(&encoding) {
+}
+
+CFrequencyEncoding::CSizeInfo::CSizeInfo(const CFrequencyEncoding* encoding)
+    : CEncoding::CSizeInfo::CSizeInfo(encoding), m_Encoding{*encoding} {
+}
```
tveasey, line 728: As above. In fact, I have the same comment for all these concrete encoding class objects.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -509,10 +772,41 @@ void COneHotEncoding::addToDocument(rapidjson::Value& parentObject,
     writer.addMember(JSON_HOT_MAP_TAG, hotMap, parentObject);
 }
 
+COneHotEncoding::CSizeInfo::CSizeInfo(const COneHotEncoding& encoding)
+    : CSizeInfo(&encoding) {
+}
```
tveasey, line 777: Maybe keep with rest of `COneHotEncoding::CSizeInfo` definitions.

lib/api/unittest/CBoostedTreeInferenceModelBuilderTest.cc
```diff
@@ -117,46 +118,120 @@ BOOST_AUTO_TEST_CASE(testIntegrationRegression) {
...
-            BOOST_REQUIRE_EQUAL(std::size_t(3), enc->hotMap().size());
-            BOOST_TEST_REQUIRE("categorical_col_cat1" == enc->hotMap()["cat1"]);
-            BOOST_TEST_REQUIRE("categorical_col_cat2" == enc->hotMap()["cat2"]);
-            BOOST_TEST_REQUIRE("categorical_col_cat3" == enc->hotMap()["cat3"]);
-            oneHot = true;
+    LOG_DEBUG(<< "Inference model definition: " << definition->jsonString());
+    std::string modelSizeDefinition{definition->sizeInfo()->jsonString()};
+    LOG_DEBUG(<< "Model size definition: " << modelSizeDefinition);
+
+    // verify model definition
```
tveasey, line 125: 👍

include/api/CInferenceModelDefinition.h
valeriy42: This is good to know! :rocket:  Had to dig deep into the standard to verify this. :monocle_face:

include/api/CInferenceModelDefinition.h
valeriy42: Done

include/api/CInferenceModelDefinition.h
valeriy42: Good catch!

lib/api/CInferenceModelDefinition.cc
valeriy42: I was consistently inconsistent in this file :see_no_evil: It is fixed now.

lib/api/CInferenceModelDefinition.cc
valeriy42: I aded the comment, but if I remove the scope and change `-2` to `-1` I fail the tests.

lib/api/CInferenceModelDefinition.cc
valeriy42: You are right, it further simplifies things.

lib/api/CInferenceModelDefinition.cc
valeriy42: Good point. I fixed this.

valeriy42:
Thank you @tveasey for your comments. I tried to integrate all of them in the code. It would be great if you could make another pass.

lib/api/CInferenceModelDefinition.cc
tveasey: That seems weird to me since I thought all the destructor did was close the array, but maybe some detail of rapidjson, but happy to leave as is.

include/api/CInferenceModelDefinition.h
tveasey: You missed this one :).

lib/api/CInferenceModelDefinition.cc
droberts195: A better option might be to add an optional argument to the `CJsonOutputStreamWrapper` constructor to say whether to output JSON or ND-JSON.

About 6-7 years ago JSON purists were in the ascendency and ND-JSON was considered impure.  So we got into the situation where our programs output multi-gigabyte JSON arrays just so that their entire output is a valid JSON document.  On the receiving side this has to be parsed using a stream parser rather than waiting for a complete JSON document, so the array structure is completely pointless.

Now ND-JSON is considered the sane way to deal with a stream of JSON documents, so we should use it for new types of output.

It isn't very hard to change `CJsonOutputStreamWrapper` to do this. In ND-JSON mode it would:

- Never write `JSON_ARRAY_START`
- Never write `JSON_ARRAY_END`
- Write `\n` wherever it currently writes `JSON_ARRAY_DELIMITER`

Then you don't need the inefficient string manipulation at the end of this method, and there's a side benefit that one of our core classes encourages the more modern approach for the future.

include/api/CInferenceModelDefinition.h
```diff
@@ -319,11 +412,33 @@ class API_EXPORT CInferenceModelDefinition : public CSerializableToJson {
...
+
+    class API_EXPORT CSizeInfo final : public CSerializableToJson {
+    public:
+        explicit CSizeInfo(const CInferenceModelDefinition& definition);
+        void addToDocument(rapidjson::Value& parentObject, TRapidJsonWriter& writer) const override;
+        std::string jsonString();
+
+    private:
+        using TEncodingSizeUPtr = std::unique_ptr<CEncoding::CSizeInfo>;
+        using TEncodingSizeUPtrVec = std::vector<TEncodingSizeUPtr>;
```
tveasey, line 425: One last thing. I don't think these typedefs are used any more.

lib/api/CInferenceModelDefinition.cc
valeriy42: This is a good point, I will do it in a separate PR though. See #1329

include/api/CInferenceModelDefinition.h
valeriy42: Fixed :+1:

PR #1317, opened 2020-06-16
[ML] Threading speedups for classification and regression

tveasey:
Testing showed up we are often I/O bound during boosted tree training on larger data sets. This change introduces a placeholder for the large aggregate derivatives objects which is allocated once upfront at the start of training a forest. This is particularly relevant for running multithreaded because we use one copy of this data per thread. In order to do this it was necessary to change the `CDataFrame` interface to take in a vector of functions to bind to each task so the placeholders can be bound by reference.

We also now pass in information about the minimum gain of a node that we'll keep. This allows us to lazily create storage for these and avoid some work if they aren't needed. 

Testing also showed up degraded performance running multi-threaded when the thread count was set too high. One factor effecting this is the amount of work performed per thread when computing aggregate derivatives. I've introduced a minimum amount of work we'll perform per thread and this operation uses fewer threads accordingly. 

Finally, I've inlined some functions for which it is worthwhile.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -59,7 +59,9 @@
 * Checkpoint state to allow efficient failover during coarse parameter search
   for classification and regression. (See {ml-pull}1300[#1300].)
 * Improve data access patterns to speed up classification and regression.
-  (See {ml-pull}1312[#1312].) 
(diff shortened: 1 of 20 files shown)


tveasey:
This failed CI with the perennial JVM crash on ARM.

PR #1323, opened 2020-06-17
[ML] Remove volatile and update thread-safe static initialisation

droberts195:
There were a few places in the code where we were inappropriately
using the volatile keyword.  Some of these should have been using
atomics, and others were simply unnecessary.

Additionally, the "magic statics" feature of C++11 finally made it
into Visual Studio 2015.  We were using Visual Studio 2013 in 5.x
and 6.x, so could not use this feature.  However, 7.x is using
Visual Studio 2017, so we can adapt our code to take advantage of
automatic protection of static variable initialisation instead of
having to write custom code for this.

### DIFF
--- a/include/api/CDataFrameAnalysisRunner.h
+++ b/include/api/CDataFrameAnalysisRunner.h
@@ -7,7 +7,6 @@
 #ifndef INCLUDED_ml_api_CDataFrameAnalysisRunner_h
 #define INCLUDED_ml_api_CDataFrameAnalysisRunner_h
 
-#include <core/CFastMutex.h>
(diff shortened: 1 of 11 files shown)


droberts195:
The aarch64 failure was `Exception in thread "main" java.lang.AssertionError: expected version [V_7_8_1] to have id [7080199] but was [7080099]`.  This is due to the test running at the same time as an ES version bump, nothing to do with the changes.  All our C++ unit tests succeeded prior to that error.  Since we have so little aarch64 test hardware I'll merge rather than burn up more CI time.

PR #1328, opened 2020-06-18
[ML] Fix "Unable to compute percentiles: doesn't bracket root..." log errors (part 2)

tveasey:
Fix the last fallout from #1268 by more aggressively expanding the initial bracket interval: we were running into issues when this was very initialised narrow. Also, I decided to downgrade the error to a warning. It should generate a notification, since it can affect the model plot and exception handling mechanism is expensive, but it doesn't affect anomaly detection quality so I don't think this should be raised at error. As before this is a non-issue since #1268 isn't released.

### DIFF
--- a/lib/maths/CMultimodalPrior.cc
+++ b/lib/maths/CMultimodalPrior.cc
@@ -640,6 +640,14 @@ CMultimodalPrior::marginalLikelihoodConfidenceInterval(double percentage,
                      : bounds.max()};
         double fa{fMinusp(a)};
         double fb{fMinusp(b)};
+        // If we enter either of the following loops, since the c.d.f. is
+        // monotonic, we know that the interval is either right or left of
+        // the required root. We can therefore shift the original interval
+        // since a and b are, respectively, upper and lower bounds for the
+        // root. Each iteration also doubles the original interval width.
+        // Note, we can only enter one or other of these loops so provided
+        // we supply a few more than 10 iterations to the solver it'll have
+        // enough iterations to comfortably converge.
         for (std::size_t i = 0; fa > 0.0 && i < 10; ++i) {
             std::tie(a, b) = std::make_pair(b - 3.0 * (b - a), a);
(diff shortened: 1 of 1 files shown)


lib/maths/CMultimodalPrior.cc
```diff
@@ -641,12 +641,12 @@ CMultimodalPrior::marginalLikelihoodConfidenceInterval(double percentage,
         double fa{fMinusp(a)};
         double fb{fMinusp(b)};
         for (std::size_t i = 0; fa > 0.0 && i < 10; ++i) {
-            a = b - 1.25 * (b - a);
-            fa = fMinusp(a);
+            std::tie(a, b) = std::make_pair(b - 3.0 * (b - a), a);
```
droberts195, line 644: This seems to be doing more than the PR description implies.  The loop didn't change `b` before.  Now the first iteration of the loop sets `b` to the original value of `a`.  If the loop does many iterations `b` will end up far smaller than `a` started off, and this is how the second loop will then start off.  So the window doesn't just get expanded but potentially also significantly shifted.

Then the second loop will shift the range the other way, but even more aggressively because `b - a` is now bigger.

If this is intentional then some comments about why it's a good idea would be good.

My worry is that because this is just setting the bounds for a solver it could be doing something very unpredictable yet work most of the time.

tveasey: The function is monotonic. If we enter the loop we know that `[a, b]` is to the right of root and indeed this is true while `fa > 0`. Therefore, we can happily shift `b -> a` since it is still a right bracket of the root, as required. So the strategy does a slide and scale of the original interval. This means we trap the root faster but expand the interval less, i.e. we are doubling the width every iteration, but decreasing shifting `a` by 3 x width. The second part is important since it means the solver will converge faster. This was an inefficiency in the original approach, so I thought it made sense to improve since I was now potentially growing the interval more.

tveasey: It is true that the description didn't capture this. I'll add a comment to the code.

PR #1339, opened 2020-06-23
[ML] Make swap and move functions noexcept where possible

droberts195:
The standard library can use more efficient algorithms when it
is certain that swap functions, move constructors and move
assignment operators do not throw exceptions.

This change adds noexcept to the majority of such functions that
didn't already have it.

(There are still a few that I didn't add it to, as it was hard
to be sure whether the complicated layers of containers of
Eigen data structures held within them might throw exceptions.)

### DIFF
--- a/lib/core/CDataFrameRowSlice.cc
+++ b/lib/core/CDataFrameRowSlice.cc
@@ -125,7 +125,7 @@ CDataFrameRowSliceHandle::CDataFrameRowSliceHandle(const CDataFrameRowSliceHandl
     : m_Impl{other.m_Impl != nullptr ? other.m_Impl->clone() : nullptr} {
 }
 
-CDataFrameRowSliceHandle::CDataFrameRowSliceHandle(CDataFrameRowSliceHandle&& other)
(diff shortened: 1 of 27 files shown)


lib/core/CDataFrameRowSlice.cc
```diff
@@ -138,7 +138,7 @@ operator=(const CDataFrameRowSliceHandle& other) {
 }
 
 CDataFrameRowSliceHandle& CDataFrameRowSliceHandle::
-operator=(CDataFrameRowSliceHandle&& other) = default;
+operator=(CDataFrameRowSliceHandle&& other) noexcept = default;
```
droberts195, line 141: Even though an `= default` move assignment would usually be `noexcept` automatically if possible, this one needs the explicit declaration as `= default` is in the implementation file so when the compiler only sees the contents of the header it wouldn't know whether `noexcept` was possible.

tveasey: 👍 I'd missed this one with the earlier change to make move constructors and assignment `noexcept`.

PR #1340, opened 2020-06-23
[ML] Improve bit vector encoding efficiency for sparse vectors

tveasey:
We use run length encoding for the bit masks we use to cache the rows for each leaf of the boosted tree. The scheme we were previously using was to use a fixed 1 byte to encode each run length. Long runs were then represented by "run length" / 256 bytes. This scheme works well while the expected run length < 256, but becomes highly suboptimal when the expected run length >> 256.

As we grow a tree we are partitioning rows between leaves, so naturally induce sparseness into the bit masks which corresponds to longer expected run lengths (of zero bits). For roughly balanced splits, our current scheme works well until the tree depth is 8 and then performance tails off for deeper trees. And in particular, the worst case for both memory usage and run time, many deep trees, are when the encoding scheme is least effective.

This switches to using 1-4 bytes to encode each run length with two header bits to define the number bytes used to encode each run. This gives similar performance to the current approach for short runs. In the worst case it uses around 50% more bytes when the expected run length is 256. However, it gives dramatically better performance for longer runs. The figure below summarises the behaviour. The black line represents optimal run length encoding using log2(E[run length]) / E[run length] / 8 bytes per bit and the green line represents a std::bitset, i.e. one bit per bit. The current scheme is shown in red and the new scheme in blue. (We see new scheme asymptotes towards optimal for very long expected runs.)

Screenshot 2020-06-23 at 10 59 52

Because we use 2 bits out of 32 for the header we can only represent vectors up to length 2^30 = 1.07B. This comfortably exceeds the maximum size of training data we'd want to have on a single node, so doesn't represent a significant limitation. Using 3 bits for encoding the run length bytes represents a significant degradation in performance for the typical case.

This doesn't change the results, but represents a memory and performance (since operations on the bit vector are O(number bytes)) improvement for classification and regression.

### DIFF
--- a/lib/core/unittest/CPackedBitVectorTest.cc
+++ b/lib/core/unittest/CPackedBitVectorTest.cc
@@ -4,8 +4,6 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
-#include "core/CStopWatch.h"
-#include <boost/test/tools/old/interface.hpp>
 #include <core/CContainerPrinter.h>
 #include <core/CLogger.h>
 #include <core/CPackedBitVector.h>
@@ -844,20 +842,46 @@ BOOST_AUTO_TEST_CASE(testCompression) {
 }
 
 BOOST_AUTO_TEST_CASE(testPersist) {
+
+    // Test persist + restore is idempotent.
+
     TBoolVec bits{true,  true,  false, false, true,
                   false, false, false, true,  true};
 
     for (std::size_t t = 0; t < bits.size(); ++t) {
-        core::CPackedBitVector origVector(bits);
+        core::CPackedBitVector originalVector(bits);
 
-        std::string origXml = origVector.toDelimited();
-        LOG_DEBUG(<< "xml = " << origXml);
+        std::string originalDelimited{originalVector.toDelimited()};
+        LOG_DEBUG(<< "delimited = " << originalDelimited);
 
         core::CPackedBitVector restoredVector;
-        restoredVector.fromDelimited(origXml);
+        restoredVector.fromDelimited(originalDelimited);
 
-        BOOST_REQUIRE_EQUAL(origVector.checksum(), restoredVector.checksum());
+        BOOST_REQUIRE_EQUAL(originalVector.checksum(), restoredVector.checksum());
     }
+}
+
+BOOST_AUTO_TEST_CASE(testUpgrade) {
+
+    // Test restoring old state produces the correct vector.
+
+    std::string state{"1310:1:1:0:255:1:5:19:7:52:255:255:206"};
+
+    core::CPackedBitVector expected;
+    expected.extend(true, 255);
+    expected.extend(false, 256);
+    expected.extend(true, 5);
+    expected.extend(false, 19);
+    expected.extend(true, 7);
+    expected.extend(false, 52);
+    expected.extend(true, 716);
+
+    core::CPackedBitVector actual;
+    actual.fromDelimited(state);
+
+    LOG_DEBUG(<< "upgraded = " << actual.toDelimited());
+
+    BOOST_REQUIRE_EQUAL(expected.checksum(), actual.checksum());
 }
 
 BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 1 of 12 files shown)


lib/core/unittest/CPackedBitVectorTest.cc
```diff
@@ -42,6 +44,200 @@ std::string toBitString(const T& v) {
...
+        return core::CPackedBitVector::readLastRunLength(lastRunBytes, runLengthBytes);
+    }
+    static std::size_t readRunLength(TUInt8VecCItr runLengthBytes) {
+        return core::CPackedBitVector::readRunLength(runLengthBytes);
+    }
+    static std::size_t popRunLength(TUInt8VecCItr& runLengthBytes) {
+        return core::CPackedBitVector::popRunLength(runLengthBytes);
+    }
+    static void writeRunLength(std::size_t runLength, TUInt8VecItr runLengthBytes) {
+        core::CPackedBitVector::writeRunLength(runLength, runLengthBytes);
```
valeriy42, line 82: Is there a reason for redefining the methods? It doesn't look like you are mocking anything particular.

lib/core/unittest/CPackedBitVectorTest.cc
```diff
@@ -42,6 +44,200 @@ std::string toBitString(const T& v) {
...
+    BOOST_TEST_REQUIRE(2, CPackedBitVectorInternals::bytes(
+                              CPackedBitVectorInternals::maximumOneByteRunLength() + 1));
+    BOOST_TEST_REQUIRE(2, CPackedBitVectorInternals::bytes(
+                              CPackedBitVectorInternals::maximumTwoByteRunLength()));
+    BOOST_TEST_REQUIRE(3, CPackedBitVectorInternals::bytes(
+                              CPackedBitVectorInternals::maximumTwoByteRunLength() + 1));
+    BOOST_TEST_REQUIRE(2, CPackedBitVectorInternals::bytes(
+                              CPackedBitVectorInternals::maximumThreeByteRunLength()));
+    BOOST_TEST_REQUIRE(3, CPackedBitVectorInternals::bytes(
+                              CPackedBitVectorInternals::maximumThreeByteRunLength() + 1));
```
valeriy42, line 105: Is this correct? It looks like it should be compared to 3 and 4.

lib/core/unittest/CPackedBitVectorTest.cc
```diff
@@ -623,21 +796,92 @@ BOOST_AUTO_TEST_CASE(testZeroLengthRunProblemCase) {
...
+        // expected memory usage of log2(E[run length]) / E[run length] / 8
+        // bytes per vector bit.
+
+        double minimumMeanBytesPerBit{density * std::log2(1.0 / density) / 8.0};
+        double meanBytesPerBit{static_cast<double>(meanMemoryUsage) / 100000.0};
+        LOG_DEBUG(<< "minimum bytes per bit = " << minimumMeanBytesPerBit);
+        LOG_DEBUG(<< "bytes per bit         = " << meanBytesPerBit);
+
+        averageOverhead += meanBytesPerBit / minimumMeanBytesPerBit;
+        BOOST_TEST_REQUIRE(meanBytesPerBit < 15.0 * minimumMeanBytesPerBit);
```
valeriy42, line 834: :+1:

lib/core/unittest/CPackedBitVectorTest.cc
tveasey: Just to make them public in the tests: they're protected.

lib/core/unittest/CPackedBitVectorTest.cc
tveasey: No! The test assertion should, of course` have been `BOOST_REQUIRE_EQUAL` and the numbers are wrong! Good catch.

tveasey: Fixed in [018f0b3](https://github.com/elastic/ml-cpp/pull/1340/commits/018f0b3baa8390601ea1750f175d513778451883).

PR #1344, opened 2020-06-23
[7.x][ML] Generate memory size info for compressed model definition

valeriy42:
This PR adds the ability to generate model_size_info objects used in Java to estimate the heap size for the inference models. I added a slim subclass CSizeInfo to all relevant classes which contains reference to the definition objects and generates Json with the necessary information when called.

We do a double work for counting nodes and leaves of the tree when computing numOperations and generating JSON. I did it to avoid memory overhead.

I am not writing the model_size_info in the results yet, I intend to do it in a subsequent PR. Hence this PR is marked as a non-issue.

Backport of #1316.

### DIFF
--- a/include/api/CInferenceModelDefinition.h
+++ b/include/api/CInferenceModelDefinition.h
@@ -113,8 +113,20 @@ public:
 
     enum ETargetType { E_Classification, E_Regression };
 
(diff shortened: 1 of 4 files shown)


valeriy42:
Something went wrong, I try again in #1346

PR #1345, opened 2020-06-23
[ML] Add memory_status to memory_stats

### LINKED ISSUE #1310: [ML] Add memory_status to memory_stats
Similar to anomaly detection, we should have status "ok" and "hard-limit". Additionally we should have an optional field for suggested new memory limit.

valeriy42:
This PR adds `status` field to the `analytics_memory_usage` instrumentation object for DGA. The status can be "ok" and "hard-limit". Additionally I added the field `advised_limit_bytes` containing new recommended memory limit for simple usage, e.g. by UI.

Since this is closely related to #1298, I don't document this separately and mark as non-issue.

Closes #1310.

### DIFF
--- a/include/api/CDataFrameAnalysisInstrumentation.h
+++ b/include/api/CDataFrameAnalysisInstrumentation.h
@@ -109,9 +109,6 @@ public:
     static void monitor(CDataFrameAnalysisInstrumentation& instrumentation,
                         core::CRapidJsonConcurrentLineWriter& writer);
 
-    std::int64_t advisedMemoryLimit() const;
-    EMemoryStatus memoryStatus() const;
-
 protected:
     using TWriter = core::CRapidJsonConcurrentLineWriter;
     using TWriterUPtr = std::unique_ptr<TWriter>;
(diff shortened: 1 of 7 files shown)


include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -103,16 +106,19 @@ class API_EXPORT CDataFrameAnalysisInstrumentation
...
 
 protected:
     using TWriter = core::CRapidJsonConcurrentLineWriter;
     using TWriterUPtr = std::unique_ptr<TWriter>;
+    using TOptionalInt64 = boost::optional<std::int64_t>;
 
 protected:
     virtual counter_t::ECounterTypes memoryCounterType() = 0;
     TWriter* writer();
+    void advisedMemoryLimit(std::int64_t advisedMemoryLimit);
```
tveasey, line 120: I'm not sure about calling this `advisedMemoryLimit` it feels a bit vague about what purpose it serves, especially since we have estimated upfront memory usage. I think this should capture something about the fact it relates to out-of-memory and so is an increase, the best I came up with was `suggestedMemoryIncrease`.

valeriy42:
retest

include/api/CDataFrameAnalysisInstrumentation.h
valeriy42: Renamed in `increasedMemoryEstimate`

dimitris-athanasiou:
@valeriy42 We had a last minute change of hearts and renamed this to `memory_reestimate_bytes`. I'll make the change in the java side shortly.

PR #1349, opened 2020-06-24
[ML] Output inference definition in compressed format

valeriy42:
This PR compresses inference definition and outputs it in the format
```
{"compressed_inference_model": {
  "doc_num": <number>,
   "definition": <gzip|base64 string>,
   "eos": <optional flag>
}}
```

Additionally it activates the output of the model size definition implemented in #1316. 

The compressed definition is split in chunks of 16MB each such that larger models will be returns as a sequence of `compressed_inference_model` objects as shown in [here](https://github.com/elastic/ml-cpp/pull/1349#issuecomment-650157471).


lib/api/CInferenceModelDefinition.cc
```diff
@@ -380,6 +386,41 @@ std::string CInferenceModelDefinition::jsonString() {
     return resultString;
 }
 
+std::string CInferenceModelDefinition::jsonStringCompressedFormat() {
```
benwtrent, line 389: Does this also split up the compressed string into a specified length?

I think we need to make guarantees around String length size so we don't ever reach the situation where even the compressed string is too large. 

This will prevent us from hitting Elasticsearch limits in:

- maximum over the wire document lengths
- maximum document size
- OOM issues for large Java strings

valeriy42: You are right, I missed the chunking part so far.

valeriy42:
So far it generates the following output in results:

benwtrent:
@valeriy42 
`definition` in the `compressed_inference_model` should not be an array. Each document sent back contains a single string of a specific size (along with the total length of ALL the chunks) and we persist the documents in order.

```
"compressed_inference_model": {
"definition": "compressed_string",
"total_definition_length": <length of ALL the string chunks>
}
```

valeriy42:
New output example:
```
{"model_size_info":{"preprocessors":[{"frequency_encoding":{"field_length":15,"feature_name_length":25,"field_value_lengths":[1,1,1]}},{"one_hot_encoding":{"field_length":15,"field_value_lengths":[1,1],"feature_name_lengths":[17,17]}}],"trained_model_size":{"ensemble_model_size":{"num_classification_weights":2,"num_classes":2,"feature_name_lengths":[17,25,11],"tree_sizes":[{"num_nodes":0,"num_leaves":1},{"num_nodes":1,"num_leaves":2},{"num_nodes":1,"num_leaves":2},{"num_nodes":2,"num_leaves":3},{"num_nodes":1,"num_leaves":2},{"num_nodes":2,"num_leaves":3},{"num_nodes":3,"num_leaves":4},{"num_nodes":3,"num_leaves":4},{"num_nodes":2,"num_leaves":3}],"num_output_processor_weights":9,"num_operations":22}}}}
,{"compressed_inference_model":{"definition":"H4sIAAAAAAAA/91Y224bNxB972fssyzMhRyS/pUiWGxkShawltTVqm0Q+N97qKRptdzWilA7TXyFufRwOHPOmZn92ByGfBj2q3w87odjc//zx2Y95F9Oebf60OLH/mG72zT3WN3m/qG5b1bdmDf7Ybvq+na175tFs87deBpyu+uecr2h/WKubP1i+qk7FKvU3NMyuhiY1ZGQcfRh0fD8spRl0ufn58XHZr/L7eN+vNLJsvOvMysnqTkfWq1zg8PeLZpx6La7/NA+7R9yX4zk3TE/ve/z+dS/BaCEcNb6v4Vld3rKeHJ2FYet+u543K6xddzud23fvc/92S5jL83s+C1vN49j2cJLWtAymBeXmEhVmDSUC3TDJo/t+OFwztHF/zeT+32CwTjk2du9zvE5t8dxOK3KUefjd/Ck3e4e8u9I+jlE7/PQHrunQ1/cEMJin7t1+2vXn3IBhnC0mPAgmYuCH8/vzkD5p4u8EPZvdqvjod+O7Wdvz9s+rWyQouaelxaU1buY2Nh5Mxz0OOTj475A/46XSqKBWDSJi5zconnI6+7Uj22f12NzD4dyWVttj+Wen73vx6YEdD22q8dtscSLZigR+PNvKbH8u/9c+x/TZVLuQNcoZiGqejp/Tq1IbYWZp7klc2KRgkVPztTzD5pbioHEWwrwwjTIRW5tCXQX9zxHTY5SesPUcqyTQiTmVB0QqQzPfLgmu6m241hTBKphyKLp1yS3VtbXUshXSDgtU8myJjVJiWO8TDhqoMBBHzUmAt9J3jLhZJX7XOGVnTmUZ2ZITWA/cd/f6K9O/HVX4KoI3SWu4Ft0pFG8GgeEcWpFZ25tFToTh2jBILfiia3CuJvxhSodZJ+CRFghJQNd4ksg/45hzUs2CLYHoVNyzlH8j3BxE44rvRFfZCsArpGSRL5GtKbqVyqbmDcu1TikyN6lr83o9S3g/zbPCCPKMpsLYEmYprkEGM/QppCIoht5w7RblS+H/qFwOEoMTlKlBrO1yr2sgQyjQiC3JJhN","total_definition_length":1696}}
,{"compressed_inference_model":{"definition":"serHcCSknZK6QOnVJHFGzKJWksgRrnpD+0go2aGyMiNmVTtWsG9iapjM0KWomr2oZt8pvjkROB4TdCwkny7r250tHTkK6kMUYS/8hvjWGplVc1HUTQrkJQVMA34CzbT0SSkQIo1Z292KTD/x3q4hltXNReW+pTLmnCVaFMPOhfcBzXDwjiNJCBgsbo39LbyC2NfEQt+uEEGJnr3qNV0CS8Urg0rKGW8RDNDKGT8DhEsj0BvEJGEy9MGUyNLUhs3YqBnuTAAMwMN5Jxq/qif/bgiOgKMji+Yg9AhVqBoV0B+CiWAm/IbifeP+u3IfZUwTWlzg3xO63depPbcwfKZyVt4rAlu+2JLHGJQuvHdoH8CpgEKFes0Ub512b2G4VmOACz4WqCTFoGaueo8xw++it9MpwIA0D25SSEYqlUzMEHz6UgW67kFO0SLraLNdqFqZGYpzqJzxrrxQQTeERjbyj8lwUAA3FDmXaSWJlyC706W5FGJEEMmJib4hw/3MmxDAC9O0Qz9NJTN6TS11/DLVGDCBSJRXDOj+gL5pGHxCD8DIlTMDL9+Qa9MqiJQBM6FM7wEOabiGa6x1MCXEkAIyi4s5oXQe0IDDbrMZ8gbDWLs/jYfTWNDe7zfb4wh04wnCUi5Zli9R/cI3rOPjpz8Au73uplQZAAA=","total_definition_length":1696}}
```

.gitignore
```diff
@@ -65,6 +65,8 @@ gcovr-unit-report.xml
 junit_results.xml
 cppunit_results.xml
 boost_test_results.xml
+**/.objs
```
droberts195, line 68: The `.objs` sub-directories are supposed to be committed to the repo.  The build system doesn't create them.  I guess it could, but that would require more changes to the make rules.

The reason you saw an untracked `.objs` directory was that the whole `autoconfig` program was removed in #1320.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -380,6 +392,48 @@ std::string CInferenceModelDefinition::jsonString() {
...
+        outFilter.push(core::CBase64Encoder());
+        outFilter.push(compressedStream);
+        outFilter << modelDefinitionStr;
+        outFilter.flush();
+    }
+    return compressedStream;
+}
+
+void CInferenceModelDefinition::addToDocumentCompressed(TRapidJsonWriter& writer) const {
+    std::stringstream compressedString{jsonStringCompressedFormat()};
```
tveasey, line 411: Ok this is probably fine, but again you are creating a copy of the entire model before chunking.

Chunking is a streaming operation: i.e. you need to buffer only one chunk then flush that chunk to the writer. It strikes me that this can be handled as a new stream sink: you just need to move the logic below (roughly) into a class with a `std::streamsize write(const char* s, std::streamsize n)` method. This doesn't actually feel like it would be a whole lot more complicated. It would be nice to make the actual logic to write a chunk pluggable. This could be done by having a callback to actually write the chunk.

We do have `CStateCompressor::CChunkFilter` I wonder if it can be factored out and generalised for this purpose. I would also be interested in @droberts195 thoughts on this.

include/api/CInferenceModelDefinition.h
```diff
@@ -435,11 +436,12 @@ class API_EXPORT CInferenceModelDefinition : public CSerializableToJson {
     TTrainedModelUPtr& trainedModel();
     const TTrainedModelUPtr& trainedModel() const;
     void addToDocument(rapidjson::Value& parentObject, TRapidJsonWriter& writer) const override;
-    std::string jsonString();
+    void addToDocumentCompressed(TRapidJsonWriter& writer) const;
+    std::string jsonString() const;
+    std::stringstream jsonStringCompressedFormat() const;
```
droberts195, line 441: This file should now `#include <sstream>` - it must be included indirectly at the moment, but that could change in the future if other headers are changed.

lib/api/CInferenceModelDefinition.cc
droberts195: This was basically my biggest comment too.

There seems to be a fundamental mismatch in the way we are treating these model definitions in different places.  In some places we're saying they're so huge that they need compression and chunking, and in other places we're saying they're so small that you can just create a temporary copy without worrying about the memory usage.

One problem with the whole architecture is that the model definition is considered to be a type of result, not a type of state.  Results are going through the `CRapidJsonConcurrentLineWriter` which expects to take a series of small documents to be sent to the results named pipe in a thread safe way so results written from different threads don't corrupt each other.  State goes through a different named pipe and is designed to be streamed, such that entire state documents never exist in memory in their entirety.  But unfortunately the Java side is not set up to receive different types of documents from a single state pipe, so it would be hard to use it for both the state for restarts and the model definition.

So it seems like the most realistic goal would be to continue to use the `CRapidJsonConcurrentLineWriter` to write the compressed model as a single document, but use streaming internally within the C++ code to avoid ever having to create the complete uncompressed model in JSON format in memory.

> We do have `CStateCompressor::CChunkFilter`

I think something along similar lines should be used, but it's probably best to create a new bespoke class that does it for this type of document.  `CChunkFilter` is designed to work with the state persistence framework, which is already generic in that it can persist to both XML and JSON, and was designed to be able to stream that state to a variety of stores, which have varied from HTTP to named pipes over the years.  To try and make it even more generic is going to make it unbelievably complex.  So I would suggest using some of the same ideas but creating a different class to do it.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -380,6 +392,48 @@ std::string CInferenceModelDefinition::jsonString() {
...
+    rapidjson::Value definitionChunksArray{writer.makeArray()};
+    std::streamsize processed{0};
+    compressedString.seekg(0, compressedString.end);
+    std::streamsize remained{compressedString.tellg()};
+    compressedString.seekg(0, compressedString.beg);
+    std::size_t utf16Length{static_cast<std::size_t>(remained)}; // since we use base64 encoding
```
droberts195, line 417: The way this is done at the moment is not compatible with streaming into the compressor.  But it seems like the output format is a document containing an array of objects, each of which knows the overall length.  If the format could be changed to just have the overall length at the very end in a separate field outside the array then that would work with streaming.  The streamer class would keep track of the length so far, and then at the very end you'd get this and add a length field.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -380,6 +392,48 @@ std::string CInferenceModelDefinition::jsonString() {
     return resultString;
 }
 
+std::stringstream CInferenceModelDefinition::jsonStringCompressedFormat() const {
+    std::stringstream compressedStream;
+    using TFilteredOutput = boost::iostreams::filtering_stream<boost::iostreams::output>;
+    std::string modelDefinitionStr{jsonString()};
```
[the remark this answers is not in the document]
droberts195: How hard would it be to change all the `addToDocument` methods that relate to the trained model to instead write JSON directly to a `ostream`?  So instead of building up RapidJSON objects in memory that then get serialised all in one go, the fields get written one at a time?

This wouldn't be thread-safe for the `ostream` that was being written to, but that would be OK because it would be a compression stream that was only being used by one thread.  This would avoid holding all the RapidJSON objects in memory until the complete state was built up.  Instead, the compressed string would gradually be built, and at any time uncompressed JSON would only exist in tiny pieces in the compression buffer.

lib/api/CInferenceModelDefinition.cc
tveasey: I think in that case, my inclination would be to make the new sink very bespoke and just lift the actual writing out of the while loop below into the sink write method and have it local to this cc file.

lib/api/CInferenceModelDefinition.cc
tveasey: We discussed this offline. Needing the whole compressed state upfront was for a consistency check in the Java, but  we can do this in a different way, so we do think it is is possible to stream the chunked documents out. This will require a small change to the schema. It is still TBD whether we'll make these changes as a follow up.

valeriy42:
Updated schema:
```
{"model_size_info":{"preprocessors":[{"frequency_encoding":{"field_length":15,"feature_name_length":25,"field_value_lengths":[1,1,1]}},{"one_hot_encoding":{"field_length":15,"field_value_lengths":[1,1],"feature_name_lengths":[17,17]}}],"trained_model_size":{"ensemble_model_size":{"num_classification_weights":2,"num_classes":2,"feature_name_lengths":[17,25,11],"tree_sizes":[{"num_nodes":0,"num_leaves":1},{"num_nodes":1,"num_leaves":2},{"num_nodes":1,"num_leaves":2},{"num_nodes":2,"num_leaves":3},{"num_nodes":1,"num_leaves":2},{"num_nodes":2,"num_leaves":3},{"num_nodes":3,"num_leaves":4},{"num_nodes":3,"num_leaves":4},{"num_nodes":2,"num_leaves":3}],"num_output_processor_weights":9,"num_operations":22}}}}
,{"compressed_inference_model":{"doc_num":0,"definition":"H4sIAAAAAAAA/91Y224bNxB972fssyzMhRyS/pUiWGxkShawltTVqm0Q+N97qKRptdzWilA7TXyFufRwOHPOmZn92ByGfBj2q3w87odjc//zx2Y95F9Oebf60OLH/mG72zT3WN3m/qG5b1bdmDf7Ybvq+na175tFs87deBpyu+uecr2h/WKubP1i+qk7FKvU3NMyuhiY1ZGQcfRh0fD8spRl0ufn58XHZr/L7eN+vNLJsvOvMysnqTkfWq1zg8PeLZpx6La7/NA+7R9yX4zk3TE/ve/z+dS/BaCEcNb6v4Vld3rKeHJ2FYet+u543K6xddzud23fvc/92S5jL83s+C1vN49j2cJLWtAymBeXmEhVmDSUC3TDJo/t+OFwztHF/zeT+32CwTjk2du9zvE5t8dxOK3KUefjd/Ck3e4e8u9I+jlE7/PQHrunQ1/cEMJin7t1+2vXn3IBhnC0mPAgmYuCH8/vzkD5p4u8EPZvdqvjod+O7Wdvz9s+rWyQouaelxaU1buY2Nh5Mxz0OOTj475A/46XSqKBWDSJi5zconnI6+7Uj22f12NzD4dyWVttj+Wen73vx6YEdD22q8dtscSLZigR+PNvKbH8u/9c+x/TZVLuQNcoZiGqejp/Tq1IbYWZp7klc2KRgkVPztTzD5pbioHEWwrwwjTIRW5tCXQX9zxHTY5SesPUcqyTQiTmVB0QqQzPfLgmu6m241hTBKphyKLp1yS3VtbXUshXSDgtU8myJjVJiWO8TDhqoMBBHzUmAt9J3jLhZJX7XOGVnTmUZ2ZITWA/cd/f6K9O/HVX4KoI3SWu4Ft0pFG8GgeEcWpFZ25tFToTh2jBILfiia3CuJvxhSodZJ+CRFghJQNd4ksg/45hzUs2CLYHoVNyzlH8j3BxE44rvRFfZCsArpGSRL5GtKbqVyqbmDcu1TikyN6lr83o9S3g/zbPCCPKMpsLYEmYprkEGM/QppCIoht5w7RblS+H/qFwOEoMTlKlBrO1yr2sgQyjQiC3JJhN"}}
,{"compressed_inference_model":{"doc_num":1,"definition":"serHcCSknZK6QOnVJHFGzKJWksgRrnpD+0go2aGyMiNmVTtWsG9iapjM0KWomr2oZt8pvjkROB4TdCwkny7r250tHTkK6kMUYS/8hvjWGplVc1HUTQrkJQVMA34CzbT0SSkQIo1Z292KTD/x3q4hltXNReW+pTLmnCVaFMPOhfcBzXDwjiNJCBgsbo39LbyC2NfEQt+uEEGJnr3qNV0CS8Urg0rKGW8RDNDKGT8DhEsj0BvEJGEy9MGUyNLUhs3YqBnuTAAMwMN5Jxq/qif/bgiOgKMji+Yg9AhVqBoV0B+CiWAm/IbifeP+u3IfZUwTWlzg3xO63depPbcwfKZyVt4rAlu+2JLHGJQuvHdoH8CpgEKFes0Ub512b2G4VmOACz4WqCTFoGaueo8xw++it9MpwIA0D25SSEYqlUzMEHz6UgW67kFO0SLraLNdqFqZGYpzqJzxrrxQQTeERjbyj8lwUAA3FDmXaSWJlyC706W5FGJEEMmJib4hw/3MmxDAC9O0Qz9NJTN6TS11/DLVGDCBSJRXDOj+gL5pGHxCD8DIlTMDL9+Qa9MqiJQBM6FM7wEOabiGa6x1MCXEkAIyi4s5oXQe0IDDbrMZ8gbDWLs/jYfTWNDe7zfb4wh04wnCUi5Zli9R/cI3rOPjpz8Au73uplQZAAA=","eos":true}}
```

.gitignore
valeriy42: Ok. thank you for the info. I'll revert the change.

include/api/CInferenceModelDefinition.h
valeriy42: Fixed.

lib/api/CInferenceModelDefinition.cc
valeriy42: We changed the format to pass sequential  `doc_num`s and `eos` flag for the last chunk.

benwtrent:
I have tested the current format and changes against https://github.com/elastic/elasticsearch/pull/58009 and it all worked GREAT!

🚀

benwtrent:
PR to mute tests: https://github.com/elastic/elasticsearch/pull/58617

PR #1352, opened 2020-06-24
[ML] Upgrade to Visual Studio 2019 for Windows compilation

### LINKED ISSUE #1015: [ML] Upgrade compilers so that we can use C++17 in the 8.x timeframe

droberts195:
We will build version 8.x of the ML C++ using Visual Studio 2019
rather than Visual Studio 2017.  This will eventually facilitate
use of C++17 (after separate PRs to upgrade the Mac and Linux
compilers have also been merged).

Since this change involves rebuilding all dependencies, the
libxml2 dependency is also upgraded from version 2.9.4 to
version 2.9.7.

Closes #1015


droberts195:
This is a replacement for #834, which had to be reverted due to a mistake with the release manager VM setup.

PR #1356, opened 2020-06-25
[ML] Implement stop-on-warn for per-partition categorization

droberts195:
This change adds:

1. More frequent writes of categorization stats - categorizers
   now know when it's "urgent" that they update their stats,
   which is basically when their status has just changed to
   "warn".
2. Caching of some of the values used to determine whether a
   categorizer's status is "warn".  This is important because
   we now calculate this a lot more frequently, to know whether
   stats writes are "urgent".
3. Functionality to stop categorizing on "warn" status, if the
   corresponding option is set.  The option can be set by a
   command line argument when the program is started, or by a
   control message while it is running.

### DIFF
--- a/bin/autodetect/Main.cc
+++ b/bin/autodetect/Main.cc
@@ -46,12 +46,11 @@
 
 #include "CCmdLineParser.h"
 
+#include <cstdio>
+#include <cstdlib>
 #include <functional>
 #include <memory>
 #include <string>
-
-#include <stdio.h>
-#include <stdlib.h>
 
 int main(int argc, char** argv) {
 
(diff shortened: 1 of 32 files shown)


bin/autodetect/Main.cc
```diff
@@ -275,17 +282,16 @@ int main(int argc, char** argv) {
     // Chain the categorizer's output to the anomaly detector's input
     ml::api::COutputChainer outputChainer(job);
 
-    ml::api::CJsonOutputWriter fieldDataCategorizerOutputWriter(jobId, wrappedOutputStream);
```
droberts195, line 278: This is making the categorizer consistent with the anomaly detector.  In the beginning, both shared the same `CJsonOutputWriter`.  When forecasting was added the anomaly detector was given a separate `CJsonOutputWriter` that it manages internally.  So the global `CJsonOutputWriter` was then for categorization only (and maybe nothing if categorization was not in use), leading to asymmetry.  Moving it into the categorizer means that it can be finalised when the categorizer is finalised, which in turn means the "Unfortunately we cannot rely on destruction to finalise" comment and associated code smell can be removed.

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -385,43 +399,48 @@ void CTokenListDataCategorizerBase::acceptPersistInserter(
...
-    while (checkIter != m_CategoriesByCount.begin()) {
-        --checkIter;
-        if (count <= checkIter->first) {
+    while (iter != m_CategoriesByCount.begin()) {
+        --iter;
+        if (count <= iter->first) {
             break;
         }
-        swapIter = checkIter;
+        swapIter = iter;
```
tveasey, line 443: Shouldn't `m_CategoriesByCount` be in descending count order, in which case don't you need to shuffle: `iter - 1 -> iter`, `iter - 2 -> iter - 1`, ... rather than just swap the start and end of the range?

include/api/CSingleFieldDataCategorizer.h
```diff
@@ -105,6 +118,11 @@ class API_EXPORT CSingleFieldDataCategorizer {
                                       const CCategoryIdMapper& categoryIdMapper,
                                       core::CStatePersistInserter& inserter);
 
+    //! Write the latest categorizer stats, plus an annotation if the
+    //! categorization status has changed.
+    void writeStats(CJsonOutputWriter& jsonOutputWriter,
+                    CAnnotationJsonWriter& annotationJsonWriter);
```
tveasey, line 124: I feel since you now have `writeStatsIfUrgent` it would be more consistent to call this `writeStatsIfStatusChanged`.

lib/model/CTokenListDataCategorizerBase.cc
droberts195: Yes, good point.  This was intended to be a no-op refactor to avoid having 3 iterator variables but I've messed it up. 🤦

lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -385,43 +399,48 @@ void CTokenListDataCategorizerBase::acceptPersistInserter(
...
+                                                     TSizeSizePrVecItr iter) {
     m_Categories[iter->second].addString(isDryRun, str, rawStringLen, tokenIds, tokenUniqueIds);
 
     std::size_t& count{iter->first};
+    bool wasCountRare{this->isCategoryCountRare(count)};
     ++count;
+    ++m_TotalCount;
+    if (wasCountRare && this->isCategoryCountRare(count) == false) {
+        --m_NumRareCategories;
+    }
```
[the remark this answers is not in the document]
droberts195: My assumption when I wrote this was that categories can only be rare on initial creation (with count 1), and can never become rare by adding a match.  That's true with the current definition of rare, but you're right it's a breach of encapsulation to assume it.  I will change the code so it's generic to any definition of rare (which _was_ the intention as it's needlessly complex as it is with the current definition).

include/api/CSingleFieldDataCategorizer.h
droberts195: It's called from two places, and writes the stats if _any_ field within the stats has changed.  I'll change it to `writeStatsIfChanged`.

lib/model/CTokenListDataCategorizerBase.cc
tveasey: Ok now I'm wondering if this whole loop doesn't just simplify (since you are only adding one to the count):
```
if (iter != m_CategoriesByCount.begin() && count > (iter - 1)->first) {
    std::iter_swap(iter, iter - 1);
}
```
If you were to add more than one to the count you'd need the swap in the loop.

tveasey: Oh this is of course wrong since you need to jump over all categories whose count is equal: ignore.

tveasey: But interestingly you are reordering the ones which are equal. Maybe this is fine, but it is interesting. It still feels slightly surprising the swap isn't in the loop, but I guess if only the partial order matters it's fine and more efficient.

PR #1358, opened 2020-06-25
Format time period in model change annotation using CTimeUtils::durationToString

przemekwitek:
This PR makes time period used in model change annotation formatted using newly introduced `CTimeUtils::durationToString`.
The goal is to make the annotation text human-readable.

Relates https://github.com/elastic/elasticsearch/issues/55781

### DIFF
--- a/include/core/CTimeUtils.h
+++ b/include/core/CTimeUtils.h
@@ -71,6 +71,7 @@ public:
     //! whitespace before calling this function.
     static bool isDateWord(const std::string& word);
 
+    //! Formats the given duration as human-readable string.
     static std::string durationToString(core_t::TTime duration);
 
 private:
--- a/include/core/CTimeUtils.h
+++ b/include/core/CTimeUtils.h
@@ -71,6 +71,8 @@ public:
     //! whitespace before calling this function.
     static bool isDateWord(const std::string& word);
 
+    static std::string durationToString(core_t::TTime duration);
+
 private:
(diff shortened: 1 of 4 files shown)


droberts195:
If the periodicity is _exactly_ 1 day then this is going to print it as `1d0h0m0s`.  It would be nicer if that could be just `1d`.  I suspect exactly 1 day is a common case too.

przemekwitek:
> If the periodicity is exactly 1 day then this is going to print it as 1d0h0m0s. It would be nicer if that could be just 1d. I suspect exactly 1 day is a common case too.

What would you be your recommendation for 1 day and 1 second then? "1d1s" or "1d0m1s"? If we want to be concise I guess we should output "1d1s" in such a case.

droberts195:
> What would you be your recommendation for 1 day and 1 second then?

I don't think this is critical, because in practice it won't happen very often.  Periodicity is extremely likely to be a multiple of bucket spans or a multiple of periods that make sense to humans (days, weeks, etc.), so I think "1d1s" or similar is fine for the extremely unlikely cases.

przemekwitek:
> I don't think this is critical, because in practice it won't happen very often. Periodicity is extremely likely to be a multiple of bucket spans or a multiple of periods that make sense to humans (days, weeks, etc.), so I think "1d1s" or similar is fine for the extremely unlikely cases.

Done.

PR #1360, opened 2020-06-26
[ML] Don't require time is supplied when adding an annotation

tveasey:
All annotations can use the start time of the bucket in which they are added. This means that it is not necessary to pass in a time when adding an annotation it can be captured in the lambda.

This is important because there are places we may want to generate annotations, such as when the residual distribution model changes, where time is unavailable. It will be undesirable to have to make it available just for annotations.

I also made a small tweak to the way the callback is passed to `std::function`. The size of this object means it won't be able to take advantage of the small object optimisation. This triggers an allocation on every single bucket and time series update which we try to avoid. By passing the callback by reference we avoid this.

### DIFF
--- a/lib/model/CEventRateModel.cc
+++ b/lib/model/CEventRateModel.cc
@@ -332,16 +332,14 @@ void CEventRateModel::sample(core_t::TTime startTime,
                     .trendWeights(trendWeights)
                     .priorWeights(priorWeights);
                 if (this->params().s_AnnotationsEnabled) {
-                    const auto modelAnnotationCallback =
-                        [&](core_t::TTime t, const std::string& annotation) {
-                            m_CurrentBucketStats.s_Annotations.emplace_back(
-                                t, annotation, gatherer.searchKey().detectorIndex(),
-                                gatherer.searchKey().partitionFieldName(),
-                                gatherer.partitionFieldValue(),
-                                gatherer.searchKey().overFieldName(),
-                                EMPTY_STRING, gatherer.searchKey().byFieldName(),
-                                gatherer.personName(pid));
-                        };
+                    const auto modelAnnotationCallback = [&](const std::string& annotation) {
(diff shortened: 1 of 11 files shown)


lib/model/CEventRateModel.cc
```diff
@@ -325,25 +325,26 @@ void CEventRateModel::sample(core_t::TTime startTime,
...
-                                gatherer.searchKey().partitionFieldName(),
-                                gatherer.partitionFieldValue(),
-                                gatherer.searchKey().overFieldName(),
-                                EMPTY_STRING, gatherer.searchKey().byFieldName(),
-                                gatherer.personName(pid));
-                        };
-                    params.annotationCallback(modelAnnotationCallback);
-                }
+                    .priorWeights(priorWeights)
+                    .annotationCallback([&](const std::string& annotation) {
```
przemekwitek, line 345: Why is the `annotationCallback` wrapped in a separate lambda here?
Is it because:
> By passing the callback by reference we avoid this.

?

tveasey: You still have to bind it to a `std::function`. You pass by constant reference and this creates a temporary `std::function` object (for the type erasure for the lambda). If the lambda is small, typically less than or equal to 16 bytes, then this object doesn't allocate any memory. Otherwise it allocates memory to hold the lambda. By wrapping the original lambda in another one which captures it by reference the lambda for which we perform type erasure is only 8 bytes: containing only a reference to original lambda as a member rather than having members for the different values which are referenced in the original lambda. For more information see for example [this](https://blog.demofox.org/2015/02/25/avoiding-the-performance-hazzards-of-stdfunction/) discussion.

przemekwitek: Wow. Thanks for explanation. In moments like this I appreciate the fact I'm (mostly) a Java developer ;)

PR #1370, opened 2020-07-02
Add script and config for FOSSA scanning

droberts195:
FOSSA will check that we are only using code with approved
licenses.


droberts195:
> The only thing that I wonder: how do we ensure the `.fossa.yml` being updated, e.g. after a new binary/new directories got added?

Yes, I agree this is fragile at the moment.

At the moment FOSSA doesn't have a dedicated C++ plugin, so C++ projects have to use the "raw" plugin, which works one directory at a time.  Apparently they are close to releasing a dedicated C++ plugin.  I would hope that will have a better understanding of the project structure and we won't then need to list every directory individually.  (For example a Java or Go project wouldn't need to do this at the moment.)

PR #1371, opened 2020-07-02
[ML] Avoid instantiating string for writing model definition

### LINKED ISSUE #1364: [ML] Avoid instantiating string for writing model definition
As discussed [here](https://github.com/elastic/ml-cpp/pull/1349#pullrequestreview-438128635)  inference model definition code should be refactored to avoid creating a string and load it into main memory. We should use the filtering mechanism and enhance it with chunking the definition in a streaming fashion.

Implementation of #1329 is a technical requirement for this issue.

valeriy42:
To reduce memory consumption, the inference model definition is now written directly into a stream using ` core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>` instead of `core::CRapidJsonConcurrentLineWriter` used before. Therefore, to distinguish the two kinds of creating json format, we now have two internfaces:
+ `CSerializableToJsonDocument` with the function `addToJsonDocument` uses provided `CRapidJsonConcurrentLineWriter` to create a rapidjson object and then write it into the result stream directly.
+ `CSerializableToJsonStream` with the function ` addToJsonStream` uses `CRapidJsonLineWriter` to write in a stream without keeping a complete json object in memory.

Memory size info classes still implement `CSerializableToJsonDocument` since the memory overhead there is minimal and this way they can use the writer created in `CDataFrameAnalyzer`.

Furthermore, we don't need a lot of boiler plate code now. I simplified and removed the unused functions.

Closes #1364


lib/api/CInferenceModelDefinition.cc
```diff
@@ -113,6 +128,29 @@ void addJsonArray(const std::string& tag,
...
+
+void addJsonArray(const std::string& tag,
+                  const std::vector<std::string>& vector,
+                  CSerializableToJson::TGenericLineWriter& writer) {
+    writer.Key(tag);
+    rapidjson::Value array{writer.makeArray(vector.size())};
+    for (const auto& value : vector) {
+        array.PushBack(toJson(value, writer), writer.getRawAllocator());
+    }
+    writer.write(array);
```
droberts195, line 152: This is still copying the text in all the strings twice.  You can just copy it once like this:

```suggestion
    writer.StartArray();
    for (const auto& value : vector) {
        writer.String(value);
    }
    writer.EndArray();
```

lib/api/CInferenceModelDefinition.cc
```diff
@@ -377,48 +487,43 @@ CTrainedModel::TStringVec CTree::removeUnusedFeatures() {
...
     using TFilteredOutput = boost::iostreams::filtering_stream<boost::iostreams::output>;
-    std::string modelDefinitionStr{jsonString()};
     {
         TFilteredOutput outFilter;
         outFilter.push(boost::iostreams::gzip_compressor());
         outFilter.push(core::CBase64Encoder());
         outFilter.push(compressedStream);
-        outFilter << modelDefinitionStr;
+        this->jsonStream(outFilter);
         outFilter.flush();
```
droberts195, line 511: I would have thought it would automatically flush on destruction.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -857,6 +1057,20 @@ void CWeightedSum::addToDocument(rapidjson::Value& parentObject,
...
+void CWeightedSum::addToJsonStream(TGenericLineWriter& writer) const {
+    writer.StartObject();
+    writer.Key(this->stringType());
+    writer.StartObject();
+    writer.Key(JSON_WEIGHTS_TAG);
+    rapidjson::Value array{writer.makeArray(m_Weights.size())};
+    for (const auto& weight : m_Weights) {
+        array.PushBack(weight, writer.getRawAllocator());
+    }
+    writer.write(array);
```
droberts195, line 1069:
```suggestion
    writer.StartArray();
    for (const auto& weight : m_Weights) {
        writer.Double(weight);
    }
    writer.EndArray();
```

lib/api/CInferenceModelDefinition.cc
```diff
@@ -895,6 +1116,20 @@ void CLogisticRegression::addToDocument(rapidjson::Value& parentObject,
...
+void CLogisticRegression::addToJsonStream(TGenericLineWriter& writer) const {
+    writer.StartObject();
+    writer.Key(this->stringType());
+    writer.StartObject();
+    writer.Key(JSON_WEIGHTS_TAG);
+    rapidjson::Value array{writer.makeArray(m_Weights.size())};
+    for (const auto& weight : m_Weights) {
+        array.PushBack(weight, writer.getRawAllocator());
+    }
+    writer.write(array);
```
droberts195, line 1128:
```suggestion
    writer.StartArray();
    for (const auto& weight : m_Weights) {
        writer.Double(weight);
    }
    writer.EndArray();
```

lib/api/CInferenceModelDefinition.cc
```diff
@@ -113,6 +128,29 @@ void addJsonArray(const std::string& tag,
...
+void addJsonArray(const std::string& tag,
+                  const std::vector<T>& vector,
+                  CSerializableToJson::TGenericLineWriter& writer) {
+    writer.Key(tag);
+    rapidjson::Value array{writer.makeArray(vector.size())};
+    for (const auto& value : vector) {
+        array.PushBack(toJson(value), writer.getRawAllocator());
+    }
+    writer.write(array);
+}
```
droberts195, line 142: Although it seems nice to have this generic, are we actually only using it for `double` at the moment?  If so it would be more efficient to have the specific method:

```suggestion
void addJsonArray(const std::string& tag,
                  const std::vector<double>& vector,
                  CSerializableToJson::TGenericLineWriter& writer) {
    writer.Key(tag);
    writer.StartArray();
    for (const auto& value : vector) {
        writer.Double(value);
    }
    writer.EndArray();
}
```

If we _are_ writing different types of vectors then you can still avoid using the allocator by keeping this method a template but replacing `writer.Double(value);` in my suggestion with `writeSingleJsonValue(value);`, and then having multiple overloads of `writeSingleJsonValue` for all the types we need to write.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -102,6 +113,10 @@ auto toJson(std::size_t value, CSerializableToJson::TRapidJsonWriter&) {
     return toJson(value);
 }
 
+auto toJson(const unsigned int& value) {
```
droberts195, line 116: If this is for handling `uint32_t` then be aware this might be fragile for portability in the future.  On Windows `uint32_t` might be `unsigned long`, as on 64 bit Windows `long` is still 32 bits like `int`.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -90,10 +91,20 @@ auto toJson(const std::string& value, CSerializableToJson::TRapidJsonWriter& wri
...
+    return result;
+}
+
+auto toJson(double value) {
     return rapidjson::Value{value};
 }
 
+auto toJson(double value, CSerializableToJson::TRapidJsonWriter&) {
+    return toJson(value);
+}
```
droberts195, line 106: Is this needed now?  (Maybe originally you used the same template `addToJsonArray` for both `double` and `string`.)

lib/api/CInferenceModelDefinition.cc
valeriy42: Fixed

lib/api/CInferenceModelDefinition.cc
valeriy42: It was a remainder of the first iteration. I use `addJsonArray()` now.

lib/api/CInferenceModelDefinition.cc
valeriy42: Indeed, I have only a version supporting `doubles` now.

lib/api/CInferenceModelDefinition.cc
valeriy42: I removed all those `toJson` functions as we don't need templates anymore.

lib/api/CInferenceModelDefinition.cc
valeriy42: Indeed, I don't need it after refactoring.

lib/api/CInferenceModelDefinition.cc
valeriy42: Replaced with `addJsonArray`.

lib/api/CInferenceModelDefinition.cc
valeriy42: Indeed, removed it now.

PR #1375, opened 2020-07-02
[ML] Adds exponent aggregator to the inference model definition

### LINKED ISSUE #1372: [ML] Inference models for MSLE regression trees should have exp aggregators
`CMsle::transform` computes an exponent  of the prediction before writing out the result. This exponentiation is not reflected in the inference model definition, hence the prediction results are off by a huge margin.

valeriy42:
This PR adds exponent aggregator required for the regression models trained with MSLE objective function.

Closes #1372

### DIFF
--- a/include/api/CInferenceModelDefinition.h
+++ b/include/api/CInferenceModelDefinition.h
@@ -112,10 +112,10 @@ private:
     TDoubleVec m_Weights;
 };
 
-//! Allows to use logistic regression aggregation.
+//! Allows to use exponent aggregation.
 //!
 //! Given a weights vector $\vec{w}$ as a parameter and an output vector from the ensemble $\vec{x}$,
-//! it computes the logistic regression function \f$1/(1 + \exp(-\vec{w}^T \vec{x}))\f$.
+//! it computes the exponent function \f$\exp(\vec{w}^T \vec{x})\f$.
 class API_EXPORT CExponent final : public CAggregateOutput {
 public:
     using TDoubleVec = std::vector<double>;
--- a/include/api/CInferenceModelDefinition.h
+++ b/include/api/CInferenceModelDefinition.h
@@ -112,6 +112,27 @@ private:
     TDoubleVec m_Weights;
 };
 
+//! Allows to use logistic regression aggregation.
+//!
+//! Given a weights vector $\vec{w}$ as a parameter and an output vector from the ensemble $\vec{x}$,
+//! it computes the logistic regression function \f$1/(1 + \exp(-\vec{w}^T \vec{x}))\f$.
+class API_EXPORT CExponent final : public CAggregateOutput {
+public:
+    using TDoubleVec = std::vector<double>;
+
+public:
+    ~CExponent() override = default;
+    //! Construct with the \p weights vector.
+    explicit CExponent(TDoubleVec&& weights);
+    //! Construct with a weight vector of \p size with all entries equal to \p weight.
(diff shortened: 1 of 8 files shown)


lib/api/CBoostedTreeInferenceModelBuilder.cc
```diff
@@ -140,12 +140,20 @@ CRegressionInferenceModelBuilder::CRegressionInferenceModelBuilder(const TStrVec
...
     this->definition().trainedModel()->targetType(CTrainedModel::ETargetType::E_Regression);
 }
 
 void CRegressionInferenceModelBuilder::setAggregateOutput(CEnsemble* ensemble) const {
-    ensemble->aggregateOutput(std::make_unique<CWeightedSum>(ensemble->size(), 1.0));
+    if (m_LossType == TLossType::E_MsleRegression) {
+        ensemble->aggregateOutput(std::make_unique<CExponent>(ensemble->size(), 1.0));
+    } else {
+        ensemble->aggregateOutput(std::make_unique<CWeightedSum>(ensemble->size(), 1.0));
+    }
```
tveasey, line 156: My inclination would be to make this a switch: it avoids us forgetting this in future if we add a new regression type.

include/api/CInferenceModelDefinition.h
```diff
@@ -112,6 +112,27 @@ class API_EXPORT CLogisticRegression final : public CAggregateOutput {
     TDoubleVec m_Weights;
 };
 
+//! Allows to use logistic regression aggregation.
+//!
+//! Given a weights vector $\vec{w}$ as a parameter and an output vector from the ensemble $\vec{x}$,
+//! it computes the logistic regression function \f$1/(1 + \exp(-\vec{w}^T \vec{x}))\f$.
```
tveasey, line 118: This comment looks wrong. The forest is predicting `sum_i{ leaf_i } = log(1 + x)`, so for the weighted case the aggregator would be `exp(w^t x) - 1` no?

tveasey: In fact, thinking about it, 1 is the offset is a parameter which we also need to supply.

valeriy42: You are right, the aggregator computes `exp(w^t x)` cf. `CMsle::transform`.

lib/api/CBoostedTreeInferenceModelBuilder.cc
valeriy42: I added a switch and `LOG_ERROR` if we unexpectedly encounter a classification function.

valeriy42:
Thank you @tveasey for the review. I addressed your comments. It would be great if you could take another look.

include/api/CInferenceModelDefinition.h
tveasey: Ok great.

and the comment on the offset was me forgetting what we're actually doing. For posterity, we compute leaf values by `argmin_x (log(offset + exp(prediction)) - log(offset + actual))^2`, i.e. the important part is that prediction is exponentiated so the forest is effectively predicting logs. The offset just governs how we penalise relative errors between the prediction actuals at the scale of offset.

lib/api/CBoostedTreeInferenceModelBuilder.cc
tveasey: Thanks!

valeriy42:
retest

PR #1377, opened 2020-07-03
[ML] Memory correction coefficient fixed

valeriy42:
There was an error in how we were computing the correction coefficient which led to wrong scaling. This PR fixes the issues and adds a unit test to check the values.

Since memory correction is not released yet I am marking this as a non-issue.


lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1827,4 +1828,30 @@ BOOST_AUTO_TEST_CASE(testRestoreErrorHandling) {
     ml::core::CLogger::instance().reset();
 }
 
+BOOST_AUTO_TEST_CASE(testWorstCaseMemoryCorrection) {
```
tveasey, line 1831: 👍

PR #1379, opened 2020-07-03
[ML] Fix cause of SIGSEGV for classification and regression

tveasey:
We weren't accounting for the padding we include since we memory aligned columns, #1142, in `CBoostedTreeFactory` when computing the number of extra columns. This meant we were using presenting extra columns to feature selection (which would nearly always ignore them) and occasionally changing the initialisation of the learn rate.

The problem we hit which caused the SIGSEGV was when there was only one feature which itself carried no information, i.e. it was a constant. We could then erroneously select an extra column as a feature.

### DIFF
--- a/include/maths/CBoostedTreeImpl.h
+++ b/include/maths/CBoostedTreeImpl.h
@@ -326,6 +326,7 @@ private:
     EInitializationStage m_InitializationStage = E_NotInitialized;
     std::size_t m_NumberThreads;
     std::size_t m_DependentVariable = std::numeric_limits<std::size_t>::max();
+    TOptionalSize m_PaddedExtraColumns;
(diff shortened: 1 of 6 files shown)


include/maths/CBoostedTreeImpl.h
```diff
@@ -326,6 +326,7 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     EInitializationStage m_InitializationStage = E_NotInitialized;
     std::size_t m_NumberThreads;
     std::size_t m_DependentVariable = std::numeric_limits<std::size_t>::max();
+    TOptionalSize m_PaddedExtraColumns;
```
tveasey, line 329: Note that this doesn't need persisting because it is recalculated every time.

PR #1397, opened 2020-07-13
[ML] Update user instructions for memory limit surpass

valeriy42:
This tiny PR updates the instructions for the user stating explicitly that the analysis job has to be force-stop before the memory setting can be updated.

### DIFF
--- a/lib/api/CDataFrameAnalysisInstrumentation.cc
+++ b/lib/api/CDataFrameAnalysisInstrumentation.cc
@@ -191,7 +191,8 @@ void CDataFrameAnalysisInstrumentation::monitor(CDataFrameAnalysisInstrumentatio
             instrumentation.flush();
             HANDLE_FATAL(<< "Input error: required memory "
                          << bytesToString(instrumentation.memory()) << " exceeds the memory limit "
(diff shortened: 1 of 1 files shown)


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -191,7 +191,8 @@ void CDataFrameAnalysisInstrumentation::monitor(CDataFrameAnalysisInstrumentatio
             instrumentation.flush();
             HANDLE_FATAL(<< "Input error: required memory "
                          << bytesToString(instrumentation.memory()) << " exceeds the memory limit "
-                         << bytesToString(instrumentation.m_MemoryLimit) << ". Please increase the limit to at least "
+                         << bytesToString(instrumentation.m_MemoryLimit)
+                         << ". Please force-stop the analysis job, increase the limit to at least "
```
tveasey, line 195: This reads slightly better to me:
```suggestion
                         << ". Please force-stop the analysis job and increase the limit to at least "
```

tveasey: Oh I didn't see the last part of the message: ignore.

PR #1399, opened 2020-07-13
[ML] Fix a race in writing the memory usage message after circuit breaking

tveasey:
This message is new to 7.9 so I've marked this PR as a non-issue.


tveasey:
retest

PR #1407, opened 2020-07-15
[ML] New memory estimation correction function

valeriy42:
This PR changes the memory correction strategy to enforce the strongly monotonic behavior of the corrected values. Previously, the correction could leave to the situation where memory estimations could be swapped (bigger job could have a smaller memory estimation), this is not possible now.

### DIFF
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -366,9 +366,13 @@ std::size_t CBoostedTreeImpl::estimateMemoryUsage(std::size_t numberRows,
 std::size_t CBoostedTreeImpl::correctedMemoryUsage(double memoryUsageBytes) {
     // We use a piecewise linear function of the estimated memory usage to compute
     // the corrected value. The values are selected in a way to reduce over-estimation
-    // on small jobs to improve the behaviour on the trial nodes in the cloud.
-    TDoubleVec estimatedMemoryUsageMB{0., 1024., 4096., 8192., 12288., 16384.};
-    TDoubleVec correctedMemoryUsageMB{0., 179.2, 512., 819.2, 1088., 1280.};
+    // and to improve the behaviour on the trial nodes in the cloud. The high level strategy
+    // also ensures that corrected memory usage is a monotonic function of estimated memory
+    // usage and any change to the approach should preserve this property.
+    TDoubleVec estimatedMemoryUsageMB{0.0,    20.0,    1024.0, 4096.0,
+                                      8192.0, 12288.0, 16384.0};
+    TDoubleVec correctedMemoryUsageMB{0.0,   20.0,   179.2, 512.0,
+                                      819.2, 1088.0, 1280.0};
     maths::CSpline<> spline(maths::CSplineTypes::E_Linear);
     spline.interpolate(estimatedMemoryUsageMB, correctedMemoryUsageMB,
                        maths::CSplineTypes::E_ParabolicRunout);
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -26,6 +26,7 @@
 #include <maths/CQuantileSketch.h>
 #include <maths/CSampling.h>
 #include <maths/CSetTools.h>
+#include <maths/CSpline.h>
 #include <maths/CTreeShapFeatureImportance.h>
 
 #include <boost/circular_buffer.hpp>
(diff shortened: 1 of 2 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -362,12 +364,15 @@ std::size_t CBoostedTreeImpl::estimateMemoryUsage(std::size_t numberRows,
...
 std::size_t CBoostedTreeImpl::correctedMemoryUsage(double memoryUsageBytes) {
-    // We compute the correction coefficient as a sigmoid function: ca. 1.0 until
-    // 10mb, ca 16.0 after 1000mb to this end we need to shift and scale using the
-    // magic numbers below.
-    double correctionCoefficient{
-        CTools::logisticFunction(memoryUsageBytes / (1024 * 1024), 100, 550) * 15 + 1};
-    return static_cast<std::size_t>(memoryUsageBytes / correctionCoefficient);
+    // We use a piecewise linear function of the estimated memory usage to compute
+    // the corrected value. The values are selected in a way to reduce over-estimation
+    // on small jobs to improve the behaviour on the trial nodes in the cloud.
```
tveasey, line 369: I don't think this just improves over estimates on small jobs (intended the reduction is larger on large jobs). I'd also be inclined to add something about monotonicity. How about:
```suggestion
    // to improve the behaviour on the trial nodes in the cloud. The high level strategy
    // also ensures that corrected memory usage is a monotonic function of estimated
    // memory usage and any change to the approach should preserve this property.
```

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -4,6 +4,8 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
+#include "core/CContainerPrinter.h"
+#include "maths/CSpline.h"
```
tveasey, line 8: Stray headers.

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1829,29 +1833,23 @@ BOOST_AUTO_TEST_CASE(testRestoreErrorHandling) {
...
+            (BYTES_IN_MB),
+        96.25, 2.0);
     // test for 1000mb
-    BOOST_REQUIRE_CLOSE(static_cast<double>(maths::CBoostedTreeImpl::correctedMemoryUsage(
-                            1000.0 * 1024 * 1024)) /
-                            (1024 * 1024),
-                        1000.0 / 15.84, 2.0);
+    BOOST_REQUIRE_CLOSE(
+        static_cast<double>(maths::CBoostedTreeImpl::correctedMemoryUsage(1024.0 * BYTES_IN_MB)) / BYTES_IN_MB,
+        179.2, 2.0);
```
tveasey, line 1852: How about adding in a test for monotonicity, i.e. randomly generate 1000 pairs and check `(estimatedMemoryUsage.first < estimatedMemoryUsage.second) == (correctedMemoryUsage.first < correctedMemoryUsage.second)`?

lib/maths/CBoostedTreeImpl.cc
valeriy42: changed

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -362,12 +364,15 @@ std::size_t CBoostedTreeImpl::estimateMemoryUsage(std::size_t numberRows,
...
+    TDoubleVec estimatedMemoryUsageMB{0., 1024., 4096., 8192., 12288., 16384.};
+    TDoubleVec correctedMemoryUsageMB{0., 179.2, 512., 819.2, 1088., 1280.};
```
tveasey, line 371: On reflection, I think it is worth having this corrected directly proportional to the estimate for small data sets. I think this will also fix the failing unit tests. How about: 
```suggestion
    TDoubleVec estimatedMemoryUsageMB{0., 20., 1024., 4096., 8192., 12288., 16384.};
    TDoubleVec correctedMemoryUsageMB{0., 20., 179.2, 512., 819.2, 1088., 1280.};
```

lib/maths/unittest/CBoostedTreeTest.cc
valeriy42: Good idea! I added the test.

lib/maths/CBoostedTreeImpl.cc
valeriy42: Fixed.

lib/maths/unittest/CBoostedTreeTest.cc
```diff
@@ -1829,29 +1831,41 @@ BOOST_AUTO_TEST_CASE(testRestoreErrorHandling) {
...
+                        1355.75, 1.0);
+
+    // test for monotonicity
+    std::size_t numberSamples{1000};
+    TDoubleVec lhs;
+    lhs.reserve(numberSamples);
+    TDoubleVec rhs;
+    rhs.reserve(numberSamples);
+    test::CRandomNumbers rng;
+    rng.generateUniformSamples(0, 20000, numberSamples, lhs);
```
tveasey, line 1863:
```suggestion
    rng.generateUniformSamples(0.0, 20000.0, numberSamples, lhs);
    rng.generateUniformSamples(0.0, 20000.0, numberSamples, rhs);
```

PR #1413, opened 2020-07-16
[7.9][ML] New memory estimation correction function

valeriy42:
This PR changes the memory correction strategy to enforce the strongly monotonic behavior of the corrected values. Previously, the correction could leave to the situation where memory estimations could be swapped (bigger job could have a smaller memory estimation), this is not possible now.

Backport of #1407


droberts195:
retest

valeriy42:
retest

PR #1415, opened 2020-07-17
[ML] Testing for no change should not use the change trend model

tveasey:
Change #1391 introduced a subtle error: when computing the likelihood of no change we should use a reference to the current time series trend model, not the copy of the model which assumes a change is occurring. This has not yet been released so I've marked this as a non-issue.


tveasey:
We'd like this to make 7.8.1. Linux failed to download the docker image and we don't support aarch yet so I'm going to go ahead and merge this.

PR #1427, opened 2020-07-21
Fix deprecated Gradle API usage

breskeby:
This PR fixes some deprecated gradle api usages in the gradle build. It also enforces the failure of the build from now on if a deprecated api is used. That makes it easy to keep the build clean and not let deprecations sneak into the build over time. 

This PR is related to a discussion with the elasticsearch infra team at https://elastic.slack.com/archives/C8UUBNASY/p1595288611366200 after a failure in a 7.8.1 BC

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,11 +126,11 @@ def zipSpec = copySpec {
 
 task buildZip(type: Zip) {
   dependsOn strip
-  baseName = artifactName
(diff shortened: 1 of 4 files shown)


droberts195:
> Please could you also make this change in `upload.gradle`

Actually there were more things than that.  I just pushed another commit with an attempt at fixing them all.

PR #1428, opened 2020-07-21
[ML] Changing the message for memory limit failure

valeriy42:
When the job fails due to exceeded memory limit, I issue a `LOG_ERROR` with the details about the final memory usage and a `HANDLE_FATAL` message which is more concise and will be used as an audit message for the user. 

The original message is confusing for the user since they would need to double memory requirement even though the limit might be exceed by a few megs.

### DIFF
--- a/lib/api/CDataFrameAnalysisInstrumentation.cc
+++ b/lib/api/CDataFrameAnalysisInstrumentation.cc
@@ -193,9 +193,9 @@ void CDataFrameAnalysisInstrumentation::monitor(CDataFrameAnalysisInstrumentatio
             LOG_ERROR(<< "Input error: required memory " << bytesToString(memory) << " exceeds the memory limit "
                       << bytesToString(memoryLimit) << ". New estimated limit is "
                       << bytesToString(memoryReestimateBytes) << ".");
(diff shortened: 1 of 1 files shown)


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -190,9 +190,11 @@ void CDataFrameAnalysisInstrumentation::monitor(CDataFrameAnalysisInstrumentatio
             instrumentation.memoryReestimate(static_cast<std::int64_t>(memoryReestimateBytes));
             instrumentation.memoryStatus(E_HardLimit);
             instrumentation.flush();
-            HANDLE_FATAL(<< "Input error: required memory " << bytesToString(memory)
-                         << " exceeds the memory limit " << bytesToString(memoryLimit)
-                         << ". Please force-stop the analysis job, increase the limit to at least "
+            LOG_ERROR(<< "Input error: required memory " << bytesToString(memory) << " exceeds the memory limit "
```
dimitris-athanasiou, line 193: Should we wrap the memory values in square brackets? This is something done in ES side. We did not use to do it in c++ but it might be good to start doing so.

valeriy42:
This change was integrated in #1432

PR #1432, opened 2020-07-21
[ML] Fixing memory_status output on fatal error

valeriy42:
When the job is failing due to the memory limit exceeding, the job fails to emit the final memory_status message. This PR fixes this. 

There was a type in the "hard_limit" tag which I fixed. Also I tweaked the parameters of the unit test `CDataFrameAnalyzerTrainingTest.testMemoryLimitHandling` to reduce runtime.

Additionally I took the new messaging from #1428 and added it here to avoid merge conflicts.


include/api/CDataFrameAnalyzer.h
```diff
@@ -99,6 +101,7 @@ class API_EXPORT CDataFrameAnalyzer {
     TDataFrameUPtr m_DataFrame;
     TTemporaryDirectoryPtr m_DataFrameDirectory;
     TJsonOutputStreamWrapperUPtrSupplier m_ResultsStreamSupplier;
+    TRapidJsonConcurrentLineWriterPtr m_ResultsWriter;
```
tveasey, line 104: I don't think this can work: `CDataFrameAnalyzer` has automatic storage and is created at this [line](https://github.com/elastic/ml-cpp/blob/master/bin/data_frame_analyzer/Main.cc#L207). Handle fatal calls `std::exit` which states "[The] stack is not unwound: destructors of variables with automatic storage duration are not called", see [here](https://en.cppreference.com/w/cpp/utility/program/exit). Objects with static storage are destructed by calling `std::exit` so one possibility would be temporarily register a callback to call the destructor with with an object with static storage.

valeriy42:
@tveasey thank you for your review. At the end it turned out enough to have one additional `flush` call to fix the code. It would be great if you could take another look.

PR #1438, opened 2020-07-24
[7.x][ML] Fixing memory_status output on fatal error 

valeriy42:
When the job is failing due to the memory limit exceeding, the job fails to emit the final memory_status message. This PR fixes this.

There was a type in the "hard_limit" tag which I fixed. Also I tweaked the parameters of the unit test CDataFrameAnalyzerTrainingTest.testMemoryLimitHandling to reduce runtime.

Additionally I took the new messaging from #1428 and added it here to avoid merge conflicts.

Backport of #1432


valeriy42:
retest

PR #1443, opened 2020-08-05
[ML] Fix progress on resume after final training has completed for classification and regression

tveasey:
We previously showed progress stuck at zero for the final training step.


tveasey:
retest

PR #1445, opened 2020-08-06
[ML] Fixed errors identified by valgrind memcheck

edsavage:
`Valgrind`'s `memcheck` tool  was run over the unit test suite. The tool picked up
an error related to unit test code only and another in production code
itself. Labelling as a non-issue because of the relatively minor nature of the problems identified.


droberts195:
Please could you also backport this to 7.x?  We should be backporting all minor fixes back to 7.x unless there's a really good reason not to.

PR #1458, opened 2020-08-18
[ML] Replace CSleep with standard equivalent

### LINKED ISSUE #559: [ML] Switch core::CSleep to this_thread::sleep_for()
We have a class that provides a portable sleep, in `include/core/CSleep.h`.

Portable sleeping is now standard C++, using `std::this_thread::sleep_for()`.

Therefore all uses of `ml::core::CSleep` can be removed from the codebase and replaced with `std:: this_thread::sleep_for()`, and then the implementation and unit tests of `ml::core::CSleep` can also be removed.

edsavage:
Remove the CSleep class and replace functionality with the modern,
portable standard library equivalent.

Closes #559


bin/controller/unittest/CCommandProcessorTest.cc
```diff
@@ -5,19 +5,19 @@
...
-#include <core/CSleep.h>
 #include <core/CStringUtils.h>
 
 #include "../CCommandProcessor.h"
 
 #include <boost/range.hpp>
 #include <boost/test/unit_test.hpp>
 
+#include <chrono>
+#include <cstdio>
```
droberts195, line 16: There are `::remove()` calls that should be changed to `std::remove()` if we're using the C++ header.

lib/core/unittest/CDetachedProcessSpawnerTest.cc
```diff
@@ -6,14 +6,15 @@
...
 #include <core/CStringUtils.h>
 
 #include <boost/range.hpp>
 #include <boost/test/unit_test.hpp>
 
-#include <stdio.h>
-#include <stdlib.h>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
```
droberts195, line 16: => `::getenv()` -> `std::getenv()`

lib/core/unittest/CDetachedProcessSpawnerTest.cc
```diff
@@ -6,14 +6,15 @@
...
-#include <core/CSleep.h>
 #include <core/CStringUtils.h>
 
 #include <boost/range.hpp>
 #include <boost/test/unit_test.hpp>
 
-#include <stdio.h>
-#include <stdlib.h>
+#include <chrono>
+#include <cstdio>
```
droberts195, line 15: => `::remove()` -> `std::remove()`

lib/core/unittest/CDualThreadStreamBufTest.cc
```diff
@@ -15,11 +14,12 @@
 #include <boost/test/unit_test.hpp>
 
 #include <algorithm>
+#include <chrono>
+#include <cstdint>
```
droberts195, line 18: => `uint32_t` -> `std::uint32_t`

lib/core/unittest/CDualThreadStreamBufTest.cc
```diff
@@ -15,11 +14,12 @@
...
+#include <cstring>
```
droberts195, line 19: => `::strlen()` -> `std::strlen()`

lib/core/unittest/CReadWriteLockTest.cc
```diff
@@ -12,15 +12,15 @@
...
 #include <core/CThread.h>
 #include <core/CTimeUtils.h>
 
 #include <boost/test/unit_test.hpp>
 
 #include <atomic>
-
-#include <stdint.h>
+#include <chrono>
+#include <cstdint>
```
droberts195, line 22: => `uint32_t` -> `std::uint32_t`

lib/core/unittest/CStopWatchTest.cc
```diff
@@ -5,10 +5,11 @@
  */
 
 #include <core/CLogger.h>
-#include <core/CSleep.h>
 #include <core/CStopWatch.h>
 
-#include <stdint.h>
+#include <chrono>
+#include <cstdint>
```
droberts195, line 11: => `uint32_t` -> `std::uint32_t`

PR #1463, opened 2020-08-26
[ML] Resolves the predicted class type when writing feature importance values

benwtrent:
This commit addresses two bugs:

- For individual class feature importance values, we now resolve the predicted class type name. e.g. we will now write JSON boolean `true` instead of `"1"`.
- For total feature importance, the predicted class type is now resolved as well.

### DIFF
--- a/include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
+++ b/include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
@@ -29,6 +29,12 @@ public:
     using TReadPredictionFunc = std::function<TDouble2Vec(const TRowRef&)>;
     using TReadClassScoresFunc = std::function<TDouble2Vec(const TRowRef&)>;
 
+    enum EPredictionFieldType {
+        E_PredictionFieldTypeString,
+        E_PredictionFieldTypeInt,
+        E_PredictionFieldTypeBool
+    };
+
 public:
     static const std::size_t MAX_NUMBER_CLASSES;
     static const std::string NUM_CLASSES;
@@ -69,11 +75,6 @@ public:
 
     //! \return A serialisable metadata of the trained regression model.
     TOptionalInferenceModelMetadata inferenceModelMetadata() const override;
-
-    static void
-    writePredictedCategoryValueForType(const std::string& categoryValue,
-                                       EPredictionFieldType predictionFieldType,
-                                       core::CRapidJsonConcurrentLineWriter& writer);
 
 private:
     static TLossFunctionUPtr loss(std::size_t numberClasses);
--- a/include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
+++ b/include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
@@ -29,12 +29,6 @@ public:
     using TReadPredictionFunc = std::function<TDouble2Vec(const TRowRef&)>;
     using TReadClassScoresFunc = std::function<TDouble2Vec(const TRowRef&)>;
 
-    enum EPredictionFieldType {
-        E_PredictionFieldTypeString,
-        E_PredictionFieldTypeInt,
-        E_PredictionFieldTypeBool
-    };
-
 public:
     static const std::size_t MAX_NUMBER_CLASSES;
     static const std::string NUM_CLASSES;
@@ -75,6 +69,11 @@ public:
 
     //! \return A serialisable metadata of the trained regression model.
     TOptionalInferenceModelMetadata inferenceModelMetadata() const override;
+
+    static void
+    writePredictedCategoryValueForType(const std::string& categoryValue,
+                                       EPredictionFieldType predictionFieldType,
+                                       core::CRapidJsonConcurrentLineWriter& writer);
 
 private:
     static TLossFunctionUPtr loss(std::size_t numberClasses);
--- a/include/api/CInferenceModelMetadata.h
+++ b/include/api/CInferenceModelMetadata.h
@@ -17,12 +17,6 @@
 namespace ml {
 namespace api {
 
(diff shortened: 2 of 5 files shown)


include/api/CInferenceModelMetadata.h
```diff
@@ -35,12 +41,16 @@ class API_EXPORT CInferenceModelMetadata {
...
     using TRapidJsonWriter = core::CRapidJsonConcurrentLineWriter;
+    using TPredictionFieldTypeResolverWriter =
+        std::function<void(const std::string&, EPredictionFieldType, TRapidJsonWriter&)>;
 
 public:
     //! Writes metadata using \p writer.
     void write(TRapidJsonWriter& writer) const;
     void columnNames(const TStrVec& columnNames);
     void classValues(const TStrVec& classValues);
+    void predictionFieldType(EPredictionFieldType predictionFieldType);
```
tveasey, line 52: I wouldn't bother to pass this in: where the callback is supplied the type is known so just wrap it into the callback.

lib/api/CInferenceModelMetadata.cc
```diff
@@ -43,7 +43,12 @@ void CInferenceModelMetadata::writeTotalFeatureImportance(TRapidJsonWriter& writ
             for (std::size_t j = 0; j < m_ClassValues.size(); ++j) {
                 writer.StartObject();
                 writer.Key(JSON_CLASS_NAME_TAG);
-                writer.String(m_ClassValues[j]);
+                if (m_PredictionFieldTypeResolverWriter) {
```
tveasey, line 46: It always feels cleaner to me to default these sorts of call backs to avoid having if not null in the code. It also protects against accidental misuse. You can declare this very easily with a lambda inline in the class definition or you could create a `defaultPredictionFieldTypeResolverWriter` private member function to default initialise in the same way. For example:
```
TPredictionFieldTypeResolverWriter m_PredictionFieldTypeResolverWriter = [](const std::string& value, TRapidJsonWriter& writer) { writer.String(m_ClassValues[j]); };
```

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -165,6 +165,9 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
         std::size_t numberClasses{classValues.size()};
         m_InferenceModelMetadata.columnNames(featureImportance->columnNames());
         m_InferenceModelMetadata.classValues(classValues);
+        m_InferenceModelMetadata.predictionFieldType(m_PredictionFieldType);
+        m_InferenceModelMetadata.predictionFieldTypeResolverWriter(
+            CDataFrameTrainBoostedTreeClassifierRunner::writePredictedCategoryValueForType);
```
tveasey, line 170: I think this should use a lambda (or std::bind can also work) and then you can ditch the new `writePredictedCategoryValueForType` method altogether. Since this is owned by this object you don't need to worry about the life time of `this` so this can just be:
```
m_InferenceModelMetadata.predictionFieldTypeResolverWriter(
        [this](const std::string& categoryValue,
               core::CRapidJsonConcurrentLineWriter& writer) {
             this->writePredictedCategoryValue(categoryValue, writer);
        });
```

include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
```diff
@@ -76,6 +70,11 @@ class API_EXPORT CDataFrameTrainBoostedTreeClassifierRunner final
     //! \return A serialisable metadata of the trained regression model.
     TOptionalInferenceModelMetadata inferenceModelMetadata() const override;
 
+    static void
+    writePredictedCategoryValueForType(const std::string& categoryValue,
+                                       EPredictionFieldType predictionFieldType,
+                                       core::CRapidJsonConcurrentLineWriter& writer);
```
tveasey, line 76: This can be removed, see other comments.

include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
```diff
@@ -29,12 +29,6 @@ class API_EXPORT CDataFrameTrainBoostedTreeClassifierRunner final
     using TReadPredictionFunc = std::function<TDouble2Vec(const TRowRef&)>;
     using TReadClassScoresFunc = std::function<TDouble2Vec(const TRowRef&)>;
 
-    enum EPredictionFieldType {
-        E_PredictionFieldTypeString,
-        E_PredictionFieldTypeInt,
-        E_PredictionFieldTypeBool
-    };
```
tveasey, line 36: This can stay here if we don't bother to pass to inference model writer directly.

tveasey (commented):
Overall looks good. I have some comments regarding the implementation details, which I think can be made a bit more concise. I'd also like to avoid `m_PredictionFieldTypeResolverWriter` being null by having it default initialised. Although you handle this in the code you added 👍, bugs of this sort have bitten us before and it would be easy to add a new call and forget the check.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
benwtrent: Ah, passing `this` is ok in this context. @tveasey my C++ lambda skills are lacking.

include/api/CDataFrameTrainBoostedTreeClassifierRunner.h
benwtrent: 👍

benwtrent:
jenkins retest this please

tveasey (approved):
LGTM. Thanks for iterating! Let's merge as soon as CI is green.

PR #1465, opened 2020-08-31
[ML] do not fail to start analytics process if memory estimate is lower than predicted memory usage

benwtrent:
From the management side, we allow users to update configured memory limits after the analytics configuration has started.

To help users with their workflow, we should allow analytics processes to attempt to run even if the configured memory limit is lower than our memory estimate. 

related to: https://github.com/elastic/elasticsearch/pull/61505

### DIFF
--- a/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
+++ b/lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
@@ -95,12 +95,8 @@ BOOST_AUTO_TEST_CASE(testComputeAndSaveExecutionStrategyDiskUsageFlag) {
                         .diskUsageAllowed(false)
                         .outlierSpec();
 
-        // single error is registered that the memory limit is to low
-        LOG_DEBUG(<< "errors = " << core::CContainerPrinter::print(errors));
-        core::CRegex re;
-        re.init("Input error: memory limit.*");
-        BOOST_REQUIRE_EQUAL(1, static_cast<int>(errors.size()));
-        BOOST_TEST_REQUIRE(re.matches(errors[0]));
+        // no error should be registered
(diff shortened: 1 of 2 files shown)


benwtrent:
This change seemed simple to me. Hopefully I didn't miss any "gotchas"

Our instrumentation will automatically detect when memory usage is higher than the configured limit and throw a fatal.

See: https://github.com/elastic/ml-cpp/blob/877a0d87a9ad9b22f22ac58783c80514f36f2b9d/lib/api/CDataFrameAnalysisInstrumentation.cc#L188

droberts195 (approved):
LGTM

lib/api/unittest/CDataFrameAnalysisRunnerTest.cc
```diff
@@ -95,12 +95,8 @@ BOOST_AUTO_TEST_CASE(testComputeAndSaveExecutionStrategyDiskUsageFlag) {
...
                         .outlierSpec();
 
-        // single error is registered that the memory limit is to low
-        LOG_DEBUG(<< "errors = " << core::CContainerPrinter::print(errors));
-        core::CRegex re;
-        re.init("Input error: memory limit.*");
-        BOOST_REQUIRE_EQUAL(1, static_cast<int>(errors.size()));
-        BOOST_TEST_REQUIRE(re.matches(errors[0]));
+        // no error should be registered
+        BOOST_REQUIRE_EQUAL(0, static_cast<int>(errors.size()));
```
droberts195, line 99: Casting to `int` is a leftover from CppUnit, which required the LHS and RHS types to be exactly the same.  Boost.Test doesn't, so the cast is just noise now.  But no need to make another commit just to change this, as it had the cast before too.

PR #1468, opened 2020-09-01
[ML] Fix peak memory usage reporting

valeriy42:
In the memory stats we reported the current memory usage instead of the peak memory usage. This PR fixes this.

Furthermore, we now check if the memory usage exceeds memory limit every time memory usage is updated instead of ca. once every second like it was before.


lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -103,7 +103,8 @@ CDataFrameAnalysisInstrumentation::CDataFrameAnalysisInstrumentation(const std::
 }
 
 void CDataFrameAnalysisInstrumentation::updateMemoryUsage(std::int64_t delta) {
-    std::int64_t memory{m_Memory.fetch_add(delta)};
+    m_Memory.fetch_add(delta);
+    std::int64_t memory{m_Memory.load()};
```
droberts195, line 107: Splitting it over two lines means the peak could be missed, if another thread added a negative delta in between the two lines.  I think the only way to make it genuinely atomic is:

```suggestion
    std::int64_t memory{m_Memory.fetch_add(delta) + delta};
```

It's a shame there isn't an `add_fetch` method.

droberts195 (approved):
LGTM

PR #1473, opened 2020-09-04
[ML] Refactor foreground persistence

edsavage:
Refactor foreground persistence to allow for using supplied descriptors when snapshotting model state.
This paves the way for upgrading model state.


lib/api/CAnomalyJob.cc
```diff
@@ -1094,9 +1105,13 @@ bool CAnomalyJob::runBackgroundPersist(TBackgroundPersistArgsPtr args,
         return false;
     }
 
+    core_t::TTime snapshotTimestamp(core::CTimeUtils::now());
+    const std::string snapshotId(core::CStringUtils::typeToString(snapshotTimestamp));
+    const std::string description{"Periodic background persist at " + snapshotId};
```
droberts195, line 1110:
```suggestion
    const std::string description{"Periodic background persist at " + core::CTimeUtils::toIso8601(snapshotTimestamp)};
```

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -287,8 +293,104 @@ class CTestFixture {
...
+            // Check that the snapshot description and Id reported by the "persist complete"
+            // handler match those supplied to the persist function
+            BOOST_REQUIRE_EQUAL(description, description_);
+            BOOST_REQUIRE_EQUAL(snapshotId, snapshotId_);
+
+            std::string state{dataStream->str()};
+
+            // Compare snapshot ID embedded in the state string with the supplied value.
+            const std::string expectedId{"job_model_state_" +
+                                         std::to_string(snapshotTimestamp) + "#1"};
```
droberts195, line 384:
```suggestion
                                         snapshotId + "#1"};
```

It's just coincidence that the snapshot ID is the snapshot timestamp in this case.  The document ID is supposed to include the snapshot ID regardless of whether it's the same as the snapshot timestamp.

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -287,8 +293,104 @@ class CTestFixture {
...
+                [firstProcessor](const ml::api::CDataProcessor::TStrStrUMap& dataRowFields) {
+                    return firstProcessor->handleRecord(
+                        dataRowFields, ml::api::CDataProcessor::TOptionalTime{});
+                }));
+
+            // Ensure the model size stats are up to date
+            job.finalise();
+
+            ml::core_t::TTime snapshotTimestamp{1283524206};
+            const std::string snapshotId{ml::core::CStringUtils::typeToString(snapshotTimestamp)};
```
droberts195, line 369: It's probably best to set the snapshot ID to something different to the snapshot timestamp, just to prove that the code being tested doesn't mix up the snapshot ID and the snapshot timestamp.

```suggestion
            const std::string snapshotId{"my_special_snapshot"};
```

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -35,9 +35,11 @@ BOOST_AUTO_TEST_SUITE(CPersistenceManagerTest)
 namespace {
 
 void reportPersistComplete(ml::api::CModelSnapshotJsonWriter::SModelSnapshotReport modelSnapshotReport,
+                           std::string& description,
                            std::string& snapshotIdOut,
                            std::size_t& numDocsOut) {
     LOG_DEBUG(<< "Persist complete with description: " << modelSnapshotReport.s_Description);
+    description = modelSnapshotReport.s_Description;
```
droberts195, line 42: This could also capture the snapshot timestamp so tests can assert that is as expected too.

lib/api/unittest/CPersistenceManagerTest.cc
```diff
@@ -287,8 +293,104 @@ class CTestFixture {
...
+            const std::string snapshotId{ml::core::CStringUtils::typeToString(snapshotTimestamp)};
+            const std::string description{"Supplied description for snapshot at " +
+                                          ml::core::CTimeUtils::toIso8601(snapshotTimestamp)};
+            BOOST_TEST_REQUIRE(job.doPersistStateInForeground(
+                dataAdder, description, snapshotId, snapshotTimestamp));
+
+            // Check that the snapshot description and Id reported by the "persist complete"
+            // handler match those supplied to the persist function
+            BOOST_REQUIRE_EQUAL(description, description_);
+            BOOST_REQUIRE_EQUAL(snapshotId, snapshotId_);
```
droberts195, line 378: Also assert that the snapshot timestamp is as expected.

droberts195 (approved):
LGTM

droberts195:
The CI failed because of the change of Gradle task name in the ES repo.  To get it to work you need to merge latest master into the PR branch.

PR #1476, opened 2020-09-07
[ML] Freshen CRestorePreviousStateTest

edsavage:
* Add state persistence files for more recent versions
* Check for when a restore/persist cycle should/should not
result in differing state documents.


droberts195 (approved):
LGTM

PR #1480, opened 2020-09-09
[ML] Fix some edge cases in new seasonal decomposition

tveasey:
Testing on our full QA suite showed up some edge cases, mainly around numerics, which were generating log errors.

We had some cases of 0/0 in the test statistics. Also we weren't checking the denominator degrees freedom for the F distribution was > 4 before getting variance (for which it is undefined). Finally, I've improved handling of the case that almost all values are either missing or constant in the test window.

Follow on from #1451.


valeriy42 (approved):
LGTM

PR #1488, opened 2020-09-15
[ML] Parse enhanced persist control message

edsavage:
Parse and validate optional arguments to the 'persist' control message.
If present these arguments are used to perform foreground persistence,
else background persistence is requested.

As further work is still required to compose and send such a message from a suitable API, labelling this PR as a non-issue.

### DIFF
--- a/lib/api/CAnomalyJob.cc
+++ b/lib/api/CAnomalyJob.cc
@@ -349,9 +349,7 @@ bool CAnomalyJob::handleControlMessage(const std::string& controlMessage) {
         this->doForecast(controlMessage);
         break;
     case 'w':
-        if (m_PersistenceManager != nullptr && this->isPersistenceNeeded("state")) {
-            m_PersistenceManager->startPersist(core::CTimeUtils::now());
-        }
+        this->doPersist(controlMessage.substr(1));
         break;
     default:
         LOG_WARN(<< "Ignoring unknown control message of length "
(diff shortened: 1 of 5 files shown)


lib/api/unittest/CAnomalyJobTest.cc
```diff
@@ -751,4 +751,106 @@ BOOST_AUTO_TEST_CASE(testRestoreFailsWithEmptyStream) {
...
+              << expectedSnapshotDescription;
+
+        const std::string& invalidPersistControlMessage{ostrm.str()};
+
+        ml::core_t::TTime snapshotTimestamp;
+        std::string snapshotId;
+        std::string snapshotDescription;
+        BOOST_TEST_REQUIRE(ml::api::CAnomalyJob::parsePersistControlMessage(
+                               invalidPersistControlMessage, snapshotTimestamp,
+                               snapshotId, snapshotDescription) == false);
```
droberts195, line 825: This one _should_ be valid, because nothing stops a user setting an empty description.

lib/api/CAnomalyJob.cc
```diff
@@ -349,9 +349,7 @@ bool CAnomalyJob::handleControlMessage(const std::string& controlMessage) {
         this->doForecast(controlMessage);
         break;
     case 'w':
-        if (m_PersistenceManager != nullptr && this->isPersistenceNeeded("state")) {
-            m_PersistenceManager->startPersist(core::CTimeUtils::now());
-        }
+        this->doPersist(controlMessage.substr(1));
```
droberts195, line 352: I would rename `doPersist` to `processPersistControlMessage`.  We already have many similarly named methods related to persistence, so it would be good if the names of new ones make totally clear what they're for.

edsavage:
retest

PR #1498, opened 2020-09-22
[ML] Consume restore state stream to end

dimitris-athanasiou:
Data frame analytics state may spread over multiple
documents. But unlike autodetect, where state is accompanied
by a model snapshot that allows us to know exactly how many
docs it comprises of, or categorizer, where we don't have
a snapshot equivalent but state can only grow, DFA state
may spread less docs than the previous one for a job.
This means that when the state is restored and streamed
into the c++ process, we might get more documents than
just the necessary ones.

We correctly restore state as we detect the `eos` marker.
However, we also need to consume the stream to its end
so that the thread writing to the stream may complete.

### DIFF
--- a/include/api/CSingleStreamSearcher.h
+++ b/include/api/CSingleStreamSearcher.h
@@ -39,14 +39,14 @@ public:
     CSingleStreamSearcher(const TIStreamP& stream);
 
     //! Ensure destructor is called to consume stream to its end
-    ~CSingleStreamSearcher();
+    ~CSingleStreamSearcher() override;
 
     //! Get the stream to retrieve data from.
     //! \return Pointer to the input stream.
     //! Some errors cannot be detected by this call itself, and are
     //! indicated by the stream going into the "bad" state as it is
     //! read from.
-    virtual TIStreamP search(size_t currentDocNum, size_t limit);
+    TIStreamP search(size_t currentDocNum, size_t limit) override;
 
 private:
     void consumeStream();
--- a/include/api/CSingleStreamSearcher.h
+++ b/include/api/CSingleStreamSearcher.h
@@ -38,12 +38,18 @@ public:
     //! called.
     CSingleStreamSearcher(const TIStreamP& stream);
 
+    //! Ensure destructor is called to consume stream to its end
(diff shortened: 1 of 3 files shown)


include/api/CSingleStreamSearcher.h
```diff
@@ -38,13 +38,19 @@ class API_EXPORT CSingleStreamSearcher : public core::CDataSearcher {
     //! called.
     CSingleStreamSearcher(const TIStreamP& stream);
 
+    //! Ensure destructor is called to consume stream to its end
+    ~CSingleStreamSearcher();
```
droberts195, line 42:
```suggestion
    ~CSingleStreamSearcher() override;
```

(this doesn't affect the generated code, it's just a clue for readers like `@Override` in Java)

include/api/CSingleStreamSearcher.h
```diff
@@ -38,13 +38,19 @@ class API_EXPORT CSingleStreamSearcher : public core::CDataSearcher {
...
+
     //! Get the stream to retrieve data from.
     //! \return Pointer to the input stream.
     //! Some errors cannot be detected by this call itself, and are
     //! indicated by the stream going into the "bad" state as it is
     //! read from.
     virtual TIStreamP search(size_t currentDocNum, size_t limit);
```
droberts195, line 49:
```suggestion
    TIStreamP search(size_t currentDocNum, size_t limit) override;
```

(because the compiler will moan if `override` is used for one overridden virtual function and `virtual` for a different one in the same class)

lib/api/CSingleStreamSearcher.cc
```diff
@@ -14,6 +14,17 @@ CSingleStreamSearcher::CSingleStreamSearcher(const TIStreamP& stream)
     : m_Stream(stream) {
 }
 
+CSingleStreamSearcher::~CSingleStreamSearcher() {
+    this->consumeStream();
```
droberts195, line 18: It would be good to add a comment here to say this is because the Elasticsearch side might send more docs than are required, and will block if we don't read them.

droberts195 (approved):
LGTM

Issue #1477, opened 2020-09-08
[ML] Factor out duplicated code in model library unit tests


droberts195:
Many unit tests in the `model` library start off by creating a model, and there is a lot of duplicated code that does this, both within the suites for the different types of models and between the suites.

The affected files are:

```
CCountingModelTest.cc
CEventRateModelTest.cc
CEventRatePopulationModelTest.cc
CMetricModelTest.cc
CMetricPopulationModelTest.cc
```

For example there are:

- Large blocks of code that call `addArrival`
- Large blocks of type definitions
- Free functions within the different test files that help with model setup
- All these classes use a test fixture that has a protected resource monitor

There should be scope to refactor this to reduce duplication and verbosity.  Some ideas are:

- Create a separate file containing a test fixture base class that all the model tests can share (in Boost.Test a test fixture is a class that every unit test that uses it inherits from, so can use its protected methods and data as a code reuse mechanism)
- Make some of the local setup helper functions methods of this base class so that they can be shared between the different suites of model tests
- Move commonly used type definitions into this shared test fixture - then they will be available to every test without needing to be redefined within it
- Add some methods that make it easier to add batches of arrivals
- Modernise the code using initializer lists rather than batches of `push_back` calls
- Add a test setup factory class to encapsulate some aspects of creating a model for unit testing

edsavage:
This list will be used to check off each constituent piece of work as it is completed and updated with the corresponding PR number. 

 - [ ] Create model test fixture base class
 - [ ] Move common type definitions to base class
 - [ ] Move resource monitor instance to base class
 - [ ] Modernize code where possible e.g. initializer lists
 - [ ] Refactor local helper functions to be members of base class
 - [ ] Add factory class to create appropriate models

PR #1499, opened 2020-09-22
[ML] Refactor model lib unit tests

edsavage:
Rework a number of model library unit tests to factor out:

 * A commonly used resource monitor instance
 * Many multiply defined types 

These will now reside in a newly defined test fixture base class.

This is the initial stage of work of that discussed in #1477

### DIFF
--- a/lib/model/unittest/CCountingModelTest.cc
+++ b/lib/model/unittest/CCountingModelTest.cc
@@ -3,6 +3,7 @@
  * or more contributor license agreements. Licensed under the Elastic License;
  * you may not use this file except in compliance with the Elastic License.
  */
+#include "CModelTestFixtureBase.h"
 
 #include <core/CContainerPrinter.h>
 #include <core/CLogger.h>
@@ -74,9 +75,7 @@ makeScheduledEvent(const std::string& description, double start, double end) {
 const std::string EMPTY_STRING;
 }
 
-class CTestFixture {
-protected:
-    CResourceMonitor m_ResourceMonitor;
+class CTestFixture : public CModelTestFixtureBase {
 };
 
 BOOST_FIXTURE_TEST_CASE(testSkipSampling, CTestFixture) {
--- a/lib/model/unittest/CCountingModelTest.cc
+++ b/lib/model/unittest/CCountingModelTest.cc
@@ -18,12 +18,12 @@
 
 #include <test/CRandomNumbers.h>
 
+#include "CModelTestFixtureBase.h"
+
 #include <boost/test/unit_test.hpp>
 
 #include <string>
 #include <vector>
-
-#include "CModelTestFixtureBase.h"
 
 BOOST_AUTO_TEST_SUITE(CCountingModelTest)
 
--- a/lib/model/unittest/CCountingModelTest.cc
+++ b/lib/model/unittest/CCountingModelTest.cc
@@ -3,7 +3,6 @@
  * or more contributor license agreements. Licensed under the Elastic License;
  * you may not use this file except in compliance with the Elastic License.
  */
-#include "CModelTestFixtureBase.h"
 
 #include <core/CContainerPrinter.h>
 #include <core/CLogger.h>
@@ -23,6 +22,8 @@
 
 #include <string>
 #include <vector>
+
+#include "CModelTestFixtureBase.h"
 
 BOOST_AUTO_TEST_SUITE(CCountingModelTest)
 
@@ -75,8 +76,7 @@ makeScheduledEvent(const std::string& description, double start, double end) {
 const std::string EMPTY_STRING;
 }
 
-class CTestFixture : public CModelTestFixtureBase {
-};
+class CTestFixture : public CModelTestFixtureBase {};
 
 BOOST_FIXTURE_TEST_CASE(testSkipSampling, CTestFixture) {
     core_t::TTime startTime(100);
--- a/lib/model/unittest/CMetricModelTest.cc
+++ b/lib/model/unittest/CMetricModelTest.cc
@@ -4,6 +4,8 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
+#include "CModelTestFixtureBase.h"
+
 #include <core/CContainerPrinter.h>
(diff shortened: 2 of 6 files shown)


lib/model/unittest/CCountingModelTest.cc
```diff
@@ -3,6 +3,7 @@
  * or more contributor license agreements. Licensed under the Elastic License;
  * you may not use this file except in compliance with the Elastic License.
  */
+#include "CModelTestFixtureBase.h"
```
droberts195, line 6: We usually put same directory headers after our library headers.  (This applies to all 5 files.)

It relates to this sentence from the style guide:

> It is best practice to list the ML include files in the build order of the libraries they relate to, as this helps to catch accidental circular dependencies.

It would be especially true for this header as it currently defines a whole load of types in the global namespace.  I am not sure that's ideal, which is the subject of a different comment.

lib/model/unittest/CModelTestFixtureBase.h
```diff
@@ -0,0 +1,62 @@
+#ifndef INCLUDED_CModelTestFixtureBase_h
```
droberts195, line 1: Needs the Elastic license copyright header.

lib/model/unittest/CModelTestFixtureBase.h
```diff
@@ -0,0 +1,62 @@
...
+using TStrVecVec = std::vector<TStrVec>;
+
+using TTimeDoublePr = std::pair<ml::core_t::TTime, double>;
+using TOptionalTimeDoublePr = boost::optional<TTimeDoublePr>;
+using TTimeStrVecPr = std::pair<ml::core_t::TTime, TStrVec>;
+using TTimeStrVecPrVec = std::vector<TTimeStrVecPr>;
+using TTimeVec = std::vector<ml::core_t::TTime>;
+
+using TUInt64Vec = std::vector<uint64_t>;
+using TUIntVec = std::vector<unsigned int>;
```
droberts195, line 55: I don't really like all these types being loose in the global namespace.  If they were `public` members of the test fixture base class they'd be available to every test, but not the helper functions in unnamed namespaces.  Maybe it's OK to leave this as-is as part of the first phase of refactoring but have a plan to make them `public` members of the test fixture base class in the next phase and make the free functions that currently use them members of either the test fixture base class or the per-test test fixture classes.

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -59,36 +61,8 @@ using namespace model;
...
-using TDouble2Vec = core::CSmallVector<double, 2>;
-using TDouble4Vec = core::CSmallVector<double, 4>;
-using TDouble4Vec1Vec = core::CSmallVector<TDouble4Vec, 1>;
-using TSizeDoublePr = std::pair<std::size_t, double>;
-using TSizeDoublePr1Vec = core::CSmallVector<TSizeDoublePr, 1>;
-using TStrVec = std::vector<std::string>;
-using TTimeStrVecPr = std::pair<core_t::TTime, TStrVec>;
-using TTimeStrVecPrVec = std::vector<TTimeStrVecPr>;
+using TMinAccumulator = ml::maths::CBasicStatistics::SMin<double>::TAccumulator;
+using TMaxAccumulator = ml::maths::CBasicStatistics::SMax<double>::TAccumulator;
```
droberts195, line 65: There's a `using namespace ml;` above so the `ml::` isn't needed here (or maybe the `using namespace ml;` isn't needed after adding these, although I doubt it).

lib/model/unittest/CMetricPopulationModelTest.cc
```diff
@@ -53,24 +55,9 @@ using namespace model;
...
-using TSizeVec = std::vector<std::size_t>;
-using TSizeVecVec = std::vector<TSizeVec>;
-using TMeanAccumulator = maths::CBasicStatistics::SSampleMean<double>::TAccumulator;
-using TMinAccumulator = maths::CBasicStatistics::COrderStatisticsStack<double, 1u>;
+using TMinAccumulator = ml::maths::CBasicStatistics::COrderStatisticsStack<double, 1u>;
 using TMaxAccumulator =
-    maths::CBasicStatistics::COrderStatisticsStack<double, 1u, std::greater<double>>;
-using TDouble1Vec = core::CSmallVector<double, 1>;
-using TDouble2Vec = core::CSmallVector<double, 2>;
+ml::maths::CBasicStatistics::COrderStatisticsStack<double, 1u, std::greater<double>>;
```
droberts195, line 60: There's a `using namespace ml;` above so the `ml::` isn't needed here (or maybe the `using namespace ml;` isn't needed after adding these, although I doubt it).

lib/model/unittest/CModelTestFixtureBase.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef INCLUDED_CModelTestFixtureBase_h
+#define INCLUDED_CModelTestFixtureBase_h
+
+#include <core/CoreTypes.h>
+
+#include <maths/CBasicStatistics.h>
+#include <maths/CModel.h>
+#include <maths/CMultivariatePrior.h>
+
+#include <model/CResourceMonitor.h>
```
tveasey, line 16: We probably ought not to rely on transitive includes of `std` and `boost` functionality here, i.e. `#include <utility>`, `#include <vector>`, etc

tveasey (commented):
Looks like a good first step. Aside from Dave's comment, I have one minor suggestion to not rely on transitive includes in `CModelTestFixtureBase.h`. Otherwise, looks good to go.

lib/model/unittest/CCountingModelTest.cc
```diff
@@ -23,6 +23,8 @@
 #include <string>
 #include <vector>
 
+#include "CModelTestFixtureBase.h"
```
droberts195, line 26: This should go after all the ML library includes but before the 3rd party includes, so after line 20 in this file (and similarly in the other 4 files).

droberts195 (approved):
I just left one more comment about the header ordering, but happy to merge without further review when that's moved.

PR #1501, opened 2020-09-23
[7.x][ML] Consume restore state stream to end (#1498)

dimitris-athanasiou:
Data frame analytics state may spread over multiple
documents. But unlike autodetect, where state is accompanied
by a model snapshot that allows us to know exactly how many
docs it comprises of, or categorizer, where we don't have
a snapshot equivalent but state can only grow, DFA state
may spread less docs than the previous one for a job.
This means that when the state is restored and streamed
into the c++ process, we might get more documents than
just the necessary ones.

We correctly restore state as we detect the `eos` marker.
However, we also need to consume the stream to its end
so that the thread writing to the stream may complete.

Backport of #1498

### DIFF
--- a/include/api/CSingleStreamSearcher.h
+++ b/include/api/CSingleStreamSearcher.h
@@ -38,12 +38,18 @@ public:
     //! called.
     CSingleStreamSearcher(const TIStreamP& stream);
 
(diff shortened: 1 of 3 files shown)


dimitris-athanasiou:
retest

PR #1502, opened 2020-09-23
[ML] Fix memory estimation for outlier detections

valeriy42:
In order to estimate the memory requirement we now use the maximum number of nearest neighbors instead of the expected average of the neighbors. This makes the new model memory estimation more robust.

I also adjusted the unit test assertions since now the estimated memory is higher.


edsavage (approved):
LGTM

PR #1506, opened 2020-09-24
[ML] Investigate total importance CI failure

valeriy42:
This PR instruments the unit tests that occasionally fail. Next time the test suppose to fail because an actual feature importance variable is 0, the test will dump the result json into the log so I can investigate the causes of failure.

I will remove this instrumentation once I've found the root cause of failure.


droberts195 (approved):
LGTM

droberts195:
Most of the failures for this test have been on the 7.x branch, so it's probably a good idea to backport to 7.x too.  We can revert the backport PR on the 7.10 branch after that is created if the problem hasn't been solved by then.

PR #1507, opened 2020-09-24
[ML] Change Docker push target

droberts195:
The host to push our Docker images to is changing from
push.docker.elastic.co to simply docker.elastic.co.

This change adjusts our scripts to reflect this.


dimitris-athanasiou (approved):
LGTM

PR #1509, opened 2020-09-24
[ML] Modernize and tidy model test code where possible

edsavage:
With respect to

```
CCountingModelTest.cc
CEventRateModelTest.cc
CEventRatePopulationModelTest.cc
CMetricModelTest.cc
CMetricPopulationModelTest.cc
```

* Use initializer lists to populate containers where possible
* Remove unnecessarily included header files

Related to #1477

### DIFF
--- a/lib/model/unittest/CEventRateModelTest.cc
+++ b/lib/model/unittest/CEventRateModelTest.cc
@@ -298,7 +298,7 @@ public:
 
 protected:
     using TInterimBucketCorrectorPtr = std::shared_ptr<CInterimBucketCorrector>;
-    using TEventRateModelFactoryPtr = boost::shared_ptr<CEventRateModelFactory>;
+    using TEventRateModelFactoryPtr = std::shared_ptr<CEventRateModelFactory>;
 
 protected:
     TInterimBucketCorrectorPtr m_InterimBucketCorrector;
--- a/lib/model/unittest/CEventRateModelTest.cc
+++ b/lib/model/unittest/CEventRateModelTest.cc
@@ -8,16 +8,12 @@
 #include <core/CRapidXmlParser.h>
 #include <core/CRapidXmlStatePersistInserter.h>
 #include <core/CRapidXmlStateRestoreTraverser.h>
-#include <core/CSmallVector.h>
 #include <core/Constants.h>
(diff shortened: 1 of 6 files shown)


lib/model/unittest/CEventRateModelTest.cc
```diff
@@ -41,7 +36,6 @@
 
 #include "CModelTestFixtureBase.h"
 
-#include <boost/foreach.hpp>
 #include <boost/range.hpp>
```
droberts195, line 39: You don't necessarily have to do it in this PR, but just an observation that this is here for `boost::size` applied to various types of container.  The loops that use it can probably be changed to range based for loops, and then this header can go too.

Also, while checking that I noticed there's a stray `boost::shared_ptr` in this file that wasn't changed to `std::shared_ptr` for some reason.

lib/model/unittest/CMetricPopulationModelTest.cc
```diff
@@ -1472,10 +1439,10 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
...
 
     messages.clear();
-    messages.push_back(SMessage(startTime + 10, "p1", "c1", TDouble1Vec(1, 21.0)));
-    messages.push_back(SMessage(startTime + 10, "p1", "c2", TDouble1Vec(1, 21.0)));
-    messages.push_back(SMessage(startTime + 10, "p2", "c1", TDouble1Vec(1, 21.0)));
-    messages.push_back(SMessage(startTime + 10, "p2", "c2", TDouble1Vec(1, 21.0)));
+    messages.push_back({startTime + 10, "p1", "c1", {1, 21.0}});
+    messages.push_back({startTime + 10, "p1", "c2", {1, 21.0}});
+    messages.push_back({startTime + 10, "p2", "c1", {1, 21.0}});
+    messages.push_back({startTime + 10, "p2", "c2", {1, 21.0}});
```
droberts195, line 1445: You could probably use `emplace_back` and get rid of one layer of braces.

droberts195 (approved):
LGTM

I saw a couple of things, but since you're doing multiple rounds of refactoring and have a green build you can leave them for the next round if you prefer.

Issue #1504, opened 2020-09-23
[ML] Processes that fail to connect to the JVM within a reasonable time should exit


droberts195:
Debugging the problems related to #1503 and https://github.com/elastic/elasticsearch/issues/62823 have highlighted that if an ML `autodetect`, `normalize` or `data_frame_analyzer` process fails to connect to the JVM (for example because the JVM gives up trying to open its named pipes) then that process will hang indefinitely.  Nothing wakes up the C++ process's wait for something to connect to the other end of its named pipes except something connecting to the other end of its named pipes.

`CThread` has functionality for waking up blocking calls that we can use to avoid this situation.  Processes that connect named pipes should start a thread before attempting to connect them that will interrupt the wait for a connection to the other end of the pipe after a certain time.  The question is what timeout to use:

1. Ideally this would be based on the `xpack.ml.process_connect_timeout` setting.  That would have to be passed to each C++ process in a command line argument.
2. The alternative would be a hardcoded timeout higher than any reasonable value for `xpack.ml.process_connect_timeout`, say 10 minutes.

droberts195:
#1514 adds a command line argument that a followup Java PR can set based on `xpack.ml.process_connect_timeout`.

Issue #1459, opened 2020-08-18
Investigate replacing strptime() with std::chrono::parse


droberts195:
Currently we are using the `strptime()` C function for time parsing.  This is a pain, as it requires workaround for Linux and Windows, so is one of our least platform independent pieces of code.

In theory [`std::chrono::parse`](https://en.cppreference.com/w/cpp/chrono/parse) might be better.  However some investigation is required to find out if `std::chrono::parse` is as portable as it should be.  Given what we found with `strptime()` Linux and Windows are the platforms most likely to not implement the standard fully, with timezone support being the likely problematic area.

If the portability _is_ better then we should remove our `strptime()` wrapper code and all uses of it and replace them with `std::chrono::parse`.  If workarounds would still be required then we might as well stick with what we've got until the C++ library _is_ consistent.

edsavage:
As `std::chrono::parse` will only be available from the `C++20` standard (https://en.cppreference.com/w/cpp/chrono/parse), which is not yet supported by our current compiler set, this investigation will be revisited at such a time when they do.

PR #1514, opened 2020-09-28
[ML] Add timeouts to named pipe connections

droberts195:
This PR adds timeouts to the named pipe connections of the
autodetect, normalize and data_frame_analyzer processes.
(The controller process already had a different mechanism,
tied to the ES JVM lifetime.)

Before this change if the ES JVM didn't connect to one of
the named pipes of an autodetect, normalize or
data_frame_analyzer process then it would hang forever.

After this change if nothing connects to the other end of
a named pipe that the C++ process has created within the
timeout period then it will delete the named pipe and exit.

The timeout will be supplied on the command line by a
followup Java PR.  In the meantime, (unreleased snapshot)
builds that have the C++ change but not the Java change will
use a timeout of 5 minutes, which is better than nothing and
highly unlikely to be less than the configured process
connect timeout in the ES settings.

Relates #1504


include/core/CBlockingCallCancellingTimer.h
```diff
@@ -0,0 +1,75 @@
...
+//! 2) This process creates named pipes and waits for Java to connect
+//! 3) Java never connects for some reason
+//!
+//! We want to unblock this process when this happens so that it can
+//! delete the named pipe and exit.
+//!
+class CBlockingCallCancellingTimer : public CBlockingCallCancellerThread {
+public:
+    //! Default timeout if not specified.
+    static const ml::core_t::TTime DEFAULT_TIMEOUT_SECONDS;
```
droberts195, line 42: I decided to stick with `TTime` here rather than switch to `std::chrono::duration`.  The reason is that all the existing command line argument parsing is currently using `TTime` and taking whole numbers of seconds, and it would be inconsistent if the new command line argument tried to do something clever with generic durations.

I think at some point we do need to move over to the `std::chrono` stuff throughout the codebase.  This is obviously a huge project but unless we do it we're going to have messy boundary points like this.  It would also allow us to support sub-second bucket spans more naturally than the scaling approach previously envisaged.  However, the time to do it is probably once we've upgraded to C++20 when the standard library will have support for parsing durations from text (as well as support for parsing timestamps - see #1459).

edsavage (approved):
LGTM

PR #1516, opened 2020-09-30
[ML] Prefer smaller models with similar performance

valeriy42:
For regression and classification, during hyperparameter optimization we prefer smaller models if the loss functions are otherwise comparable.

To this end, we add  0.01 * "forest number nodes" * E[GP] / "average forest number nodes" as an additional penalty.


valeriy42:
@wwang500 just a heads-up: in some cases this may lead to a regression of the results by <1% (in terms of MSE, MSLE, Huber) or less (for classification).

tveasey (approved):
LGTM

PR #1517, opened 2020-09-30
[ML] Replace boost::size in model unit tests

edsavage:
Use modern C++ data structures, initialisation to remove the need for
boost::size.

Use uniform braced initialisation of variables consistently.

Relates to #1477

### DIFF
--- a/lib/model/unittest/CModelTestFixtureBase.h
+++ b/lib/model/unittest/CModelTestFixtureBase.h
@@ -22,6 +22,8 @@
 #include <utility>
 #include <vector>
 
+using TBoolVec = std::vector<bool>;
+
 using TDouble1Vec = ml::core::CSmallVector<double, 1>;
 using TDouble2Vec = ml::core::CSmallVector<double, 2>;
 using TDouble4Vec = ml::core::CSmallVector<double, 4>;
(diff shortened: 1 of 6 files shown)


lib/model/unittest/CModelTestFixtureBase.h
```diff
@@ -54,11 +56,17 @@ using TSizeSizePrUInt64Map = std::map<TSizeSizePr, uint64_t>;
 using TSizeVec = std::vector<std::size_t>;
 using TSizeVecVec = std::vector<TSizeVec>;
 using TSizeVecVecVec = std::vector<TSizeVecVec>;
+using TStrSizePr = std::pair<std::string, std::size_t>;
+using TStrSizePrVec = std::vector<TStrSizePr>;
+using TStrSizePrVecVec = std::vector<TStrSizePrVec>;
+using TStrSizePrVecVecVec = std::vector<TStrSizePrVecVec>;
+using TStrUInt64Map = std::map<std::string, uint64_t>;
```
droberts195, line 63: It would still be good to work towards having all these types as public members of the test fixture rather than in the global namespace, but it can wait for a subsequent PR in the series.

droberts195 (approved):
LGTM

PR #1519, opened 2020-10-01
[ML] Fix total feature importance unit tests

valeriy42:
We have an occasional failure in the total feature importance unit tests like [this one](https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+7.x+windows/109/). It happens while for some reason the mean value accumulator fails to produce correct results:
```json
            {
                "feature_name": "c2",
                "classes": [
                    {
                        "class_name": "foo",
                        "importance": {
                            "mean_magnitude": 0,
                            "min": -3.288343526748277,
                            "max": 3.288343526748277
                        }
                    },
                    {
                        "class_name": "bar",
                        "importance": {
                            "mean_magnitude": 0,
                            "min": -3.288343526748277,
                            "max": 3.288343526748277
                        }
                    }
                ]
            }
```
In an attempt to fix this I changed the mean accumulator from using the `eigen` vector type to an explicit vector of doubles same as the min/max accumulator. 

I keep the instrumentation of the unit tests for now so we have enough info if the tests continue to fail.


droberts195 (approved):
LGTM

valeriy42:
retest

PR #1520, opened 2020-10-01
[ML] Make controller send responses for each command received

droberts195:
This change makes the controller process respond to each command
it receives with a document indicating whether that command was
successfully executed or not.

This response will be used by the Java side of the connection to
determine when it is appropriate to move on to the next phase of
the action that the controller command was part of.  For example,
when starting a process and connecting named pipes to it it is
best that the named pipe connections are not attempted until the
process is confirmed to be started.

Relates elastic/elasticsearch#62823
Relates elastic/elasticsearch#63542


edsavage (approved):
LGTM

droberts195:
retest

PR #1523, opened 2020-10-02
[ML] Move model test helper functions to base class

edsavage:
Move existing model test helper functions to base class and share functionality where possible.

While this goes some way to reducing code duplication there may still be scope to
 * generalise the existing helper functions further to reduce their number
 * add further helper functions to encapsulate code duplicated in multiple suites 

Relates to #1477

### DIFF
--- a/lib/model/unittest/CEventRateModelTest.cc
+++ b/lib/model/unittest/CEventRateModelTest.cc
@@ -1110,9 +1110,6 @@ BOOST_FIXTURE_TEST_CASE(testPrune, CTestFixture) {
                         clonedModel->dataGatherer().numberActivePeople());
 }
 
-
-
-
 BOOST_FIXTURE_TEST_CASE(testKey, CTestFixture) {
     function_t::TFunctionVec countFunctions{function_t::E_IndividualCount,
                                             function_t::E_IndividualNonZeroCount,
@@ -1126,8 +1123,9 @@ BOOST_FIXTURE_TEST_CASE(testKey, CTestFixture) {
     std::string overFieldName;
 
     generateAndCompareKey(countFunctions, fieldName, overFieldName,
-                          [](CSearchKey expectedKey, CSearchKey actualKey){
-                              BOOST_TEST_REQUIRE(expectedKey == actualKey);});
+                          [](CSearchKey expectedKey, CSearchKey actualKey) {
+                              BOOST_TEST_REQUIRE(expectedKey == actualKey);
+                          });
 }
 
 BOOST_FIXTURE_TEST_CASE(testModelsWithValueFields, CTestFixture) {
--- a/lib/model/unittest/CEventRateModelTest.cc
+++ b/lib/model/unittest/CEventRateModelTest.cc
@@ -124,25 +124,6 @@ public:
         }
     }
 
-    CEventData makeEventData(core_t::TTime time, std::size_t pid) {
-        CEventData eventData;
-        eventData.time(time);
-        eventData.person(pid);
-        eventData.addAttribute(std::size_t(0));
-        eventData.addValue(TDoubleVec(1, 0.0));
-        return eventData;
-    }
-
-    CEventData makeEventData(core_t::TTime time, std::size_t pid, const std::string value) {
-        CEventData eventData;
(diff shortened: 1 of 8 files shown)


lib/model/unittest/CEventRateModelTest.cc
```diff
@@ -44,218 +44,163 @@
...
+BOOST_TEST_DONT_PRINT_LOG_VALUE(CModelTestFixtureBase::TStrVec::iterator)
 
 BOOST_AUTO_TEST_SUITE(CEventRateModelTest)
 
 using namespace ml;
 using namespace model;
 
 namespace {
-
 const std::string EMPTY_STRING;
```
droberts195, line 55: This is in an unnamed namespace in several of the files.  Could it be made a public or protected static constant on the test fixture?  Then it would still be usable in the tests without qualification, but wouldn't be duplicated.

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -57,176 +57,35 @@ namespace {
...
-                   SAnnotatedProbability& probability) {
-    for (std::size_t i = 0u; i < bucket.size(); ++i) {
-        addArrival(gatherer, resourceMonitor, time, "p", bucket[i],
-                   TOptionalStr(influencerValues[i]));
+class CTestFixture : public CModelTestFixtureBase {
+public:
+    CEventData makeEventData(core_t::TTime time,
+                             std::size_t pid,
+                             double value,
+                             const TOptionalStr& influence = TOptionalStr()) {
```
tveasey, line 68: I think it should be possible to move this into the base class. We have `makeEventData` in `CEventRateModelTest` which I think this could be combined with.

lib/model/unittest/CModelTestFixtureBase.cc
```diff
@@ -0,0 +1,310 @@
...
+        orderedAnomalies.push_back(anomalies[i].second);
+    }
+
+    std::sort(orderedAnomalies.begin(), orderedAnomalies.end());
+
+    LOG_DEBUG(<< "orderedAnomalies = "
+              << ml::core::CContainerPrinter::print(orderedAnomalies));
+}
+
+void CModelTestFixtureBase::generateAndCompareKey(const ml::model::function_t::TFunctionVec& countFunctions,
```
tveasey, line 280: This seems like it should be static.

lib/model/unittest/CModelTestFixtureBase.cc
```diff
@@ -0,0 +1,310 @@
...
+    TBoolVec useNull{true, false};
+    TStrVec byFields{"", "by"};
+    TStrVec partitionFields{"", "partition"};
+
+    ml::model::CAnomalyDetectorModelConfig config =
+        ml::model::CAnomalyDetectorModelConfig::defaultConfig();
+
+    int detectorIndex{0};
+    for (const auto& countFunction : countFunctions) {
+        for (bool usingNull : useNull) {
```
tveasey, line 293: Probably just moved here, but this can be more concise:
```suggestion
        for (bool usingNull : {true, false}) {
```
then delete `useNull`. Similar for others below.

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -1140,8 +997,8 @@ BOOST_FIXTURE_TEST_CASE(testInfluence, CTestFixture) {
 
         core_t::TTime time{startTime};
         for (std::size_t i = 0u; i < values.size(); ++i) {
-            processBucket(time, bucketLength, values[i], influencers[i], *gatherer,
-                          m_ResourceMonitor, model, annotatedProbability);
+            processBucket(time, bucketLength, values[i], influencers[i],
```
tveasey, line 1000: For consistency it feels like all the calls to `processBucket` should be qualified (it's our usual style and you've done it for the other methods you've added):
```suggestion
            this->processBucket(time, bucketLength, values[i], influencers[i],
```

tveasey (commented):
Overall looks good. I made a few minor comments and also wonder if we can pull one more bit of functionality up to the base class.

tveasey (approved):
LGTM

PR #1526, opened 2020-10-07
[ML] Allow unbounded num_top_classes in classification analysis 

przemekwitek:
This PR makes the classification analysis accept `-1` as the value of `num_top_classes` parameter. When `num_top_classes` is set to `-1`, all the predicted classes will be emitted in the `top_classes` array in the results.

Relates https://github.com/elastic/elasticsearch/issues/63306

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
@@ -81,7 +81,7 @@ CDataFrameTrainBoostedTreeClassifierRunner::CDataFrameTrainBoostedTreeClassifier
     : CDataFrameTrainBoostedTreeRunner{
           spec, parameters, loss(parameters[NUM_CLASSES].as<std::size_t>())} {
 
-    m_NumTopClasses = parameters[NUM_TOP_CLASSES].fallback(std::size_t{0});
+    m_NumTopClasses = parameters[NUM_TOP_CLASSES].as<std::ptrdiff_t>();
     m_PredictionFieldType =
         parameters[PREDICTION_FIELD_TYPE].fallback(E_PredictionFieldTypeString);
     this->boostedTreeFactory().classAssignmentObjective(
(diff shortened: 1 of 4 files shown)


lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
```diff
@@ -138,14 +138,16 @@ void CDataFrameTrainBoostedTreeClassifierRunner::writeOneRow(
...
         TSizeVec classIds(scores.size());
         std::iota(classIds.begin(), classIds.end(), 0);
         std::sort(classIds.begin(), classIds.end(),
                   [&scores](std::size_t lhs, std::size_t rhs) {
                       return scores[lhs] > scores[rhs];
                   });
-        classIds.resize(std::min(classIds.size(), m_NumTopClasses));
+        // -1 is a special value meaning "output all the classes"
+        classIds.resize(m_NumTopClasses == -1 ? classIds.size()
+                                              : std::min(classIds.size(), m_NumTopClasses));
```
tveasey, line 150: I expect that this comparison generates a compiler warning. My inclination would be to change to type of `m_NumTopClasses` to `std::ptrdiff_t`. You'll then have to change `std::min(classIds.size(), m_NumTopClasses)` to `std::min(classIds.size(), static_cast<std::ptrdiff_t>(m_NumTopClasses)`.

tveasey (commented):
Functionally this looks fine but the signed unsigned comparison will generate a compiler warning with our settings so I think you should change the type of `m_NumTopClasses`.

lib/api/CDataFrameTrainBoostedTreeClassifierRunner.cc
przemekwitek: Done.
Changed to `std::min(classIds.size(), static_cast<std::size_t>(m_NumTopClasses)`

tveasey (approved):
LGTM

PR #1530, opened 2020-10-12
[ML] Anomaly detection residual model initialisation improvements

tveasey:
This addresses some of the fallout from #1202 in our QA suite. It improves reinitialisation of the residual model after detecting a new component and sets the amount of model history based on the window length (rather than a fixed short interval). This improves stability after reinitialisation. I've also improved the interaction between the long and short window tests.

I'm going to make a couple of related changes, but I'm splitting them into separate PRs. I've just disabled any failing unit tests and will reenable when everything is committed. As before, this will not be documented separately.


valeriy42 (approved):
LGTM

PR #1534, opened 2020-10-13
[7.x][ML] Make controller send responses for each command received

droberts195:
This change makes the controller process respond to each command
it receives with a document indicating whether that command was
successfully executed or not.

This response will be used by the Java side of the connection to
determine when it is appropriate to move on to the next phase of
the action that the controller command was part of. For example,
when starting a process and connecting named pipes to it it is
best that the named pipe connections are not attempted until the
process is confirmed to be started.

Backport of #1520


droberts195:
retest

PR #1535, opened 2020-10-14
[ML] Retry pulling of Docker images to work around network issues

droberts195:
Adds a function to retry pulling Docker images up to 5 times, and
calls it to pre-fetch every Docker image that's used as a "FROM"
image by our "docker build" commands.

"docker build" will pull the "FROM" image if it's not present on
the local machine, but will only try once, so is vulnerable to a
transient problem with the Docker registry failing the whole build.
By retrying "docker pull" several times before running "docker build"
we ensure that the image is present locally when "docker build" runs
and hence "docker build" never needs to pull it.

### DIFF
--- a/dev-tools/docker/build_check_style_image.sh
+++ b/dev-tools/docker/build_check_style_image.sh
@@ -21,7 +21,10 @@ set -e
 
 cd `dirname $0`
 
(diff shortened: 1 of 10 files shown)


valeriy42 (approved):
LGTM

PR #1536, opened 2020-10-15
[ML] Amalgamate addArrival methods

edsavage:
Provide a common interface to addArrival by extending the existing SMessage structure.

This reduces the amount of code duplicated between multiple addArrival methods at the expense of some added verbosity when specifying default arguments.

Relates to #1477


droberts195 (approved):
LGTM

PR #1537, opened 2020-10-15
[ML] Gain upper bound estimation for classification and regression

valeriy42:
In this PR we start computing an upper bound on the potential gain from splitting a node. If the upper bound of the gain is lower than the currently smallest gain among all candidates, we ignore the node and this way prevent computations that are especially expensive on the large datasets.

Since we avoid computation of the splits that we wouldn't be added to the tree anyway, this PR does not change the qualitative results.

At the moment, we can only compute the upper bound for regression and binary classification. For multiclass classification we proceed as before.

Note: this PR contains additional instrumentation to assess the performance improvement. I will remove this instrumentation in  a follow-up PR  after tests.


lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -267,6 +335,20 @@ void CBoostedTreeLeafNodeStatistics::addRowDerivatives(const CEncodedDataFrameRo
 
     auto derivatives = readLossDerivatives(row.unencodedRow(), m_ExtraColumns,
                                            m_NumberLossParameters);
+    if (derivatives.size() == 2) {
```
tveasey, line 338: This is the hottest path in the training code. Whilst this condition is predictable we know outside of this function if it true or not. It might be worth having another version of this function, you could have say `updateDerivatives` and `updateDerivativesAndBounds`, and pulling the check out of the loop over rows.

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -267,6 +335,20 @@ void CBoostedTreeLeafNodeStatistics::addRowDerivatives(const CEncodedDataFrameRo
...
+        if (derivatives(0) >= 0.0) {
+            splitsDerivatives.m_PositiveDerivativesGSum += derivatives(0);
+            splitsDerivatives.m_PositiveDerivativesHSum += derivatives(1);
```
tveasey, line 341: I would have `m_PositiveDerivatives`, `m_NegativeDerivatives`, `m_PositiveDerivativesMin` and `m_PositiveDerivativesMax` and have `splitsDerivatives.addPositiveDerivatives(derivatives)` and `splitsDerivatives.addNegativeDerivatives(derivatives)`. For example, `addPositiveDerivatives` would look like:
```
  m_PositiveDerivatives += derivatives;
  m_PositiveDerivativesMin.cwiseMin(derivatives);
  m_PositiveDerivativesMax.cwiseMax(derivatives);
```
When you return the value you can do std::max(m_PositiveDerivativesMax(1), 0.0) in place of taking `fabs` here. I would make `m_PositiveDerivatives`, `m_NegativeDerivatives` type Eigen::Matrix<double, 2, 1> no heap allocations. And `m_PositiveDerivativesMin` and `m_PositiveDerivativesMax` type Eigen::Matrix<CFloatStorage, 2, 1>. This way the coefficientwise min and max operations should use float comparison and be vectorised. You just need to initialise `m_PositiveDerivativesMin` to `std::numeric_limits<float>::max()`, etc. The accessors you have have can remain, i.e. you could have a `::negativeDerivativesGSum() const` which reads the value out of the appropriate vector coefficient so the rest of the code doesn't need to now how they're stored and updated.

tveasey (commented):
I know this is still work in progress and needs a little tidying up. I made a couple of comments on the implementation to make the update of the bound related information as efficient as possible.

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -399,29 +503,43 @@ CBoostedTreeLeafNodeStatistics::computeBestSplitStatistics(const TRegularization
...
+                                                minLossRight[ASSIGN_MISSING_TO_RIGHT];
+            }
 
             if (gain[ASSIGN_MISSING_TO_LEFT] > maximumGain) {
                 maximumGain = gain[ASSIGN_MISSING_TO_LEFT];
                 splitAt = m_CandidateSplits[feature][split];
                 leftChildRowCount = cl[ASSIGN_MISSING_TO_LEFT];
                 assignMissingToLeft = true;
+                maxGainLeftChild = C - (minLossLeft[ASSIGN_MISSING_TO_LEFT]);
+                maxGainRightChild = C - (minLossRight[ASSIGN_MISSING_TO_LEFT]);
```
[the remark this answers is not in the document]
valeriy42: Good observation :+1:  Fixed it

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: I tried this change in there was so difference in the performance :man_shrugging: I guess the compiler is already good enough to restructure/optimize the loop. Therefore, I will not create two versions of the function to avoid code duplication.

valeriy42: Done

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: I implemented the suggested change. It doesn't seem to make the code significantly faster, but the implementation is cleaner now.

valeriy42: I don't use `CFloatStorage` though to avoid in the hot loop.

tveasey: > I don't use CFloatStorage though to avoid in the hot loop.

That's interesting. `CFloatStorage` will use float comparison operations. Since the underlying types are float it therefore shouldn't coerce them to doubles before comparing. I've definitely found switching min/max operations to use them speed up other code before. Note this was only being proposed for `m_*DerivativesMin` and `m_*DerivativesMax` and not the values you accumulate into.

valeriy42:
@tveasey I finished up this PR. It would be great if you could take another look at it.

valeriy42:
@tveasey Also, do you think we need to update the memory estimation?

tveasey:
> Also, do you think we need to update the memory estimation?

I don't think so: the extra memory is accounted for in the sizeof of the derivatives object so we should get it accounted for already.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -808,16 +817,18 @@ CBoostedTreeImpl::trainTree(core::CDataFrame& frame,
...
 
     // Sampling transforms the probabilities. We use a placeholder outside
     // the loop adding nodes so we only allocate the vector once.
     TDoubleVec featureSampleProbabilities{m_FeatureSampleProbabilities};
     TSizeVec featureBag;
     this->featureBag(featureSampleProbabilities, featureBag);
 
-    TLeafNodeStatisticsPtrQueue leaves(maximumTreeSize / 2 + 3);
-    leaves.push_back(std::make_shared<CBoostedTreeLeafNodeStatistics>(
+    TLeafNodeStatisticsPtrQueue splitCandidateTreeNodes(maximumNumberInternalNodes / 2 + 3);
```
tveasey, line 830: This is the current working set of tree leaves' statistics, i.e. notice we always pop the node after splitting it, alternatively we only ever consider splitting a leaf of the current tree. Therefore, I wouldn't drop this from the name. Since you know have a constraint such that it isn't all leaves how about calling it `splittableLeaves`?
```suggestion
    TLeafNodeStatisticsPtrQueue splittableLeaves(maximumNumberInternalNodes / 2 + 3);
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -878,32 +891,43 @@ CBoostedTreeImpl::trainTree(core::CDataFrame& frame,
         featureSampleProbabilities = m_FeatureSampleProbabilities;
         this->featureBag(featureSampleProbabilities, featureBag);
 
+        std::size_t n{splitCandidateTreeNodes.size()};
```
tveasey, line 894:
```suggestion
        std::size_t numberSplittableLeaves{splitCandidateTreeNodes.size()};
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -878,32 +891,43 @@ CBoostedTreeImpl::trainTree(core::CDataFrame& frame,
...
+        std::size_t currentNumberInternalNodes{(tree.size() - 1) / 2};
+        std::size_t lastCandidateIdx{n - (maximumNumberInternalNodes - currentNumberInternalNodes)};
```
tveasey, line 896: Doesn't this underflow in a bad way? I think this needs to be:
```suggestion
        auto smallestCurrentCandidateGainIndex = static_cast<std::ptrdiff_t>(numberSplittableLeaves) - static_cast<std::ptrdiff_t>(maximumNumberInternalNodes - currentNumberInternalNodes);
```

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -24,6 +24,27 @@ using TRowItr = core::CDataFrame::TRowItr;
 namespace {
 const std::size_t ASSIGN_MISSING_TO_LEFT{0};
 const std::size_t ASSIGN_MISSING_TO_RIGHT{1};
+
+void incrementStatsComputed(CBoostedTreeLeafNodeStatistics::TAnalysisInstrumentationPtr instrumentation) {
```
tveasey, line 28: +1 on having these utility functions (improves readability).

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -125,29 +146,66 @@ CBoostedTreeLeafNodeStatistics::TPtrPtrPr
...
+        }
+    } else {
+        incrementStatsNotComputed(instrumentation, this->m_BestSplit.s_RightChildRowCount);
+        if (this->m_BestSplit.s_LeftChildMaxGain > gainThreshold) {
+            leftChild = std::make_shared<CBoostedTreeLeafNodeStatistics>(
+                leftChildId, *this, numberThreads, frame, encoder, regularization,
+                featureBag, true /*is left child*/, split, workspace);
+            incrementStatsComputed(instrumentation);
+        } else {
+            incrementStatsNotComputed(instrumentation, this->m_BestSplit.s_RightChildRowCount);
```
tveasey, line 206:
```suggestion
            incrementStatsNotComputed(instrumentation, this->m_BestSplit.s_LeftChildRowCount);
```

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -398,30 +479,47 @@ CBoostedTreeLeafNodeStatistics::computeBestSplitStatistics(const TRegularization
...
-            gain[ASSIGN_MISSING_TO_RIGHT] =
-                cl[ASSIGN_MISSING_TO_RIGHT] == 0 || cl[ASSIGN_MISSING_TO_RIGHT] == c
-                    ? -INF
-                    : minimumLoss(gl[ASSIGN_MISSING_TO_RIGHT], hl[ASSIGN_MISSING_TO_RIGHT]) +
-                          minimumLoss(gr[ASSIGN_MISSING_TO_RIGHT],
-                                      hr[ASSIGN_MISSING_TO_RIGHT]);
+            if (cl[ASSIGN_MISSING_TO_LEFT] == 0 || cl[ASSIGN_MISSING_TO_LEFT] == c) {
+                gain[ASSIGN_MISSING_TO_LEFT] = -INF;
+            } else {
+                minLossLeft[ASSIGN_MISSING_TO_LEFT] = minimumLoss(
```
tveasey, line 485: Does it matter that we only set `minLossLeft` in this branch?  I don't think so because they are only used if `gain > -INF`, but this is not immediately obvious from the code. I think it warrants a comment.

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -434,24 +532,72 @@ CBoostedTreeLeafNodeStatistics::computeBestSplitStatistics(const TRegularization
...
+    }
 
     LOG_TRACE(<< "best split: " << result.print());
 
     return result;
 }
+
+double CBoostedTreeLeafNodeStatistics::childMaxGain(double gChild,
+                                                    double minLossChild,
+                                                    double lambda) const {
```
tveasey, line 579: I think it might be worthwhile writing a brief outline of what this is doing, i.e. something along the lines of: "This computes the maximum possible gain we can expect splitting a child node given we know the sum of the positive (g^+) and negative gradients (g^-) at its parent, the minimum curvature on the positive and negative gradient set (hmin^+ and hmin^-) and largest and smallest gradient (gmax and gmin, respectively). The highest possible gain consistent with these constraints can be shown to be: (g^+)^2 / (hmin^+ g^+ / gmax + lambda) + (g^-)^2 / (hmin^- g^- / gmin + lambda)".

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -434,24 +532,72 @@ CBoostedTreeLeafNodeStatistics::computeBestSplitStatistics(const TRegularization
...
+        result.s_LeftChildMaxGain = INF;
+        result.s_RightChildMaxGain = INF;
+    } else if (result.s_Gain > 0) {
+        double childPenaltyForDepth{regularization.penaltyForDepth(m_Depth + 1)};
+        double childPenaltyForDepthPlusOne{regularization.penaltyForDepth(m_Depth + 2)};
+        double childPenalty{regularization.treeSizePenaltyMultiplier() +
+                            regularization.depthPenaltyMultiplier() *
+                                (2.0 * childPenaltyForDepthPlusOne - childPenaltyForDepth)};
+        result.s_LeftChildMaxGain =
+            0.5 * childMaxGain(childrenGainStatsGlobal.s_GLeft,
```
tveasey, line 562:
```suggestion
            0.5 * this->childMaxGain(childrenGainStatsGlobal.s_GLeft,
```

lib/maths/CBoostedTreeLeafNodeStatistics.cc
```diff
@@ -434,24 +532,72 @@ CBoostedTreeLeafNodeStatistics::computeBestSplitStatistics(const TRegularization
...
+                               childrenGainStatsGlobal.s_MinLossLeft, lambda) -
+            childPenalty;
+
+        result.s_RightChildMaxGain =
+            0.5 * childMaxGain(childrenGainStatsGlobal.s_GRight,
```
tveasey, line 567:
```suggestion
            0.5 * this->childMaxGain(childrenGainStatsGlobal.s_GRight,
```

tveasey (commented):
Over all looks great. There are a couple of potential issues I think. One is just in the temporary stats you compute, but the other looks like it is a real issue caused by unsigned integer underflow. I also made some minor suggestions for readability and consistency.

include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -614,12 +712,18 @@ class MATHS_EXPORT CBoostedTreeLeafNodeStatistics final {
                          double splitAt,
                          std::size_t minimumChildRowCount,
                          bool leftChildHasFewerRows,
-                         bool assignMissingToLeft)
+                         bool assignMissingToLeft,
+                         std::size_t leftChildRowCount, // TODO remove after stats measurement
+                         std::size_t rightChildRowCount, // TODO remove after stats measurement
```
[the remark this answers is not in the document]
valeriy42: I remove them in a follow-up PR.

include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -182,16 +184,32 @@ class MATHS_EXPORT CBoostedTreeLeafNodeStatistics final {
     class MATHS_EXPORT CSplitsDerivatives {
     public:
         using TDerivativesVec = std::vector<CDerivatives>;
+        using TDerivativesMappedVec = boosted_tree_detail::TAlignedMemoryMappedFloatVector;
```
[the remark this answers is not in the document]
valeriy42: I removed the typedef as you suggested.

include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -338,9 +378,51 @@ class MATHS_EXPORT CBoostedTreeLeafNodeStatistics final {
...
+        }
+
+        double negativeDerivativesHMin() const {
+            return m_NegativeDerivativesMin(1);
+        }
+
     private:
         using TDerivativesVecVec = std::vector<TDerivativesVec>;
         using TAlignedDoubleVec = std::vector<double, core::CAlignedAllocator<double>>;
+        using TDerivatives2D = Eigen::Matrix<double, 2, 1>;
```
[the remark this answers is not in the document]
valeriy42: Renamed

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -295,6 +295,15 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
     m_Instrumentation->updateProgress(1.0);
     m_Instrumentation->updateMemoryUsage(
         static_cast<std::int64_t>(this->memoryUsage()) - lastMemoryUsage);
+
+    if (m_Instrumentation != nullptr) {
+        LOG_INFO(<< "Statistics computed: " << m_Instrumentation->statisticsComputed()
```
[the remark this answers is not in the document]
valeriy42: I added a TODO comment and will remove this output completely together wit the instrumentation code.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -808,16 +817,18 @@ CBoostedTreeImpl::trainTree(core::CDataFrame& frame,
     workspace.reinitialize(m_NumberThreads, candidateSplits, m_Loss->numberParameters());
 
     TNodeVec tree(1);
-    tree.reserve(2 * maximumTreeSize + 1);
+    // since number of leaves in a perfect binary tree is (numberInternalNodes+1)
+    // the total number of nodes in a tree is (2*numberInternalNodes+1)
```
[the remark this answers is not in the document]
valeriy42: Fixed

lib/maths/CBoostedTreeImpl.cc
valeriy42: Fixed

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: Good catch!

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: Renamed

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: Renamed

lib/maths/CBoostedTreeImpl.cc
valeriy42: Renamed

lib/maths/CBoostedTreeImpl.cc
valeriy42: You are right, it was `int` in the original version, I changed it to `size_t` for clean-up, but it should be `ptrdiff_t` of course.

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: You are right, I added a comment.

lib/maths/CBoostedTreeLeafNodeStatistics.cc
valeriy42: I added the explanation comment as you suggested.

tveasey (approved):
LGTM

Issue #1253, opened 2020-05-15
[ML] Change way the job config is passed to autodetect


droberts195:
For historical reasons the job config that `autodetect` uses is passed to it in a mixture of 2 Windows `.ini` file style config files and a few command line options.

It would make life much easier and would make the code clearer if it was passed as a single config file in JSON format containing the exact job config document from the `.ml-config` index.

Benefits would be:

1. Much less code on the Java side - currently it has to translate the job config into a completely different format to pass to `autodetect`
2. Less of a learning curve for new joiners - when looking at the config handling code in `autodetect` they will be able to look at job config examples from our public documentation
3. Easier reproduction of problems directly using the `autodetect` process from the command line - currently when we get a job config from a support diag we have to manually translate it to the `.ini` formats and command line arguments in order to try to reproduce problems

droberts195:
One tricky thing about this change is how to keep the way Java passes the inputs to C++ working while the changes are made.

I think the way to do it is as follows:

1. Add an option to the C++ code to accept a job config in JSON format.  This could be a `--config=/path/to/config.json` argument.  Initially nothing needs to be done with the supplied config, not even confirming that the file contains JSON.
2. Change the Java code to pass `--config=/path/to/config.json` _as well as_ all the arguments it currently passes.  The JSON file it should save to disk (in the same temporary directory where it currently writes the limits, model plot and field config files) should contain the job config JSON in UTF-8 format (as created by `Job.toXContent()`) _with no adjustments_.  No adjustments is critical to making debugging easier in the future.
3. Change the C++ code to get its config information from the JSON config specified to `--config` instead of the mishmash of options it gets them from at the moment.  This is obviously the main part of the work.  Additionally the options that are no longer needed should be made optional (if they aren't already) but not completely removed.
4. Change the Java code not to pass the redundant options: `--limitconfig`, `--fieldconfig`, `--bucketspan`, `--jobid`, `--modelplotconfig`, `--latency`, `--summarycountfield`, `--delimiter`, `--multivariateByFields`, `--stopCategorizationOnWarnStatus`, `--maxQuantileInterval`, `--persistInterval`.  This should allow quite a lot of Java code to be deleted.
[...]

PR #1540, opened 2020-10-16
[ML] Add argument for config file to autodetect

edsavage:
Add an argument to autodetect to enable specifying a config file.

Eventually this will be used to read an autodetect job config in JSON
format but in this initial step it is unused. 

Relates to #1253


bin/autodetect/CCmdLineParser.cc
```diff
@@ -62,6 +63,8 @@ bool CCmdLineParser::parse(int argc,
         desc.add_options()
             ("help", "Display this information and exit")
             ("version", "Display version information and exit")
+            ("config", boost::program_options::value<std::string>(),
+             "The configuration file")
```
droberts195, line 67: It's probably worth saying "The job configuration file", to hint that what's required is the job config

droberts195 (approved):
LGTM

Issue #1545, opened 2020-10-26
[ML] min_version for anomaly detector model state is wrong


droberts195:
The `min_version` field for anomaly detector model state that stops model snapshots being loaded on nodes that are too old to understand then is currently incorrect.  It is [set to version 6.4.0](https://github.com/elastic/ml-cpp/blob/7d1e67b5a8669ad0028b87f1ca4df2122a69d210/lib/api/CAnomalyJob.cc#L88), when we have updated the model state format more recently than that.

It's too late to correct released versions, but we can make some improvements:

1. For 7.10.0 we can update the constant to the version where the previous breaking change to model state was made
2. For 7.11.0 we can update the constant to version 7.11.0 to reflect the changes being made for time series decomposition and change point detection

droberts195:
> 1. For 7.10.0 we can update the constant to the version where the previous breaking change to model state was made

I believe this was the change made in 7.9.0 in #1340

PR #1546, opened 2020-10-26
[ML] Correcting model state min_version

droberts195:
The model state min_version should indicate the minimum
node version required to restore the model state.  This
has not been updated since 6.4.0, meaning we have not had
the intended protection against model states being used on
nodes that are too old.  This change updates the
min_version to the correct version at present, which is
7.9.0 for latest state.

Relates #1545

### DIFF
--- a/lib/api/CAnomalyJob.cc
+++ b/lib/api/CAnomalyJob.cc
@@ -85,8 +85,11 @@ const std::string INTERIM_BUCKET_CORRECTOR_TAG("k");
 
 //! The minimum version required to read the state corresponding to a model snapshot.
 //! This should be updated every time there is a breaking change to the model state.
-//! (The last breaking change was in 7.9 in lib/core/CPackedBitVector.cc in
-//! https://github.com/elastic/ml-cpp/pull/1340)
+//! Newer versions are able to read the model state of older versions, but older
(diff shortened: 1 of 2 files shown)


lib/api/CAnomalyJob.cc
```diff
@@ -85,7 +85,9 @@ const std::string INTERIM_BUCKET_CORRECTOR_TAG("k");
 
 //! The minimum version required to read the state corresponding to a model snapshot.
 //! This should be updated every time there is a breaking change to the model state.
-const std::string MODEL_SNAPSHOT_MIN_VERSION("6.4.0");
+//! (The last breaking change was in 7.9 in lib/core/CPackedBitVector.cc in
+//! https://github.com/elastic/ml-cpp/pull/1340)
+const std::string MODEL_SNAPSHOT_MIN_VERSION("7.9.0");
```
tveasey, line 90: Can we add a comment that this is not breaking backwards compatibility but forwards compatibility as well? We probably should have a QA test: i.e. export model and try to load into the build with MODEL_SNAPSHOT_MIN_VERSION version. I suspect other changes have broken forwards compatibility before this one (although this will be the last).

tveasey (approved):
One minor comment, but LGTM.

PR #1549, opened 2020-10-27
[ML] Short circuit seasonal component gradient calculation

tveasey:
Profiling showed up the calculation for computing seasonal component gradients is surprisingly expensive. The effect is more pronounced for longer bucket lengths, but it was accounting for nearly 7% of the entire runtime for one of our QA data sets. There is a trivial optimisation: we care only about the sign margin for the gradient (the amount by which the component gradients are all positive or all negative). As soon as we know this is zero we can short circuit the remaining work.

Before this change we have:
Screenshot 2020-10-27 at 11 35 00

After this change we have:
Screenshot 2020-10-27 at 16 03 06

This is a pure runtime optimisation and has no impact on results.


droberts195 (approved):
LGTM

PR #1544, opened 2020-10-21
[ML] Time series time shift segmentation

tveasey:
This implements segmentation of a window of times series values allowing for piecewise constant time shifts of the seasonality. This is needed to extend change detection to handle a wider range of possible time shifts. I also took the opportunity to finish up testing time series segmentation and make some improvements to documentation and variable naming.

This isn't used yet; I plan to wire it in the next change. This is marked as a non-issue because it will not be documented separately.

PR #1550, opened 2020-10-29
[ML] New change point detection

tveasey:
This reworks change point detection to makes use of the enhancements to `CTimeSeriesSegmentation` introduced in #1543 and #1544. It also revisits and ensures that all disabled tests are enabled and passing.

This is a non-issue since it will not be documented separately.

~(This is WIP since I expect to have to make some more tweaks to `CTimeSeriesTestForChange` and I have haven't yet fixed and re-enabled all unit tests.)~


lib/maths/CTimeSeriesDecompositionDetail.cc
```diff
@@ -292,7 +292,7 @@ const core::TPersistenceTag MEAN_OFFSET_7_11_TAG{"c", "mean_offset"};
 const core::TPersistenceTag RESIDUAL_MOMENTS_7_11_TAG{"d", "residual_moments"};
 const core::TPersistenceTag LARGE_ERROR_FRACTION_7_11_TAG{"e", "large_error_fraction"};
 const core::TPersistenceTag LAST_TEST_TIME_7_11_TAG{"f", "last_test_time"};
-const core::TPersistenceTag LAST_CHANGE_TIME_7_11_TAG{"g", "last_change_time"};
+const core::TPersistenceTag LAST_UNCOMMITTED_CHANGE_TIME_7_11_TAG{"g", "last_uncommitted_change_time"};
```
droberts195, line 295: I know it's not directly related to this change, but if you haven't done so already please can you update `api::CAnomalyJob::MODEL_SNAPSHOT_MIN_VERSION` to `7.11.0` on your feature branch so we won't try to load the new format on old nodes.

tveasey: Good point: it's on my list. I was going to bump it at the point I merged to master in case this slipped, but I think that's unlikely now so I can add in this PR.

lib/maths/CTimeSeriesSegmentation.cc
```diff
@@ -665,21 +673,20 @@ void CTimeSeriesSegmentation::fitPiecewiseLinearScaledSeasonal(
 
     auto computeScales = [&] {
         for (std::size_t j = 1; j < segmentation.size(); ++j) {
-            TMeanAccumulator projection;
-            TMeanAccumulator Z;
+            TMeanAccumulator xp;
+            TMeanAccumulator pp;
```
valeriy42, line 677: Do the new variable names `xp` and `pp` refer to some formula? For me personally, it was easier to understand what `projection` and `Z` meant.

lib/maths/CTimeSeriesTestForSeasonality.cc
```diff
@@ -374,6 +376,7 @@ CSeasonalDecomposition CTimeSeriesTestForSeasonality::decompose() const {
                     trend.add(static_cast<double>(j), CBasicStatistics::mean(values[j]),
                               CBasicStatistics::count(values[j]));
                 }
+                trend.parameters(parameters);
```
valeriy42, line 379: I am probably mistaken, but it seems to me that `parameters` is getting set but not read from here on. Why do we need to store it if we don't use the value?

lib/maths/unittest/CTimeSeriesModelTest.cc
```diff
@@ -30,6 +30,7 @@
 #include <test/CTimeSeriesTestData.h>
 
 #include "TestUtils.h"
+#include "maths/MathsTypes.h"
```
valeriy42, line 33:
```suggestion
#include <TestUtils.h>
#include <maths/MathsTypes.h>
```

valeriy42 (commented):
Looks good altogether :+1: I am wondering if as a part of this change in anomaly detection, we could move the relevant files (e.g. `CTimeSeries*`) into a subdirectory to have more structure?

lib/maths/unittest/CTimeSeriesTestForChangeTest.cc
```diff
@@ -0,0 +1,242 @@
...
+
+    test::CRandomNumbers rng;
+
+    core_t::TTime startTime{100000};
+
+    double truePositives{0.0};
+    TMeanAccumulator meanError;
+    TMeanAccumulator meanTimeError;
+
+    TDoubleVec samples;
```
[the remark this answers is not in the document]
tveasey: The random number generator calls the reserve for us when we actually fill in the values.

include/maths/CTimeSeriesDecompositionDetail.h
```diff
@@ -85,10 +92,13 @@ class MATHS_EXPORT CTimeSeriesDecompositionDetail : private CTimeSeriesDecomposi
         double s_Seasonal;
         //! The calendar component prediction at the value's time.
         double s_Calendar;
-        //! The window values predictor.
-        TPredictor s_Predictor;
-        //! The window values preconditioner to use for seasonality testing.
-        TFilteredPredictor s_SeasonalityTestPreconditioner;
+        //! The time series decomposition.
+        CTimeSeriesDecomposition* s_Decomposition;
```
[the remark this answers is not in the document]
tveasey: This is a temporary object and there is no ownership involved, it just needs a reference back to the decomposition. The constructor takes by reference but it is stored as a pointer so this type is copyable.

include/maths/CTimeSeriesDecompositionDetail.h
```diff
@@ -181,8 +204,111 @@ class MATHS_EXPORT CTimeSeriesDecompositionDetail : private CTimeSeriesDecomposi
         THandlerRefVec m_Handlers;
     };
 
-    //! \brief Scans through increasingly low frequencies looking for custom
-    //! diurnal and any other large amplitude seasonal components.
+    //! \brief Checks for sudden change or change events.
+    class MATHS_EXPORT CChangePointTest : public CHandler {
```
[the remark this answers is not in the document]
tveasey: For definite this needs restructuring. I want to create subdirectories in `maths` as you suggest. Then this can all move into a details subdirectory, for time series modelling, and be split over several files. However, I don't want to make this change in this PR, which is already large. It also needs us to do some work to get the build system to accept the new directory arrangement. So whilst I agree with this comment, I would propose making the change in a separate PR. WDYT?

lib/maths/CTimeSeriesSegmentation.cc
tveasey: It was just to make these consistent with other code which does a similar thing, i.e. I went for consistency. I'm happy to instead standardise on these names.

lib/maths/CTimeSeriesTestForSeasonality.cc
tveasey: This is correct, but the parameters are referenced by the predictor. This is actually a significant performance optimisation since the runtime is dominated by computing these parameters, I'll add a comment to explain what is going on here.

lib/maths/unittest/CTimeSeriesTestForChangeTest.cc
```diff
@@ -0,0 +1,242 @@
...
+
+#include <core/Constants.h>
+
+#include <maths/CBasicStatistics.h>
+#include <maths/CTimeSeriesTestForChange.h>
+
+#include <test/CRandomNumbers.h>
+
+#include "TestUtils.h"
+#include "core/CContainerPrinter.h"
```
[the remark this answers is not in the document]
tveasey: Local files to test directory are normally included like this, but I'll correct `CContainerPrinter`.

lib/maths/CTimeSeriesDecompositionDetail.cc
tveasey: Done in [3617328](https://github.com/elastic/ml-cpp/pull/1550/commits/36173280d512decd1340e2e2e3b7d25771efeabf).

tveasey:
>  we could move the relevant files (e.g. CTimeSeries*) into a subdirectory to have more structure?

Definitely, however I would prefer to make this in a separate PR.

lib/maths/unittest/CTimeSeriesTestForChangeTest.cc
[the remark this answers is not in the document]
tveasey: See [079a076](https://github.com/elastic/ml-cpp/pull/1550/commits/079a07614ed72f3b9562f478b634eaedbcd68f77).

lib/maths/unittest/CTimeSeriesTestForChangeTest.cc
```diff
@@ -0,0 +1,242 @@
...
+    test::CRandomNumbers rng;
+
+    core_t::TTime startTime{100000};
+
+    double truePositives{0.0};
+    TMeanAccumulator meanError;
+    TMeanAccumulator meanTimeError;
+
+    TDoubleVec samples;
+    for (std::size_t test = 0; test < 100; ++test) {
```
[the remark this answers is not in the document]
tveasey: Done in [079a076](https://github.com/elastic/ml-cpp/pull/1550/commits/079a07614ed72f3b9562f478b634eaedbcd68f77).

lib/maths/CTimeSeriesSegmentation.cc
tveasey: Reverted in [079a076](https://github.com/elastic/ml-cpp/pull/1550/commits/079a07614ed72f3b9562f478b634eaedbcd68f77).

lib/maths/CTimeSeriesTestForSeasonality.cc
tveasey: See comment in [079a076](https://github.com/elastic/ml-cpp/pull/1550/commits/079a07614ed72f3b9562f478b634eaedbcd68f77).

lib/maths/unittest/CTimeSeriesModelTest.cc
tveasey: See [079a076](https://github.com/elastic/ml-cpp/pull/1550/commits/079a07614ed72f3b9562f478b634eaedbcd68f77).

tveasey:
Testing has shown up a couple of refinements I need to make. I'm going to prepare one more change which addresses these issues. I'll let you know as soon as its ready for review again.

include/maths/CTimeSeriesDecompositionDetail.h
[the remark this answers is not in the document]
valeriy42: +1 Let's have it in a separate PR

valeriy42 (approved):
LGTM

tveasey:
This passes CI on windows and linux and we'd like to kick off another test over so I'm going to go ahead and merge.

PR #1551, opened 2020-11-02
[ML] Trap edge case causing model bounds to blow up to max double

tveasey:
If a mode has too few samples to model reliably we short circuit various calculations, such as the model plot bounds, which return (x, max double), with x depending on the distribution support. This happens very rarely, but is triggered by an intentional difference in the value weights used for clustering and distribution modelling. This is exacerbated by #1550 and the issue showed up testing this change.

In such cases it is completely fine to just remove the cluster. We'll reform the cluster if we get more data for it. This is the change this PR makes. I also took the opportunity to make some changes to bring the code more inline with our guidelines and to remove dead functionality related to forecasting in `CMultivariateMultimodalPrior`.

This most noticeably affects model plot, which would blow up, for example:

Screenshot 2020-11-02 at 20 36 14

but it could have a small impact on anomaly detection as well.


include/maths/CMultivariateMultimodalPrior.h
```diff
@@ -386,14 +386,21 @@ class CMultivariateMultimodalPrior : public CMultivariatePrior {
         // where w(i) is its weight we can achieve this by multiplying
         // all weights by some factor f in the range [0, 1].
 
-        if (!this->isForForecasting()) {
```
droberts195, line 389: There is nothing in the PR description about removing the "is for forecasting" functionality.  I think it would be good to mention it in the description and explain why it was done because one day somebody might want to search for when and why this functionality was removed.

include/maths/CMultivariateMultimodalPrior.h
```diff
@@ -386,14 +386,21 @@ class CMultivariateMultimodalPrior : public CMultivariatePrior {
...
 
+        // Remove any mode which is non-informative.
+        while (m_Modes.size() > 1) {
+            auto i = std::find_if(m_Modes.begin(), m_Modes.end(), [](const auto& mode) {
+                return mode.s_Prior->isNonInformative();
+            });
+            if (i == m_Modes.end() || m_Clusterer->remove(i->s_Index) == false) {
+                break;
+            }
+        }
```
droberts195, line 402: I think it would be worth adding a comment that `m_Modes` gets modified by the merge callback function that `m_Clusterer->remove()` calls.  Without this knowledge it seems that the `find_if` will always find the same mode.

lib/maths/CMultimodalPrior.cc
```diff
@@ -444,6 +444,16 @@ void CMultimodalPrior::propagateForwardsByTime(double time) {
...
 
+    // Remove any mode which is non-informative.
+    while (m_Modes.size() > 1) {
+        auto i = std::find_if(m_Modes.begin(), m_Modes.end(), [](const auto& mode) {
+            return mode.s_Prior->isNonInformative();
+        });
+        if (i == m_Modes.end() || m_Clusterer->remove(i->s_Index) == false) {
+            break;
+        }
+    }
```
droberts195, line 455: As above, I think it would be worth adding a comment that `m_Modes` gets modified by the merge callback function that `m_Clusterer->remove()` calls.  Without this knowledge it seems that the `find_if` will always find the same mode.

lib/model/CModelDetailsView.cc
```diff
@@ -16,6 +16,7 @@
 #include <model/CEventRatePopulationModel.h>
 #include <model/CMetricModel.h>
 #include <model/CMetricPopulationModel.h>
+#include <model/CSearchKey.h>
```
droberts195, line 19: Did this need to be added?

include/maths/CMultivariateMultimodalPrior.h
tveasey: This is dead functionality. The code to set this flag to true is now never called, so this condition is always true.  I spotted it because it is different to the univariate case: it was removed from there but missed from here. I'll add a comment in PR description.

include/maths/CMultivariateMultimodalPrior.h
tveasey: Good suggestion! I added a comment in [3cc7624](https://github.com/elastic/ml-cpp/pull/1551/commits/3cc762409bd1cc4df40328ea01765dc617bfd14c).

lib/maths/CMultimodalPrior.cc
tveasey: As above. See [3cc7624](https://github.com/elastic/ml-cpp/pull/1551/commits/3cc762409bd1cc4df40328ea01765dc617bfd14c).

lib/model/CModelDetailsView.cc
tveasey: No this was accidental while testing with some debug. Removed in [a1742db](https://github.com/elastic/ml-cpp/pull/1551/commits/a1742db52970e90876639780cf33f8d7a14c7d80).

droberts195 (approved):
LGTM

PR #1552, opened 2020-11-03
[ML] Parse anomaly job JSON config file

edsavage:
Add a parser for the recently added anomaly job configuration file, which
is in JSON format.

As an initial step to the configuration file replacing a number of
command line arguments, modify autodetect to use the jobId extracted by
the parser.

Relates to #1253


lib/core/unittest/CStringUtilsTest.cc
```diff
@@ -969,4 +969,263 @@ BOOST_AUTO_TEST_CASE(testRoundtripMaxDouble) {
...
+    bool parsedOk{false};
+    std::size_t memorySizeMB{0};
+    const std::size_t defaultValue{1};
+
+    {
+        // All valid and specifying a whole number of MB
+        std::tie(memorySizeMB, parsedOk) =
+            ml::core::CStringUtils::memorySizeStringToMB("4194304b", defaultValue);
+        BOOST_TEST_REQUIRE(parsedOk);
+        BOOST_TEST_REQUIRE(4, memorySizeMB);
```
droberts195, line 982: All the `BOOST_TEST_REQUIRE` macros in this file that have two arguments need to be changed.  Either the comma can be changed to `==`, e.g. `BOOST_TEST_REQUIRE(4 == memorySizeMB)`, or the macro can be changed to `BOOST_REQUIRE_EQUAL`, e.g. `BOOST_REQUIRE_EQUAL(4, memorySizeMB)`.  (I think as it stands the comma is acting as a comma operator, so all these assertions are checking the second argument is not zero, null, false, etc.)

droberts195 (commented):
The overall structure of this is fine.

I left quite a few minor comments, but there's nothing fundamentally wrong with this, just small niggles.

lib/core/unittest/CTimeUtilsTest.cc
```diff
@@ -509,4 +509,155 @@ BOOST_AUTO_TEST_CASE(testDurationToString) {
...
+    bool parsedOk{false};
+    ml::core_t::TTime durationSeconds{0};
+    const ml::core_t::TTime defaultValue{1};
+
+    {
+        // All valid and specifying a whole number of seconds
+        std::tie(durationSeconds, parsedOk) =
+            ml::core::CTimeUtils::timeDurationStringToSeconds("14d", defaultValue);
+        BOOST_TEST_REQUIRE(parsedOk);
+        BOOST_TEST_REQUIRE(1209600, durationSeconds);
```
droberts195, line 522: There are still places where `BOOST_TEST_REQUIRE` is being used with the comma operator.

This one needs to be either `BOOST_REQUIRE_EQUAL(1209600, durationSeconds)` or `BOOST_TEST_REQUIRE(1209600 == durationSeconds)`.

But there are others too - please scan all the assertions in the PR to check they are not inadvertently using the comma operator.

droberts195 (approved):
LGTM

PR #1555, opened 2020-11-04
[ML] Fix error logging for foreground persistence

edsavage:
Ensure an error is logged when persistence fails.


droberts195:
It's reasonable for this to be `>non-issue`, because although the code it affects is in 7.10 it is never triggered by the Java side in 7.10.  So no end user will observe the problem in 7.10 and hence it's not necessary to mention in the release notes.

benwtrent (approved):

Issue #1554, opened 2020-11-04
[ML] Pass minimum required disk space for forecasting as a forecast parameter


droberts195:
Currently there is a hardcoded requirement to have 4GB of disk space available for forecasting to run: https://github.com/elastic/ml-cpp/blob/436567b555cf4857606431d823a31fea67b9f1cd/lib/api/CForecastRunner.cc#L498-L502

On the Java side there is a setting to control the disk space requirement, `xpack.ml.min_disk_space_off_heap`.  This should be passed through to the C++ as an optional forecast request field instead of the C++ using a hardcoded value.

An example of how a new field can be added to forecast requests is in #1238 and elastic/elasticsearch#57254.

Then, for consistency with how things worked with the original defaults, the C++ should enforce disk space of 80% of what the Java enforces.  (The reasoning behind this was that some disk space may have been used in between the Java and C++ checks, so the Java code was more conservative than the C++ code.)

Relates to https://github.com/elastic/elasticsearch/issues/58806#issuecomment-659393024

PR #1556, opened 2020-11-04
[ML] Adding a minimum disk space forecast parameter

droberts195:
This new parameter will allow the value of the
xpack.ml.min_disk_space_off_heap Elasticsearch setting
to be propagated to the autodetect process.  It will
prevent problems where xpack.ml.min_disk_space_off_heap
has been set to a low value but the previous hardcoded
C++ check still prevented forecasting from working.

Relates #1554


droberts195:
This is a `>non-issue` because `xpack.ml.min_disk_space_off_heap` is intentionally undocumented.

edsavage (approved):
LGTM

PR #1561, opened 2020-11-09
[ML] Fixing move constructor/assignment for CForecastRunner::SForecast

droberts195:
The hand-crafted move constructor and move assignment operator
for CForecastRunner::SForecast were unnecessarily copying some
members instead of moving them.  The defaults will do the
correct thing and also avoid the need for future maintenance
when adding extra members.  (The hand-crafted versions were
created when we were using an old compiler that had a bug
related to default move assignment.)

### DIFF
--- a/include/api/CForecastRunner.h
+++ b/include/api/CForecastRunner.h
@@ -174,10 +174,10 @@ public:
 
 private:
     struct API_EXPORT SForecast {
-        SForecast() {}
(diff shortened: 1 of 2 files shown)


hendrikmuhs (approved):
LGTM

edsavage (approved):
LGTM

PR #1569, opened 2020-11-12
[ML] Remove references to the .ml-state index from C++ code

droberts195:
The ML C++ code contains several references to the .ml-state
index.  These date back to the days when the C++ code decided
which index to store state in.  Now this decision is made in
the Java code, and the index used isn't even .ml-state any
more, so it was highly confusing to have this name embedded
in the C++ code.  This PR also changes a few interfaces to
make clearer that the C++ isn't choosing the index to store
state in.


dimitris-athanasiou (approved):
LGTM

PR #1570, opened 2020-11-12
[ML] Migrate away from inheriting from std::iterator

tveasey:
C++17 deprecates inheriting from `std::iterator` for iterator trait typedefs. Our use of this generates a lot of deprecation warnings from our VS compiler. This change migrates the few instances we have to explicit typedefs.


droberts195 (approved):
LGTM

PR #1571, opened 2020-11-12
[ML] A few minor fixes and warning removals

droberts195:
None of these changes are critical to production use,
but they avoid compiler warnings and fix trace output.


tveasey (approved):
LGTM

PR #1575, opened 2020-11-12
[7.x][ML] Include peak memory usage in model snapshots (#1572)

edsavage:
Ensure that the static counter tracking the peak memory usage for a job
is included in the limited set to be persisted/restored from model state
snapshots.

Relates to elastic/elasticsearch#64154
Backports #1572


edsavage:
Merging now without waiting for aarch64 CI build as there is currently a major backlog and this change is minor

PR #1578, opened 2020-11-12
[ML] Work around erroneous reference requirement for std::vector::assign

tveasey:
It seems that some versions of libc++ require that an iterator has a non void reference type to be used with `std::vector::assign`. This is an annoying unnecessary extra type requirement, since the code doesn't actually need operator* to return a reference, but never mind. Defining them - as we had implicitly before via `std::iterator` - should be harmless. Follows on from #1570.


droberts195 (approved):
LGTM

PR #1579, opened 2020-11-16
[ML] Fixing the debug build

droberts195:
When building without optimisation RapidJson's templates
cause problems on Windows when used in different libraries.
This change works around the problem by telling the api
library to use the definition from the core library.  (We
don't notice this when optimisation is used because the
problematic classes end up being inlined away.)


tveasey (approved):
LGTM

PR #1580, opened 2020-11-16
[ML] Fix test fallout from new change point detection

tveasey:
This reenables all disabled unit tests and makes the following improvements to do so:
1. Adds the bucket residual variance when performing the significance test to decide whether to select a more complex model for seasonality. This also changes the definition of the bucket residual variance to be the truncated variance of value - modelled seasonality. This is important for long buckets and seasonal patterns with sudden changes because otherwise the test is too pessimistic. (This change addresses a separate issue I've observed where if a component is detected for a long bucket length we'd reinitialise the residual model with too much variance.)
2. We account better for the total number of repeats we observe in the window for interference pattern of all seasonal components when selecting selecting the seasonal model.
3. We correctly compute degrees of freedom when one seasonal component nests in another (and so can explain no additional variance).
4. Forbid negative scales for both seasonality and change detection (they are a route to false positives) and improve the autocorrelation calculation when the seasonality is scaled.
5. We only test for time shifts and scaling change points if we have seasonality.
6. I've tightened up the update of the probabilistic model for step changes.
7. Improves the accuracy of predictors we use for testing for change.

1 and 2 showed up as cases where we were wrongly using interference between seasonal components to fit a change, such as a time translation of the seasonal pattern. These changes require that we have significantly more evidence to accept the model than we would otherwise. 3 was just an oversight and avoided a regression from 1 and 2. 4 helps avoid false positives where the trend is changing. 5 showed up in poor forecasts for staircase and saw tooth functions.

The remaining test failures were due to mismatch between test code and the new model behaviour, for example causing mismatch in model checksums, or needed small parameter tweaks.

Marking this change as a non issue since it will not be documented separately.


lib/maths/CTimeSeriesTestForChange.cc
```diff
@@ -487,20 +479,59 @@ CTimeSeriesTestForChange::timeShift(double varianceH0,
...
+    // The following code is reverse engineering the average of the predictions
+    // which fall in each bucket of the time window. We have samples at some
+    // approximately fixed interval l and a bucket length L >= l. We also know
+    //
+    //   "values start - buckets start" = l / L sum_{0<=i<n}{ t0 + i l }    (1)
+    //
+    // where n = L / l. This summation is n t0 + n(n - 1)/2 l and we can use this
+    // to solve (1) for t0. The expected prediction is then
+    //
+    //   1 / n sum_{0<=i<n}{ p(t + t0 + i * l) }
```
valeriy42, line 502: good explanation :+1:

valeriy42 (approved):
LGTM

PR #1583, opened 2020-11-17
[ML] Make sure we don't try and test significance when there is insufficient data

tveasey:
We aren't guaranteed that H0 always has more degrees of freedom H1 in all cases where we were testing explained variance. This could therefore sometimes trigger log errors. In the cases this happens it is appropriate to treat the p-value of the alternative as 1, i.e. unproven.

Showed up running #1550 through the QA. Marking as a non issue since it fixes an issue in unreleased code.


przemekwitek (approved):
LGTM

PR #1584, opened 2020-11-17
[ML] Correct the test to select the trend model to use to forecast

tveasey:
Whilst reviewing possible causes of the log errors in I addressed in #1583, I noticed that the test statistic for selecting the trend model complexity for forecasting isn't right: it's not accounting for degrees of freedom properly. This effect will be small since number values from which the MSE is computed is expected to be much larger than number of parameters.

I don't think this is worth a release note so marking as a non-issue. Also note that it can only affect forecasting.


tveasey:
This is interacting with state upgrade, which needs investigating before this is merged. I'll marking as work in progress.

tveasey:
retest

valeriy42 (approved):
LGTM

PR #1585, opened 2020-11-17
[ML] Recalculate memory usage before upgrading model state

droberts195:
When a persist control message with arguments is received
by the anomaly detector it doesn't go through the standard
chain of persistence calls, as it unconditionally rewrites
the state (even if no data has been seen) and includes only
the anomaly detector state rather than the categorizer state
too.  Because of this the memory usage was not being
recalculated prior to persisting the state as would normally
happen.  This PR rectifies that omission.

Fixes one of the problems detailed in
https://github.com/elastic/elasticsearch/pull/64665#pullrequestreview-524548875

### DIFF
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -102,6 +102,9 @@ public:
     //! Recalculate the memory usage regardless of whether there is a memory limit
     void forceRefresh(CMonitoredResource& resource);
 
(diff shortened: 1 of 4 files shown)


droberts195:
I've labelled this `>non-issue` because the fixed functionality is logically part of elastic/elasticsearch#64665, and that's already going to be release noted for 7.11.

tveasey (approved):
LGTM

PR #1586, opened 2020-11-17
[ML] Trap potential cause of log errors from CXMeansOnline1d

tveasey:
We are getting log errors like
```
[CXMeansOnline1d.cc@462] Bad category = (0.0017123, 2.27255e+10, 0): Error in function boost::math::normal_distribution<double>::normal_distribution...
```
from a long running QA job.

Whilst this has proven difficult to reproduce the underlying cause seems to be that numerical errors have resulted in a variance being around `-eps`. Since we take the `sqrt` of this in the offending code it generates a NaN which triggers this and other log errors. The code which does this should be defensive.

The only mechanism I can see causing this is if the weight used for update is somehow negative. This also shouldn't be possible and it is a large task to trace back how this might be occurring. I simply now also trap this case and discard the sample.

This can affect results, but it is an edge case and we've only seen these errors in one job, so I would assume that nearly all the time this change will have no impact.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -54,6 +54,10 @@
 * Estimate upper bound of potential gains before splitting a decision tree node to avoid 
   unnecessary computation. (See {ml-pull}1537[#1537].)
 
+=== Bug Fixes
+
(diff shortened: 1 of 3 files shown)


droberts195 (approved):
LGTM

PR #1589, opened 2020-11-20
[ML] Addressing a QA regression due to 1451 and 1580 

tveasey:
Some more refinements following on from #1451 and #1580 aimed at fixing a specific QA regression.

This does the following:
1. Rolls out the refinement to the bucket predictor (for already modelled components) from change detection to the test for seasonal components. The previous approach introduced spurious errors for very long bucket lengths (which we use when testing for long, i.e. monthly, yearly, etc, seasonal components). These could cause us to detect spurious seasonality at multiples of the modelled components.
2. Precondition before testing for seasonal components by explicitly removing any modelled seasonality which is too short to test given the window bucket length. The intent is similar to 1 and works around any error in the seasonal model.
3. Don't adjust the gradient when applying scaling events. The window we use to test for change is generally not long enough to reliably estimate changes in gradient and changing it can cause instability and transient anomalies after a change is applied, particularly if the scale factor is large.

Note I've also tried to standardise terminology across different classes where there were discrepancies. The code isn't released so marking as a non-issue.

### DIFF
--- a/include/maths/CExpandingWindow.h
+++ b/include/maths/CExpandingWindow.h
@@ -59,7 +59,7 @@ public:
     using TPredictor = std::function<double(core_t::TTime)>;
 
 public:
-    CExpandingWindow(core_t::TTime bucketLength,
+    CExpandingWindow(core_t::TTime sampleInterval,
                      TTimeCRng bucketLengths,
                      std::size_t size,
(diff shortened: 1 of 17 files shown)


valeriy42 (approved):
LGTM

PR #1591, opened 2020-11-23
[ML] Avoid misleading error message

droberts195:
When the logging named pipe is not connected because process
startup gets cancelled, the named pipe classes intentionally
do not log errors, as this would be confusing for people
reading the logs.  However, the logger class was still logging
an error message that implied a problem with the file system
rather than cancellation of startup.

This change stops that error message from being logged, using
the same mechanism that is used in the named pipe classes.

Relates elastic/elasticsearch#64578


dimitris-athanasiou (approved):
LGTM

PR #1595, opened 2020-11-30
[ML] Avoid potential integer overflow

tveasey:
Reworks the calculation of bucket average sample times to avoid potential integer overflow.

### DIFF
--- a/lib/maths/CExpandingWindow.cc
+++ b/lib/maths/CExpandingWindow.cc
@@ -95,10 +95,11 @@ core_t::TTime CExpandingWindow::sampleAverageOffset() const {
 }
 
 core_t::TTime CExpandingWindow::beginValuesTime() const {
-    return this->sampleAverageOffset() +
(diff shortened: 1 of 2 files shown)


droberts195 (approved):
LGTM

Realistically there was only a risk of overflow for 32 bit programs though right?  For 64 bit the sample times would need to be a billion years in the future for an overflow to occur.

I am sure that since we stopped building 32 bit for any platform we'll have introduced other places that would overflow.  When we built 32 bit versions we used to do it occasionally and catch it in the tests.  So since we stopped building/testing 32 bit I am sure we will have done it somewhere else.  I am not sure it really matters though now that every computer we'd possibly want to run on is 64 bit - even phones nowadays.

tveasey:
> Realistically there was only a risk of overflow for 32 bit programs though right?

Yes it is true I agree we shouldn't run into this in practice. My thoughts were since `core_t::TTime` is a typedef and it was in new code and it was no more complicated to never unnecessarily overflow it felt like it was worthwhile.

PR #1596, opened 2020-11-30
[ML] Handle edge case where missing values mean we can't test the existing seasonality

tveasey:
We were running into an issue where [this](https://github.com/elastic/ml-cpp/compare/long-bucket-modelling-improvements...tveasey:untestable-edge-case?expand=1#diff-7b9b36a6bec1003216b0e96287772112bc8515fd2968580f7ceb55a4dfc357c2L986) check means that we can't compute test statistics for one of the components of the seasonality we're already modelling. In this case we choose not to change seasonality.


droberts195 (approved):
LGTM

PR #1597, opened 2020-11-30
[ML] Initialise CLimits from JSON config

edsavage:
Initialise the max categorisation examples and model memory limit
settings in `CLimits` with values extracted from the JSON anomaly job
config file.

Relates to #1253

### DIFF
--- a/lib/model/unittest/CLimitsTest.cc
+++ b/lib/model/unittest/CLimitsTest.cc
@@ -24,18 +24,38 @@ BOOST_AUTO_TEST_CASE(testTrivial) {
 }
 
 BOOST_AUTO_TEST_CASE(testValid) {
-    ml::model::CLimits config;
-    BOOST_TEST_REQUIRE(config.init("testfiles/mllimits.conf"));
+    {
+        ml::model::CLimits config;
+        BOOST_TEST_REQUIRE(config.init("testfiles/mllimits.conf"));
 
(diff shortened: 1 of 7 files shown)


lib/model/unittest/CLimitsTest.cc
```diff
@@ -24,18 +24,38 @@ BOOST_AUTO_TEST_CASE(testTrivial) {
...
 
-    BOOST_REQUIRE_EQUAL(0.005, config.unusualProbabilityThreshold());
+        // This also should be defaulted (as a percentage)
+        BOOST_REQUIRE_EQUAL(ml::model::CLimits::DEFAULT_RESULTS_UNUSUAL_PROBABILITY_THRESHOLD / 100,
+                            config.unusualProbabilityThreshold());
 
-    BOOST_REQUIRE_EQUAL(size_t(4567), config.memoryLimitMB());
+        // These two should be as specified in the constructor.
+        BOOST_REQUIRE_EQUAL(size_t(2), config.maxExamples());
+        BOOST_REQUIRE_EQUAL(size_t(4096), config.memoryLimitMB());
```
droberts195, line 57: nit: there's no need to cast the expected values with Boost.Test.

There are many examples in old code of us casting expected values, because CppUnit required identical types as well as values.  But Boost.Test doesn't, so the casts can be omitted.

droberts195 (approved):
LGTM

PR #1600, opened 2020-12-01
[ML] Address spurious anomalies after reinitialising time series decomposition

tveasey:
QA regressions associated #1451 and #1580 have shown we've increased false positives after changing the seasonal components included in the decomposition. This change injects a small amount of noise into the window values we test for seasonal components. This has two advantages:
1. If the signal if very stable we don't want to model seasonality if it's relative amplitude is too small
2. It avoids underestimating the variance when reinitialising the residual model which is the cause of the false positives


droberts195 (approved):
LGTM, but it looks like 3 more test assertions need tweaking as a result

PR #1602, opened 2020-12-01
[ML] Init AD model config from JSON config

edsavage:
Initialise CAnomalyDetectorModelConfig with values from
the JSON config file parsed by CAnomalyJobConfig.

Relates to #1253

### DIFF
--- a/bin/autodetect/Main.cc
+++ b/bin/autodetect/Main.cc
@@ -297,6 +297,7 @@ int main(int argc, char** argv) {
     // The anomaly job knows how to detect anomalies
     ml::api::CAnomalyJob job{jobId,
                              limits,
+                             jobConfig,
                              fieldConfig,
                              modelConfig,
                              wrappedOutputStream,
--- a/bin/autodetect/Main.cc
+++ b/bin/autodetect/Main.cc
@@ -91,9 +91,6 @@ int main(int argc, char** argv) {
     std::string modelPlotConfigFile;
     std::string logProperties;
     std::string logPipe;
-    ml::core_t::TTime bucketSpan{0};
-    ml::core_t::TTime latency{0};
-    std::string summaryCountFieldName;
     char delimiter{'\t'};
     bool lengthEncodedInput{false};
     std::string timeField{ml::api::CAnomalyJob::DEFAULT_TIME_FIELD_NAME};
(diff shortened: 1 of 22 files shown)


bin/autodetect/Main.cc
```diff
@@ -212,12 +207,19 @@ int main(int argc, char** argv) {
...
+    ml::core_t::TTime latency = analysisConfig.latency();
+    bool multivariateByFields = analysisConfig.multivariateByFields();
+
     ml::model_t::ESummaryMode summaryMode{
         summaryCountFieldName.empty() ? ml::model_t::E_None : ml::model_t::E_Manual};
     ml::model::CAnomalyDetectorModelConfig modelConfig{ml::model::CAnomalyDetectorModelConfig::defaultConfig(
         bucketSpan, summaryMode, summaryCountFieldName, latency, multivariateByFields)};
     modelConfig.detectionRules(ml::model::CAnomalyDetectorModelConfig::TIntDetectionRuleVecUMapCRef(
-        fieldConfig.detectionRules()));
+        analysisConfig.detectionRules()));
```
droberts195, line 222: In the long term this could be encapsulated in a method of `CAnomalyJobConfig`, say `CAnomalyJobConfig::makeModelConfig()`.  But since you're doing a series of PRs you can save this for the next one.

include/api/CAnomalyJob.h
```diff
@@ -424,6 +426,10 @@ class API_EXPORT CAnomalyJob : public CDataProcessor {
     //! Object to which the output is passed
     CJsonOutputWriter m_JsonOutputWriter;
 
+    //! Configuration settings for the analysis parsed from
+    //! JSON configuration file.
```
droberts195, line 430: It would be good to say why this is non-const.  It's because it gets updated by job updates right?  And of course since it's a reference that will also affect everywhere else that has a reference.

include/api/CAnomalyJobConfig.h
```diff
@@ -68,7 +67,7 @@ class API_EXPORT CAnomalyJobConfig {
             std::string m_PartitionFieldName{};
             std::string m_ExcludeFrequent{};
             std::string m_DetectorDescription{};
-            CDetectionRulesJsonParser::TDetectionRuleVec m_CustomRules{};
+            std::size_t m_DetectorIndex{};
```
droberts195, line 70: Although probably not the best choice originally, detector index is an `int` everywhere else, so it's best if it's an `int` here too.  (I think originally it was some generic "identifier" instead of an array index, and -1 indicated "unknown".  That's so far in the past that -1 never exists now, but it's still probably best to keep the types consistent across the codebase.)

include/api/CAnomalyJobConfig.h
```diff
@@ -234,7 +257,7 @@ class API_EXPORT CAnomalyJobConfig {
 
     std::string jobId() const { return m_JobId; }
     std::string jobType() const { return m_JobType; }
-    const CAnalysisConfig& analysisConfig() const { return m_AnalysisConfig; }
```
droberts195, line 237: You could provide _both_ const and non-const accessors.  In future as this class gets used more widely some places may hold a const reference to the `CAnomalyJobConfig`, and it's annoying for those places to have to cast away constness just to get a const reference to one of the sub-objects.

droberts195 (approved):
LGTM

I saw a few very minor things, but am happy to merge without further review if you could fix those.

PR #1603, opened 2020-12-02
[ML] Addressing a QA regression due to 1451 and 1580

tveasey:
There was a bug in `fitAndRemoveUntestableModelledComponents` which should check the component length against the minimum component length in *window buckets* not time units. For very long bucket lengths we also need more values in the sliding window to test for changes. This is unreleased code so a non-issue.


tveasey:
retest

tveasey:
The `integration.DetectionRulesIT.testCondition` failure is caused by a multi-bucket anomaly. It should be made more robust by only counting records with low probability. I'll make this change at the point I merge to master.

tveasey:
retest

droberts195 (approved):
LGTM

edsavage (approved):
LGTM

PR #1604, opened 2020-12-03
[ML] Avoid overfit when reinitialising the residual model with new seasonal components

tveasey:
We're running into issues where we estimate the residual distribution after removing seasonality using a larger more accurate model than we eventually choose. This means we can underestimate its variance and generate false transient positives. This change migrates to matching the model predictions to the model size we select so that our residual distribution is correctly initialised. The code isn't released so this is a non-issue.

### DIFF
--- a/include/maths/CSignal.h
+++ b/include/maths/CSignal.h
@@ -426,6 +426,7 @@ public:
     //! of \p component at granularity \p size. This is done by convolving with
     //! a function of width "component size" / "size" so \p component is not
     //! actually resized.
+    //! \warning \p size should be greater than zero.
     static TMeanAccumulatorVec smoothResample(std::size_t size, TMeanAccumulatorVec component);
 
     //! Restrict to the windows defined by \p period.
--- a/include/maths/CSignal.h
+++ b/include/maths/CSignal.h
@@ -418,6 +418,16 @@ public:
     static std::size_t selectComponentSize(const TFloatMeanAccumulatorVec& values,
                                            std::size_t period);
 
+    //! Apply a smooth resampling of \p component at \p size granularity.
+    //!
+    //! \param[in] size The granularity at which to resample.
+    //! \param[in] component The component to resample.
+    //! \return A resampling of \p component whose values are smoothed averages
+    //! of \p component at granularity \p size. This is done by convolving with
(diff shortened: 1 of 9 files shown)


tveasey:
The `integration.DetectionRulesIT.testCondition` failure is caused by a multi-bucket anomaly. It should be made more robust by only counting records with low probability. I'll make this change at the point I merge to master.

tveasey:
retest

lib/maths/CSignal.cc
```diff
@@ -910,6 +910,38 @@ std::size_t CSignal::selectComponentSize(const TFloatMeanAccumulatorVec& values,
     return size;
 }
 
+CSignal::TMeanAccumulatorVec
+CSignal::smoothResample(std::size_t size, TMeanAccumulatorVec component) {
+
+    if (size >= component.size()) {
```
droberts195, line 916: Is it worth adding `|| size == 0` to the condition here to defend against division by zero later on if this method is ever reused in the future in a place where `size` could be zero?

edsavage (approved):
LGTM

include/maths/CSignal.h
```diff
@@ -418,6 +418,16 @@ class MATHS_EXPORT CSignal {
     static std::size_t selectComponentSize(const TFloatMeanAccumulatorVec& values,
                                            std::size_t period);
 
+    //! Apply a smooth resampling of \p component at \p size granularity.
+    //!
+    //! \param[in] size The granularity at which to resample.
```
droberts195, line 423: If you don't think it's worth making the method more defensive then you could specifically note here that zero doesn't make sense and will cause a crash.

droberts195 (approved):
LGTM, although I saw a place where the code could be made more defensive

lib/maths/CSignal.cc
tveasey: I think it is probably worth both. We can use a best guess which is `size == 1`, but it isn't really clear what the user intention is so I think it is worth documenting that it shouldn't be called like this. I made these changes in [e45d65d](https://github.com/elastic/ml-cpp/pull/1604/commits/e45d65d0e3f08c31bfdc69d3bbebe8e486f14a12).

PR #1607, opened 2020-12-04
[ML] Init AD job config with scheduled events

edsavage:
* Scheduled events configuration exists outside of the job config.
As an interim step pass the scheduled events parsed from
the old-style field config to the new-style JSON parser.

* Ensure that updates to scheduled event rules are correctly handled.

* Tidy the code to create the anomaly detector model config.

Relates to #1253

### DIFF
--- a/lib/api/CAnomalyJobConfig.cc
+++ b/lib/api/CAnomalyJobConfig.cc
@@ -12,6 +12,7 @@
 
 #include <api/CAnomalyJobConfigReader.h>
 
+#include <model/CAnomalyDetectorModelConfig.h>
 #include <model/CLimits.h>
 
 #include <rapidjson/document.h>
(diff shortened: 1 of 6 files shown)


lib/api/CAnomalyJobConfig.cc
```diff
@@ -347,17 +366,108 @@ bool CAnomalyJobConfig::CAnalysisConfig::processFilter(const std::string& key,
...
+        m_SummaryCountFieldName.empty() ? ml::model_t::E_None : ml::model_t::E_Manual};
+
+    ml::model::CAnomalyDetectorModelConfig modelConfig{ml::model::CAnomalyDetectorModelConfig::defaultConfig(
+        m_BucketSpan, summaryMode, m_SummaryCountFieldName, m_Latency, m_MultivariateByFields)};
+
+    modelConfig.detectionRules(
+        ml::model::CAnomalyDetectorModelConfig::TIntDetectionRuleVecUMapCRef(m_DetectorRules));
+
+    modelConfig.scheduledEvents(
+        ml::model::CAnomalyDetectorModelConfig::TStrDetectionRulePrVecCRef(m_ScheduledEvents));
```
droberts195, line 466: The `ml::` qualifiers aren't needed here, as the whole class is in the `ml` namespace.

droberts195 (commented):
LGTM

I saw one minor thing, but since you're doing a series of PRs you can change it in the next one since this has already passed CI.

droberts195 (approved):

PR #1610, opened 2020-12-04
[ML] Fix incorrect forecast bounds

### LINKED ISSUE #1606: Forecast bounds are too thin
**Found in version**
- 7.10.1-bc2

**Steps to reproduce**
- Run a forecast for an anomaly detection job.

**Expected result**
- Predicted value and forecast bounds are sensible.

**Actual result**
- The forecast bounds are super "thin":

farequote_forecast_7 10 0_vs_7 10 1
image

**Additional information**
- This is a regression, in 7.10.0 the forecast bounds are looking good.

droberts195:
Due to a mistake in #1559 the forecast bounds were
reduced from 95% to 0.95%.

This change reinstates the correct bounds, and also
moves the definition of the constant to the maths
library, where it can be used in a unit test that
will fail if it is ever inadvertently changed again.

Fixes #1606

### DIFF
--- a/include/api/CForecastRunner.h
+++ b/include/api/CForecastRunner.h
@@ -101,10 +101,6 @@ public:
     //! minimum time between stat updates to prevent to many updates in a short time
     static const std::uint64_t MINIMUM_TIME_ELAPSED_FOR_STATS_UPDATE = 3000ul; // 3s
 
-    //! default bounds percentile
-    //! (not defined inline because we need its address)
-    static const double DEFAULT_BOUNDS_PERCENTILE;
-
 private:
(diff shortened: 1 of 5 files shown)


droberts195:
I'm waiting to see what happens with 7.10.1 before deciding whether to label this `>bug` or `>non-issue`.

This problem was also fixed on the 7.10 branch by reverting commits, and it remains to be seen which release that will go into.

benwtrent (approved):

droberts195:
> I'm waiting to see what happens with 7.10.1 before deciding whether to label this `>bug` or `>non-issue`.

It's confirmed that 7.10.1 reverts the commits that introduced this bug, so it's never been in a released version and hence this PR is a `>non-issue`.

PR #1611, opened 2020-12-04
[ML] Fix invalid writes and increase robustness

tveasey:
This primarily fixes two potential edge cases leading to invalid writes:
1. We weren't checking that the time supplied to `shiftTime` for the expanding window was after the start of the window. This led to unsigned integer underflow and bad arguments supplied to `std::fill`. This could very occasionally happen.
2. We were applying the wrongly indexing the seasonal hypothesis vector if some of them weren't testable. This could lead to an invalid write.

As part of tracking down these memory corruptions, I also audited potentially dangerous places in the code. These were safe, but relied on rather delicate invariants or inputs being correct. I've tightened up the logic to ensure that these assumptions are now checked.

This is unreleased code so marking as a non-issue.

### DIFF
--- a/lib/maths/CTimeSeriesTestForSeasonality.cc
+++ b/lib/maths/CTimeSeriesTestForSeasonality.cc
@@ -314,8 +314,19 @@ void CTimeSeriesTestForSeasonality::fitAndRemoveUntestableModelledComponents() {
 }
 
 bool CTimeSeriesTestForSeasonality::checkInvariants() const {
-    return m_ModelledPeriods.size() == m_ModelledPeriodsSizes.size() ||
-           m_ModelledPeriodsSizes.size() == m_ModelledPeriodsTestable.size();
+    if (m_ModelledPeriods.size() != m_ModelledPeriodsSizes.size()) {
+        LOG_ERROR(<< "# modelled periods ("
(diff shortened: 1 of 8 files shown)


lib/maths/CTimeSeriesTestForSeasonality.cc
```diff
@@ -359,6 +364,11 @@ CSeasonalDecomposition CTimeSeriesTestForSeasonality::decompose() const {
     // is based on a number of criterion which are geared towards our modelling
     // techniques and are described in select.
 
+    if (this->checkInvariants() == false) {
+        LOG_ERROR(<< "Failed invariants. Not testing for seasonality.");
```
droberts195, line 368: Is it worth printing the 3 sizes involved here?  Or maybe `checkInvariants` could log them instead.

droberts195 (approved):
LGTM

Feel free to leave the suggestion for a subsequent PR if you want to get this merged sooner

lib/maths/CTimeSeriesTestForSeasonality.cc
tveasey: I think this is a good suggestion. I think it is cleanest to add the logging to `checkInvariants` see [c3bd243](https://github.com/elastic/ml-cpp/pull/1611/commits/c3bd24322f36d3183512759eb54d497eed9a9e95).

PR #1612, opened 2020-12-07
[ML] Fix hyperparameter scaling

valeriy42:
Since downsampling affects the scaling of some hyperparameter, sometimes the scaled hyperparamters were outside of the defined bounding boxes. This fix restores the boundary constraint. It allows us to better describe the error surface.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -57,6 +57,7 @@
 === Bug Fixes
 
 * Fix potential cause for log errors from CXMeansOnline1d. (See {ml-pull}1586[#1586].)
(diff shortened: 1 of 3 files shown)


tveasey (approved):
LGTM.

We discussed this a bit offline. I think it is more correct, but I'm not 100% sure it'll lead to better results. It doesn't actually matter to BO if there are points outside the bounding box which the GP is conditioned on. Constrained LBFGS will still choose a point inside the bounding box. However, we'll get a different characterisation of the loss surface because the GP parameters will be different. What we do in this (new) approach is remove the relationship we believe holds between parameter values *before* estimating the loss surface. This does seem the sensible/more correct thing to do, but it'll be interesting to see how it pans out.

valeriy42:
retest

PR #1615, opened 2020-12-08
[ML] Warn and error level log throttling

tveasey:
We have some C++ log message deduplication on the Java side, but it relies on the log messages repeating. If several different lines interleave their logging then all lines are logged. We've had situations where we get a lot of logging generated at ERROR level which can significantly slow down anomaly detection. They also spam the global Elasticsearch logs.

This proposes a simple mechanism to throttle logging at WARN and ERROR level on C++ side per process. In particular, it
1. Always outputs the first WARN or ERROR for a particular log line
2. Thereafter limits to outputting a duplicate line once per hour (wall clock)

This should be enough to allow us to check for presence of errors (useful for QA). Also, nearly always, only the first instance of an error is useful.

Since I only introduce this for WARN and ERROR logging I have not attempted to make the throttling mechanism particularly efficient: there is contention on a global hash map which caches the last log times. In normal operation this should have no overhead and if there is a spamming log line the overhead of writing all those messages should dwarf the cost. Finally, we don't have enough distinct log lines for the memory usage of this map to warrant tracking as part of memory estimation.

### DIFF
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -7,7 +7,6 @@
 #ifndef INCLUDED_ml_core_CLoggerThrottler_h
 #define INCLUDED_ml_core_CLoggerThrottler_h
 
-#include <core/CoreTypes.h>
 #include <core/ImportExport.h>
 
 #include <boost/unordered_map.hpp>
@@ -43,7 +42,9 @@ public:
     static CLoggerThrottler& instance();
 
     //! Set the minimum interval between repeated log messages.
-    void minimumLogInterval(std::int64_t minimumLogInterval);
+    //!
+    //! \note This expects the interval in milliseconds.
+    void minimumLogIntervalMs(std::int64_t minimumLogIntervalMs);
 
     //! Should we skip logging of \p line in \p file?
     //!
@@ -68,7 +69,7 @@ private:
 
 private:
     static CLoggerThrottler ms_Instance;
-    std::int64_t m_MinimumLogInterval;
+    std::int64_t m_MinimumLogIntervalMs;
     std::mutex m_Mutex;
     TConstCharPtrIntPrInt64SizePrUMap m_LastLogTimesAndCounts;
 };
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -62,9 +62,6 @@ private:
         boost::unordered_map<TConstCharPtrIntPr, TInt64SizePr>;
 
 private:
-    TInt64SizePr& lookup(const TConstCharPtrIntPr& key);
-
-private:
     std::int64_t m_MinimumLogIntervalMs;
     std::mutex m_Mutex;
     TConstCharPtrIntPrInt64SizePrUMap m_LastLogTimesAndCounts;
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -33,7 +33,7 @@ namespace core {
 //! little contention. Furthermore, the overhead of locking and unlocking the mutex
 //! should be neglible compared to the work done if the log line were actually
 //! emitted. So this should actually give a significant performance improvement
-//! if an log line is spamming.
+//! if a log line is spamming.
 class CORE_EXPORT CLoggerThrottler {
 public:
     CLoggerThrottler(const CLoggerThrottler&) = delete;
@@ -46,6 +46,14 @@ public:
     void minimumLogInterval(std::int64_t minimumLogInterval);
 
     //! Should we skip logging of \p line in \p file?
+    //!
+    //! \param[in] file The file containing the log line.
+    //! \param[in] line The log line.
+    //! \return A pair comprising the count of log lines since the log line was
+    //! last output (skip returned false) and whether to output the line or not.
+    //! \note This is expected to be used in conjuction with __FILE__ and __LINE__
+    //! macros although any unique identifiers which are a string literal and an
+    //! integer are permitted.
     std::pair<std::size_t, bool> skip(const char* file, int line);
 
 private:
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -53,11 +53,11 @@ public:
     //! \note This is expected to be used in conjuction with __FILE__ and __LINE__
     //! macros although any unique identifiers which are a string literal and an
     //! integer are permitted.
-    //! \warning We make no attempt to deduplicate errors for which __FILE__
-    //! resolves to a different string literal; this might conceivably happen
-    //! for log lines in a header. This is not expected to be important (it might
-    //! result in slightly more logging) and it is possible to work around this
-    //! by explicitly wrapping the logging, for example:
+    //! \warning We make no attempt to deduplicate logging from the same line for
+    //! which __FILE__ resolves to a different string literal; this might conceivably
+    //! happen for logging in a header. This is not expected to be important (it
+    //! might result in slightly more logging) and it is possible to work around
+    //! this by explicitly wrapping the logging, for example:
     //! \code
     //! if (core::CLogger::instance().throttler().skip("my unique string", 0) == false) {
     //!     LOG_ERROR(<< ...)
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -14,6 +14,7 @@
 
 #include <atomic>
 #include <cstdint>
+#include <mutex>
 
 namespace ml {
 namespace core {
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -18,7 +18,7 @@
 namespace ml {
 namespace core {
 
-//! \brief Implements per log line throttling. 
+//! \brief Implements per log line throttling.
 //!
 //! DESCRIPTION:\n
 //! This implements global throttling for a program per line log line. By default
@@ -45,10 +45,11 @@ public:
     void minimumLogInterval(std::int64_t minimumLogInterval);
 
     //! Should we skip logging of \p line in \p file?
-    bool skip(const char* file, int line);
+    std::pair<std::size_t, bool> skip(const char* file, int line);
 
 private:
-    using TCharIntInt64UMap = boost::unordered_map<std::pair<const char*, int>, std::int64_t>; 
+    using TCharIntPrInt64SizePrUMap =
+        boost::unordered_map<std::pair<const char*, int>, std::pair<std::int64_t, std::size_t>>;
 
 private:
     CLoggerThrottler();
@@ -57,9 +58,8 @@ private:
     static CLoggerThrottler ms_Instance;
     std::int64_t m_MinimumLogInterval;
     std::mutex m_Mutex;
-    TCharIntInt64UMap m_LastLogTimes;
+    TCharIntPrInt64SizePrUMap m_LastLogTimesAndCounts;
 };
-
 }
 }
 
--- a/include/core/CLoggerThrottler.h
+++ b/include/core/CLoggerThrottler.h
@@ -53,6 +53,16 @@ public:
     //! \note This is expected to be used in conjuction with __FILE__ and __LINE__
     //! macros although any unique identifiers which are a string literal and an
     //! integer are permitted.
+    //! \warning We make no attempt to deduplicate errors for which __FILE__
+    //! resolves to a different string literal; this might conceivably happen
+    //! for log lines in a header. This is not expected to be important (it might
+    //! result in slightly more logging) and it is possible to work around this
+    //! by explicitly wrapping the logging, for example:
+    //! \code
+    //! if (core::CLogger::instance().throttler().skip("my unique string", 0) == false) {
+    //!     LOG_ERROR(<< ...)
+    //! }
+    //! \endcode
     std::pair<std::size_t, bool> skip(const char* file, int line);
 
 private:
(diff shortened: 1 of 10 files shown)


lib/core/CLoggerThrottler.cc
```diff
@@ -0,0 +1,44 @@
...
+    if (now > lastTime->second + m_MinimumLogInterval) {
+        lastTime->second = now;
+        return false;
+    }
+    return true;
+}
+
+CLoggerThrottler::CLoggerThrottler() : m_MinimumLogInterval{3600 * 1000} {}
+
+CLoggerThrottler CLoggerThrottler::ms_Instance;
```
droberts195, line 46: Code that runs during static initialisation might contain log messages, so I think this will have to be constructed using a pattern that isn't vulnerable to the order of static initialisation, for example:

```
CLoggerThrottler& CLoggerThrottler::instance() {
    static CLoggerThrottler instance;
    return instance;
}
```

And then get rid of the `ms_Instance` variable.

The other way to do it would be to make this class not a singleton but a non-static member of the `CLogger` class, so that if the `CLogger` singleton exists then it's guaranteed that its corresponding `CLoggerThrottler` also does.  So the macros would call `ml::core::CLogger::instance().throttler()` similarly to how they currently call `ml::core::CLogger::instance().logger()` to get the low level logger object.

lib/core/CLoggerThrottler.cc
```diff
@@ -0,0 +1,44 @@
...
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#include <core/CLoggerThrottler.h>
+
+#include <core/CTimeUtils.h>
+
+#include <limits>
+#include <iostream>
```
droberts195, line 11: Not needed?

lib/core/unittest/CLoggerTest.cc
```diff
@@ -199,6 +204,53 @@ BOOST_FIXTURE_TEST_CASE(testNonAsciiJsonLogging, CTestFixture) {
...
+        if (line.empty()) {
+            continue;
+        }
+        rapidjson::Document doc;
+        doc.Parse<rapidjson::kParseDefaultFlags>(line);
+        BOOST_TEST_REQUIRE(!doc.HasParseError());
+        BOOST_TEST_REQUIRE(doc.HasMember("message"));
+        const rapidjson::Value& messageValue = doc["message"];
+        std::string messageString(messageValue.GetString(), messageValue.GetStringLength());
+        std::cout << messageString << std::endl;
```
droberts195, line 243: Should this be committed?

include/core/CLoggerThrottler.h
```diff
@@ -0,0 +1,66 @@
...
+    static CLoggerThrottler& instance();
+
+    //! Set the minimum interval between repeated log messages.
+    void minimumLogInterval(std::int64_t minimumLogInterval);
+
+    //! Should we skip logging of \p line in \p file?
+    bool skip(const char* file, int line);
+
+private:
+    using TCharIntInt64UMap = boost::unordered_map<std::pair<const char*, int>, std::int64_t>;
```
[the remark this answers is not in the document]
droberts195: This suggestion is outdated now, but still it's a `char*`, not a `char`.

include/core/LogMacros.h
```diff
@@ -73,25 +75,30 @@
...
 #ifdef LOG_FATAL
 #undef LOG_FATAL
 #endif
-#define LOG_FATAL(message)                                                                   \
-    BOOST_LOG_STREAM_SEV(ml::core::CLogger::instance().logger(), ml::core::CLogger::E_Fatal) \
-    LOG_LOCATION_INFO                                                                        \
-    message
-
+#define LOG_FATAL(message)                                                                       \
+    if (ml::core::CLoggerThrottler::instance().skip(__FILE__, __LINE__) == false) {              \
```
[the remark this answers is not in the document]
tveasey: I agree and decided the same thing: see [bfa62ba](https://github.com/elastic/ml-cpp/pull/1615/commits/bfa62ba1976acfa7742b590a6a389d76509e0fe6).

include/core/CLoggerThrottler.h
```diff
@@ -0,0 +1,66 @@
...
+
+private:
+    using TCharIntInt64UMap = boost::unordered_map<std::pair<const char*, int>, std::int64_t>; 
+
+private:
+    CLoggerThrottler();
+
+private:
+    static CLoggerThrottler ms_Instance;
+    std::int64_t m_MinimumLogInterval;
```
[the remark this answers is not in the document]
tveasey: I think I'll leave the `chrono` change for the time being. That can be addressed when we migrate `CTimeUtils` as per #1459.

include/core/CLoggerThrottler.h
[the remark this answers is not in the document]
tveasey: Yes sorry type names were sloppy. I fixed this in [b3631bb](https://github.com/elastic/ml-cpp/pull/1615/commits/b3631bb328d262aeefe85ee1ccfbb2d33a024330).

include/core/CLoggerThrottler.h
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License;
+ * you may not use this file except in compliance with the Elastic License.
+ */
+
+#ifndef INCLUDED_ml_core_CLoggerThrottler_h
+#define INCLUDED_ml_core_CLoggerThrottler_h
+
+#include <core/CoreTypes.h>
```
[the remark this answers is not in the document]
tveasey: +1 it was a hangover from an earlier version.

lib/core/unittest/CLoggerTest.cc
tveasey: No hangover.

tveasey: See [65caeb0](https://github.com/elastic/ml-cpp/pull/1615/commits/65caeb0d5cedead7d382e82993dca0959a5cfd37).

lib/core/CLoggerThrottler.cc
tveasey: See [65caeb0](https://github.com/elastic/ml-cpp/pull/1615/commits/65caeb0d5cedead7d382e82993dca0959a5cfd37).

include/core/CLoggerThrottler.h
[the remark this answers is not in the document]
tveasey: See [65caeb0](https://github.com/elastic/ml-cpp/pull/1615/commits/65caeb0d5cedead7d382e82993dca0959a5cfd37).

lib/core/CLoggerThrottler.cc
```diff
@@ -0,0 +1,44 @@
...
+
+bool CLoggerThrottler::skip(const char* file, int line) {
+
+    auto key = std::make_pair(file, line);
+    auto now = CTimeUtils::nowMs();
+
+    std::unique_lock<std::mutex> lock{m_Mutex};
+    auto lastTime = m_LastLogTimes.emplace(key, std::numeric_limits<std::int64_t>::min()).first;
+
+    if (now > lastTime->second + m_MinimumLogInterval) {
```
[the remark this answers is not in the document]
tveasey: See [65caeb0](https://github.com/elastic/ml-cpp/pull/1615/commits/65caeb0d5cedead7d382e82993dca0959a5cfd37).

include/core/CLoggerThrottler.h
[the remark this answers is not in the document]
tveasey: I renamed in [65caeb0](https://github.com/elastic/ml-cpp/pull/1615/commits/65caeb0d5cedead7d382e82993dca0959a5cfd37).

lib/core/CLoggerThrottler.cc
tveasey: I like the idea of this being a member of the logger class see [6c93df4](https://github.com/elastic/ml-cpp/pull/1615/commits/6c93df43771cced3678e08982ffc89508e46806f).

(I don't think we would currently suffer from static initialisation order issues, because I don't think we do anything complicated enough at static variable initialisation time, but do agree we should be defensive to this case.)

tveasey:
Thanks for the review @droberts195: good suggestions. I've addressed all your comments. Can you take another look?

include/core/LogMacros.h
```diff
@@ -73,25 +75,42 @@
...
+    {                                                                           \
+        std::size_t countOfWarnMessages;                                        \
+        bool skipWarnMessage;                                                   \
+        std::tie(countOfWarnMessages, skipWarnMessage) =                        \
+            ml::core::CLogger::instance().throttler().skip(__FILE__, __LINE__); \
+        if (skipWarnMessage == false) {                                         \
+            BOOST_LOG_STREAM_SEV(ml::core::CLogger::instance().logger(),        \
+                                 ml::core::CLogger::E_Warn)                     \
+            LOG_LOCATION_INFO                                                   \
+            message << " [" << countOfWarnMessages << "]";                      \
```
droberts195, line 88: The format here is different to what's used for other repeated messages - see https://github.com/elastic/elasticsearch/blob/4de4c14b5b9e0ecb6be53dfe7bacf945e50efd55/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/process/logging/CppLogMessageHandler.java#L338

It will look inconsistent to people looking at the overall log file.

To make it consistent the format would be this:

```suggestion
            message << " | repeated [" << countOfWarnMessages << "]";                      \
```

include/core/LogMacros.h
```diff
@@ -73,25 +75,42 @@
...
+    {                                                                           \
+        std::size_t countOfErrorMessages;                                       \
+        bool skipErrorMessage;                                                  \
+        std::tie(countOfErrorMessages, skipErrorMessage) =                      \
+            ml::core::CLogger::instance().throttler().skip(__FILE__, __LINE__); \
+        if (skipErrorMessage == false) {                                        \
+            BOOST_LOG_STREAM_SEV(ml::core::CLogger::instance().logger(),        \
+                                 ml::core::CLogger::E_Error)                    \
+            LOG_LOCATION_INFO                                                   \
+            message << " [" << countOfErrorMessages << "]";                     \
```
droberts195, line 104:
```suggestion
            message << " | repeated [" << countOfErrorMessages << "]";                     \
```

lib/core/CLoggerThrottler.cc
```diff
@@ -0,0 +1,45 @@
...
+CLoggerThrottler::CLoggerThrottler() : m_MinimumLogIntervalMs{3600 * 1000} {
+}
+
+void CLoggerThrottler::minimumLogIntervalMs(std::int64_t minimumLogIntervalMs) {
+    m_MinimumLogIntervalMs = minimumLogIntervalMs;
+}
+
+std::pair<std::size_t, bool> CLoggerThrottler::skip(const char* file, int line) {
+
+    auto key = std::make_pair(file, line);
```
droberts195, line 25: One interesting question about using `__FILE__` as part of the key here is whether it's the same if a log message is in a header file and that header file is included in different implementation files.  For example, `CLinearAlgebra.h` contains a `LOG_ERROR` macro, so is the `__FILE__` for `CLinearAlgebra.h` the same in both `CSeasonalComponentAdaptiveBucketing.cc` and `CTimeSeriesDecompositionDetail.cc`?  The answer might vary between compilers too.

If it does vary I don't think that's a big enough problem to warrant changing to a less efficient implementation, but it would deserve a comment to say it's a known problem and sufficiently minor to be acceptable.

lib/core/CLoggerThrottler.cc
droberts195: > (I don't think we would currently suffer from static initialisation order issues, because I don't think we do anything complicated enough at static variable initialisation time, but do agree we should be defensive to this case.)

There is currently one place: https://github.com/elastic/ml-cpp/blob/685b55cfd04b9490ff923417801a272ab148ff43/lib/core/CTimezone_Windows.cc#L54-L55

We used to have more of these, which is the origin of the `DO_NOT_USE_THIS_VARIABLE` variables.  To solve the static initialization order problem we had to use the Meyers singleton pattern for our singletons, but then in the days where function-local static initialization wasn't thread safe we had to ensure they were initialized before `main()` ran.  (In Visual Studio 2015 function-local static initialization was finally made thread safe, so all the `DO_NOT_USE_THIS_VARIABLE` variables can be deleted now - I will do that in a separate PR once this is merged.)

lib/core/CLoggerThrottler.cc
tveasey: Thinking about logging I've seen in the passed I think `__FILE__` nearly always resolves to the header name. However, I added a warning. There is also a work around that the line is explicitly wrapped in a check supplying a custom unique string. I think this is another good reason for not trying to do any more complex association of log messages. See [80fc1ca](https://github.com/elastic/ml-cpp/pull/1615/commits/80fc1cad6711f5fbf6050c173a87ce8657934487).

droberts195: > `__FILE__` nearly always resolves to the header name

Yes, it does, but you're keying on a pointer to the header name, which might well vary between compiler invocations.

tveasey: I'd also wondered this, and in fairness I didn't try other compilers, but clang certainly uses a unique string literal. I think it would be poor compiler behaviour to duplicate strings in this fashion as it would lead to unnecessary executable bloat.

droberts195 (approved):
LGTM

PR #1614, opened 2020-12-08
[ML] Assorted time series modelling improvements

tveasey:
This merges a long running branch which:
1. Improves modularity, robustness and testing of code to select seasonal components to model.
2. Improves re-initialisation of seasonal components after detecting new or different components.
3. Reworks and unifies the code to detect change points in time series and extend to detect new types of change, such as a broader range of  time shifts.
4. Improves handling of extended anomalous regions which subsequently revert to normal behaviour (things like bank holidays) that should have minimal impact on the model.
5. Improves handling of very long bucket lengths.

PR #1618, opened 2020-12-09
[7.11][ML] Assorted time series modelling improvements

tveasey:
Backport #1614.


tveasey:
retest

droberts195:
`std::size` is C++17 - that's why it doesn't work in 7.x, where we're still on C++14.  I guess some compilers might not be following what's in each standard exactly though, hence letting it work ahead of time.

tveasey:
retest

tveasey:
retest

PR #1620, opened 2020-12-09
[ML] Switch to using global memory constants

tveasey:
We have global constants for memory in KB, MB, etc, but we were not reliably using them. This is a small tidy up to migrate to using them consistently. Since this is a refactor I don't think it needs documentation and I'm marking as a non-issue.


valeriy42 (approved):
LGTM

PR #1623, opened 2020-12-09
[ML] Include assignment_memory_basis in model_size_stats

droberts195:
This change is the companion to elastic/elasticsearch#65561.
It calculates the new model_size_stats field,
assignment_memory_basis, to indicate whether an actual model
size or the model memory limit should be used when assigning
jobs to nodes.  The switch only goes one way: once we have
switched from model memory limit to an actual model size we
never revert to using the limit.  This is a change compared
to 7.10, because with autoscaling flipping backwards and
forwards between model memory limit and actual model memory
could (in bad edge cases) cause the cluster to repeatedly
scale up and down.

Relates elastic/elasticsearch#65561
Closes elastic/elasticsearch#63163


droberts195:
This is currently draft because it needs more unit tests

edsavage (approved):
LGTM

PR #1630, opened 2020-12-15
[ML] Improve test to decide whether to drop seasonal components

tveasey:
Our regression test suite showed up some places where we are now over eager to drop all seasonal components. This turns out to be because we weren't properly considering whether the time window we test is suitable for making this decision. This rolls out the checks we were already performing per component to decide if we can drop seasonality modelling altogether. I've also included some consideration of the fraction of values in the window we've observed: we require more evidence to drop seasonality modelling if we're observing a fraction of the window.

This affects results, but it only impacts the decision to keep modelling specific seasonal components so the impact is small. Since this improves unreleased code I've marked it as a non-issue.


droberts195 (approved):
LGTM

PR #1631, opened 2020-12-15
[ML] Make coefficient of variation formula more robust

droberts195:
The previous formula was vulnerable to dividing by zero
in unlikely edge cases.  This change rearranges the formula
to avoid that, and also adds a test to prove no crashes
occur in edge cases.

Relates #1623

### DIFF
--- a/lib/model/CResourceMonitor.cc
+++ b/lib/model/CResourceMonitor.cc
@@ -9,6 +9,8 @@
 #include <core/CProgramCounters.h>
 #include <core/Constants.h>
 
+#include <maths/CTools.h>
+
 #include <model/CMonitoredResource.h>
 #include <model/CStringStore.h>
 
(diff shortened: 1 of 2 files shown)


droberts195:
`>non-issue` because logically this is part of #1623

lib/model/CResourceMonitor.cc
```diff
@@ -330,12 +338,14 @@ bool CResourceMonitor::isMemoryStable(core_t::TTime bucketLength) const {
...
-              << maths::CBasicStatistics::count(m_ModelBytesMoments) << ", mean = " << mean
-              << ", variance = " << variance << ", coefficient of variation = " << cv);
-    if (cv > ESTABLISHED_MEMORY_CV_THRESHOLD) {
+              << ": bucket count = " << bucketCount
+              << ", sample count = " << maths::CBasicStatistics::count(m_ModelBytesMoments)
+              << ", mean = " << mean << ", variance = " << variance
+              << ", coefficient of variation = " << (std::sqrt(variance) / mean));
+    // Instead of literally testing the coefficient of variation it's more
+    // robust against zeroes and NaNs to rearrange it as follows
+    if (variance > maths::CTools::pow2(ESTABLISHED_MEMORY_CV_THRESHOLD * mean)) {
```
tveasey, line 348: Regarding your comment on NaNs: is it actually better to return `false` in the case of NaN? Perhaps it is. Unlike 754, C++ doesn't guaranty comparison operator results on NaNs as I recall, so should we replace
```
if (variance > maths::CTools::pow2(ESTABLISHED_MEMORY_CV_THRESHOLD * mean)) {
    return false;
}

return true;
```
with
```
return CMathsFuncs::isNan(variance) &&
       variance <= maths::CTools::pow2(ESTABLISHED_MEMORY_CV_THRESHOLD * mean);
```
Even if we don't explicitly check NaN, it feels slightly cleaner not to bother with the `if`.

tveasey (approved):
Just a minor suggestion/question regarding NaN. Otherwise, LGTM. Also, no need for another review.

PR #1632, opened 2020-12-15
[7.x][ML] Make coefficient of variation formula more robust

droberts195:
The previous formula was vulnerable to dividing by zero
in unlikely edge cases.  This change rearranges the formula
to avoid that, and also adds a test to prove no crashes
occur in edge cases.

Backport of #1631


droberts195:
retest

PR #1635, opened 2020-12-17
[ML] Use JSON config for categorization

edsavage:
Replace the use of the old-style ini file parser to initialize
categorization jobs with the equivalent JSON configuration parser.

Remove the use of the CFieldConfig parser wherever possible. Effectively
this leaves its use to parse filters and scheduled events in autodetect 
(these will be supplied in JSON formatted configuration files in due
course).

#relates to #1253


lib/api/CAnomalyJobConfig.cc
```diff
@@ -634,9 +672,7 @@ void CAnomalyJobConfig::CAnalysisConfig::CDetectorConfig::parse(
 
     // The detector index is of type int for historical reasons
     // and for consistency across the code base.
-    // The explicit cast is required here as our JSON parser does not support
-    // 32 bit integer values.
-    m_DetectorIndex = static_cast<int>(parameters[DETECTOR_INDEX].as<std::ptrdiff_t>());
+    m_DetectorIndex = detectorIndex;
```
droberts195, line 675: I am not sure about this.  It means both the Java and C++ now rely on the assumption that the detectors array contains the detector indices in order.  It might create a problem in the future, for example if we decided to let users drag their detectors into a different order for display purposes but we wanted to remember their original indices so that we could still tie them up with the old results.  Somebody might change the logic for assigning detector indices in the Java code but not the C++, and then that would cause a subtle, hard-to-detect discrepancy between the detector index reported in C++ results and which detector it related to.

I can see why you've done it - it means you don't need to specify detector index in all the unit test config files.  And it's true that technically detector index doesn't have to be set for configs in the index that date back to 5.x and have never been rewritten.  However, these would get rewritten by the Java as they're written to the input file of the C++.

Would it be possible to take the parsed value if it exists and only use the array index if there is no parsed value?

Maybe something like:

```
    m_DetectorIndex = parameters[DETECTOR_INDEX].as<int>().fallback(fallbackDetectorIndex);
```

I imagine it wouldn't be hard to add support for `int` to the parser either to avoid the cast and comment.

lib/api/CAnomalyJobConfig.cc
```diff
@@ -448,16 +447,48 @@ void CAnomalyJobConfig::CAnalysisConfig::parse(const rapidjson::Value& analysisC
     // The Job config has already been validated by Java before being passed to
     // the C++ backend. So we can safely assume that the detector config is a
     // non-null array - hence this check isn't strictly necessary.
+    this->seenField(m_SummaryCountFieldName);
+    this->seenField(m_CategorizationFieldName);
     if (detectorsConfig != nullptr && detectorsConfig->IsArray()) {
         m_Detectors.resize(detectorsConfig->Size());
+        int detectorIndex{0};
```
droberts195, line 454:
```suggestion
        int fallbackDetectorIndex{0};
```

This suggestion relates to the comment about not introducing a very subtle dependency on the Java implementation of detector index.

bin/autodetect/Main.cc
```diff
@@ -194,16 +194,16 @@ int main(int argc, char** argv) {
     limits.init(analysisLimits.categorizationExamplesLimit(),
                 analysisLimits.modelMemoryLimitMb());
 
-    bool doingCategorization{fieldConfig.fieldNameSuperset().count(
+    const ml::api::CAnomalyJobConfig::CAnalysisConfig& analysisConfig =
+        jobConfig.analysisConfig();
+
+    bool doingCategorization{analysisConfig.fieldNameSuperset().count(
```
droberts195, line 200: It looks like the only use of `fieldNameSuperset` is to check if it contains `mlcategory`.  This way of testing dates back to the days when it was not compulsory to configure `categorization_field_name`, and if you used `mlcategory` in a detector without explicitly specifying `categorization_field_name` then `categorization_field_name` defaulted to `_raw`.  But now it's compulsory to configure `categorization_field_name` if you want categorization you could replace this check with:

```
    bool doingCategorization{analysisConfig.categorizationFieldName().empty() == false};
```

If you do that everywhere where `fieldNameSuperset` is being called then you'll avoid having to populate the set in the first place.

(Ignore this comment if there's some other usage of `fieldNameSuperset` that I missed.)

droberts195 (commented):
Not long now before 1000s of lines of code and test configs can be deleted.

I just had a few minor comments.

bin/autodetect/Main.cc
edsavage: Yes that's correct. This construct is only used to check to see if categorization is wanted - your suggestion is far simpler and makes the desired intent much clearer.

lib/api/CAnomalyJobConfigReader.cc
```diff
@@ -77,6 +77,16 @@ bool CAnomalyJobConfigReader::CParameter::fallback(bool value) const {
     return m_Value->GetBool();
 }
 
+int CAnomalyJobConfigReader::CParameter::fallback(int value) const {
+    if (m_Value == nullptr) {
+        return value;
+    }
+    if (m_Value->IsInt64() == false) {
```
droberts195, line 84:
```suggestion
    if (m_Value->IsInt() == false) {
```

lib/api/CAnomalyJobConfigReader.cc
```diff
@@ -77,6 +77,16 @@ bool CAnomalyJobConfigReader::CParameter::fallback(bool value) const {
...
+        this->handleFatal();
+    }
+    return static_cast<int>(m_Value->GetInt64());
```
droberts195, line 87:
```suggestion
    return m_Value->GetInt();
```

droberts195 (approved):
LGTM

I think a couple of simplifications to the `int` parameter method are possible, but unless there's some complexity there I am not seeing feel free to merge without further review.

PR #1639, opened 2020-12-18
[ML] Avoid applying scale change points when the scale can't be accurately estimated

tveasey:
Scaling has little impact on values close to zero. This means if the window we test for change points overlap a region where the time series is small we cannot estimate scale points well. In such cases it is appropriate to wait to observe more data before committing the change. For example, this can easily happen with say an unusual Friday if the data display typical elevated values during trading days:
before
Screenshot 2020-12-18 at 18 58 59
after
Screenshot 2020-12-18 at 19 03 05

This checks the *predicted* values before and after a candidate scale change point and waits longer to accept the change if the values after are relatively closer to zero.

Since this only affects the decision to apply a scale change point and then only sometimes delays making the choice the scope should be limited. This affects unreleased code and I plan to port this correction to the 7.11 branch, so marking as a non-issue.

### DIFF
--- a/include/maths/CTimeSeriesTestForChange.h
+++ b/include/maths/CTimeSeriesTestForChange.h
@@ -44,7 +44,7 @@ public:
 
     virtual TChangePointUPtr undoable() const = 0;
     virtual bool largeEnough(double threshold) const = 0;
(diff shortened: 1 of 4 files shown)


droberts195 (approved):
LGTM

PR #1640, opened 2020-12-22
[ML] Parse rule filters and scheduled events from JSON config files

edsavage:
Parse JSON config files supplied on the autodetect command line to
extract rule filters and scheduled events configuration.

The new functionality is currently disabled in autodetect until the Java
code to supply the new JSON configuration files has been committed.

It is imperative that these C++ changes be committed ahead of the
corresponding Java changes as the new command line options must be
recognised.

Relates to elastic/elasticsearch#66736
Relates to #1253

### DIFF
--- a/lib/api/CAnomalyJobConfig.cc
+++ b/lib/api/CAnomalyJobConfig.cc
@@ -205,6 +205,12 @@ const std::size_t CAnomalyJobConfig::CAnalysisLimits::DEFAULT_MEMORY_LIMIT_BYTES
 const std::string CAnomalyJobConfig::CDataDescription::TIME_FIELD{"time_field"};
 const std::string CAnomalyJobConfig::CDataDescription::TIME_FORMAT{"time_format"};
 
+const std::string CAnomalyJobConfig::CEventConfig::DESCRIPTION{"description"};
+const std::string CAnomalyJobConfig::CEventConfig::RULES{"rules"};
+
+const std::string CAnomalyJobConfig::CFilterConfig::FILTER_ID{"filter_id"};
+const std::string CAnomalyJobConfig::CFilterConfig::ITEMS{"items"};
+
 namespace {
 const std::string EMPTY_STRING;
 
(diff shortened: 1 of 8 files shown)


lib/api/CAnomalyJobConfig.cc
```diff
@@ -335,6 +359,166 @@ bool CAnomalyJobConfig::initFromFile(const std::string& configFile) {
...
+    }
+
+    try {
+        if (doc.IsArray()) {
+            m_Filters.resize(doc.Size());
+            for (unsigned int i = 0; i < doc.Size(); ++i) {
+                if (!doc[i].IsObject()) {
+                    LOG_ERROR(<< "Could not parse filters: expected filters array to contain objects. JSON: "
+                              << doc.GetParseError());
+                    ;
```
droberts195, line 492: nit: empty statement

lib/core/CPatternSet.cc
```diff
@@ -26,6 +26,42 @@ CPatternSet::CPatternSet()
...
+        if (pattern[0] == WILDCARD) {
+            if (length > 2 && pattern[length - 1] == WILDCARD) {
+                std::string middle = pattern.substr(1, length - 2);
+                containsPatterns.push_back(middle);
+            } else if (length > 1) {
+                std::string suffix = pattern.substr(1);
+                suffixPatterns.push_back(std::string(suffix.rbegin(), suffix.rend()));
+            }
+        } else if (length > 1 && pattern[length - 1] == WILDCARD) {
+            prefixPatterns.push_back(pattern.substr(0, length - 1));
```
droberts195, line 49: It's probably not necessary to optimize for performance in this code, but (apart from the one that reverses order) instead of creating extra copies using `substr` you could use `emplace_back` with the `std::string` constructor that copies a substring of another string.

droberts195 (approved):
I left a couple of minor comments, but since there needs to be a second round of changes to this code after the Java PR is merged feel free to merge this immediately and make extra tweaks in the second round.  You'll need to type `Jenkins run elasticsearch-ci/2` on the Java PR after this PR is merged _and the resulting snapshot build that gets triggered is published_ - you can see when that happens on from the Slack notification.  If you retest the Java side PR too soon it will fail again in the same way.

PR #1644, opened 2021-01-05
[ML] Remove CFieldConfig and related files

edsavage:
* Remove trace of the CFieldConfig class and any related files (e.g. test
configuration files).

* Enable the code to use JSON formatted files to obtain configuration
settings.

* Use the model plot configuration settings present in the anomaly job
configuration to initialise model plot properties.

Note that these changes must not be committed until after the
corresponding changes on the Java side have been.


droberts195 (commented):
The end of this project is getting close now!

lib/model/CAnomalyDetectorModelConfig.cc
```diff
@@ -372,6 +372,29 @@ bool CAnomalyDetectorModelConfig::init(const boost::property_tree::ptree& propTr
     return result;
 }
 
+void CAnomalyDetectorModelConfig::configureModelPlot(bool modelPlotEnabled,
+                                                     bool annotationsEnabled,
+                                                     const std::string& terms) {
+    if (modelPlotEnabled) {
+        m_ModelPlotBoundsPercentile = ml::maths::CModel::DEFAULT_BOUNDS_PERCENTILE;
```
droberts195, line 379: nit: don't need the `ml::` here as it's in the `ml` namespace.

droberts195 (approved):
LGTM

I saw one tiny thing, but since there will be another PR in the series you can leave it to that to get through PR CI quicker.

PR #1646, opened 2021-01-07
[ML] Improve time series model stability

tveasey:
This makes the following changes which affect the stability of the time series decomposition:
1. We cache and interpolate the seasonal model predictions (for performance reasons). We must ensure that the times we compute the predictions match the times at which we will request the predictions next or we induce an error which is proportional to the gradient multiplied by the time offset. We weren't using the correct time for time windowed models, for example weekday and weekend specific models. The same considerations apply when we apply change points to the model.
2. When we apply a scale event we preserve the gradient across the scaling. This was done to avoid amplifying gradient estimation errors. In fact, we only really run into this issue if we scale up. If we scale down it is more appropriate to scale the gradient as well since we do expect this to also be scaled and by preserving the old gradient (estimation error) we are effectively amplifying errors afterwards.

We've had a number of issues related poor anomaly detection after a partition stops receiving data some of which have already been fixed. This came up (although the scope is broader) investigating another case in our QA suite:
Before
Screenshot 2021-01-07 at 13 05 33

After
Screenshot 2021-01-07 at 13 05 41

The situation is still not perfect and I plan to make a follow on change, but it represents a significant improvement (notice the y-axis scale) and this change is relatively self contained.


lib/maths/CSeasonalComponentAdaptiveBucketing.cc
```diff
@@ -179,21 +179,32 @@ void CSeasonalComponentAdaptiveBucketing::shiftLevel(double shift) {
...
+            double gradientBefore{gradient(m_Buckets[i].s_Regression)};
+            m_Buckets[i].s_Regression.linearScale(scale);
+            double gradientAfter{gradient(m_Buckets[i].s_Regression)};
+
+            core_t::TTime bucketTime{time + static_cast<core_t::TTime>(centres[i] + 0.5)};
+            m_Buckets[i].s_Regression.shiftGradient(gradientBefore - gradientAfter);
+            m_Buckets[i].s_Regression.shiftOrdinate(
+                -(gradientBefore - gradientAfter) * m_Time->regression(bucketTime));
+        }
+        m_Buckets[i].s_Variance += CTools::truncate(0.02 * (scale - 1.0), 0.0, 0.1) *
```
droberts195, line 206: This formula seems to contain some magic numbers.  It might be clearer if they were symbolic constants or if there was a comment explaining the reasoning for them.

tveasey: The functional form isn't magic, but the numbers are somewhat. In practice anomaly detection is not super sensitive to the exact choice of numbers. These were chosen as the smallest correction needed to essentially avoid spurious anomalies after applying a change point in some representative data sets. I think the key idea is to set the scale based on prediction^2. This means we effectively stop transient anomalies associated with small **relative** errors in our predictions. I've added a comment to this effect.

tveasey: See [51019a6](https://github.com/elastic/ml-cpp/pull/1646/commits/51019a6da2e720e304bb9dad4b400c53ab655b0f).

tveasey: I changed my mind and just reverted the non-essential parts of this PR. I think it makes sense to keep this to purely addressing causes of instability and any changes of behaviour can be made in a separate PR.

droberts195 (approved):
LGTM

PR #1647, opened 2021-01-07
[ML] Always use Boost unordered maps/sets

droberts195:
The memory instrumentation in core::CMemory exists for
boost::unordered_set and boost::unordered_map, but not
the std equivalents.  It is better that we switch all
uses in the codebase at the same time (or even switch
to completely different hash container implementations).

This change switches the few cases of std::unordered_map
and std::unordered_set to the boost versions.  This may
cause some increases in reported memory usage, where these
containers are in memory instrumented classes.  It won't
increase actual memory usage (at least not significantly),
just make the reporting more accurate.

### DIFF
--- a/include/api/CBoostedTreeInferenceModelBuilder.h
+++ b/include/api/CBoostedTreeInferenceModelBuilder.h
@@ -14,6 +14,8 @@
 
 #include <rapidjson/document.h>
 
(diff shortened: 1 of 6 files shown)


tveasey (approved):
LGTM

PR #1650, opened 2021-01-08
[ML] Fix scale initialization

valeriy42:
This PR completes the assignment of hyperparameters using enumeration, which was started in  #1627. Using enums instead of fixed order allows preventing bugs if new hyperparameters will be added.


tveasey (approved):
LGTM

PR #1654, opened 2021-01-11
[ML] Make sure we remove seasonality modelling when data flatlines

tveasey:
While fixing #1646, I noticed we weren't reliably removing seasonal components from our time series model when the data flatlined. This happens for example, when a count anomaly detector stops receiving new data.

The behaviour was due to a bug in how we computed statistics related to the components we already modelled. Specifically, we would discard the correct statistics because `SHypothesisStats::isBetter` didn't consider whether or not the type was default initialised. In such cases, discarding seasonality leads to smaller, more accurate and more stable models.

I took the opportunity to remove the previous approach of perturbing the values to deal with low variation data, because this doesn't work when the values are zero, and otherwise should no longer be needed.


droberts195 (approved):
LGTM

PR #1655, opened 2021-01-12
[ML] Fix wrong model metadata label

valeriy42:
When I was writing model metadata hyperparameters in #1627 , I was outputting `eta` values and was using `downsample_factor` label. This RP fixes this.


edsavage (approved):
LGTM

PR #1656, opened 2021-01-12
[ML] Improve program counter resets between unit tests

droberts195:
Program counters are implemented as a singleton object,
which means their values spill between unit tests unless
action is taken to prevent this.  Often it doesn't matter,
as most unit tests do not care about program counter
values.

Two unit test suites already reset program counters in
between tests: CProgramCountersTest and
CRestorePreviousStateTest.

It turns out that a third test suite should have been
resetting counters but wasn't:
CDataFrameAnalysisInstrumentationTest.

This PR rectifies that omission and also factors out the
code used to reset counters into the test library which
makes it easier to reuse if other suites also need the
functionality in the future.

### DIFF
--- a/include/core/CProgramCounters.h
+++ b/include/core/CProgramCounters.h
@@ -18,7 +18,6 @@
 #include <vector>
 
 namespace CProgramCountersTest {
-class CTestFixture;
 class CProgramCountersTestRunner;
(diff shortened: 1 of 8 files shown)


edsavage (approved):
LGTM

PR #1661, opened 2021-01-12
[ML] Revert "Improve time series model stability"

tveasey:
#1646 created some QA regressions. It also doesn't fully fix the underlying issues with time windowed components. I'm going to prepare a new PR which targets just this issue and introduces better unit testing and revert this in the mean time.


droberts195 (approved):
LGTM

PR #1662, opened 2021-01-13
[ML] Leniency parsing background_persist_interval

edsavage:
Be more lenient when parsing the background_persist_interval field as it is: 
1. optional and 
2. not present in the job config in the majority of cases


droberts195 (approved):
LGTM

edsavage:
retest

PR #1663, opened 2021-01-13
[ML] Remove redundant autodetect options

edsavage:
Remove the now redundant `autodetect` command line options and
exit with error message on seeing unknown command line arguments

### DIFF
--- a/bin/autodetect/CCmdLineParser.cc
+++ b/bin/autodetect/CCmdLineParser.cc
@@ -17,7 +17,7 @@
 namespace ml {
 namespace autodetect {
 
-const std::string CCmdLineParser::DESCRIPTION = "Usage: autodetect [options] [<fieldname>+ [by <fieldname>]]\n"
+const std::string CCmdLineParser::DESCRIPTION = "Usage: autodetect [options]]\n"
                                                 "Options:";
 
 bool CCmdLineParser::parse(int argc,
@@ -45,29 +45,21 @@ bool CCmdLineParser::parse(int argc,
                            bool& isPersistFileNamedPipe,
                            bool& isPersistInForeground,
                            std::size_t& maxAnomalyRecords,
-                           bool& memoryUsage,
-                           bool& stopCategorizationOnWarnStatus,
-                           TStrVec& unknownTokens) {
+                           bool& memoryUsage) {
     try {
         boost::program_options::options_description desc(DESCRIPTION);
         // clang-format off
         desc.add_options()
             ("help", "Display this information and exit")
             ("version", "Display version information and exit")
-            ("config", boost::program_options::value<std::string>(),
+            ("config", boost::program_options::value<std::string>()->required(),
                     "The job configuration file")
             ("filtersconfig", boost::program_options::value<std::string>(),
-             "The filters configuration file")
+                    "The filters configuration file")
             ("eventsconfig", boost::program_options::value<std::string>(),
-             "The scheduled events configuration file")
-            ("limitconfig", boost::program_options::value<std::string>(),
-                    "Optional limit config file")
+                    "The scheduled events configuration file")
             ("modelconfig", boost::program_options::value<std::string>(),
                     "Optional model config file")
-            ("fieldconfig", boost::program_options::value<std::string>(),
-                    "Optional field config file")
-            ("modelplotconfig", boost::program_options::value<std::string>(),
-                    "Optional model plot config file")
             ("logProperties", boost::program_options::value<std::string>(),
                     "Optional logger properties file")
             ("logPipe", boost::program_options::value<std::string>(),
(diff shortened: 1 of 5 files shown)


bin/autodetect/Main.cc
```diff
@@ -152,6 +151,12 @@ int main(int argc, char** argv) {
     // statically links its own version library.
     LOG_DEBUG(<< ml::ver::CBuildInfo::fullInfo());
 
+    if (unknownTokens.size()) {
+        LOG_ERROR(<< "Unkown command line options: "
+                  << ml::core::CContainerPrinter::print(unknownTokens));
+        return EXIT_FAILURE;
+    }
```
droberts195, line 158: You won't need this check if you configure Boost Program Options to ban unknown arguments.

bin/autodetect/CCmdLineParser.cc
```diff
@@ -69,24 +68,14 @@ bool CCmdLineParser::parse(int argc,
                     "Optional field config file")
             ("modelplotconfig", boost::program_options::value<std::string>(),
                     "Optional model plot config file")
```
droberts195, line 70: Can `modelplotconfig`, `fieldconfig` and `limitconfig` also be removed now?

Also (nit) I just noticed the help text for `filtersconfig` and `eventsconfig` is not indented the same amount as all the other help text.

bin/autodetect/CCmdLineParser.cc
```diff
@@ -107,19 +96,13 @@ bool CCmdLineParser::parse(int argc,
...
-            ("maxQuantileInterval", boost::program_options::value<core_t::TTime>(),
-                    "Optional interval at which to periodically output quantiles if they have not been output due to an anomaly - if not specified then quantiles will only be output following a big anomaly")
+                    "Optional number of buckets after which to periodically persist model state.")
             ("maxAnomalyRecords", boost::program_options::value<std::size_t>(),
                     "The maximum number of records to be outputted for each bucket. Defaults to 100, a value 0 removes the limit.")
             ("memoryUsage",
                     "Log the model memory usage at the end of the job")
-            ("multivariateByFields",
-                    "Optional flag to enable multi-variate analysis of correlated by fields")
             ("stopCategorizationOnWarnStatus",
```
droberts195, line 106: This one can come from the job config too now, and shouldn't be supported as a command line option.

bin/autodetect/CCmdLineParser.cc
```diff
@@ -48,7 +47,7 @@ bool CCmdLineParser::parse(int argc,
                            std::size_t& maxAnomalyRecords,
                            bool& memoryUsage,
                            bool& stopCategorizationOnWarnStatus,
```
droberts195, line 49: Shouldn't be needed any more.

bin/autodetect/CCmdLineParser.cc
```diff
@@ -107,19 +96,13 @@ bool CCmdLineParser::parse(int argc,
...
                     "Optional file to persist state to - not present means no state persistence")
             ("persistIsPipe", "Specified persist file is a named pipe")
-            ("persistInterval", boost::program_options::value<core_t::TTime>(),
-                    "Optional time interval at which to periodically persist model state (Mutually exclusive with bucketPersistInterval)")
             ("persistInForeground", "Persistence occurs in the foreground. Defaults to background persistence.")
             ("bucketPersistInterval", boost::program_options::value<std::size_t>(),
-                    "Optional number of buckets after which to periodically persist model state (Mutually exclusive with persistInterval)")
-            ("maxQuantileInterval", boost::program_options::value<core_t::TTime>(),
-                    "Optional interval at which to periodically output quantiles if they have not been output due to an anomaly - if not specified then quantiles will only be output following a big anomaly")
+                    "Optional number of buckets after which to periodically persist model state.")
```
droberts195, line 101: Is it now that if `bucketPersistInterval` is specified on the command line then it takes precedence over the persist interval in the config?

bin/autodetect/CCmdLineParser.h
```diff
@@ -60,7 +59,7 @@ class CCmdLineParser {
                       std::size_t& maxAnomalyRecords,
                       bool& memoryUsage,
                       bool& stopCategorizationOnWarnStatus,
-                      TStrVec& clauseTokens);
+                      TStrVec& unknownTokens);
```
droberts195, line 62: This argument can be removed.

bin/autodetect/CCmdLineParser.cc
edsavage: Yes, we don't expect the `bucketPersistInterval` to ever be used in production code (it was introduced to support the `model_extractor` project). So if present it should take precedence over the background persist interval.

bin/autodetect/Main.cc
```diff
@@ -110,16 +109,15 @@ int main(int argc, char** argv) {
     bool isPersistInForeground{false};
     std::size_t maxAnomalyRecords{100};
     bool memoryUsage{false};
-    bool stopCategorizationOnWarnStatus{false};
-    TStrVec clauseTokens;
+    TStrVec unknownTokens;
```
droberts195, line 112: I think this variable is now redundant.

droberts195 (approved):
LGTM if you could just delete 1 more redundant line

PR #1668, opened 2021-01-14
[ML] Fix bug persisting and restoring time series model decay rate controller

tveasey:
We were failing to persist and restore the checks we perform when controlling decay rate for time series models. The upshot is that we would disable decay rate control after a round of persist and restore. This could happen on failover or if the job is closed and restarted.

The checksum for the class was also incomplete which should have picked this up in the unit testing. I have however also added a test that the behaviour is preserved after persist and restore.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -99,6 +99,7 @@ double aggregateFeatureProbabilities(const TDouble4Vec& probabilities, double co
 const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_5_TAG("6.5");
 const std::string VERSION_7_3_TAG("7.3");
+const std::string VERSION_7_11_TAG("7.11");
 
 // Models
 // Version >= 6.3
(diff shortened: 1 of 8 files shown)


droberts195 (approved):
LGTM

lib/maths/unittest/CDecayRateControllerTest.cc
```diff
@@ -109,4 +108,47 @@ BOOST_AUTO_TEST_CASE(testPersist) {
     }
 }
 
+BOOST_AUTO_TEST_CASE(testBehaviourAfterPersistAndRestore) {
```
droberts195, line 111: This will need to be a CppUnit test in 6.8, with a method declaration added to the header too.

lib/maths/CTimeSeriesModel.cc
```diff
@@ -99,6 +99,7 @@ double aggregateFeatureProbabilities(const TDouble4Vec& probabilities, double co
 const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_5_TAG("6.5");
 const std::string VERSION_7_3_TAG("7.3");
+const std::string VERSION_7_11_TAG("7.11");
```
droberts195, line 102: One more thing to consider from a backport perspective is the naming of this tag.  It's going to look a bit weird to have "VERSION_7_11_TAG" in 6.8.14.  So maybe:

```suggestion
const std::string VERSION_7_11_AND_6_8_14_TAG("7.11and6.8.14");
```

tveasey: I think my inclination is just to rename this in 6.8 since I'll need to make changes to get this to merge anyway and we shouldn't get conflicts here. I'll get a review on that PR.

tveasey:
The aarch failure is a CI worker setup issue.

PR #1674, opened 2021-01-18
[6.8][ML] Fix bug persisting and restoring time series model decay rate controller

tveasey:
Backport #1668.

### DIFF
--- a/lib/maths/CTimeSeriesModel.cc
+++ b/lib/maths/CTimeSeriesModel.cc
@@ -112,6 +112,7 @@ double aggregateFeatureProbabilities(const TDouble4Vec& probabilities, double co
 
 const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_5_TAG("6.5");
+const std::string VERSION_7_11_TAG("6.8");
 
 // Models
 // Version >= 6.3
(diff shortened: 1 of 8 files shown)


lib/maths/CTimeSeriesModel.cc
```diff
@@ -112,6 +112,7 @@ double aggregateFeatureProbabilities(const TDouble4Vec& probabilities, double co
 
 const std::string VERSION_6_3_TAG("6.3");
 const std::string VERSION_6_5_TAG("6.5");
+const std::string VERSION_6_8_TAG("6.8");
```
droberts195, line 115: I think this should make clear that it starts from 6.8.14 and not 6.8.0.

```suggestion
const std::string VERSION_6_8_14_TAG("6.8.14");
```

droberts195 (approved):
LGTM apart from one variable name but if that's changed I am happy to merge without further review if it passes CI

lib/maths/CTimeSeriesModel.cc
tveasey: Done in [6add0d2](https://github.com/elastic/ml-cpp/pull/1674/commits/6add0d230149bdfab359c78a32925eec43da26ed).

PR #1675, opened 2021-01-18
[ML] Assorted time series modelling stability fixes

tveasey:
This addresses various sources of instability in the time series modelling for anomaly detection:
1. Our update dynamics were stable (just). However, we do other things which aren't considered in this analysis, such as apply different weights to different updates. I added a small margin.
2. There was an error in the times at which we asked for model predictions when component models were restricted to time windows, i.e. weekdays and weekends.
3. We could use stale predictions when there were long gaps in the data.
4. Changing the seasonal derivative too fast can create transient instability after data changes.

I've also added some more unit testing. Particular, in the area of when we refresh predictions and an explicit testing long term stability.


droberts195 (approved):
LGTM

PR #1676, opened 2021-01-19
[ML] Early stopping for hyperparameter optimization procedure

### LINKED ISSUE #1669: [ML] Early stopping for hyperparameter optimization procedure
There are several scenarios when hyperparameter optimization doesn't seem to work. When the CV loss variation is too high, model performance basically varies from the fold to fold so much that the algorithm scales the whole GP down to a constant (\theta_0 -> 0), In this case, we can interrupt the `fine_parameter_tuning` phase without a problem, since any potential improvement is due to randomness and not due to a particular choice of the parameters.

To this end, in order to introduce early stopping, in `selectNextHyperparameters()` one can check after the random sampling phase if the 0-s element of `m_KernelParameters` is very small (e.g.  1e-9) or alternatively if the total variation of the GP is very small. Then the method should return `false` and the `fine_parameter_tuning` phase will be interrupted as intended.

valeriy42:
This PR implements early stopping of fine parameter tuning if the search process if the loss surfaces appear to be flat (very low total variation of the loss surface of hyperparameters). 
It also replaces uniform sampling of the hyperparameter in the first step of exploration by Sobol sampling which has a higher discrepancy.

I introduced the new configuration parameter `early_stopping_allowed` similar to the `disk_usage_allowed`. In the internal routines and for persist/restore, I call this flag `stopHyperparameterOptimizationEarly` similar to `stopCrossValidationEarly` that is already used.

Closes  #1669

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -47,9 +47,11 @@
   if the time series has non-diurnal seasonality. (See {ml-pull}1634[#1634].)
 * Compute importance of hyperparameters optimized in the fine parameter tuning step.
   (See {ml-pull}1627[#1627].)
-* Early stopping for the fine parameter tuning step. (See {ml-pull}1676[#1676].)
+* Early stopping for the fine parameter tuning step  of classification and regression
+  model training. (See {ml-pull}1676[#1676].)
 * Correct upgrade for pre-6.3 state for lat_long anomaly anomaly detectors. (See
   {ml-pull}1681[#1681].)
+
 == {es} version 7.11.0
 
 === Enhancements
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -47,6 +47,7 @@
   if the time series has non-diurnal seasonality. (See {ml-pull}1634[#1634].)
 * Compute importance of hyperparameters optimized in the fine parameter tuning step.
   (See {ml-pull}1627[#1627].)
+* Early stopping for the fine parameter tuning step. (See {ml-pull}1676[#1676].)
 
 == {es} version 7.11.0
 
--- a/lib/maths/CBoostedTreeImpl.cc
+++ b/lib/maths/CBoostedTreeImpl.cc
@@ -1297,7 +1297,8 @@ bool CBoostedTreeImpl::selectNextHyperparameters(const TMeanVarAccumulator& loss
                   m_HyperparameterSamples[m_CurrentRound].end(), parameters.data());
         parameters = minBoundary + parameters.cwiseProduct(maxBoundary - minBoundary);
     } else {
-        if (m_EarlyStoppingAllowed && m_BayesianOptimization->anovaTotalVariance() < 1e-9) {
+        if (m_StopHyperparameterOptimizationEarly &&
+            m_BayesianOptimization->anovaTotalVariance() < 1e-9) {
             return false;
         }
         std::tie(parameters, std::ignore) = bopt.maximumExpectedImprovement();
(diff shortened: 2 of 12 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1286,12 +1293,14 @@ bool CBoostedTreeImpl::selectNextHyperparameters(const TMeanVarAccumulator& loss
 
     bopt.add(parameters, meanLoss, lossVariance);
     if (3 * m_CurrentRound < m_NumberRounds) {
-        std::generate_n(parameters.data(), parameters.size(), [&]() {
-            return CSampling::uniformSample(m_Rng, 0.0, 1.0);
-        });
-
+        std::move(m_HyperparameterSamples[m_CurrentRound].begin(),
+                  m_HyperparameterSamples[m_CurrentRound].end(), parameters.data());
```
tveasey, line 1297: I'd change the entry condition for this block to `if (m_CurrentRound < m_HyperparameterSamples.size())` otherwise we can easily get undefined behaviour if we change `initializeHyperparameterSamples` and miss here.

tveasey: Also I'd just use a copy here since the moved types are doubles, i.e.
```
std::copy(m_HyperparameterSamples[m_CurrentRound].begin(),
                  m_HyperparameterSamples[m_CurrentRound].end(), parameters.data());
```

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -341,6 +343,11 @@ std::size_t CBoostedTreeImpl::estimateMemoryUsage(std::size_t numberRows,
     std::size_t hyperparametersMemoryUsage{numberColumns * sizeof(double)};
     std::size_t tunableHyperparametersMemoryUsage{m_TunableHyperparameters.size() *
                                                   sizeof(int)};
+    std::size_t hyperparameterSamplesMemoryUsage{
+        m_HyperparameterSamples.empty()
+            ? 0
+            : m_HyperparameterSamples.size() *
+                  m_HyperparameterSamples[0].size() * sizeof(double)};
```
tveasey, line 350: This function gets called before train so I don't think this works. I'd propose:
```
    std::size_t hyperparameterSamplesMemoryUsage{
        (m_NumberRounds / 3 + 1) * m_TunableHyperparameters.size() * sizeof(double)};
```
I'd be inclined to pull `m_NumberRounds / 3 + 1` into a function as well so this properly encapsulated.

tveasey: Except that I think this should also be using `this->numberHyperparametersToTune()` since `m_TunableHyperparameters.size()` also won't be initialised at the point this function is called.

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -203,6 +203,8 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
 
     this->startProgressMonitoringFineTuneHyperparameters();
 
+    this->initializeHyperparameterSamples();
```
tveasey, line 206: nit: logically this seems to belong right before `while (m_CurrentRound < m_NumberRounds) {` together with call to `this->initializePerFoldTestLosses();`.

lib/maths/CBoostedTreeImpl.cc
tveasey: Also just realised we probably want to make:
```
TVector parameters{this->numberHyperparametersToTune()};
```
at the start of this function
```
TVector parameters{m_TunableHyperparameters.size()};
```
for the same reason.

tveasey (commented):
Overall looks good. I made some minor suggestions. Also the memory accounting is slightly off (I don't suppose the errors are large enough to actually matter, but I think it makes sense to correct).

docs/CHANGELOG.asciidoc
```diff
@@ -47,6 +47,7 @@
   if the time series has non-diurnal seasonality. (See {ml-pull}1634[#1634].)
 * Compute importance of hyperparameters optimized in the fine parameter tuning step.
   (See {ml-pull}1627[#1627].)
+* Early stopping for the fine parameter tuning step. (See {ml-pull}1676[#1676].)
```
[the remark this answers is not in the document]
valeriy42: changed

lib/maths/CBoostedTreeImpl.cc
valeriy42: :+1:  Great comments! I made the change.

lib/maths/CBoostedTreeImpl.cc
valeriy42: :+1: Thank you for the suggestion. Fixed now.

tveasey (approved):
LGTM. Nice improvement!

PR #1678, opened 2021-01-19
[ML] Add libtorch Linux DockerFile

davidkyle:
Builds and installs Python 3.7.9 and CMake.  
Installs the Python dependencies for PyTorch and builds libtorch.

Source is downloaded and built in the `${build_dir}=/usr/src` directory. The existing `RUN` commands have been changed to use this directory rather than the file system root. 

This PR is a draft because the change is still being tested and those test cycles are long.


dev-tools/docker/linux_image/Dockerfile
```diff
@@ -82,3 +90,51 @@ RUN \
...
+  cd ${build_dir} && \
+  git -c advice.detachedHead=false clone --depth=1 --branch=v1.7.1 https://github.com/pytorch/pytorch.git && \
+  cd pytorch && \
+  git submodule sync && \
+  git submodule update --init --recursive 
+
+# Build LibTorch
+RUN \
+  cd ${build_dir}/pytorch && \
+  /usr/local/bin/python3.7 setup.py install && \
```
droberts195, line 132:
```suggestion
  export USE_DISTRIBUTED=OFF
  /usr/local/bin/python3.7 setup.py install && \
```

This needs to be explicitly specified on Linux, whereas it's the default on macOS.

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -82,3 +90,51 @@ RUN \
...
+  make altinstall && \
+  cd .. && rm -rf Python-3.7.9
+
+# Install Python dependencies 
+RUN \
+  /usr/local/bin/pip3.7 install numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses  
+
+# Install CMake
+# v3.5 minimum is required
+ARG cmake_install_dir=/usr/local/cmake
```
droberts195, line 109: I am not sure this should be an `ARG`.  As things stand there doesn't seem to be any benefit in changing it, as we don't use CMake outside of this Dockerfile.  But if we _do_ start using it later on in our build process, if somebody changed it to a different value here then that would just create complication for our downstream build scripts.  The way our build environment is specified is that you _must_ install build tools into certain directories and then downstream scripts don't have to deal with uncertainty about where they are.

droberts195 (approved):
I saw a couple of issues, but due to the build time am happy to merge this to the feature branch as-is if it built a usable image.  The reason is that this will need further iteration before we ship, so `USE_DISTRIBUTED` can be incorporated then.  There will probably be other flags to set differently too, similar to the ones I mentioned in #1677.

A couple of things to check about the libraries that got built are:

1. What does [`hardening-check`](http://www.linuxcertif.com/man/1/hardening-check/) say about them?  The comments in `setup.py` say that `CLFAGS` get propagated, so if that's true then the libraries should be hardened.  But it would be good to confirm.
2. What does `ldd -v` say about the libraries?  On Windows `fbgemm.dll` and `asmjit.dll` were hard dependencies of `torch_cpu.dll`.  Does `ldd` show that `libfbgemm.so` and `libasmjit.so` are hard dependencies of `libtorch_cpu.so` on Linux?  If so we'll need to ship them too.  But maybe the linking is done differently on Linux.

But even if one or both of these things need attention it's OK to merge this PR and just keep a list of all the things that need doing in the next iteration before we merge the feature branch.

dev-tools/docker/linux_image/Dockerfile
davidkyle: I used an ARG to avoid copy and paste programming but yes I suppose it shouldn't be updatable by the outside world. 

ARG `build_dir` is ok as that directory is wiped in the final command anyway

davidkyle:
I build the pytorch app in this image and ran the smoke test which worked fine. This is good to merge now.

The only problem was that `libmkl` (specifically libmkl_intel_lp64.so.1) was not on `LD_LIBRARY_PATH`, once I added `/usr/local/lib` all was good.

PR #1680, opened 2021-01-20
[6.8][ML] Fix bug persisting and restoring time series model decay rate controller (part 2)

tveasey:
Unfortunately, using the state tag to determine if the state is incomplete is not forward compatible. This breaks upgrading state from the 6.8 branch to 7.x branches. I could introduce a new item (which would be ignored) to carry this information, but since this parameter is currently not user configurable I can simply check its value to determine if it is missing. This fixes fallout from #1674 so marking as a non-issue since it logically belongs with that change.


droberts195 (approved):
LGTM

PR #1681, opened 2021-01-20
[ML] Correct a bug in upgrade from pre 6.3 state for the lat_long function

tveasey:
Whilst investigating #1680 I noticed another issue in state upgrade. This one is much more limited in scope and only affects pre 6.3 lat_long models.


droberts195 (approved):
LGTM

PR #1683, opened 2021-01-20
[ML] Provide the ability to override eta growth rate hyperparameter

tveasey:
The eta growth rate was historically controlled by the choice for the eta parameter, but setting them independently might sometimes yield better results. This adds the ability to override it manually. This isn't user facing yet (in the Java) so marking as a non-issue and not documenting.

### DIFF
--- a/lib/maths/CBoostedTreeFactory.cc
+++ b/lib/maths/CBoostedTreeFactory.cc
@@ -1234,7 +1234,7 @@ CBoostedTreeFactory& CBoostedTreeFactory::eta(double eta) {
 
 CBoostedTreeFactory& CBoostedTreeFactory::etaGrowthRatePerTree(double etaGrowthRatePerTree) {
     if (etaGrowthRatePerTree < MIN_ETA) {
-        LOG_WARN(<< "Truncating supplied learning rate " << etaGrowthRatePerTree
+        LOG_WARN(<< "Truncating supplied learning rate growth rate " << etaGrowthRatePerTree
                  << " which must be no smaller than " << MIN_ETA);
         etaGrowthRatePerTree = std::max(etaGrowthRatePerTree, MIN_ETA);
     }
--- a/lib/maths/CBoostedTreeFactory.cc
+++ b/lib/maths/CBoostedTreeFactory.cc
@@ -233,7 +233,8 @@ void CBoostedTreeFactory::initializeHyperparameterOptimisation() const {
             }
             break;
         case E_EtaGrowthRatePerTree:
-            if (m_TreeImpl->m_EtaOverride == boost::none) {
+            if (m_TreeImpl->m_EtaOverride == boost::none &&
+                m_TreeImpl->m_EtaGrowthRatePerTreeOverride == boost::none) {
                 double rate{m_TreeImpl->m_EtaGrowthRatePerTree - 1.0};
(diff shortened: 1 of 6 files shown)


lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1381,7 +1381,9 @@ void CBoostedTreeImpl::restoreBestHyperparameters() {
 std::size_t CBoostedTreeImpl::numberHyperparametersToTune() const {
     return m_RegularizationOverride.countNotSet() +
            (m_DownsampleFactorOverride != boost::none ? 0 : 1) +
-           (m_EtaOverride != boost::none ? 0 : 2) +
+           (m_EtaOverride != boost::none
```
dimitris-athanasiou, line 1384: it seems like here in the scenario where eta is overridden but eta_growth is not, we will calculate zero params to tune. Is that right? I would expect we'd count 1 param to tune (eta_growth).

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -1225,6 +1232,16 @@ CBoostedTreeFactory& CBoostedTreeFactory::eta(double eta) {
     return *this;
 }
 
+CBoostedTreeFactory& CBoostedTreeFactory::etaGrowthRatePerTree(double etaGrowthRatePerTree) {
+    if (etaGrowthRatePerTree < MIN_ETA) {
+        LOG_WARN(<< "Truncating supplied learning rate " << etaGrowthRatePerTree
```
dimitris-athanasiou, line 1237: should we explicitly mention "growth rate" here to know which one it is?

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -1512,6 +1514,7 @@ const std::string DOWNSAMPLE_FACTOR_TAG{"downsample_factor"};
 const std::string ENCODER_TAG{"encoder"};
 const std::string ETA_GROWTH_RATE_PER_TREE_TAG{"eta_growth_rate_per_tree"};
 const std::string ETA_OVERRIDE_TAG{"eta_override"};
+const std::string ETA_GROWTH_PER_TREE_OVERRIDE_TAG{"eta_growth_per_tree"};
```
dimitris-athanasiou, line 1517: is `rate` missing intentionally?

tveasey: No. Agreed it makes sense to be consistent.

tveasey: Fixed in [e9ba4f5](https://github.com/elastic/ml-cpp/pull/1683/commits/e9ba4f50624d2e832ce41e0e8ec459c2855bbaf8).

lib/maths/CBoostedTreeFactory.cc
tveasey: Corrected in [1bd2a41](https://github.com/elastic/ml-cpp/pull/1683/commits/1bd2a414085496269f06478b5e2e9d957dbccdd9).

dimitris-athanasiou (approved):
LGTM

valeriy42 (approved):
LGTM. I am bumping the persist/restore format version of `CBoosteTreeImpl` in my PR, so it's fine that you didn't.

PR #1685, opened 2021-01-20
[ML] Adjusting macOS build instructions for PyTorch

droberts195:
1. Use vanilla Python instead of Anaconda
2. Don't build all the possible libraries, as we
   don't need them (there may be more to do here
   in the future too)
3. Move instructions to main build instructions now
   we are on the road to incorporating this into
   core product


davidkyle (approved):
LGTM

PR #1692, opened 2021-01-24
[ML] Improving PyTorch build on Windows

droberts195:
This change sets extra flags during the PyTorch build to
avoid building unnecessary components, which dramatically
cuts the build time.  It also changes the location of
debug symbols from the libraries themselves to separate
.pdb files.

A couple of tweaks to the 1.7.1 build files are required
to get these options to work.  One of these is already
in the PyTorch master branch code.  There may be time to
get the other one incorporated in time for 1.8.0 too.

### DIFF
--- a/build-setup/windows.md
+++ b/build-setup/windows.md
@@ -233,15 +233,39 @@ git submodule sync
 git submodule update --init --recursive
 ```
 
+In `caffe2/utils/CMakeLists.txt` delete the following line (line 4):
+
+```
+    utils/threadpool/pthreadpool-cpp.cc
+```
+
+In `torch/CMakeLists.txt` remove the guards around `onnx_library`, i.e. change lines 99-101 from:
+
+```
+    if(BUILD_TEST)
+      list(APPEND TORCH_PYTHON_LINK_LIBRARIES onnx_library)
+    endif(BUILD_TEST)
+```
+
+to:
+
+```
+    list(APPEND TORCH_PYTHON_LINK_LIBRARIES onnx_library)
+```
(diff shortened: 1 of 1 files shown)


build-setup/windows.md
```diff
@@ -233,16 +233,40 @@ git submodule sync
 git submodule update --init --recursive
 ```
 
+In `caffe2/utils/CMakeLists.txt` delete the following line (line 4):
+
+```
+    utils/threadpool/pthreadpool-cpp.cc
+```
```
droberts195, line 240: This one is already in the PyTorch master branch, so shouldn't be necessary for us to do locally when we upgrade to version 1.8.0 - see https://github.com/pytorch/pytorch/pull/45610/files#diff-7a428c42f928ff643f0eae8a0b095bb6af88289aae4ba334ab0a62ce617c64acL4

build-setup/windows.md
```diff
@@ -233,16 +233,40 @@ git submodule sync
...
+    if(BUILD_TEST)
+      list(APPEND TORCH_PYTHON_LINK_LIBRARIES onnx_library)
+    endif(BUILD_TEST)
+```
+
+to:
+
+```
+    list(APPEND TORCH_PYTHON_LINK_LIBRARIES onnx_library)
+```
```
droberts195, line 254: This one is still a problem in the current master branch code of PyTorch - see https://github.com/pytorch/pytorch/issues/45060#issuecomment-766332179

davidkyle (approved):
LGTM

PR #1694, opened 2021-01-25
[ML] Additional logging for the feature importance test failure

valeriy42:
This PR adds debug information that can be used next time elastic/elasticsearch#59413 fails. The changes will be removed once we have collected this information.


droberts195 (approved):
LGTM

To make this show up in the ES logs when the problem occurs it will be necessary to open an ES PR to change the log level for the `CppLogMessageHandler` class to `DEBUG` for the affected test.  Alternatively the log level could be set to `INFO` in this PR and then it will show up without an ES side change.  (But obviously then we need to be really careful to remove it before 7.12 is released.)

PR #1695, opened 2021-01-25
[ML] Unify CBoostedTreeTest in 7.x and master

valeriy42:
As I was backporting unit tests on `7.x`, I noticed that due to numerical properties for different platforms and compilers, the unit test checks diverged somewhat. This PR synchronizes the code of `CBoostedTreeTest.cc` for `master` and `7.x`.


przemekwitek (approved):
LGTM

PR #1696, opened 2021-01-25
[ML] Use Eigen BLAS library when building PyTorch for Linux

davidkyle:
Following https://github.com/elastic/ml-cpp/pull/1678#issuecomment-766327160 it is preferable to not have a dependency on libMKL for now at least. There are a number of alternative libraries that could be used in place of libMKL we have opted to use the header-only Eigen library. 

This uses the version of Eigen included in [PyTorch](https://github.com/pytorch/pytorch/tree/master/third_party) not the version in this repo.

Also sets the PyTorch library version number.

I've built and tested the docker image.


droberts195 (approved):
LGTM

PR #1697, opened 2021-01-26
[ML] Windows adjustments for PyTorch

droberts195:
1. The IO manager needs to open streams with the binary flag,
   otherwise the stream will stop reading binary files that
   contain an EOF character prematurely.
2. Upload the Windows build tools to S3 so CI can download
   them.
3. Adjust the test script so it uses regular files instead
   of named pipes.  Named pipes are completely different on
   Windows and hard to use from Python.  It's easier to make
   a Python script portable if it uses regular files.


bin/pytorch_inference/evaluate.py
```diff
@@ -1,181 +1,116 @@
...
+    parser.add_argument('--restore_file', default='restore_file')
+    parser.add_argument('--input_file', default='input_file')
+    parser.add_argument('--output_file', default='output_file')
 
     return parser.parse_args()
 
-
 def path_to_app():
 
-	os_platform = platform.system()
```
droberts195, line 40: I changed tabs to spaces in this file as it had a mix before - best to review with whitespace differences hidden

davidkyle (approved):
LGTM

PR #1699, opened 2021-01-28
[ML] Rename the early stopping configuration parameter

valeriy42:
We agreed that the parameter should be renamed to `early_stopping_enabled` for consistency. 

Within `CBoostedTreeImpl` I would still keep the differentiated set of flags to enable/disable early stopping in different stages. All this flags will be initialized in `CBoostedTreeFactory` using the value from `early_stopping_enabled` configuration from the API.

Relates #1676.

### DIFF
--- a/include/maths/CBoostedTreeFactory.h
+++ b/include/maths/CBoostedTreeFactory.h
@@ -261,7 +261,7 @@ private:
     //! Stubs out persistence.
     static void noopRecordTrainingState(CBoostedTree::TPersistFunc);
 
-    //! Stop hyperparameter optimization early if the the process is not promising.
+    //! Stop hyperparameter optimization early if the process is not promising.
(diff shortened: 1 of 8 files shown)


include/maths/CBoostedTreeFactory.h
```diff
@@ -261,6 +261,9 @@ class MATHS_EXPORT CBoostedTreeFactory final {
     //! Stubs out persistence.
     static void noopRecordTrainingState(CBoostedTree::TPersistFunc);
 
+    //! Stop hyperparameter optimization early if the the process is not promising.
```
przemekwitek, line 264:
```suggestion
    //! Stop hyperparameter optimization early if the process is not promising.
```

przemekwitek (approved):
LGTM

PR #1705, opened 2021-01-28
[7.x][ML] Rename the early stopping configuration parameter

valeriy42:
We agreed that the parameter should be renamed to early_stopping_enabled for consistency.

Within CBoostedTreeImpl I would still keep the differentiated set of flags to enable/disable early stopping in different stages. All this flags will be initialized in CBoostedTreeFactory using the value from early_stopping_enabled configuration from the API.

Backport of #1699.


valeriy42:
retest

PR #1707, opened 2021-01-28
[ML] Fix seasonal model update immediately after initialisation

tveasey:
#1675 introduced a subtle bug in the updating of newly initialised seasonal components. It switched to using state to track when it refreshes the seasonal model predictions to reflect new data (to correct behaviour when there are long gaps in the data). However, we test for seasonal components on a window which can sometimes extend beyond the current time (with future values set to "null"). As a result we were sometimes setting the time to refresh into the future and the model wasn't reflecting new data immediately after a seasonal component was added. The change has not been released so marking as a non-issue.

### DIFF
--- a/lib/maths/CSeasonalComponent.cc
+++ b/lib/maths/CSeasonalComponent.cc
@@ -13,6 +13,7 @@
 #include <core/Constants.h>
 #include <core/RestoreMacros.h>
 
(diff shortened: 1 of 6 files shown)


droberts195 (approved):
LGTM

PR #1712, opened 2021-01-29
[ML] Remove mkl dependency and improve docker cleanup

davidkyle:
Linux uses the Eigen BLAS library and macOS the Accelerate framework so MKL is not required.

This also fixes an intermittent problem building the Linux build image where the pytorch repo was not cleaned leaving a 23GB image. This fix takes the image size down to 5.4GB. 2GB of that is in `/usr/local/lib/python3.7` which could also be removed.


droberts195 (approved):
LGTM

PR #1714, opened 2021-02-01
[ML] Build libtorch for Linux aarch64

davidkyle:
The main difference between the x86 and aarch builds is that FBGEMM requires [AVX2](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#Advanced_Vector_Extensions_2) instructions. FBGEMM cannot be compiled on aarch64, Caffe2 is used instead. 

x86 builds set `BUILD_CAFFE2=OFF` but it is required here and left unset.  

I've built and run the smoketest `evaluate.py` script on native linux aarch64. 

## Multi-stage DockerFile
The process of building the libraries ml uses creates a lot of files that are not required in the final build image. The simple solution is to use a multi-stage build where the final image copies all the required dependencies built in the intermediate image. Primarily it is the contents of `/usr/local/gcc93` and `/usr/include` that must be copied and also some programs (make, git, etc) must then be installed in the final image.

There is one unexpected requirement: `/lib64/crti.o` and `/lib64/crtn.o` are needed to link the ml libs and these files are installed with `gcc` so `gcc` must be installed again even though we have just copied the contents of  `/usr/local/gcc93`. This may not be the optimal solution but it is the one I've tested.

### DIFF
--- a/dev-tools/docker/linux_image/Dockerfile
+++ b/dev-tools/docker/linux_image/Dockerfile
@@ -145,4 +145,4 @@ COPY --from=builder /usr/local/gcc93 /usr/local/gcc93
 COPY --from=builder /usr/include /usr/include
 RUN \
   rm /var/lib/rpm/__db.* && \
-  yum install -y bzip2 git make unzip which zip zlib-devel
+  yum install -y bzip2 gcc git make unzip which zip zlib-devel
--- a/dev-tools/docker/linux_image/Dockerfile
+++ b/dev-tools/docker/linux_image/Dockerfile
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
-FROM centos:7
+FROM centos:7 AS builder
 
 # This is basically automating the setup instructions in build-setup/linux.md
 
@@ -139,3 +139,10 @@ RUN \
   cp torch/lib/libtorch_cpu.so /usr/local/gcc93/lib && \
   cp torch/lib/libc10.so /usr/local/gcc93/lib && \
   cd .. && rm -rf ${build_dir}/*
+
+FROM centos:7
+COPY --from=builder /usr/local/gcc93 /usr/local/gcc93
+COPY --from=builder /usr/include /usr/include
+RUN \
+  rm /var/lib/rpm/__db.* && \
+  yum install -y bzip2 git make unzip which zip zlib-devel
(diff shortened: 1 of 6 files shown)


dev-tools/docker/linux_image/Dockerfile
```diff
@@ -133,10 +133,13 @@ RUN \
...
   mkdir /usr/local/gcc93/include/pytorch && \
   cp -r torch/include/* /usr/local/gcc93/include/pytorch/ && \
   cp torch/lib/libtorch_cpu.so /usr/local/gcc93/lib && \
   cp torch/lib/libc10.so /usr/local/gcc93/lib && \
-  rm -rf ${build_dir}/*
+  cd .. && rm -rf ${build_dir}/*
+
+FROM centos:7
+COPY --from=0 /usr/local/gcc93/lib /usr/local/gcc93/lib
+COPY --from=0 /usr/local/gcc93/include /usr/local/gcc93/include
```
droberts195, line 145: You'll need more than this.  The compiler itself is in `/usr/local/gcc93/bin` and `/usr/local/gcc93/lib64`.  It's probably best to just copy the whole `gcc93` directory structure:

```suggestion
COPY --from=0 /usr/local/gcc93 /usr/local/gcc93
```

Same for aarch64.

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -133,10 +133,13 @@ RUN \
...
-  cd ${build_dir}/pytorch && \
   /usr/local/bin/python3.7 setup.py install && \
   mkdir /usr/local/gcc93/include/pytorch && \
   cp -r torch/include/* /usr/local/gcc93/include/pytorch/ && \
   cp torch/lib/libtorch_cpu.so /usr/local/gcc93/lib && \
   cp torch/lib/libc10.so /usr/local/gcc93/lib && \
-  rm -rf ${build_dir}/*
+  cd .. && rm -rf ${build_dir}/*
+
+FROM centos:7
```
droberts195, line 143: Also, we probably need some of the optional packages that got installed by the `yum` command.

The build steps require these:

```
bzip2 gcc gcc-c++ git libffi-devel make openssl-devel texinfo unzip wget which zip zlib-devel
```

At a guess, the product build would require these:

```
bzip2 git make unzip which zip zlib-devel
```

droberts195 (approved):
LGTM

PR #1716, opened 2021-02-01
[ML] Use the clang flag to ignore unknown warning options

droberts195:
New versions of clang often bring new warnings that trigger
repetitively on our codebase and need to be silenced.  But
then older versions of clang complain about the options that
silence the new warnings.

This commit adds the flag that tells the older versions of
clang to ignore warning options they don't understand.  This
works nicely with our approach of enabling all warnings and
then disabling the ones that cause useless noise.


przemekwitek (approved):
LGTM

PR #1717, opened 2021-02-01
[ML] Check invariants after state restoration

edsavage:
Abort early if expected object invariants do not hold true after state
restoration.

Provide new macros to check violation of invariants and to abort with a
clear, consistently formatted messages.

This commit follows an audit of the maths library only. More to follow.

### DIFF
--- a/lib/maths/CQDigest.cc
+++ b/lib/maths/CQDigest.cc
@@ -73,7 +73,13 @@ bool CQDigest::acceptRestoreTraverser(core::CStateRestoreTraverser& traverser) {
         }
     } while (traverser.next());
 
+    this->checkRestoredInvariants();
+
     return true;
+}
+
+void CQDigest::checkRestoredInvariants() const {
+    VIOLATES_INVARIANT_NO_EVALUATION(m_Root, ==, nullptr);
 }
 
 void CQDigest::add(uint32_t value, uint64_t n) {
@@ -103,7 +109,7 @@ void CQDigest::add(uint32_t value, uint64_t n) {
         this->compress(compress);
     }
 
-    //this->checkInvariants();
+    this->checkInvariants();
 }
 
 void CQDigest::merge(const CQDigest& digest) {
@@ -123,7 +129,7 @@ void CQDigest::merge(const CQDigest& digest) {
     // Compress the whole tree.
     this->compress();
 
-    //this->checkInvariants();
+    this->checkInvariants();
 }
 
 void CQDigest::propagateForwardsByTime(double time) {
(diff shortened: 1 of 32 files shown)


lib/maths/CQDigest.cc
```diff
@@ -103,7 +109,7 @@ void CQDigest::add(uint32_t value, uint64_t n) {
         this->compress(compress);
     }
 
-    //this->checkInvariants();
+    this->checkInvariants();
```
droberts195, line 112: I don't think it's necessary to check here, as the input to this method didn't come from user input.

The check was previously commented out due to the performance impact.  Uncommenting it doesn't just do the check once per run like it would after restoring from persisted state, but potentially millions of times.

lib/maths/CQDigest.cc
```diff
@@ -123,7 +129,7 @@ void CQDigest::merge(const CQDigest& digest) {
     // Compress the whole tree.
     this->compress();
 
-    //this->checkInvariants();
+    this->checkInvariants();
```
droberts195, line 132: I don't think it's necessary to check here, as the input to this method didn't come from user input.

The check was previously commented out due to the performance impact.  Uncommenting it doesn't just do the check once per run like it would after restoring from persisted state, but potentially millions of times.

lib/maths/CQDigest.cc
```diff
@@ -73,9 +73,15 @@ bool CQDigest::acceptRestoreTraverser(core::CStateRestoreTraverser& traverser) {
...
     } while (traverser.next());
 
+    this->checkRestoredInvariants();
+
     return true;
 }
 
+void CQDigest::checkRestoredInvariants() const {
+    VIOLATES_INVARIANT_NO_EVALUATION(m_Root, ==, nullptr);
+}
```
droberts195, line 83: Maybe this can also call the pre-existing `this->checkInvariants();` and abort if it returns false.

lib/maths/CQDigest.cc
```diff
@@ -73,9 +73,16 @@ bool CQDigest::acceptRestoreTraverser(core::CStateRestoreTraverser& traverser) {
...
     } while (traverser.next());
 
+    this->checkRestoredInvariants();
+
     return true;
 }
 
+void CQDigest::checkRestoredInvariants() const {
+    VIOLATES_INVARIANT_NO_EVALUATION(m_Root, ==, nullptr);
+    this->checkInvariants();
```
droberts195, line 83:
```suggestion
    if (this->checkInvariants() == false) {
        LOG_ABORT(<< "Invariance check failed for Q digest");
    }
```

droberts195 (approved):
I'm happy to merge this if you could just make the one extra suggested change.

Without that the Q digest invariant check will just print an error, but won't abort.

PR #1725, opened 2021-02-05
[ML] Allow a user to override weights for assigning labels to classes

### LINKED ISSUE #1719: [ML] Add the ability to override the threshold at which we assign a class for binary classification
We would like to allow users to set the decision threshold for assigning a class by examining the ROC curve. In order to do this we first need to support supplying this parameter in the configuration of a classification job.

tveasey:
We plan to allow users to also set weights for class assignment via choosing a point on the ROC curve. In this case, the config should allow overriding class weights to reflect the user's error preferences if retraining. This introduces a new value in the parameters section of the JSON object of the form
```
"classification_weights": [{"class":"<string>", "weight": <double>}]
```
and a new class assignment objective `custom`.

Closes #1719. I've marked this as a non-issue because it isn't really user facing and we should only document when the Java side of the changes to support overriding the decision threshold is implemented.

### DIFF
--- a/lib/test/CDataFrameAnalysisSpecificationFactory.cc
+++ b/lib/test/CDataFrameAnalysisSpecificationFactory.cc
@@ -206,6 +206,12 @@ CDataFrameAnalysisSpecificationFactory::predictionFieldType(const std::string& t
 }
 
 CDataFrameAnalysisSpecificationFactory&
+CDataFrameAnalysisSpecificationFactory::classificationWeights(const TStrDoublePrVec& weights) {
+    m_ClassificationWeights = weights;
+    return *this;
+}
+
+CDataFrameAnalysisSpecificationFactory&
 CDataFrameAnalysisSpecificationFactory::regressionLossFunction(TLossFunctionType lossFunction) {
     m_RegressionLossFunction = lossFunction;
     return *this;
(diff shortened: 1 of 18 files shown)


tveasey:
retest

tveasey:
retest

tveasey:
retest

lib/test/CDataFrameAnalysisSpecificationFactory.cc
```diff
@@ -272,97 +278,116 @@ std::string
...
+    if (analysis == classification()) {
+        writer.Key(TClassificationRunner::NUM_CLASSES);
+        writer.Uint64(m_NumberClasses);
+        writer.Key(TClassificationRunner::NUM_TOP_CLASSES);
+        writer.Uint64(m_NumberTopClasses);
+        if (m_PredictionFieldType.empty() == false) {
+            writer.Key(TClassificationRunner::PREDICTION_FIELD_TYPE);
+            writer.String(m_PredictionFieldType);
+        }
+        if (m_ClassificationWeights.size() > 0) {
```
valeriy42, line 362:
```suggestion
        if (m_ClassificationWeights..empty() == false) {
```
nit: for consistency with the code above and also to prevent clang-tidy readability complaints.

valeriy42 (approved):
LGTM

lib/test/CDataFrameAnalysisSpecificationFactory.cc
tveasey: Agreed. Done in [7809056](https://github.com/elastic/ml-cpp/pull/1725/commits/7809056cdf48bc2ddd5db8148da3e33d75cb4c39).

PR #1733, opened 2021-02-09
[ML] Use a per tree feature bag for boosted tree training

tveasey:
We already downsample candidate split features at node. However, because we can use different feature bags along a branch, cache split derivatives and use the parent and sibling derivatives to compute them for the child with more training examples, we compute them for all features. If we were to use a single feature bag for the entire tree then we could just compute these.

This change introduces a tree wide feature bag which contains the node feature bag. I explored replacing the per node feature bag altogether, but found evidence that enforcing some diversity in selected features by sampling within a tree was beneficial. However, I didn't find any evidence to warrant an extra hyperparameter and simply use a slightly, 1.5x times, larger feature bag for the tree.

There is some additional work which could be saved and it might make sense from a cache performance perspective to restrict the `CBoostedTreeLeafNodeStatistics::CSplitsDerivatives` to only store tree feature bag derivatives. However, this complicates the change and as it is it tackles the hottest path which is `CBoostedTreeLeafNodeStatistics::addRowDerivatives`. I get a 10-20% performance improvement for normal cases  and importantly we also get the benefit that we can scale better to large numbers of features.

### DIFF
--- a/include/maths/CBoostedTreeLeafNodeStatistics.h
+++ b/include/maths/CBoostedTreeLeafNodeStatistics.h
@@ -605,7 +605,8 @@ public:
                                    const CDataFrameCategoryEncoder& encoder,
                                    const TRegularization& regularization,
                                    const TImmutableRadixSetVec& candidateSplits,
-                                   const TSizeVec& featureBag,
+                                   const TSizeVec& treeFeatureBag,
+                                   const TSizeVec& nodeFeatureBag,
                                    std::size_t depth,
                                    const core::CPackedBitVector& rowMask,
                                    CWorkspace& workspace);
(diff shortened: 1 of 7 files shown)


include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -740,8 +737,6 @@ class MATHS_EXPORT CBoostedTreeLeafNodeStatistics final {
         std::size_t s_Feature = std::numeric_limits<std::size_t>::max();
         double s_SplitAt = boosted_tree_detail::INF;
         std::uint32_t s_MinimumChildRowCount = 0;
-        std::uint32_t s_LeftChildRowCount = 0; // TODO remove after stats measurement
-        std::uint32_t s_RightChildRowCount = 0; //  TODO remove after stats measurement
```
valeriy42, line 744: Thanks for cleaning this up :pray:

valeriy42 (approved):
LGTM and thank you for cleaning up the remainings from the stats measurements.

PR #1737, opened 2021-02-10
[ML] Check invariants after restoration

edsavage:
Add checking of invariants after restoration of several classes in the
model library.

Also change a number of conditions where an error was emitted during the
restoration process to be hard stops with LOG_ABORT.

### DIFF
--- a/lib/model/CDataGatherer.cc
+++ b/lib/model/CDataGatherer.cc
@@ -778,7 +778,8 @@ bool CDataGatherer::acceptRestoreTraverser(const std::string& summaryCountFieldN
             int feature(-1);
             if (core::CStringUtils::stringToType(traverser.value(), feature) == false ||
                 feature < 0) {
-                LOG_ABORT(<< "Invalid feature in " << traverser.value());
+                LOG_ERROR(<< "Invalid feature in " << traverser.value());
+                return false;
             }
             m_Features.push_back(static_cast<model_t::EFeature>(feature));
             continue;
@@ -819,14 +820,16 @@ bool CDataGatherer::restoreBucketGatherer(const std::string& summaryCountFieldNa
                 *this, summaryCountFieldName, personFieldName, attributeFieldName,
                 valueFieldName, influenceFieldNames, traverser);
             if (m_BucketGatherer == nullptr) {
-                LOG_ABORT(<< "Failed to create event rate bucket gatherer");
+                LOG_ERROR(<< "Failed to create event rate bucket gatherer");
+                return false;
             }
         } else if (name == CBucketGatherer::METRIC_BUCKET_GATHERER_TAG) {
             m_BucketGatherer = std::make_unique<CMetricBucketGatherer>(
                 *this, summaryCountFieldName, personFieldName, attributeFieldName,
                 valueFieldName, influenceFieldNames, traverser);
             if (m_BucketGatherer == nullptr) {
-                LOG_ABORT(<< "Failed to create metric bucket gatherer");
+                LOG_ERROR(<< "Failed to create metric bucket gatherer");
+                return false;
             }
         }
     } while (traverser.next());
--- a/lib/model/CDataGatherer.cc
+++ b/lib/model/CDataGatherer.cc
@@ -778,8 +778,7 @@ bool CDataGatherer::acceptRestoreTraverser(const std::string& summaryCountFieldN
             int feature(-1);
             if (core::CStringUtils::stringToType(traverser.value(), feature) == false ||
                 feature < 0) {
-                LOG_ERROR(<< "Invalid feature in " << traverser.value());
-                return false;
+                LOG_ABORT(<< "Invalid feature in " << traverser.value());
             }
             m_Features.push_back(static_cast<model_t::EFeature>(feature));
             continue;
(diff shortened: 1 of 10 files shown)


lib/model/CDataGatherer.cc
```diff
@@ -820,16 +819,14 @@ bool CDataGatherer::restoreBucketGatherer(const std::string& summaryCountFieldNa
...
+                LOG_ABORT(<< "Failed to create event rate bucket gatherer");
             }
         } else if (name == CBucketGatherer::METRIC_BUCKET_GATHERER_TAG) {
             m_BucketGatherer = std::make_unique<CMetricBucketGatherer>(
                 *this, summaryCountFieldName, personFieldName, attributeFieldName,
                 valueFieldName, influenceFieldNames, traverser);
             if (m_BucketGatherer == nullptr) {
-                LOG_ERROR(<< "Failed to create metric bucket gatherer");
-                return false;
+                LOG_ABORT(<< "Failed to create metric bucket gatherer");
```
tveasey, line 829: I'm not convinced that this is correct. `std::make_unique...` isn't normally going to be returning null and we're missing the case that the state doesn't have a `EVENTRATE_BUCKET_GATHERER_TAG` or a `METRIC_BUCKET_GATHERER_TAG`. I'd just have a check that it's non-null after the loop.

tveasey (commented):
I'm slightly uneasy about some of the new aborts. I think we need to decide about whether each possible case is recoverable or not and if it is whether we want to try and recover from them. I think it would be worth discussing this a bit offline.

lib/model/CEventRateBucketGatherer.cc
```diff
@@ -223,23 +223,20 @@ bool restoreAttributePeopleData(core::CStateRestoreTraverser& traverser, TSizeUS
...
+            if (seenCid == false) {
+                LOG_ABORT(<< "Incorrect format - person ID before attribute ID in "
                           << traverser.value());
-                return false;
             }
             std::size_t pid = 0;
             if (core::CStringUtils::stringToType(traverser.value(), pid) == false) {
-                LOG_ERROR(<< "Invalid person ID in " << traverser.value());
-                return false;
+                LOG_ABORT(<< "Invalid person ID in " << traverser.value());
```
tveasey, line 239: I don't think this needs an abort at present

lib/model/CEventRateBucketGatherer.cc
```diff
@@ -223,23 +223,20 @@ bool restoreAttributePeopleData(core::CStateRestoreTraverser& traverser, TSizeUS
...
+                LOG_ABORT(<< "Invalid attribute ID in " << traverser.value());
             }
             seenCid = true;
             if (lastCid >= data.size()) {
                 data.resize(lastCid + 1);
             }
         } else if (name == PERSON_TAG) {
-            if (!seenCid) {
-                LOG_ERROR(<< "Incorrect format - person ID before attribute ID in "
+            if (seenCid == false) {
```
tveasey, line 233: This should be checking `lastCid < data.size()` as well.

lib/model/CEventRateBucketGatherer.cc
```diff
@@ -685,8 +680,7 @@ bool restoreInfluencerUniqueStrings(core::CStateRestoreTraverser& traverser,
         } else if (name == UNIQUE_WORD_TAG) {
             CUniqueStringFeatureData::TWord value;
             if (value.fromDelimited(traverser.value()) == false) {
-                LOG_ERROR(<< "Failed to restore word " << traverser.value());
-                return false;
+                LOG_ABORT(<< "Failed to restore word " << traverser.value());
```
tveasey, line 683: I wouldn't make this an abort

lib/model/CMetricBucketGatherer.cc
```diff
@@ -1002,74 +1003,62 @@ bool CMetricBucketGatherer::acceptRestoreTraverserInternal(core::CStateRestoreTr
     if (name == MEAN_TAG) {
         CRestoreFeatureData<model_t::E_Mean> restore;
         if (restore(traverser, 1, isCurrentVersion, *this, m_FeatureData) == false) {
-            LOG_ERROR(<< "Invalid mean data in " << traverser.value());
-            return false;
+            LOG_ABORT(<< "Invalid mean data in " << traverser.value());
```
tveasey, line 1006: I don't think any of these are aborts

lib/model/CTokenListCategory.cc
```diff
@@ -93,91 +94,85 @@ bool CTokenListCategory::acceptRestoreTraverser(core::CStateRestoreTraverser& tr
...
+                LOG_ABORT(<< "Invalid base raw string length in " << traverser.value());
             }
         }
     } while (traverser.next());
 
+    // Verify that m_CommonUniqueTokenIds is sorted in ascending order.
+    bool isSortedAsc = std::is_sorted(
+        m_CommonUniqueTokenIds.begin(), m_CommonUniqueTokenIds.end(),
+        [](const auto& a, const auto& b) { return a.first < b.first; });
+    VIOLATES_INVARIANT(isSortedAsc, ==, false);
```
tveasey, line 174: Let's just sort the collection here if this is an important invariant.

tveasey (commented):
Left some more concrete suggestions

lib/model/CEventRateBucketGatherer.cc
edsavage: > This should be checking `lastCid < data.size()` as well.

I think that this check is already covered by

```
            if (lastCid >= data.size()) {
                data.resize(lastCid + 1);
            }
```
above

lib/model/CMetricBucketGatherer.cc
```diff
@@ -1057,7 +1057,6 @@ bool CMetricBucketGatherer::acceptRestoreTraverserInternal(core::CStateRestoreTr
         CRestoreFeatureData<model_t::E_MultivariateMin> restore;
         if (restore(traverser, dimension, isCurrentVersion, *this, m_FeatureData) == false) {
             LOG_ERROR(<< "Invalid multivariate min data in " << traverser.value());
-            return false;
```
droberts195, line 1060: I think this line should be reverted too since it's the only change in this file now.

droberts195 (approved):
LGTM, but please wait until the 7.12 branch has been created before merging the 7.x backport PR

lib/model/CEventRateBucketGatherer.cc
tveasey: Yes, good point.

lib/model/CTokenListCategory.cc
```diff
@@ -178,6 +178,10 @@ bool CTokenListCategory::acceptRestoreTraverser(core::CStateRestoreTraverser& tr
         }
     } while (traverser.next());
 
+    // Ensure that m_CommonUniqueTokenIds is sorted in ascending order.
+    std::sort(m_CommonUniqueTokenIds.begin(), m_CommonUniqueTokenIds.end(),
+              [](const auto& a, const auto& b) { return a.first < b.first; });
```
tveasey, line 183: Currently this is the same but presumably if `CTokenIdLess` changed this would need to too:
```suggestion
             CTokenIdLess{});
```

tveasey (approved):
Just one minor point, but basically LGTM now

edsavage:
retest

edsavage:
CI build failed due to known issue with the new macOS on aarch64 worker.

PR #1739, opened 2021-02-11
[ML] Add pytorch_inference to Makefile and controller

### LINKED ISSUE #1702: [ML] [PyTorch] Add ability for the controller process to spawn the modelling process
Make the native controller aware of the `pytorch_inference` app and add ability to spawn the process.

droberts195:
Now we have built libraries for all the different platforms
pytorch_inference can be added to the Makefile without
breaking CI.

If CI fails for this PR it will smoke out any mistakes in
Docker images or dependency bundles.

Also adding to controller as that's a one-liner.

Closes #1702

### DIFF
--- a/bin/pytorch_inference/.gitignore
+++ b/bin/pytorch_inference/.gitignore
@@ -1,0 +1,3 @@
+input_file
+output_file
+restore_file
--- a/bin/Makefile
+++ b/bin/Makefile
@@ -16,6 +16,7 @@ COMPONENTS= \
             controller \
             data_frame_analyzer \
             normalize \
+            pytorch_inference \
 
 include $(CPP_SRC_HOME)/mk/toplevel.mk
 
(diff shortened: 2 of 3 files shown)


bin/pytorch_inference/.gitignore
```diff
@@ -0,0 +1,3 @@
+input_file
```
davidkyle, line 1: If `evaluate.py` runs successfully these files should be cleaned up. 

I'll raise a change to always cleanup these files as the actual names are arguments to the script anyway. Happy to merge this as is I will revert this file when I make the PR

davidkyle (approved):
LGTM

I expect we will rename `pytorch_inference` at some point before beta

droberts195:
The failure of the first CI run was a transient download failure in Gradle, so we didn't get to find out if these changes work on Windows 😿

droberts195:
retest

droberts195:
🎉 green CI, so we can build programs that link to libtorch on all platforms 🚀

PR #1740, opened 2021-02-11
[ML] Delete temp files created by evaluate.py

davidkyle:
Running the `evaluate.py` smoketest script leaves files behind that can be cleaned up.

The diff makes this change look much bigger than it is, I wrapped the code in a `try.. finally` statement and indented it. The only non-whitespace changes are the try keyword and finally section.


droberts195 (approved):
LGTM

PR #1746, opened 2021-02-17
[ML] Line search feature bag fraction for classification and regression model training

tveasey:
Following on from #1733, we can get further speedups by line searching for the best feature bag fraction for data sets where we only need a fraction of the features per tree. For example, training time on Higgs 1M drops from 2585s to 1742s and we actually get a small improvement in accuracy because our hyperparameter search region is better initialised.

This makes three changes:
1. Adds a line search for the best initial feature bag fraction to use.
2. Adds a small linear penalty at most 1% minimum loss to encourage larger down sample factors and smaller feature bag fractions.
3. Handles better the case we have many features and relatively few training examples.

Some of the variable naming in `CBoostedTreeFactory` is also now misleading. I've split this change out into a non-functional commit. The functional commit is [f9eacc2](https://github.com/elastic/ml-cpp/commit/f9eacc23559e5b921fb421b1f7a61cc6dd2668d1).

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -39,6 +39,13 @@
 * The Linux build platform for the {ml} C++ code is now CentOS 7 running gcc 9.3. (See
   {ml-pull}1170[#1170].)
 
+== {es} version 7.13.0
+
+=== Enhancements
+
+* Speed up training of regression and classification model training for data sets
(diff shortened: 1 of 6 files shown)


valeriy42 (approved):
LGTM

tveasey:
CI on macos is expected to fail at the moment so I'll go ahead and merge.

PR #1749, opened 2021-02-18
[ML] Do nothing on macOS ARM workers on old branches

droberts195:
PR builds will now fan out to workers that include macOS ARM
workers. But for old branches we cannot build on this architecture,
so just make such builds a no-op.

The approach is similar to #1203, and because the 6.8 branch
already contains #1203 this change is only needed for 7.11 and
7.12.


droberts195:
retest

edsavage (approved):
LGTM

PR #1751, opened 2021-02-19
[ML] Replace BOOST_REQUIRE with BOOST_TEST_REQUIRE

droberts195:
BOOST_TEST_REQUIRE is better than BOOST_REQUIRE for assertions
that contain a single comparison operator and where both the
left hand side and right hand side of it are printable.  The
advantage is that on failure it prints the values that failed
the test.


tveasey (approved):
LGTM

droberts195:
Only CI failure is for macOS ARM

PR #1755, opened 2021-02-19
[ML] Scale regularisers for final train

tveasey:
As we move towards training for hyperparameter tuning on a small fraction of the data set and final training on more we will suffer issues with overfitting if we don't address the bias this introduces estimating regularisers. Interestingly, we already see a mismatch in train and test errors on larger data sets where we only use two-folds. I tested this correction, which is the one we use when we downsample, on a variety of data sets and we ended up with lower mismatch between train and test errors.


valeriy42 (approved):
LGTM

droberts195:
This is labelled `v7.13.0` but is not yet backported to the 7.x branch - please don't forget!

PR #1745, opened 2021-02-16
[ML] Adding CI for macOS on ARM

droberts195:
This change adds coverage for macOS on ARM to CI.

There are 3 parts to this:

1. Download the pre-built 3rd party dependencies at the
   beginning of the build.  Unlike Linux builds which can
   get these dependencies by building in a Docker container,
   the macOS CI machines need to get them via a more
   primitive mechanism.
2. Adjust the Jenkins CI shell script to account for the
   possibility of having to build natively on macOS as well
   as in Docker on Linux.
3. Switch to the newer S3 API from Amazon.  The old S3 API
   didn't work in Java 9 or above, which we worked around by
   ensuring Java 8 was installed on the CI workers.  But
   that's not possible for macOS on ARM, so it's finally
   forced an upgrade.

PR #1758, opened 2021-02-22
[ML] Further increase tolerance for logger throttler test

droberts195:
In the backport of #1745 the tolerance on the logger
throttler test had to be increased again due to a
problem with sleep calls sleeping for longer than
specified when running under Jenkins on macOS on ARM.
The change was in the 3rd commit on #1752.

This change makes the same increase on the master
branch, to keep the branches in sync.


edsavage (approved):
LGTM

PR #1759, opened 2021-02-22
[ML] Correct the macOS ARM dependency bundle for 7.x

droberts195:
When #1745 was backported to the 7.x branch in #1752
the version of the dependency bundle should have been
adjusted to the one that includes the correct dependencies
for the 7.x branch, but this step was missed. Although
the master dependency bundle works, the correct bundle
should be used for consistency across the whole 7.x
distribution.


edsavage (approved):

edsavage:
LGTM

PR #1760, opened 2021-02-22
[ML] Increase tolerance for read/write lock test

droberts195:
In #1759 the tolerance on the read/write lock test
had to be increased again due to a problem with sleep
calls sleeping for longer than specified when running
under Jenkins on macOS on ARM.  The change was in the
2nd commit on #1759.

This change makes the same increase on the master
branch, to keep the branches in sync.


edsavage (approved):
LGTM

PR #1768, opened 2021-02-23
[ML] Upgrade to Gradle 6.8.3

droberts195:
Soon we will upgrade to version 7, and being on the
latest in the 6.x series reduces the risk of missing
a deprecation warning.


edsavage (approved):
LGTM

PR #1770, opened 2021-02-23
[ML] PyTorch Command Processor 

### LINKED ISSUE #1700: [ML] [PyTorch] Communications with between ES and the native process
## Request ID 
The design has to accommodate multiple concurrent inference requests to the server and a mechanism to tie a specific request to a model output is required. This could be inferred from the processing order which is strictly FIFO but adding a ID token to each request provides additional context for development and debugging. The token has no semantics and is purely passed through the C++. Anomaly Detector flushID is the prior art here. 

## Payload
The inference payload is a series of numeric tokens. An individual inference request will consist of the request ID, the payload tokens and a marker to delineate each request.

Anomaly Detection uses a concise length encoded binary protocol because of the high volume of data sent across pipes. Compared with Anomaly Detection the input is small so a more verbose input format can be used which has the advantage of being descriptive. 

## Input Format
A JSON document:
```
{
  “request_id” : “string”,  
  “token_ids” : [int, int,...],
  “attention_mask” : [int, int,...],
  “token_type_ids” : [int, int,...],
  “position_ids” : [int, int,...]
}
```

`token_ids` and `attention_mask` are required for all uses, `token_type_ids` and `position_ids` are optional depending on the model type.

## Output Format
A JSON document for flexibility containing the request ID token, the result tensor and optionally the predicted tokens depending on the model type:
```
{
  “request_id” : “string”,  
  “predictions” : [float, float,...],
  “tokens” : [int, int,...]
}
```

### LINKED ISSUE #1701: [ML] [PyTorch] Create command processor for inference app
The current implementation (#1638) and the test script perform single shot inference then exit, a production version must be capable of loading a model and completing multiple inference calls before exiting.  

The first iteration will parse the input defined in #1700 and called evaluate on the model, there not be any explicit control commands in the first instance but by structuring the code in this way it will be easier to add commands in the future. The app will exit when the input pipe is closed the same as `autodetect`. 

I see the command processor code living in the `bin/pytorch_inference` rather than adding it to an existing library or creating a new library.

davidkyle:
Defines the input and output documents for the PyTorch 3rd party model app and adds a command processor which parses JSON documents from a stream then calls a handler function for each request. This all happens in a single thread, the output will be written before the next request document is parsed. 

### Input
Models accept a variable number of arguments depending the purpose. In Python PyTorch these are named arguments, in LibTorch an array of input tensors is used. All BERT models take a list of tokens, the other parameters are passed in the fields `arg_1`, `arg_2` etc. This program knows nothing about the expected number of arguments it simply consumes all fields starting with `arg_` and forwards them to the model.

```
{
  "request_id": "foo", 
  "tokens": [101, 20164, 10932],
  "arg_1": [1, 1, 1],
  "arg_2": [0, 0, 0]
}
```

RapidJSON supports reading multiple documents from a stream if the `kParseStopWhenDoneFlag` flag is used. The docs don't have to have a common root (e.g. in an array or nested inside a _wrapper_ object). Docs can optionally be separated by whitespace but any other separator is invalid.  

### Input Validation 
The input will come from Elasticsearch never from a client, we control the comms with Elasticsearch so minimal validation is required. If the request is not correctly formed then something catastrophic has happened (broken pipe). 

Typically the model throws a `std::runtime_error` if the input is not right, this is caught and returned to the  caller.

### Output
All BERT models output a tuple the first element of which is the output tensor. The remaining elements are model dependent (might be logits or labels) we have not found a use case requiring the full tuple yet so the output response will only contain the tensor (for now). The tensor must have 2 dimensions or be reducible to 2 dimensions. 

The output is a JSON document containing the tensor as an array of arrays.

```
{
  "request_id": "foo",
  "inference": [[9.265243530273438,-2.515533685684204,-1.3969738483428956,],[1.2384196519851685,-1.9405957460403443,-1.775153398513794]]
}
```

In the case of an error the output doc has an `error` field. It is envisaged any errors will be returned directly to the client.
```
{
  "request_id": "foo",
  "error": "what went wrong"
}
```

### Tests
The inputs to `evaluate.py` have been reworked so that a single JSON file contains both the input and expected output. Invoking the test and adding new examples is now much easier:

```
python3 evaluate.py /path/todistilbert-base-uncased-finetuned-sst-2-english.pt examples/sentiment_analysis/test_run.json
python3 evaluate.py /path/to/conll03_traced_ner.pt examples/ner/test_run.json
```

Closes #1700 
Closes #1701


davidkyle:
The macOS aarch64 build failed because PyTorch has not been setup on the machine

```
In file included from Main.cc:15:
./CBufferedIStreamAdapter.h:10:10: fatal error: 'caffe2/serialize/read_adapter_interface.h' file not found
#include <caffe2/serialize/read_adapter_interface.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
gnumake[2]: *** [.objs/Main.o] Error 1
gnumake[1]: *** [pytorch_inference/targetdirectory] Error 2
gnumake[1]: *** Waiting for unfinished jobs....
```

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+        varArgName = VAR_ARG_PREFIX + std::to_string(varCount);
+    }
+
+    return true;
+}
+
+CCommandParser::SRequest CCommandParser::jsonToRequest(const rapidjson::Document& doc) const {
+    TUint32Vec tokens;
+    const rapidjson::Value& arr = doc[TOKENS];
+    for (auto itr = arr.Begin(); itr != arr.End(); ++itr) {
```
droberts195, line 111: You should be able to reserve the vector before the loop to prevent repeated reallocations:

```
   tokens.reserve(arr.Size());
```

Same for the `arg` vector in the loop below.

bin/pytorch_inference/CCommandParser.h
```diff
@@ -0,0 +1,72 @@
...
+
+#ifndef INCLUDED_ml_torch_CCommandParser_h
+#define INCLUDED_ml_torch_CCommandParser_h
+
+#include <functional>
+#include <iosfwd>
+#include <string>
+#include <vector>
+
+#include <rapidjson/document.h>
```
droberts195, line 15: nit: move before the standard headers

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
 
 #include <rapidjson/ostreamwrapper.h>
 #include <torch/script.h>
 
-#include <boost/optional.hpp>
-
 #include <memory>
 #include <string>
 
-// For ntohl
```
droberts195, line 30: Removing this means you can also remove the `USE_NET` in `Makefile`.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
+#include <api/CIoManager.h>
```
droberts195, line 7: We usually put our headers in the dependency order of the libraries, so core first, api last.  clang-format has changed this because you deleted the blank lines in between.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
+                     const std::string& requestId,
+                     std::ostream& outputStream) {
 
-boost::optional<TFloatVec> readTokens(std::istream& inputStream) {
-    if (inputStream.eof()) {
-        LOG_ERROR(<< "Unexpected end of stream reading tokens");
-        return boost::none;
+    torch::Tensor view;
+    auto sizes = prediction.sizes();
+    if (sizes.size() == 3 && sizes[0] == 1) {
```
droberts195, line 61: This could do with a comment to say what this special case represents.

bin/pytorch_inference/Makefile
```diff
@@ -25,6 +25,7 @@ SRCS= \
     Main.cc \
     CBufferedIStreamAdapter.cc \
     CCmdLineParser.cc \
+    CCommandParser.cc \
 
 NO_TEST_CASES=1
```
droberts195, line 30: This line should be removed so that `make test` will run the unit tests.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
-    inputs.push_back(torch::arange(static_cast<std::int64_t>(data.size())).to(torch::kInt64)); // position ids
+
+    for (auto args : request.s_SecondaryArguments) {
+        torch::Tensor tensor =
+            torch::from_blob(static_cast<void*>(args.data()),
+                             {1, static_cast<std::int64_t>(args.size())},
+                             at::dtype(torch::kInt32))
+                .to(torch::kInt64);
+
+        inputs.push_back(tensor);
```
droberts195, line 47: This is copying the tensors at the moment.  It may be possible to emplace:

```suggestion
        inputs.emplace_back(
            torch::from_blob(static_cast<void*>(args.data()),
                             {1, static_cast<std::int64_t>(args.size())},
                             at::dtype(torch::kInt32))
                .to(torch::kInt64));
```

Or if that doesn't work for some reason, you could at least move when adding, i.e. `inputs.push_back(std::move(tensor));`.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+        tokens.push_back(itr->GetUint());
+    }
+
+    std::uint64_t varCount{1};
+    std::string varArgName = VAR_ARG_PREFIX + std::to_string(varCount);
+    TUint32VecVec args;
+
+    while (doc.HasMember(varArgName)) {
+        TUint32Vec arg;
```
tveasey, line 120: Probably worth pulling this out of the loop and clearing so you aren't constantly reallocating.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
-torch::Tensor infer(torch::jit::script::Module& module, TFloatVec& data) {
     torch::Tensor tokensTensor =
-        torch::from_blob(data.data(), {1, static_cast<std::int64_t>(data.size())})
+        torch::from_blob(static_cast<void*>(request.s_Tokens.data()),
+                         {1, static_cast<std::int64_t>(request.s_Tokens.size())},
+                         at::dtype(torch::kInt32))
             .to(torch::kInt64);
+
     std::vector<torch::jit::IValue> inputs;
     inputs.push_back(tokensTensor);
```
droberts195, line 38: As below, it looks like this is copying the tensor.  If it compiles this should be more efficient:

```
    std::vector<torch::jit::IValue> inputs;
    inputs.emplace_back(
        torch::from_blob(static_cast<void*>(request.s_Tokens.data()),
                          {1, static_cast<std::int64_t>(request.s_Tokens.size())},
                          at::dtype(torch::kInt32))
             .to(torch::kInt64));
```

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+
+        ++varCount;
+        varArgName = VAR_ARG_PREFIX + std::to_string(varCount);
+    }
+
+    return true;
+}
+
+CCommandParser::SRequest CCommandParser::jsonToRequest(const rapidjson::Document& doc) const {
+    TUint32Vec tokens;
```
tveasey, line 109: All the vectors in this function are reservable. But if you expect to process a stream of JSON it is probably better to make them members and clear before use so you aren't repeatedly reallocating.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
 #include <string>
 
-// For ntohl
-#ifdef Windows
-#include <WinSock2.h>
-#else
-#include <netinet/in.h>
-#endif
+static const std::string INFERENCE{"inference"};
+static const std::string ERROR{"error"};
```
droberts195, line 26: The modern way is to put file scope constants in an unnamed namespace instead of using `static`.

```suggestion
namespace {
const std::string INFERENCE{"inference"};
const std::string ERROR{"error"};
}
```

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+namespace ml {
+namespace torch {
+
+namespace {
+void debug(const rapidjson::Document& doc) {
+    rapidjson::StringBuffer buffer;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+    doc.Accept(writer);
+    LOG_TRACE(<< buffer.GetString());
+}
```
tveasey, line 27: I wonder if we should move this to `core/CRapidJsonUtils.h`. This pattern comes up a certain amount and it would be good to have a single definition for this.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+const std::string CCommandParser::REQUEST_ID{"request_id"};
+const std::string CCommandParser::TOKENS{"tokens"};
+const std::string CCommandParser::VAR_ARG_PREFIX{"arg_"};
+
+CCommandParser::CCommandParser(std::istream& strmIn) : m_StrmIn(strmIn) {
+}
+
+bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler) const {
+
+    rapidjson::IStreamWrapper isw(m_StrmIn);
```
tveasey, line 39: nit: can you converge on braces initialisation. You currently use a mixture `()`, ` = ` and `{}`.

droberts195:
> The macOS aarch64 build failed because PyTorch has not been setup on the machine

Hopefully this will be possible in PyTorch 1.8, which is not too far off now - see https://github.com/pytorch/pytorch/issues/51886#issuecomment-781688459

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,133 @@
...
+
+    std::uint64_t varCount{1};
+    std::string varArgName = VAR_ARG_PREFIX + std::to_string(varCount);
+    TUint32VecVec args;
+
+    while (doc.HasMember(varArgName)) {
+        TUint32Vec arg;
+        const rapidjson::Value& v = doc[varArgName];
+        for (auto itr = v.Begin(); itr != v.End(); ++itr) {
+            arg.push_back(itr->GetUint());
```
tveasey, line 123: I wonder if we should be checking the type is int here since this is effectively unknown input.

bin/pytorch_inference/CCommandParser.h
```diff
@@ -0,0 +1,72 @@
...
+
+    //! Pass input to the processor until it's consumed as much as it can.
+    bool ioLoop(const TRequestHandlerFunc& requestHandler) const;
+
+    CCommandParser(const CCommandParser&) = delete;
+    CCommandParser& operator=(const CCommandParser&) = delete;
+
+private:
+    bool validateJson(const rapidjson::Document& doc) const;
+    SRequest jsonToRequest(const rapidjson::Document& doc) const;
```
tveasey, line 64: I would avoid creating temporary request objects. This could be quite large and you probably get throughput benefit by keeping a single request object which is created outside the IO loop. I don't think it really complicates matters to do this so doesn't feel like premature optimisation.

bin/pytorch_inference/Main.cc
```diff
@@ -4,104 +4,117 @@
...
+
+    for (auto args : request.s_SecondaryArguments) {
+        torch::Tensor tensor =
+            torch::from_blob(static_cast<void*>(args.data()),
+                             {1, static_cast<std::int64_t>(args.size())},
+                             at::dtype(torch::kInt32))
+                .to(torch::kInt64);
+
+        inputs.push_back(tensor);
+    }
```
tveasey, line 48: `inputs` elements are large; you can reserve this vector given you know all the sizes involved.

tveasey (commented):
I did a pass through. Overall looks nice and clean. I made some minor suggestions. My main observation is since this is long running and you are going to be streaming stuff to this executable I'd try and avoid all the temporary large heap objects. I don't feel this would complicate matters and as is feels like premature pessimization.

bin/pytorch_inference/Main.cc
davidkyle: `torch::tensor` is a wrapper around `at::tensor` and internally has a smart pointer to its data. Passing by value/copying is the way to use it by design

Some discussion here https://discuss.pytorch.org/t/tensor-move-semantics-in-c-frontend/77901/5

bin/pytorch_inference/CCommandParser.cc
davidkyle: I added `operator<<` to this file. It is a trivial function that has to live in the `rapidjson` namespace so I don't think it belongs in `core/CRapidJsonUtils.h`

bin/pytorch_inference/CCommandParser.cc
davidkyle: This is not user input it will be generated by Elasticsearch, in fact I was tempted to remove all validation as we control the input. If the JSON string parses it can be considered well formed, only a catastrophic failure such as a broken pipe or simultaneous writes would invalidate the input.

bin/pytorch_inference/Main.cc
davidkyle: Depending on the use case I would expect no more than 4 elements in this vector. Those are the tokens, segment Ids, attention mask and positional Ids. Each vector must be the same length or PyTorch will throw an exception.

Of course uses change so it makes sense to reserve. As this code develops I expect this code will be extracted out of `main.cc` into a dedicated class.

bin/pytorch_inference/CCommandParser.cc
davidkyle: I did think quite a lot about the validation considered copying the pattern used for AD job configuration but felt that was unnecessary complex for this relatively simple document.

tveasey: I think this warrants a comment, because it isn't clear in this context that this would never be under a user control. (It is also worth considering whether it could be under a users control if abused.)

bin/pytorch_inference/CCommandParser.cc
tveasey: It's this stuff I meant:
```
rapidjson::StringBuffer buffer;
rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
doc.Accept(writer);
```
i.e. conversion to string which I'm sure I've put somewhere local myself in the past. I find I have to remind myself every time how to convert a `rapidjson::Document` to a string and thought perhaps it was time to make this a utility somewhere. That said you don't have to make this change and perhaps we should hunt for other cases address them all in one go.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,138 @@
...
+
+void CCommandParser::jsonToRequest(const rapidjson::Document& doc) {
+
+    m_Request.s_RequestId = doc[REQUEST_ID].GetString();
+    const rapidjson::Value& arr = doc[TOKENS];
+    // overwrite any previous
+    m_Request.s_Tokens.resize(arr.Size());
+
+    for (auto itr = arr.Begin(); itr != arr.End(); ++itr) {
+        m_Request.s_Tokens.push_back(itr->GetUint());
```
tveasey, line 116: I think this probably can't be `push_back` anymore, maybe:
```
auto token = m_Request.s_Tokens.begin();
for (auto itr = arr.Begin(); itr != arr.End(); ++itr, ++token) {
    *token = itr->GetUint();
}
```

tveasey (approved):
I think an issue got introduced in the refactor, plus I do think input validation warrants a code comment.

bin/pytorch_inference/Main.cc
```diff
@@ -8,100 +8,121 @@
...
     inputs.push_back(tokensTensor);
-    inputs.push_back(torch::ones({1, static_cast<std::int64_t>(data.size())})); // attention mask
-    inputs.push_back(
-        torch::zeros({1, static_cast<std::int64_t>(data.size())}).to(torch::kInt64)); // token type ids
-    inputs.push_back(torch::arange(static_cast<std::int64_t>(data.size())).to(torch::kInt64)); // position ids
 
-    torch::NoGradGuard noGrad;
-    auto tuple = module.forward(inputs).toTuple();
-    auto predictions = tuple->elements()[0].toTensor();
+    for (auto args : request.s_SecondaryArguments) {
```
tveasey, line 46: One more thing...
```suggestion
    for (const auto& args : request.s_SecondaryArguments) {
```

davidkyle: 👍  reference yes but it can't be const because later there is non-const access to `.data()`

tveasey: Interestingly, this should cause undefined behaviour because the tensor is then referencing memory from the loop copy of request vector. I wonder if this points to a missing test of non-empty secondary arguments.

tveasey (approved):
Thanks for iterating. LGTM.

bin/pytorch_inference/CCommandParser.h
```diff
@@ -0,0 +1,83 @@
...
+
+//! \brief
+//! Reads JSON documents from a stream calling the request handler
+//! for each parsed document.
+//!
+//! DESCRIPTION:\n
+//! Validation on the input documents is light. It is expected the input
+//! comes from another process which tightly controls what is sent.
+//! Input from an outside source that has not been sanitized should never
+//! be sent.
```
droberts195, line 28: This is saying to a hacker, "If you can manage to send dodgy input to this process we'll give you a shell prompt on the system."

I _think_ the input _is_ actually validated to the extent of preventing array bounds overwrites.  So instead the comment could be more along the lines of, "Validation exists to prevent memory violations from malicious input, but no more.  The caller is responsible for sending input that will not result in errors from libTorch and will produce meaningful results."

bin/pytorch_inference/Main.cc
tveasey: Actually, calling `to()` is probably what saves you since it'll create a copy. This makes me wonder, should we be doing this here? I'd have thought it would be better to write the values into `std::vector<std::uint64_t>` and keep with just having a reference to this memory for the Tensor. It may need some alignment shenanigans to make the most of library optimisations but that should be manageable. Something to investigate in a follow up anyway.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -0,0 +1,139 @@
...
+            LOG_ERROR(<< "Request handler forced exit");
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool CCommandParser::validateJson(const rapidjson::Document& doc) const {
+    if (doc.HasMember(REQUEST_ID) == false) {
```
droberts195, line 75: Actually there is one security hole in the validation, which is that we need to confirm `doc[REQUEST_ID].IsString()`.  Without this additional check, sending an integer for this field instead would be a way to get a pointer of choice dereferenced.

droberts195 (approved):
LGTM if you could just change a couple of things related to securing the input.

bin/pytorch_inference/CCommandParser.cc
tveasey: I'd missed the general comment in the class description, that's good enough.

bin/pytorch_inference/CCommandParser.h
davidkyle: 👍

bin/pytorch_inference/CCommandParser.cc
davidkyle: I added this and also the checks that the token arrays contain unsigned ints

droberts195:
Windows is showing up a problem:

```
CBufferedIStreamAdapter.obj : error LNK2019: unresolved external symbol __imp_ntohl referenced in function "private: bool __cdecl ml::torch::CBufferedIStreamAdapter::parseSizeFromStream(unsigned __int64 &)" (?parseSizeFromStream@CBufferedIStreamAdapter@torch@ml@@AEAA_NAEA_K@Z)
.objs\pytorch_inference.exe : fatal error LNK1120: 1 unresolved externals
```

It seems that:

```
// For ntohl
#ifdef Windows
#include <WinSock2.h>
#else
#include <netinet/in.h>
#endif
```

is still used in `CBufferedIStreamAdapter.cc` which wasn't touched in this PR.

So I was wrong when I said:

```
USE_NET=1
```

could be removed from the `Makefile`.  Sorry about that.

davidkyle:
I merged this before CI was green because the macOS ARM build will not pass until PyTorch has been added to the build machine, Windows was failing for unrelated reasons. Given this is a feature branch PR it is fine to merge

PR #1771, opened 2021-02-24
[ML] Add clang-tidy configuration

valeriy42:
This PR specifies the properties for clang-tidy static code analysis tool. 

The full list of checks is here: [List of clang-tidy checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html)


droberts195:
retest

droberts195:
I tried these out in CLion and they seem like a pretty good subset of inspections.

One thing that I think warrants further investigation is whether we can tell clang-tidy only to make suggestions that are appropriate for a certain C++ standard (without having to list every one individually).

In CLion by far the most frequent clang-tidy suggestion with these rules is `Function 'blah' should be marked [[nodiscard]]`.  The problem with that is that `[[nodiscard]]` is a C++17 feature, and we are using C++14 in 7.x.  So if we accepted all these suggestions then backporting would be a pain, and the benefit of `[[nodiscard]]` is not high enough to justify that backporting pain.

So if there was a way to tell clang-tidy to only make suggestions that make sense for C++14 (even though the master branch is set to use C++17) then that would be good.  Then once 7.last is released we'd just bump that argument to C++17 and reenable the extra inspections.

If it's not possible to do that then I think `modernize-use-nodiscard` should be added to the exclusions.  The worry is that there are others that will also suggest use of C++17 features and cause backporting pain.

valeriy42:
Thank you for having a look @droberts195! I went through all `moderinize-*` rules and excluded those that require C++17 support. There were only 4 of them.

.clang-tidy
```diff
@@ -0,0 +1,58 @@
...
+  -google-runtime-references,
+  misc-*,
+  
+  modernize-*,
+  -modernize-use-trailing-return-type,
+  -modernize-concat-nested-namespaces,
+  -modernize-use-nodiscard,
+  -modernize-replace-random-shuffle,
+  -modernize-unary-static-assert,
+  -modernize-use-uncaught-exceptio
```
droberts195, line 22: Should this end with an "n"?

droberts195 (approved):
LGTM

I am happy to merge the rules file now - just one question about a possible typo.

Please wait for Dave K to review the setup instructions before merging though - I haven't looked at these.

.clang-tidy
valeriy42: You are right, fixed this typo.

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,202 @@
...
+[Install `clangd`](https://clangd.llvm.org/installation.html). Version 11.0 had a significant performance improvement,
+so it is worse to go an extra mile and install the latest stable version. Install
+[Visual Studio Code](https://code.visualstudio.com/) with
+[clangd plugin](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd).
+
+In **Settings**, make sure *C_Cpp:Autocomplete* and *C_Cpp:Intelli Sense Engine* from the native C++ plugin of VS Code
+are disabled. Note that we don't want to disable the complete C/C++ Extension, since we are still going to use the
+debugging capabilities.
+
+Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** und *Clangd: Path*.
```
davidkyle, line 24:
```suggestion
Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** and *Clangd: Path*.
```

und -> and

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,202 @@
...
+            "label": "Build ml-cpp lib",
+            "type": "shell",
+            "command": "cd ${workspaceFolder} && make ML_DEBUG=1 -j6",
+            "problemMatcher": [
+                "$gcc"
+            ],
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            }
```
davidkyle, line 67: VSCode on macOS does not pick the `CPP_SRC_HOME` env var. To build I had to explicitly add this to the 'Build ml-cpp lib' task json

```
            "options": {
                "env": {
                    "CPP_SRC_HOME": "${env:CPP_SRC_HOME}"
                }
            }  
```

valeriy42:
I removed the Visual Studio Code instructions and will merge only the `.clang-tidy` file.

Issue #1721, opened 2021-02-03
[ML] Boosted tree incremental training


tveasey:
This covers the work selectively rip up and regrow and/or recompute leaf values for a decision tree for fine tuning. It should be possible to treat this as just another round of boosting. The work breaks down into the following tasks:
- [x] Leaf node statistics should be hierarchy with concrete implementations from train from scratch and incremental training
- [x] New leaf node statistics for incremental training
- [x] New loss function for MSE for incremental training (other loss functions will be the subject of a separate issue)
- [x] Select trees to retrain based on how much they can reduce improve loss
- [x] Retrain functionality for selected trees 
- [x] Cross-validate additional hyperparameters targeting prediction stability
- [x] Update interfaces to support incremental training

PR #1772, opened 2021-02-24
[ML] Incremental train (part 1)

tveasey:
This makes some mainly interface changes in anticipation of incremental training. The plan is to adjust the loss function to include penalties for modifying the tree topology and results. It should be possible to manage this in `CBoostedTreeLeafNodeStatistics` and reuse our existing training code. As such I've made this a hierarchy. The current implementation moves into `CBoostedTreeLeafNodeStatisticsScratch` and I'll have a new implementation `CBoostedTreeLeafNodeStatisticsIncremental` (placeholder). There will likely need to be further tweaks down the line particularly to the workspace object, but I'll make those in the context of working on the implementation `CBoostedTreeLeafNodeStatisticsIncremental`. This feels like a useful place to break up the change before I move into implementation details.

A step towards #1721.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeRunner.cc
@@ -358,7 +358,7 @@ std::size_t CDataFrameTrainBoostedTreeRunner::estimateBookkeepingMemoryUsage(
     std::size_t totalNumberRows,
     std::size_t /*partitionNumberRows*/,
     std::size_t numberColumns) const {
-    // TODO Incremental training.
+    // TODO https://github.com/elastic/ml-cpp/issues/1790.
     return m_BoostedTreeFactory->estimateMemoryUsageTrain(
         static_cast<std::size_t>(static_cast<double>(totalNumberRows) * m_TrainingPercent + 0.5),
         numberColumns);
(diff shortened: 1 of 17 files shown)


lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -358,7 +358,8 @@ std::size_t CDataFrameTrainBoostedTreeRunner::estimateBookkeepingMemoryUsage(
     std::size_t totalNumberRows,
     std::size_t /*partitionNumberRows*/,
     std::size_t numberColumns) const {
-    return m_BoostedTreeFactory->estimateMemoryUsage(
+    // TODO Incremental training.
```
valeriy42, line 361: Do you want to reference the issue, so this TODO won't get lost?

valeriy42 (approved):
LGTM

lib/api/CDataFrameTrainBoostedTreeRunner.cc
tveasey: [f55e90f](https://github.com/elastic/ml-cpp/pull/1772/commits/f55e90f939a5e446e72435c87fd0a38007525164)

tveasey:
Windows CI failure is known worker issue.

PR #1774, opened 2021-02-25
[ML] Ensure program counters cache always cleared

edsavage:
The collection of static program counters is cached prior to
persistence. This provides the background persistence thread access to a
consistent set of counters as they are being written.

As it is desired to persist the program counters only the once for each
model state snapshot, their persistence, and the clearing of the cache,
is coupled to the persistence of the simple count detector, which is
assumed to always exist.

However there is a scenario where persistence operates on an empty
collection of detectors. This occurs when no data has been seen but time
has advanced (see #393 for more details). In this situation the program
counter cache is populated but not cleared. A subsequent persistence
operation will lead to a warning that the counter cache is being
overwritten.

To avoid the warning message, this PR takes the approach of ensuring
that the program counter cache is always cleared at the end of the
persistence operation, regardless of its success or not.

As this is essentially a non-functional change, just an avoidance of a warning message, labelling as a non-issue.

### DIFF
--- a/lib/api/CFieldDataCategorizer.cc
+++ b/lib/api/CFieldDataCategorizer.cc
@@ -425,7 +425,7 @@ bool CFieldDataCategorizer::persistStateInForeground(core::CDataAdder& persister
         return false;
     }
 
-    LOG_DEBUG(<< "Persist categorizer state");
+    LOG_DEBUG(<< "Persist categorizer state in foreground");
 
     TStrVec partitionFieldValues;
     TPersistFuncVec dataCategorizerPersistFuncs;
@@ -469,6 +469,10 @@ bool CFieldDataCategorizer::doPersistState(const TStrVec& partitionFieldValues,
                                            const TPersistFuncVec& dataCategorizerPersistFuncs,
                                            std::size_t categorizerAllocationFailures,
                                            core::CDataAdder& persister) {
+
+    // Takes care of clearing the cache of program counters when exiting the current scope.
+    core::CProgramCounters::CCacheManager cacheMgr;
+
     // The two input vectors should have the same size _unless_ we are not
     // doing per-partition categorization, in which case partition field values
     // should be empty and there should be exactly one categorizer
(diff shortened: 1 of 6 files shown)


droberts195:
Please can you merge latest master into the PR branch.  The CI is failing because it doesn't have the latest CI script changes on master.

edsavage:
The CI build failed on macOS aarch64 due to

```./dev-tools/jenkins_ci.sh: line 71: docker: command not found```

which is unrelated to the content of this PR.

File: lib/api/CFieldDataCategorizer.cc [hunk 1]
```diff
@@ -469,6 +469,10 @@ bool CFieldDataCategorizer::doPersistState(const TStrVec& partitionFieldValues,
                                            const TPersistFuncVec& dataCategorizerPersistFuncs,
                                            std::size_t categorizerAllocationFailures,
                                            core::CDataAdder& persister) {
+
+    // Takes care of clearing the cache of program counters when exiting the current scope.
+    core::CProgramCounters::CCacheManager cacheMgr;
```
droberts195, line 474: This is here for the case of the standalone `categorizer` program right?

I think it's unnecessary in the case of categorization state being persisted as part of a joint categorization/anomaly detection job run by `autodetect`.

For the standalone `categorize` program there is a much bigger problem in that the counters will never get cached or persisted at all.  It's not a problem at the moment as development of the pure standalone categorization program was halted - it's impossible to use it in production.  Therefore I think it would be best to remove this line and replace it with a comment:

```suggestion
    // TODO: if the standalone categorize program is ever progressed, a mechanism needs
    // to be added that does the following:
    // 1. Caches program counters in the foreground before starting background persistence
    // 2. Calls core::CProgramCounters::staticsAcceptPersistInserter once and only once per persist
    // 3. Clears the program counter cache after persistence is complete
```

droberts195 (approved):
LGTM

edsavage:
Windows CI build is failing due to a known error - https://github.com/elastic/infra/issues/27149 - unrelated to this PR.

PR #1777, opened 2021-02-25
[ML] Fixes to get CI working on CentOS on AArch64

droberts195:
This has been observed to fail in Docker running
on CentOS 8.3.  It's also not necessary - those
ulimit commands were really intended for tests
running on static workers where we could log in
afterwards to look at core dumps.

Additionally, there are complexities with the Java
build used to run the subset of Elasticsearch tests.
We need to use a Java that was built on a
distribution with the same page size as the test
machine.

PR #1782, opened 2021-03-02
[7.11][ML] Fixes to get CI working on CentOS on AArch64

droberts195:
Setting shell limits has been observed to fail in
Docker running on CentOS 8.3. It's also not
necessary - those ulimit commands were really
intended for tests running on static workers
where we could log in afterwards to look at core
dumps.

Additionally, there are complexities with the Java
build used to run the subset of Elasticsearch tests.
We need to use a Java that was built on a
distribution with the same page size as the test
machine.

Backport of #1777


droberts195:
The only CI failure was the Windows Jenkins disconnect.

PR #1773, opened 2021-02-25
[ML] Add makefile dependency on static objects for tests

davidkyle:
Programs under the `bin` directory with unit tests statically link objects in the parent directory (`../.objs`). If code is changed and recompiled the unittest makefile does not have a dependency to cause a rebuild.

This means you have to run `make relink` to pick up the change then run `make test` but it is error prone and can mean pushing code that fails tests. 

This change causes the unittest to relink if the object files change. I've called those `LIB_OBJS` for want of a better name.

PR #1786, opened 2021-03-02
[ML] Use TESTED_OBJS dependency in PyTorch unittest makefile

davidkyle:
Following the changes in #1773 this causes a rebuild of the unit tests when the compiled class files change


droberts195:
In this case the Windows CI failed, but it did actually run and found an error:

```
CBufferedIStreamAdapter.obj : error LNK2019: unresolved external symbol __imp_ntohl referenced in function "private: bool __cdecl ml::torch::CBufferedIStreamAdapter::parseSizeFromStream(unsigned __int64 &)" (?parseSizeFromStream@CBufferedIStreamAdapter@torch@ml@@AEAA_NAEA_K@Z)
ml_test.exe : fatal error LNK1120: 1 unresolved externals
```

So `USE_NET=1` needs adding to the `Makefile`.

davidkyle:
The build passed on all platforms except macOS ARM as expected

droberts195 (approved):
LGTM

PR #1787, opened 2021-03-02
[ML] VSCode setup instructions

davidkyle:
Commits shamelessly copied from #1771

### DIFF
--- a/build-setup/vscode/using_vscode.md
+++ b/build-setup/vscode/using_vscode.md
@@ -21,7 +21,7 @@ In **Settings**, make sure *C_Cpp:Autocomplete* and *C_Cpp:Intelli Sense Engine*
 are disabled. Note that we don't want to disable the complete C/C++ Extension, since we are still going to use the
 debugging capabilities.
 
-Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** und *Clangd: Path*.
+Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** and *Clangd: Path*.
 *Clangd: Arguments* can look something like this:
 
 ```text
--- a/build-setup/vscode/using_vscode.md
+++ b/build-setup/vscode/using_vscode.md
@@ -13,7 +13,7 @@ In this tutorial, we assume that you have `gcc` or `clang` installed as describe
 described there since `clangd` also uses it.
 
 [Install `clangd`](https://clangd.llvm.org/installation.html). Version 11.0 had a significant performance improvement,
-so it is worse to go an extra mile and install the latest stable version. Install
+so it is worth going the extra mile and install the latest stable version. Install
 [Visual Studio Code](https://code.visualstudio.com/) with
 [clangd plugin](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd).
 
--- a/build-setup/vscode/using_vscode.md
+++ b/build-setup/vscode/using_vscode.md
@@ -1,0 +1,202 @@
+# Using Visual Studio Code with `ml-cpp`
+
+Previously, we explained how to [set up CLion for `ml-cpp`](../clion/using_clion.md). You may find, that CLion has some
+performance problems when indexing `ml-cpp` project: it may take very long to open the project or CLion may freeze and
+fail. In this case, [Visual Studio Code](https://code.visualstudio.com/) can be an alternative. When combined with
+`clangd`, it is significantly faster, while still being able to provide features like autocompletion, code navigation,
+static code analysis, etc.
+
+## Prerequisites
+
+In this tutorial, we assume that you have `gcc` or `clang` installed as described in the
+[CLion tutorial](../clion/using_clion.md). We also assume that you already created `compile_commands.json` file as
+described there since `clangd` also uses it.
+
+[Install `clangd`](https://clangd.llvm.org/installation.html). Version 11.0 had a significant performance improvement,
+so it is worse to go an extra mile and install the latest stable version. Install
+[Visual Studio Code](https://code.visualstudio.com/) with
+[clangd plugin](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd).
+
+In **Settings**, make sure *C_Cpp:Autocomplete* and *C_Cpp:Intelli Sense Engine* from the native C++ plugin of VS Code
+are disabled. Note that we don't want to disable the complete C/C++ Extension, since we are still going to use the
+debugging capabilities.
+
+Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** und *Clangd: Path*.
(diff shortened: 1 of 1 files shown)


build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,232 @@
...
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Format code",
+            "type": "shell",
+            "command": "docker run --rm -v ${workspaceFolder}:/ml-cpp -u $(id -u):$(id -g) docker.elastic.co/ml-dev/ml-check-style:2 /ml-cpp/dev-tools/clang-format.sh",
+            "problemMatcher": [],
+            "options": {
+                "env": {
+                    "PATH": "${env:PATH}"
```
davidkyle, line 58: This is a heavy handed fix but on macOS docker is in `/usr/local/bin/` which isn't on the default` PATH`

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,232 @@
...
+Here is an example of user tasks specified in `tasks.json`:
+
+```json
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Format code",
+            "type": "shell",
+            "command": "docker run --rm -v ${workspaceFolder}:/ml-cpp -u $(id -u):$(id -g) docker.elastic.co/ml-dev/ml-check-style:2 /ml-cpp/dev-tools/clang-format.sh",
```
davidkyle, line 54: >  -u $(id -u):$(id -g)

I'm not sure what should be passed for the login here. Could we invoke `clang-format.sh` directly outside of docker assuming it is installed locally

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,232 @@
...
+
+```json
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "data_frame_analyzer",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/build/distribution/platform/linux-x86_64/bin/data_frame_analyzer",
```
davidkyle, line 163: > /platform/linux-x86_64/

The platform specific path is the problem here. Perhaps we can define a variable for the value or maybe we just leave a comment saying edit this for your particular platform

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,232 @@
...
+            ],
+            "preLaunchTask": "Build ml-cpp bin"
+        },
+        {
+            "name": "api unit test",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/lib/api/unittest/ml_test",
+            "args": [
+                "--run_test= CInferenceModelMetadataTest/testJsonSchema"
```
davidkyle, line 187: This is a useful example but we shouldn't have it in the setup instructions as the user would be fooled into thinking all the tests have run.

davidkyle (commented):
I think we should take some things out launch.json and document them as the settings will not work for all platforms

build-setup/vscode/using_vscode.md
valeriy42: You require to use `clang-format` version 5.0.1. I am using docker container here to ensure the correct version and to avoid mingling with host dependencies.

davidkyle: Once I logged into docker.elastic.co I could run this in a shell.

I changed the command to call `run_docker_clang_format.sh` because if the docker image is updated the shell script should also be changed. Hardcoding the version here means it will go out of date eventually.

build-setup/vscode/using_vscode.md
davidkyle: I added a token for the platform and instructions to set the platform specific value

build-setup/vscode/using_vscode.md
davidkyle: The `args` option is now documented at the end of the file

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,237 @@
...
+static code analysis, etc.
+
+## Prerequisites
+
+In this tutorial, we assume that you have `gcc` or `clang` installed as described in the
+[CLion tutorial](../clion/using_clion.md). We also assume that you already created `compile_commands.json` file as
+described there since `clangd` also uses it.
+
+[Install `clangd`](https://clangd.llvm.org/installation.html). Version 11.0 had a significant performance improvement,
+so it is worse to go an extra mile and install the latest stable version. Install
```
valeriy42, line 16:
```suggestion
so it is worth going the extra mile and install the latest stable version. Install
```

valeriy42: My typo

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,237 @@
...
+    "tasks": [
+        {
+            "label": "Format code",
+            "type": "shell",
+            "command": "cd ${workspaceFolder} && dev-tools/docker/run_docker_clang_format.sh",
+            "problemMatcher": [],
+            "options": {
+                "env": {
+                    "PATH": "${env:PATH}",
+                    "CPP_SRC_HOME": "${env:CPP_SRC_HOME}"
```
valeriy42, line 59:
```suggestion
                    "CPP_SRC_HOME": "${workspaceFolder}"
```
If we have this option, I think it is better to use `${workspaceFolder}` here. For instance, if you have two copies of `ml-cpp` locally, this will make using VSCode with both of them easier.

build-setup/vscode/using_vscode.md
```diff
@@ -0,0 +1,237 @@
...
+}
+```
+
+## Debugging
+
+To use [debugging](https://code.visualstudio.com/docs/editor/debugging), you need to specify the binary and the
+arguments in a json file. You can use `preLaunchTask` to specify the user tasks that should be executed before the start
+(e.g. building project).
+
+Here is a example of the `launch.json` file. Please replace `YOUR_PLATFORM` in the `program` path with your
```
valeriy42, line 154:
```suggestion
Here is an example of the `launch.json` file. Please replace `YOUR_PLATFORM` in the `program` path with your 
```

valeriy42 (approved):
LGTM. Thank you for cleaning up this tutorial! :pray: 

I noticed a couple of typos I missed initially. Also, I have a small suggestion to define `CPP_SRC_HOME` using `{$workspaceFolder}` instead of using `env`.

build-setup/vscode/using_vscode.md
davidkyle: ++ It makes more sense as `CPP_SRC_HOME` needn't be defined to build in VSCode

davidkyle:
@dimitris-athanasiou ^ instructions for building ml-cpp with VSCode

PR #1792, opened 2021-03-05
[ML] Address clang tidy integer literal warnings

tveasey:
Since we introduced clang tidy rules we get warnings for lower case u suffices for integer literals. There are a lot of these in the code base and it makes reading the code with common setups (such as clangd for VSCode) harder. We could automatically replace them with uppercase U, but they aren't really needed anyway and the qualification is inconsistently applied in the codebase. I therefore just scripted up a change to replace `([0-9])u;` with `\1;`. For reference here is the script:
```
import argparse
import in_place
import re

parser = argparse.ArgumentParser(description='clang tidy integer literals')
parser.add_argument('--files', nargs='+', help='the source files')

args = parser.parse_args()

for file_name in args.files:
    with in_place.InPlace(file_name) as file:
        for line in file:
            line = re.sub('([0-9]+)u;', r'\1;', line)
            file.write(line)
```
Run from `ml-cpp/lib` as follows:
```
% python3 script.py --files $(find . -name '*.cc' | xargs)
```
and similarly from `ml-cpp/include`.


droberts195 (approved):
LGTM

tveasey:
The windows failure is a known CI issue.

PR #1795, opened 2021-03-10
[ML] Upgrading to PyTorch 1.8.0 on macOS

droberts195:
Upgrades PyTorch from version 1.7.1 to version 1.8.0 on macOS.

It is now possible to build a version that works on Apple Silicon.

The build instructions are also adjusted to remove functions that
call an external compiler to build custom extensions. Although
these would never have worked in our programs due to system call
filtering, it's best if they aren't present at all as they could
alarm heuristic virus scanners.

(Similar upgrades will be done for Windows and Linux in the near
future.)

### DIFF
--- a/build-setup/macos.md
+++ b/build-setup/macos.md
@@ -183,14 +183,15 @@ external processes.
 Build as follows:
 
 ```
+# TODO: add BLAS=vecLib when we upgrade to 1.9
 export BUILD_TEST=OFF
 export BUILD_CAFFE2=OFF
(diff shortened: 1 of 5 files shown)


davidkyle (approved):
LGTM

We can set `BLAS=vecLib` for macOS now that pytorch/pytorch#51288 has been merged. You've already built the dependencies so that can wait until the next iteration

droberts195:
retest

PR #1797, opened 2021-03-10
[ML] Upgrading to PyTorch 1.8.0 on Linux

droberts195:
The build instructions are also adjusted to remove functions that
call an external compiler to build custom extensions. Although
these would never have worked in our programs due to system call
filtering, it's best if they aren't present at all as they could
alarm heuristic virus scanners.


droberts195:
Argh!

While trying to follow these new instructions on aarch64 I ran into https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96796.

```
during RTL pass: reload
In file included from /home/ubuntu/tools_gcc93/pytorch/aten/src/ATen/Parallel.h:140,
                 from /home/ubuntu/tools_gcc93/pytorch/build/aten/src/ATen/native/cpu/UnaryOpsKernel.cpp.DEFAULT.cpp:11:
/home/ubuntu/tools_gcc93/pytorch/aten/src/ATen/ParallelOpenMP.h: In function â_ZN2at12parallel_forIZNS_3vml12_GLOBAL__N_17vlgammaIfEEvPT_PKS4_lEUlllE_EEvlllRS6_._omp_fn.0â:
/home/ubuntu/tools_gcc93/pytorch/aten/src/ATen/ParallelOpenMP.h:33:9: internal compiler error: Max. number of generated reload insns per insn is achieved (90)

   33 | #pragma omp parallel if (omp_get_max_threads() > 1 && !omp_in_parallel() && ((end - begin) > grain_size))
      |         ^~~
[ 77%] Building CXX object caffe2/CMakeFiles/torch_cpu.dir/core/context_base.cc.o
[ 77%] Building CXX object caffe2/CMakeFiles/torch_cpu.dir/core/init_denormals.cc.o
[ 77%] Building CXX object caffe2/CMakeFiles/torch_cpu.dir/core/net_async_task_graph.cc.o
0xa131f7 lra_constraints(bool)
        ../../gcc-9.3.0/gcc/lra-constraints.c:4901
0x9fd867 lra(_IO_FILE*)
        ../../gcc-9.3.0/gcc/lra.c:2472
0x9bcc57 do_reload
        ../../gcc-9.3.0/gcc/ira.c:5523
0x9bcc57 execute
        ../../gcc-9.3.0/gcc/ira.c:5707
Please submit a full bug report,
with preprocessed source if appropriate.
Please include the complete backtrace with any bug report.
```

The line numbers in the backtrace exactly match up with the [existing bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96796), so there's no point submitting a new one.  The [last comment](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96796#c9) says:

> 2020-09-11 12:24:31 UTC
>
> Fixed on trunk so far.

The most recent gcc release is version 10.2, from July 2020, so there cannot possibly be a release that fixes that problem.  The bug report says it's targeted for 9.4, which should be due pretty soon, but I cannot find any chatter about it on any public mailing lists.  So the only hope in the short term is to try adjusting the code of the problematic source file slightly.

droberts195:
retest

droberts195:
> So the only hope in the short term is to try adjusting the code of the problematic source file slightly.

This is what I did, via https://github.com/elastic/ml-cpp/pull/1797/files#diff-88435d0cdd0bfef440a7483f80bc6e75cba43032104f6cd5332049817e837c10R295-R314 and https://github.com/elastic/ml-cpp/pull/1797/files#diff-c962419ba16b09a46625af077bf69f477fbb6a11ec954af07050255aa20f0530R130.  The effect is that a few loops that ought to get parallelised don't, although nowhere near as many as if the workaround was to completely disable OPENMP.  The workaround effectively just sends a small number of loops via https://github.com/pytorch/pytorch/blob/bea3cb70693ee85d3f7435efa7232c7fd2fe50cd/aten/src/ATen/ParallelOpenMP.h#L59, which is the "no parallelisation" branch of https://github.com/pytorch/pytorch/blob/bea3cb70693ee85d3f7435efa7232c7fd2fe50cd/aten/src/ATen/ParallelOpenMP.h#L25.

Hopefully gcc 10.3 will be released soonish containing a fix for the compiler bug and we can upgrade to it before our version 8 is feature frozen.

build-setup/linux.md
```diff
@@ -245,58 +245,103 @@ CMake version 3.5 is the minimum required to build PyTorch. Download version 3.1
...
+do for security reasons. Replacing the calls to `system()` ensures that
+a heuristic virus scanner looking for potentially dangerous function
+calls in our shipped product will not encounter these functions that run
+external processes.
+
+If you are building on aarch64, also edit `aten/src/ATen/cpu/vml.h` and
+change lines 88-93 from:
+
+```
+    parallel_for(0, size, 2048, [out, in](int64_t begin, int64_t end) {           \
```
davidkyle, line 299: The macro you changed is exactly the same in v.1.7.1 so I’m guessing something else has changed that  both triggers the bug and means openmp is required. We are not doing training and evaluation is single threaded (as far as I know in our uses) so we shouldn’t really need openmp. 

Is it because of the tangled dependencies in the build system? Did you try `USE_OPENMP=OFF` or `WITH_OPENMP=OFF`?

3rd_party/3rd_party.sh
```diff
@@ -68,6 +69,8 @@ case `uname` in
             XML_EXTENSION=.so.2
             GCC_RT_LOCATION=/usr/local/gcc93/lib64
             GCC_RT_EXTENSION=.so.1
+            OMP_LOCATION=/usr/local/gcc93/lib64
```
davidkyle, line 72: I'm guessing this is installed by the Python dependencies as I don't see it explicitly built in the dockerfile

droberts195: `libgomp` is part of gcc and built alongside it, like `libgcc_s` and `libstdc++`.

build-setup/linux.md
droberts195: > evaluation is single threaded (as far as I know in our uses)

Oh, I didn't know that.

So it seems that the difference between my build of 1.8.0 and your build of 1.7.1 is that mine decided to enable OPENMP by default whereas yours didn't.  That would explain both this problem and also that fact that `libgomp` was getting linked.

If we really don't want any multi-threading in the library we ship I can rebuild it all again with `USE_OPENMP=OFF`.  That should then mean that we don't need to hack the code while using gcc 9.3 and don't need to ship `libgomp`.

So I guess it depends how certain we are that multithreading isn't giving us any benefits at all in the way we're using the library...

What do you think?

droberts195: According to https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html model interference _can_ be multi-threaded.  So I think we are better off sticking with a version of the library built with `USE_OPENMP=ON` (which is the default).

davidkyle (approved):
LGTM

PR #1798, opened 2021-03-11
[ML] Improve bounds checks in state restore code

droberts195:
Fixes a few more places where array or vector bounds were not
checked during state restoration.

Follows on from #1717 and #1737

### DIFF
--- a/include/core/CPersistUtils.h
+++ b/include/core/CPersistUtils.h
@@ -977,7 +977,12 @@ private:
                         LOG_ERROR(<< "Restoration error at " << traverser.name());
                         return false;
                     }
-                    *(i++) = std::move(value);
+                    if (i == container.end()) {
+                        LOG_ERROR(<< "Too many values for size " << N
+                                  << " array during restoration");
+                    } else {
+                        *(i++) = std::move(value);
+                    }
                 }
             } while (traverser.next());
             return true;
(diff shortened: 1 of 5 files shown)


include/core/CPersistUtils.h
```diff
@@ -977,7 +977,12 @@ class CRestorerImpl<ContainerRestore> {
                         LOG_ERROR(<< "Restoration error at " << traverser.name());
                         return false;
                     }
-                    *(i++) = std::move(value);
+                    if (i == container.end()) {
+                        LOG_ERROR(<< "Too many values for size " << N
+                                  << " array during restoration");
```
tveasey, line 982: Is there a good reason not to `return false;` here?

tveasey (approved):
Changes LGTM. Just a question.

include/core/CPersistUtils.h
droberts195: Since this is being backported to old branches I was worried there might be a place where, say, we restore 4 values into a size 3 array and have never noticed because the padding means it never causes a crash.

After this is merged I'll create a followup PR just for master that makes it return `false` here.  Then there will be plenty of time for any historical bugs to be revealed by the new logging and our internal testing on master before end users get the stricter behaviour.

droberts195:
The only CI failure was the Jenkins Windows problem.

PR #1799, opened 2021-03-11
[ML] Rename pytorch log pipe arg

dimitris-athanasiou:
Renames pytorch process logging arguments to be
consistent with the other processes and with what java
provides.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -128,16 +128,20 @@ int main(int argc, char** argv) {
     std::string restoreFileName;
     bool isRestoreFileNamedPipe{false};
     std::string logFileName;
-    bool isLogFileNamedPipe{false};
+    std::string logProperties;
     ml::core_t::TTime namedPipeConnectTimeout{
         ml::core::CBlockingCallCancellingTimer::DEFAULT_TIMEOUT_SECONDS};
+
+    LOG_INFO(<< "Parsing command args");
 
     if (ml::torch::CCmdLineParser::parse(
             argc, argv, modelId, namedPipeConnectTimeout, inputFileName,
             isInputFileNamedPipe, outputFileName, isOutputFileNamedPipe, restoreFileName,
-            isRestoreFileNamedPipe, logFileName, isLogFileNamedPipe) == false) {
+            isRestoreFileNamedPipe, logFileName, logProperties) == false) {
         return EXIT_FAILURE;
     }
+
+    LOG_INFO(<< "Passed command parsing");
 
     ml::core::CBlockingCallCancellingTimer cancellerThread{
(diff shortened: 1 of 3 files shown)


bin/pytorch_inference/Main.cc
```diff
@@ -128,17 +128,21 @@ int main(int argc, char** argv) {
     std::string restoreFileName;
     bool isRestoreFileNamedPipe{false};
     std::string logFileName;
-    bool isLogFileNamedPipe{false};
+    std::string logProperties;
     ml::core_t::TTime namedPipeConnectTimeout{
         ml::core::CBlockingCallCancellingTimer::DEFAULT_TIMEOUT_SECONDS};
 
+    LOG_INFO(<< "Parsing command args");
```
droberts195, line 135: Since you've confirmed it works locally this should be removed.  In production this would go to the processes (hard to see) stderr, not to the ES log, as it's before the logging gets redirected.  This may cause some obscure problem one day.

bin/pytorch_inference/Main.cc
```diff
@@ -128,17 +128,21 @@ int main(int argc, char** argv) {
...
+
     if (ml::torch::CCmdLineParser::parse(
             argc, argv, modelId, namedPipeConnectTimeout, inputFileName,
             isInputFileNamedPipe, outputFileName, isOutputFileNamedPipe, restoreFileName,
-            isRestoreFileNamedPipe, logFileName, isLogFileNamedPipe) == false) {
+            isRestoreFileNamedPipe, logFileName, logProperties) == false) {
         return EXIT_FAILURE;
     }
 
+    LOG_INFO(<< "Passed command parsing");
```
droberts195, line 144: As above, please remove before merging.

bin/pytorch_inference/Main.cc
```diff
@@ -162,13 +166,11 @@ int main(int argc, char** argv) {
...
     }
 
-    if (isLogFileNamedPipe) {
-        if (ml::core::CLogger::instance().reconfigureLogToNamedPipe(
-                logFileName, cancellerThread.hasCancelledBlockingCall()) == false) {
-            LOG_FATAL(<< "Could not reconfigure logging");
-            cancellerThread.stop();
-            return EXIT_FAILURE;
-        }
+    if (ml::core::CLogger::instance().reconfigure(
```
davidkyle, line 169: The `evaluate.py` script does not set the `logPipe` argument so all logging is done to std out which is quite useful at this stage of development. 

I think we still need to protect the call to reconfigure with `if (logFileName.empty() == false)` at least while we are using `evaluate.py`

droberts195: That protection is already in `CLogger::reconfigure()`.  As the code stands here it will default to logging to stderr, just like the other ML processes like `autodetect`.

droberts195 (approved):
LGTM

dimitris-athanasiou:
retest

PR #1803, opened 2021-03-15
[ML] Make too many array values on restore fatal

droberts195:
This implements what was discussed in
https://github.com/elastic/ml-cpp/pull/1798#discussion_r592515203

By making this change only on the master branch we will have a
long time to see if there are any unexpected side effects, where
we were accidentally relying on the bug.


tveasey (approved):
LGTM

PR #1804, opened 2021-03-15
[ML] Reducing permitted system calls

droberts195:
mmap, munmap and mprotect are in the list of permitted
system calls because they are required by the dynamic
linker.

This PR is a test to see what will happen if we remove
them. Now that we use immediate binding (-z now) for
all linking it MAY be that the dynamic linker finishes
its work before we install the system call filters, and
hence we don't need these calls to be permitted after
that point. Or alternatively it may be that the dynamic
linker expects to be able to make these system calls
later in the program's lifetime...

### DIFF
--- a/lib/seccomp/CSystemCallFilter_Linux.cc
+++ b/lib/seccomp/CSystemCallFilter_Linux.cc
@@ -34,7 +34,7 @@ const struct sock_filter FILTER[] = {
 
 #ifdef __x86_64__
     // Only applies to x86_64 arch. Jump to disallow for calls using the x32 ABI
-    BPF_JUMP(BPF_JMP | BPF_JGT | BPF_K, UPPER_NR_LIMIT, 46, 0),
+    BPF_JUMP(BPF_JMP | BPF_JGT | BPF_K, UPPER_NR_LIMIT, 43, 0),
     // If any sys call filters are added or removed then the jump
     // destination for each statement including the one above must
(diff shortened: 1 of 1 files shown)


droberts195:
The Linux CI failures on both x86_64 and aarch64 reveal a place where `mprotect` is used: it's when pthreads starts a new thread or grows the stack for an existing thread. The places where these calls are made can be seen by searching for `__mprotect` in [the NPTL's `allocatestack.c` source file](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/allocatestack.c;h=149b999603e2a6bd54c168f17b2049f6e9d86b73;hb=HEAD). So for this reason we cannot disallow `mprotect` unless we go completely single threaded (which is unrealistic).

Additionally [a paper presented at Black Hat 2007](https://www.blackhat.com/presentations/bh-usa-07/Ferguson/Whitepaper/bh-usa-07-ferguson-WP.pdf) explains how `mmap` and `munmap` as well as `mprotect` are used by the multiple arena heap in modern glibc.  To provide better performance in multi-threaded programs glibc divides the heap into multiple arenas, and these are not all initialized on program startup.  Therefore, even if we managed to go single threaded, we might find calls to `mmap`, `munmap` or `mprotect` being made as the heap grew and shrunk.

tonymeehan:
This makes sense. Thanks for looking into it, @droberts195 .

PR #1805, opened 2021-03-15
[ML] Implement the boosted tree leaf node statistics for incremental training

tveasey:
This is a first pass implementing leaf node statistics for incremental training. This introduces an extra term in the gain calculation penalising changing the topology of the tree being retrained. It also migrates common functionality back to `CBoostedTreeLeafNodeStatistics`. Note that I have not tried to compute look ahead bounds for incremental training (because this is complicated by the split dependent penalty). Therefore, I optionally disable the extra statistics we compute to support this. I decided not to unit test this yet because I think it will be more valuable in the context of end-to-end testing, but it represents a nice self contained piece of work, so I'm breaking this out into a PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ nbactions.xml
 #vscode files
 .vscode
 .clangd
+.cache
 
 # gradle stuff
 .gradle/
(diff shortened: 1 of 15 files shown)


valeriy42 (approved):
LGTM

PR #1806, opened 2021-03-16
[ML] Document the reason for allowing some syscalls

droberts195:
mmap and munmap are used by the malloc arenas functionality
of glibc. mprotect is also used by that functionality,
together with pthreads, which uses mprotect for managing
each pthread's stack.


davidkyle (approved):
LGTM

PR #1810, opened 2021-03-16
[ML] Supplementary files for data summarization investigation

valeriy42:
This PR contains supplementary files to run data sampling and summarization in Python.

I add configurations for datasets, but not the CSV files to the repository. The dataset files can be found in the GCP bucket for datasets.

As long as we don't have a specific call to evaluate a dataset using the persisted model, I had to deactivate some invariance checks to "fool" the model to work with accept dataset. I added #1811 so we won't forget to reactivate the checks.


valeriy42:
retest

tveasey (approved):
LGTM

PR #1812, opened 2021-03-16
[ML] A couple of improvements to setting RPATH

droberts195:
1. When setting RPATH using patchelf, use the --force-rpath
   option, because otherwise it sets RUNPATH instead. This
   only makes a difference when LD_LIBRARY_PATH is set, so
   won't affect our expected production scenario where
   LD_LIBRARY_PATH is _not_ set. But preferring RPATH may
   avoid confusion in dev environments.
2. Use $ORIGIN instead of $ORIGIN/. when setting RPATH on
   the link line. The /. gets appended automatically, so
   we are actually using $ORIGIN/./. at the moment, which
   works fine but looks messy.


dimitris-athanasiou (approved):
LGTM

PR #1818, opened 2021-03-17
[ML] Handle validation errors in PyTorch requests

davidkyle:
Inference requests are JSON documents, previously the command parser would drop a malformed request with only a log message as evidence of its passing. 

This adds an error handler to the command processor that is called in the event of the bad inference request, the error is returned on the output so a client can expect a response for every request sent. 

In practice it is unlikely that a malformed request is sent during normal operation but I want to establish the contract that a client can always expect some sort of response.

### DIFF
--- a/bin/pytorch_inference/CCommandParser.cc
+++ b/bin/pytorch_inference/CCommandParser.cc
@@ -165,7 +165,7 @@ void CCommandParser::jsonToRequest(const rapidjson::Document& doc) {
     TUint64Vec arg;
     while (doc.HasMember(varArgName)) {
         const rapidjson::Value& v = doc[varArgName];
-        for (const auto* itr = v.Begin(); itr != v.End(); ++itr) {
+        for (auto itr = v.Begin(); itr != v.End(); ++itr) {
             arg.push_back(itr->GetUint64());
         }
 
--- a/bin/pytorch_inference/CCommandParser.cc
+++ b/bin/pytorch_inference/CCommandParser.cc
@@ -14,6 +14,8 @@
 #include <rapidjson/writer.h>
 
 #include <istream>
+#include <sstream>
+#include <string>
 
 namespace rapidjson {
 
@@ -31,11 +33,13 @@ namespace torch {
 const std::string CCommandParser::REQUEST_ID{"request_id"};
 const std::string CCommandParser::TOKENS{"tokens"};
 const std::string CCommandParser::VAR_ARG_PREFIX{"arg_"};
+const std::string CCommandParser::UNKNOWN_ID;
 
 CCommandParser::CCommandParser(std::istream& strmIn) : m_StrmIn(strmIn) {
 }
 
-bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler) {
+bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler,
+                            const TErrorHandlerFunc& errorHandler) {
 
     rapidjson::IStreamWrapper isw{m_StrmIn};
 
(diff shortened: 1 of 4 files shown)


bin/pytorch_inference/CCommandParser.cc
```diff
@@ -152,7 +165,7 @@ void CCommandParser::jsonToRequest(const rapidjson::Document& doc) {
     TUint64Vec arg;
     while (doc.HasMember(varArgName)) {
         const rapidjson::Value& v = doc[varArgName];
-        for (auto itr = v.Begin(); itr != v.End(); ++itr) {
+        for (const auto* itr = v.Begin(); itr != v.End(); ++itr) {
```
droberts195, line 168: This looks strange.  It's idiomatic to use iterators by value in for loops.  I think what you've changed it to is working because a RapidJSON array iterator happens to be a raw pointer, but it's a breach of encapsulation to code for that.  Explicitly stating that the `auto` has to be a pointer isn't making it more efficient.  The type `auto` referred to would have been a pointer anyway if that's what a RapidJSON array iterator is.

droberts195 (approved):
LGTM except for one question about a loop iterator

bin/pytorch_inference/CCommandParser.cc
davidkyle: The IDE suggested this change based on the clang tidy rule `clang-tidy(readability-qualified-auto)` I was surprised it compiled TBH then forgot to revert it. 

@valeriy42 do you want to re-evaluate the use of [clang-tidy(readability-qualified-auto)](https://clang.llvm.org/extra/clang-tidy/checks/readability-qualified-auto.html) given the above. Does it align with the repo's coding standards?

valeriy42: Digging into the link, the warning comes from [LLVM coding standards](https://llvm.org/docs/CodingStandards.html#beware-unnecessary-copies-with-auto) where they suggest avoiding unnecessary copies and making pointer copies explicit (for readability). 

image

In this trade-off between encapsulation and easiness to uncover unnecessary copy by value, LLVM advice towards latter. AFAICS, in our repo's coding standards this topic is not mentioned explicitly.

droberts195: The standard library classes presumably do something to avoid such encapsulation breaches.  For example, an `std::vector<int>::iterator` _could_ be an `int*`.  It's up to the library implementor whether it is or whether they use a class to implement vector iterators.

So for standard libraries where it _is_ a simple `int*`, using explicit types we could write:

```
    std::vector<int> myVec{1, 2, 3};
    for (int* iter = myVec.begin(); iter != myVec.end(); ++iter) {
        // blah
    }
```

I don't think anyone would say that was best practice.  The equivalent code with explicit types but using the typedefs instead of breaching encapsulation would be:

```
    std::vector<int> myVec{1, 2, 3};
    for (std::vector<int>::iterator iter = myVec.begin(); iter != myVec.end(); ++iter) {
        // blah
    }
```

That is exactly as efficient as using `int*` as the type in the case where `std::vector<int>::iterator` is a simple `int*`.

So if `clang-tidy` isn't recommending replacement of `auto` with `auto*` for some standard container iterators that happen to be pointers that makes me think there is some way that it knows that for iterators encapsulation is more important than revealing what is a pointer.  It makes me think that either this knowledge is hardcoded in `clang-tidy` for certain standard library classes, or else there's something RapidJSON is doing that bypasses some logic in `clang-tidy`.  Possibly it's because the RapidJSON methods for getting iterators are `Begin()` and `End()` instead of `begin()` and `end()`.

davidkyle: The second line causes a compilation failure:

```c++
    static_assert(std::is_pointer<rapidjson::GenericValue<rapidjson::UTF8<char>>::ConstValueIterator>::value, 
        "RapidJSON iterator is pointer");

    static_assert(std::is_pointer<std::vector<int>::iterator>::value, 
        "Int iterator is pointer");   
```

The RapidJSON iterator is defined as a pointer in document.h

`typedef const GenericValue* ConstValueIterator; //!< Constant value iterator for iterating in array.`

As @valeriy42 said the rule isn't explicitly defined in `.clang-tidy` it comes as part of the LLVM rule set. For the sake of a few squiggly lines in VSCode there is not point starting to pick those rules apart. 

I've reverted the iterator back to `auto itr` and I've learned something today so we will call that a win and merge as is.

PR #1819, opened 2021-03-18
[ML] Adding mremap to syscall filter

droberts195:
Under certain circumstances, the realloc() function in glibc
will make the mremap syscall. If this fails then realloc()
falls back to using malloc() and memcpy(), so the end result
is the same from the caller's perspective, it's just slower.

Prior to now we did not have the mremap syscall on our
permitted list in our seccomp filter. Not only did this mean
that our realloc() calls were slower than they could be, we
are also filling up audit logs in systems that audit syscalls
that seccomp blocks.

This change permits the mremap syscall. The risk is low,
because we already permit mmap and munmap and mremap can be
emulated using those two.


tveasey (approved):
LGTM

PR #1821, opened 2021-03-18
[ML] Abort more explicitly

droberts195:
The LOG_ABORT macro calls CLogger::fatal(), which, prior
to this change, used to throw an exception.

We have a rule of thumb that we don't use exceptions in
our code and exceptions from third party libraries are
caught as close as possible to the call into that library.
When CLogger::fatal() was originally written this rule
was pretty universally applied, and it was a reasonable
assumption that a thrown exception would not be handled
and would hence abort the program. However, we now have so
many exceptions to this rule that an exception thrown by
a LOG_ABORT macro is quite likely to get caught, logged,
and cause some sort of minor workaround code to run. This
is not the intention for LOG_ABORT; it is intended to be
used only when the program is in a fatally compromised
state due to some unforeseen bug and we do not want to
be continuing execution.

This change makes CLogger::fatal() unconditionally call
std::terminate(), i.e. the effect of an unhandled
exception. This puts the functionality back to what it
originally did before we had so many catch blocks.

### DIFF
--- a/include/core/CLogger.h
+++ b/include/core/CLogger.h
@@ -136,7 +136,7 @@ public:
     //! Access to underlying logger (must only be called from macros)
     TLevelSeverityLogger& logger();
 
-    //! Throw a fatal exception
+    //! Terminate the program.
     [[noreturn]] static void fatal();
 
     //! Register a new global fatal error handler.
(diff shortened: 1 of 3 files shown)


dimitris-athanasiou (approved):
LGTM

PR #1823, opened 2021-03-22
[ML] Make state traversal unwind more quickly on errors

droberts195:
This change makes state traversal cascade failures up the
call stack more quickly than it used to.

The biggest changes are:

1. We no longer traverse a sub-level if there's ever been
   bad state detected
2. Some constructors that used to ignore bad state now
   set the bad state flag

Also adds invariant checking after restore to the
maths::SMultimodalPriorMode and model::CSearchKey classes,
which were missed in #1717 and #1737 respectively.

### DIFF
--- a/include/core/CStateRestoreTraverser.h
+++ b/include/core/CStateRestoreTraverser.h
@@ -7,10 +7,10 @@
 #define INCLUDED_ml_core_CStateRestoreTraverser_h
 
 #include <core/CLogger.h>
-#include <core/CNonCopyable.h>
 
 #include <core/ImportExport.h>
 
+#include <exception>
 #include <string>
 
 namespace ml {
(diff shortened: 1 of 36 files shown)


include/core/CStateRestoreTraverser.h
```diff
@@ -59,16 +63,16 @@ class CORE_EXPORT CStateRestoreTraverser : private CNonCopyable {
     //! of this method.
     template<typename FUNC>
     bool traverseSubLevel(FUNC f) {
-        if (!this->hasSubLevel()) {
+        if (this->hasSubLevel() == false || this->haveBadState()) {
             return false;
         }
 
         CAutoLevel level(*this);
```
edsavage, line 70: Is the declaration of `level` still needed? (It seems to be no longer used)

edsavage (approved):
LGTM. Just one minor question.

include/core/CStateRestoreTraverser.h
droberts195: It's an RAII class, like `scoped_lock`.  Its constructor steps down a level and its destructor will step back up.

PR #1831, opened 2021-03-23
[7.12][ML] Make ML native processes work with glibc 2.33 on x86_64

droberts195:
In glibc version 2.33 the implementation of the mknod() function
was consolidated so that the underlying syscall used was mknodat
for all architectures.

Prior to glibc 2.33 the aarch64 implementation of the mknod()
function used the mknodat syscall but other architectures used
the mknod syscall.

To allow the ML C++ to work with glibc 2.33 on x86_64, this change
moves mknodat into the list of syscalls that are permitted on all
architectures.

The mknod syscall is still also permitted on x86_64.  We will need
to leave this until our oldest supported distribution is using
glibc 2.33 or higher (10+ years in the future probably).

Backport of #1828

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,12 @@
 
 //=== Regressions
 
+== {es} version 7.12.1
+
+=== Enhancements
+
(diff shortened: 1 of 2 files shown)


droberts195:
retest

droberts195:
retest

PR #1835, opened 2021-03-26
[ML] Restructure Jenkins scripts

droberts195:
This is the first phase of a refactor that will eventually enable
the same Jenkins setup in elasticsearch-ci to run:

1. PR builds
2. Branch builds that cache artifacts for developers
3. Daily debug builds with assertions enabled
4. Builds for release to end users

A secondary goal is to speed up CI by spreading the build over
more servers.

In this first stage of the changes:

1. The extra options needed by debug builds and release builds are
   passed through to the low level build commands
2. The cross compile steps are split into a separate build script,
   which, eventually, will be run on a different server

Completing this project without lengthy ML CI downtime will
require a series of changes alternating between the ml-cpp repo
and the Jenkins configuration.  This is just the first step of
many.

PR #1836, opened 2021-03-29
[ML] Make CI scripts independent of initial working directory

droberts195:
This is a followup to #1835 that makes the scripts work
reliably nomatter where they are started from.


droberts195:
This is for master only - I will incorporate it into the backports of #1835.

edsavage (approved):
LGTM

PR #1841, opened 2021-04-01
[ML] PyTorch results should be written as an array

dimitris-athanasiou:
This commit changes the way the pytorch_inference process writes
results so they fit the format java expects.


davidkyle (approved):
LGTM

PR #1843, opened 2021-04-06
[ML] Adjust parsing of inference results in python test script

davidkyle:
Following #1841 the output is now an array of objects rather than NDJSON


dimitris-athanasiou (approved):
LGTM

PR #1844, opened 2021-04-06
[ML] Cross compile solely in dedicated script

droberts195:
The changes of #1835 introduced a separate script for
cross compiling in Jenkins builds.  However, at that
time cross compiles still had to be done in the original
CI script too because Jenkins was not running the
dedicated script.

Now Jenkins _is_ running the dedicated script, and we
are in the situation where cross compiles are being done
twice for each build.  This PR tidies that up by making
jenkins_ci.sh dedicated to native compilation.

This should improve end-to-end build times.


tveasey (approved):
LGTM

PR #1848, opened 2021-04-13
[ML] Persist state for correcting loss to account for model size when training a boosted tree

tveasey:
We were missing state used to adjust the loss function to select the best hyperparameters based on model size. This would sometimes cause us to select different hyperparameters if the modelling failed and was restarted during training.

Note that this already tested by `testRunBoostedTreeRegressionTrainingWithStateRecovery` but only fails if the change causes us to select a different model so this doesn't reliably fail.


valeriy42 (approved):
LGTM

PR #1855, opened 2021-04-19
[ML] Gracefully handle insufficient training data for classification and regression model training

tveasey:
This explicitly checks we have enough data to even create a hold out set. This is a very weak condition and this isn't really enough data anyway, but before we could either throw (reserving collection with max size due to unsigned integer underflow) or hang (downsampling). This just adds an assertion at the point we create cross-validation row masks.


valeriy42 (approved):
LGTM

dimitris-athanasiou (approved):
LGTM2

PR #1856, opened 2021-04-19
[ML] Don't enable trace logging in categorizer tests

droberts195:
At some point in the distant past there was a good reason
to enable trace logging in categorizer tests. Times have
changed and this is no longer a good idea:

1. We compile out trace logging when optimisation is
   enabled, so it generally has no effect
2. When we do debug builds the extra logging generates
   huge output files, which can blow Jenkins limits

In the future if somebody needs extra detailed logging
about categorization they can enable it manually on a
local test run.


edsavage (approved):
LGTM

PR #1862, opened 2021-04-21
[ML] Pass row iterators by constant reference to callbacks

tveasey:
This migrates the callback to read and write rows on a data frame to pass the iterators by constant reference. This generates clang-tidy warnings. Although the types are small and relatively a lot of work happens for each copy, each one is 64 bytes and there is no need to copy them.

I've marked as a non-issue because I don't think there is any need to document this change.

### DIFF
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -311,12 +311,7 @@ void CDataFrameAnalyzer::writeResultsOf(const CDataFrameAnalysisRunner& analysis
     std::size_t numberThreads{1};
 
     using TRowItr = core::CDataFrame::TRowItr;
-    m_DataFrame->readRows(numberThreads, [&](TRowItr beginRows, TRowItr endRows) {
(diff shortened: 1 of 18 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -311,12 +311,7 @@ void CDataFrameAnalyzer::writeResultsOf(const CDataFrameAnalysisRunner& analysis
     std::size_t numberThreads{1};
 
     using TRowItr = core::CDataFrame::TRowItr;
-    m_DataFrame->readRows(numberThreads, [&](TRowItr beginRows, TRowItr endRows) {
-        TMeanVarAccumulator timeAccumulator;
-        core::CStopWatch stopWatch;
-        stopWatch.start();
-        std::uint64_t lastLap{stopWatch.lap()};
```
tveasey, line 318: This appears to be dead code left over from some previous experiment.

valeriy42 (approved):
LGTM! :star2:

PR #1868, opened 2021-04-23
[ML] Prediction task

### LINKED ISSUE #1811: [ML] Reactivate the checks once we have a dedicated way to evaluate persisted model on new data.
https://github.com/valeriy42/ml-cpp/blob/157916a3284beff044a40584d4987d7d932459df/lib/maths/CBoostedTreeImpl.cc#L1901-L1901

valeriy42:
This PR introduces a new task "predict" to initialize a trained model and use it only for predicting a new dataset, without updating the model itself. To this end, I implement a new factory method `CBoostedTreeFactory::buildForPredict` that handles the task-specific boosted tree generation.

This also enables us to reactivate the invariance checks, that were temporarily disabled.

I faced the problem of mapping the features from the model definition back to the encoded columns. To fix it temporarily, I deactivated the removal of unused features when the model definition is generated. However, since it is incompatible with our current model definition schema, we still need to properly fix it. I created #1869 to have a reminder.

Closes #1811.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeRunner.cc
@@ -73,9 +73,10 @@ const CDataFrameAnalysisConfigReader& CDataFrameTrainBoostedTreeRunner::paramete
                                CDataFrameAnalysisConfigReader::E_OptionalParameter);
         theReader.addParameter(EARLY_STOPPING_ENABLED,
                                CDataFrameAnalysisConfigReader::E_OptionalParameter);
-        theReader.addParameter(
-            TASK, CDataFrameAnalysisConfigReader::E_OptionalParameter,
-            {{TASK_TRAIN, int{ETask::E_Train}}, {TASK_UPDATE, int{ETask::E_Update}}});
+        theReader.addParameter(TASK, CDataFrameAnalysisConfigReader::E_OptionalParameter,
+                               {{TASK_TRAIN, int{ETask::E_Train}},
+                                {TASK_UPDATE, int{ETask::E_Update}},
+                                {TASK_PREDICT, int{ETask::E_Predict}}});
         return theReader;
     }()};
     return PARAMETER_READER;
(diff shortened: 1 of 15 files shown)


lib/maths/CBoostedTreeFactory.cc
```diff
@@ -134,6 +134,36 @@ CBoostedTreeFactory::buildForTrain(core::CDataFrame& frame, std::size_t dependen
...
 
+CBoostedTreeFactory::TBoostedTreeUPtr
+CBoostedTreeFactory::buildForPredict(core::CDataFrame& frame, std::size_t dependentVariable) {
+
+    m_TreeImpl->m_DependentVariable = dependentVariable;
+
+    skipIfAfter(CBoostedTreeImpl::E_NotInitialized,
+                [&] { this->initializeMissingFeatureMasks(frame); });
+
+    this->prepareDataFrameForTrain(frame);
```
tveasey, line 145: This looks surprising. Is this needed for the weight column? I actually wonder if there is a case for having a special setup for predict since the other rows are not needed.

lib/maths/CBoostedTreeFactory.cc
```diff
@@ -134,6 +134,36 @@ CBoostedTreeFactory::buildForTrain(core::CDataFrame& frame, std::size_t dependen
...
+
+    m_TreeImpl->m_InitializationStage != CBoostedTreeImpl::E_NotInitialized
+        ? this->skipProgressMonitoringFeatureSelection()
+        : this->startProgressMonitoringFeatureSelection();
```
tveasey, line 149: Do we need this. It seems odd to me to have feature selection popping up in monitoring for prediction tasks.

lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -357,20 +359,26 @@ void CDataFrameTrainBoostedTreeRunner::runImpl(core::CDataFrame& frame) {
             treeRestored = this->restoreBoostedTree(frame, dependentVariableColumn,
                                                     restoreSearcher);
         }
-        if (treeRestored == false) {
+        if (treeRestored == false && m_Task == E_Train) {
             m_BoostedTree = m_BoostedTreeFactory->buildForTrain(frame, dependentVariableColumn);
         }
+        if (m_Task == E_Predict) {
+            m_BoostedTree = m_BoostedTreeFactory->buildForPredict(frame, dependentVariableColumn);
+        }
```
tveasey, line 367: Can these setup tasks move into the switch statements below (seems cleaner if they can), or do they need to precede validate?

tveasey (commented):
Overall looks good. I had a couple of minor comments, but this is basically a nice clean up.

tveasey (approved):

PR #1870, opened 2021-04-26
[ML] Apply tree depth constraint to hyperparameter search bounding box setup

tveasey:
Following on from #1867, we can and should be imposing the minimum depth constraint to the hyperparameter search bounding box. This was incorrectly applied before and also fixes the issue with reproducibility based on user overrides. This is a bit cleaner than applying the constraint magically in the code to adjust hyperparameters.


valeriy42 (approved):
LGTM

PR #1872, opened 2021-04-26
[ML] Fail gracefully on encountering unexpected state

tveasey:
We were missing some checks that member state had been properly initialised to fail gracefully when restoring one of our model types for anomaly detection.

This also fixes a potential uninitialised variable problem (which generates a warning from clang), although would only happen for malformed state.


droberts195 (approved):
LGTM

PR #1879, opened 2021-04-27
[ML] Restore blob for incremental training in place

tveasey:
There were a few issues with restoring the data blob into a local member of `CBoostedTreeImpl`:
1. We needed to copy it again to the data frame we actually use for training (it may be possible to move into place, but there are other concerns),
2. For memory estimation we don't want to actually be restoring or even sending the model so we need to disable this operation when we make the runner. Passing an optional data frame is a reasonable way to achieve this,
3. We were bypassing all the logic around how the data frame is stored (this would be ok because we only ever plan to use in core storage for boosted tree training for performance reasons, but it is better to make this explicit).


include/api/CDataFrameAnalysisRunner.h
```diff
@@ -192,19 +208,41 @@ class API_EXPORT CDataFrameAnalysisRunner {
...
 
-    TRunnerUPtr make(const CDataFrameAnalysisSpecification& spec) const;
+    //! Create a new runner object from \p spec.
+    //!
+    //! \param[in] spec The analysis specification.
+    //! \param[out] frameAndDirectory If non-null a data frame is created which
+    //! is suitable for the analysis together with the directory handle, if it
+    //! is stored on disk, and written to this.
+    TRunnerUPtr make(const CDataFrameAnalysisSpecification& spec,
+                     TDataFrameUPtrTemporaryDirectoryPtrPr* frameAndDirectory = nullptr) const;
```
tveasey, line 225: A significant part of this change (particular in the test code) relates to handing off the creation of the data frame on which to run the analysis to this class. This is needed because we can restore data if we are running incremental training. This is optional because we don't want to restore or create a data frame if we are just estimating memory.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -71,14 +71,14 @@ bool getBool(const rapidjson::Value& element, const std::string& tag, bool fallb
     return element[tag].GetBool();
 }
 
-std::stringstream decompressStream(boost::iostreams::stream_buffer<Device>& buffer) {
-    std::stringstream decompressedStream;
+auto decompressStream(boost::iostreams::stream_buffer<Device>& buffer) {
+    auto decompressedStream = std::make_shared<std::stringstream>();
```
droberts195, line 75: Is this change really necessary?  The returned stringstream will be moved, so returning by value won't be copying the contents.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -117,19 +118,19 @@ void CDataSummarizationJsonSerializer::addToJsonStream(TGenericLineWriter& write
...
     writer.Key(JSON_NUM_COLUMNS_TAG);
-    writer.Uint64(m_Frame.numberColumns());
+    writer.Uint64(m_NumberColumns);
 
     writer.Key(JSON_COLUMN_NAMES_TAG);
     writer.StartArray();
-    for (const auto& columnName : m_Frame.columnNames()) {
-        writer.String(columnName);
+    for (std::size_t i = 0; i < m_NumberColumns; ++i) {
+        writer.String(m_Frame.columnNames()[i]);
```
droberts195, line 126: Why can't range for loops be used here any more?

droberts195 (approved):
LGTM

Just a couple of minor questions

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: Ultimately, I don't think we want to be copying this into a `stringstream` at all. In fact, I started to make this change, which is relatively easy, but then thought, probably we should actually be reusing `CStateDecompressor` here and using a handler rather than parsing everything into intermediate representation.

I decided not to focus on that in this change, but didn't revert this small change because it represents a step in that direction.

droberts195: Yes, agreed it would be nice to reuse the pure streaming implementation we have rather than hold intermediate states in full.

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: This was an oversight in the original implementation. The data frame contains extra columns which it uses to cache information during training. These should not be serialised. So here `m_NumberColumns < m_Frame.columnNames().size()`. On reflection, it is worth adding a comment for this.

tveasey: See [2edd458](https://github.com/elastic/ml-cpp/pull/1879/commits/2edd458f6acd57dd4da9073af66de812d78afa5d).

PR #1887, opened 2021-04-30
[ML] Fix CStringStore unit test

### LINKED ISSUE #1885: [CI] CStringStoreTest.testStringStore failed
Build: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request/380/

Branches: 7.12

Failure:

```
[Exception] - critical check strings.size() == CStringStore::names().m_Strings.size() has failed [16 != 15]
 == [File] - CStringStoreTest.cc
 == [Line] -167
```

droberts195:
We have seen numerous failures of CStringStoreTest.testStringStore
caused by pruning not working as it should.

These failures invariably occur on aarch64 (both Mac and Linux).

The problem is that the pruning relies on std::shared_ptr::unique()
and this is not accurate. (In fact C++20 removes this method.)

The reason it's not accurate is that it returns the usage count
in a relaxed manner, i.e. returning the last value visible to the
current thread, but not forcing a resync between threads first.
The x86 architecture is more forgiving of theoretical memory
ordering bugs like this than the aarch64 architecture.

The solution is to change the use count in the thread that's
checking its value to be sure that this thread has accurate
visibility of the value.  Obviously this is inefficient, so we
only do this in unit test code.

Fixes #1885

### DIFF
--- a/include/core/CStoredStringPtr.h
+++ b/include/core/CStoredStringPtr.h
@@ -51,6 +51,8 @@ public:
     explicit operator bool() const noexcept;
 
     //! Is there only one pointer for this stored string?
+    //! This method is inefficient and should only be used in unit test
+    //! code (and then only infrequently).
     bool isUnique() const noexcept;
 
     //! Equality operator for NULL.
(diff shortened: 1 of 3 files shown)


edsavage (approved):
LGTM

droberts195:
There was another failure after this PR was merged: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+master+macos-aarch64+debug/16/

After further debugging on the M1 Mac Mini I am pretty sure #1888 in conjunction with this change with resolve the problem.

I will backport both PRs together after both are merged to master, hence marking this one as `backport pending`.

PR #1888, opened 2021-05-03
[ML] Sync writing atomic more strongly in CStringStore

### LINKED ISSUE #1885: [CI] CStringStoreTest.testStringStore failed

droberts195:
Decrementing the write count with a relaxed memory order
on aarch64 can result in other threads not noticing for
a very long time.  Instead the "release" memory order
needs to be used.  (As we never saw this problem on x86
I suspect this is a case of us taking advantage of
atomics on x86 offering better synchronization than
required by the language constructs.)

Followup to #1887
Fixes #1885

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -60,6 +60,8 @@
   training. (See {ml-pull}1855[#1855].)
 * Fail gracefully on encountering unexpected state in restore from snapshot for anomaly
   detection. (See {ml-pull}1872[#1872].)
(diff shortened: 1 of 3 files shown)


tveasey (approved):
LGTM.

PR #1889, opened 2021-05-04
[7.x][ML] Fix CStringStore unit test

droberts195:
We have seen numerous failures of CStringStoreTest.testStringStore
caused by pruning not working as it should.

These failures invariably occur on aarch64 (both Mac and Linux).

The problem is that the pruning relies on std::shared_ptr::unique()
and this is not accurate. (In fact C++20 removes this method.)

The reason it's not accurate is that it returns the usage count
in a relaxed manner, i.e. returning the last value visible to the
current thread, but not forcing a resync between threads first.
The x86 architecture is more forgiving of theoretical memory
ordering bugs like this than the aarch64 architecture.

The solution is to change the use count in the thread that's
checking its value to be sure that this thread has accurate
visibility of the value.  Obviously this is inefficient, so we
only do this in unit test code.

Backport of #1887

### DIFF
--- a/include/core/CStoredStringPtr.h
+++ b/include/core/CStoredStringPtr.h
@@ -51,6 +51,8 @@ public:
     explicit operator bool() const noexcept;
 
     //! Is there only one pointer for this stored string?
+    //! This method is inefficient and should only be used in unit test
+    //! code (and then only infrequently).
     bool isUnique() const noexcept;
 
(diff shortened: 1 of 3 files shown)


droberts195:
The test failure is likely impacted by the changes of #1888, so I will merge this PR despite the failure but then quickly backport #1888 to hopefully resolve the failure.

PR #1893, opened 2021-05-10
[ML] Make atomic operations safer for aarch64

droberts195:
This PR is a followup to #1888 which makes one more relaxed
store of an atomic variable in the CStringStore class use
release memory ordering rather than relaxed.

Additionally, a relaxed load is changed to use acquire memory
ordering in the CIntegration class.

These changes only have an effect on the aarch64 architecture.

Relates #1885


edsavage (approved):
LGTM

Issue #1869, opened 2021-04-26
[ML] Map feature names to correct encoding using definition only


valeriy42:
https://github.com/valeriy42/ml-cpp/blob/121a3e51362760b248957c5cd2e13d616703e3ad/lib/api/CBoostedTreeInferenceModelBuilder.cc#L88-L88

PR #1894, opened 2021-05-10
[ML] Improve error checking when reading trainable model

tveasey:
If fields are missing or the type doesn't match then rapidjson accessors result in undefined behaviour when compiled in release mode. We do not trust our state document and so need to verify that every field we expect exists and has the appropriate type. In the retrainable model state we were not reliably doing this. In this change I migrate to wrap all accessors in methods which check that both the fields exist and their type is correct. I also renamed the classes to match naming the names used for similar classes in the api library.

### DIFF
--- a/include/api/CDataSummarizationJsonSerializer.h
+++ b/include/api/CDataSummarizationJsonSerializer.h
@@ -23,24 +23,32 @@ class CDataFrame;
 }
 namespace api {
 
-//! \brief Class serializes and deserializes data summarization using chunked JSON blob.
-//! Data summarization contains data rows as well as categorical encoding information.
-// TODO #1849 chunking is not supported yet.
-class API_EXPORT CDataSummarizationJsonSerializer final
-    : public CSerializableToJsonDocumentCompressed {
+//! \brief Class which writes the data summarization as a compressed, base64 encoded
+//! and chunked JSON blob.
+//!
+//! DESCRIPTION:\n
+//! The data summarization contains data rows as well as the feature value encoding
+//! information.
+class API_EXPORT CDataSummarizationJsonWriter final : public CSerializableToJsonDocumentCompressed {
 public:
-    CDataSummarizationJsonSerializer(const core::CDataFrame& frame,
-                                     core::CPackedBitVector rowMask,
-                                     std::size_t numberColumns,
-                                     std::stringstream encodings);
+    CDataSummarizationJsonWriter(const core::CDataFrame& frame,
+                                 core::CPackedBitVector rowMask,
+                                 std::size_t numberColumns,
+                                 std::stringstream encodings);
 
-    CDataSummarizationJsonSerializer(const CDataSummarizationJsonSerializer&) = delete;
-    CDataSummarizationJsonSerializer&
-    operator=(const CDataSummarizationJsonSerializer&) = delete;
+    CDataSummarizationJsonWriter(const CDataSummarizationJsonWriter&) = delete;
+    CDataSummarizationJsonWriter& operator=(const CDataSummarizationJsonWriter&) = delete;
 
+    //! Write the JSON data summarisation.
     void addToJsonStream(TGenericLineWriter& writer) const override;
+
+    //! Write a compressed and chunked JSON data summarisation.
     void addToDocumentCompressed(TRapidJsonWriter& writer) const override;
+
+    //! \name Test Only
+    //@{
     std::string jsonString() const;
+    //@}
 
 private:
     core::CPackedBitVector m_RowMask;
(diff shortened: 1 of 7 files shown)


include/api/CDataSummarizationJsonSerializer.h
```diff
@@ -23,24 +23,32 @@ class CDataFrame;
...
-// TODO #1849 chunking is not supported yet.
-class API_EXPORT CDataSummarizationJsonSerializer final
-    : public CSerializableToJsonDocumentCompressed {
+//! \brief Class which writes the data summarization as a compressed, base64 encoded
+//! and chunked JSON blob.
+//!
+//! DESCRIPTION:\n
+//! The data summarization contains data rows as well as the feature value encoding
+//! information.
+class API_EXPORT CDataSummarizationJsonWriter final : public CSerializableToJsonDocumentCompressed {
```
droberts195, line 32: The class has been renamed but not the file or the include guards, so now there's a mismatch.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -27,60 +28,93 @@
...
+    if (value.HasMember(tag)) {
+        try {
+            return get(value[tag]);
+        } catch (const std::runtime_error& e) {
+            LOG_ERROR(<< "Field '" << tag << "' " << e.what() << ".");
+        }
+    } else {
+        LOG_ERROR(<< "Field '" << tag << "' is missing.");
+    }
+    throw std::runtime_error{""};
```
droberts195, line 60: I think this should contain a unique message.  Even if it is never expected to be seen, if it leaks out one day because the logic is changed elsewhere then it will be nice to be able to find out where it came from.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -185,206 +218,181 @@ void CDataSummarizationJsonSerializer::addToJsonStream(TGenericLineWriter& write
...
         return nullptr;
     }
-    auto& compressedDataSummarization = doc[JSON_COMPRESSED_DATA_SUMMARIZATION_TAG];
-    if (compressedDataSummarization.HasMember(JSON_DATA_SUMMARIZATION_TAG) == false ||
-        compressedDataSummarization[JSON_DATA_SUMMARIZATION_TAG].IsString() == false) {
-        LOG_ERROR(<< "Field " << JSON_DATA_SUMMARIZATION_TAG << " not found or is not a string.");
-        return nullptr;
+    try {
+        return dataSummarizationFromJson(*istream, frame);
+    } catch (...) {
```
droberts195, line 229: This is silently eating `std::bad_alloc`.  It's also silently eating Windows structured exceptions that Windows APIs might throw (only `...` catches these using the C++ syntax).

We are in a strange situation because our coding guidelines say exceptions shouldn't be used, so we assume that in the places where we _do_ use them that we can be very simplistic in how we do it.  But the more places where we use them the less appropriate the simplistic handling becomes.

If you only expect this to ever catch `std::runtime_exception` then catching that explicitly would solve the problems of eating `std::bad_alloc` and Windows structured exceptions.  If you're expecting other types of exception like `std::logic_error` or `std::bad_cast` then it's more painful to handle nicely.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -185,206 +218,181 @@ void CDataSummarizationJsonSerializer::addToJsonStream(TGenericLineWriter& write
...
+        rapidjson::Document doc;
+        doc.ParseStream(isw);
+        auto compressedDataSummarization =
+            ifExists(JSON_COMPRESSED_DATA_SUMMARIZATION_TAG, getObject, doc);
+        TStreamBuffer buffer{ifExists(JSON_DATA_SUMMARIZATION_TAG, getString,
+                                      compressedDataSummarization),
+                             ifExists(JSON_DATA_SUMMARIZATION_TAG, getStringLength,
+                                      compressedDataSummarization)};
+        return dataSummarizationFromJsonStream(decompressStream(buffer), frame);
+    } catch (...) {
```
droberts195, line 317: Same as previous comment regarding eating low level errors.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -185,206 +218,181 @@ void CDataSummarizationJsonSerializer::addToJsonStream(TGenericLineWriter& write
...
-                  << "' is missing or has an unexpected format.");
+CRetrainableModelJsonReader::TNodeVecVecUPtr
+CRetrainableModelJsonReader::bestForestFromJsonStream(TIStreamSPtr istream) {
+    if (istream == nullptr) {
         return nullptr;
     }
-    return encoder;
+    try {
+        return bestForestFromJson(*istream);
+    } catch (...) {
```
droberts195, line 330: Same as previous comment regarding eating low level errors.

lib/api/CDataSummarizationJsonSerializer.cc
```diff
@@ -396,35 +404,30 @@ CRetrainableModelJsonDeserializer::bestForestFromJsonStream(const core::CDataSea
...
+        rapidjson::Document doc;
+        doc.ParseStream(isw);
+        auto compressedDataSummarization = ifExists(
+            CInferenceModelDefinition::JSON_COMPRESSED_INFERENCE_MODEL_TAG, getObject, doc);
+        TStreamBuffer buffer{ifExists(CInferenceModelDefinition::JSON_DEFINITION_TAG,
+                                      getString, compressedDataSummarization),
+                             ifExists(CInferenceModelDefinition::JSON_DEFINITION_TAG,
+                                      getStringLength, compressedDataSummarization)};
+        return bestForestFromJsonStream(decompressStream(buffer));
+    } catch (...) {
```
droberts195, line 427: Same as previous comment regarding eating low level errors.

include/api/CDataSummarizationJsonSerializer.h
tveasey: This header previously contained two classes one for serialisation and one for deserialisation so there was already a mismatch here really. I think we should probably rename and maybe even split out the writing and reading class into separate files. The reason I didn't make this change in this PR is that I factored this out of fixing #1869 and don't want to make merging the two PRs a pain. I therefore propose to postpone this to that PR.

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: Ok that seems reasonable. However, I'd add that in this case any exception being thrown should probably trigger returning a null summarisation which causes the program to call terminate elsewhere. This is probably slightly safer IMO than just letting the exception propagate since, as you say, we don't in general know if it might be caught elsewhere (although in this case I'm pretty sure it isn't).

I actually think for restore throwing exceptions is a good way of handling bad or missing state (it made this error checking code much cleaner to write) and we want to fail as fast as possible. So my preferred solution would be to allow any restore code to throw (including all error handling where we currently return false from restore methods) have no try catch in any individual class restore and wrap all top level calls to read state with try catch. That is beyond the scope of this change, but I guess having targeted catch here is more in keeping with that approach.

tveasey: Done in [f2c3588](https://github.com/elastic/ml-cpp/pull/1894/commits/f2c35881e9ec52f95d64cd4ba16bc45c3712fed3).

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: Done in [f2c3588](https://github.com/elastic/ml-cpp/pull/1894/commits/f2c35881e9ec52f95d64cd4ba16bc45c3712fed3).

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: Done in [f2c3588](https://github.com/elastic/ml-cpp/pull/1894/commits/f2c35881e9ec52f95d64cd4ba16bc45c3712fed3).

lib/api/CDataSummarizationJsonSerializer.cc
tveasey: Done in [f2c3588](https://github.com/elastic/ml-cpp/pull/1894/commits/f2c35881e9ec52f95d64cd4ba16bc45c3712fed3).

PR #1896, opened 2021-05-11
[ML] Fix inference model feature encoder index mismatch

### LINKED ISSUE #1869: [ML] Map feature names to correct encoding using definition only

tveasey:
When we write out the inference model we remove any features which were included for training which weren't selected. This creates a mismatch between the model feature indices and the encoder indices. This change additionally persists a map from feature name to encoder index so we can recover the correct encodings to use in the best forest when restoring for retraining.

Fixes #1869.


droberts195 (approved):
LGTM

PR #1897, opened 2021-05-12
[ML] Read summarisation and model from chunked state

### LINKED ISSUE #1849: [ML] Support data chunking of CDataSummarizationJsonSerializer
At the moment we only support compressed best forest and data summarization documents without chunking. This functionality has to be extended to support chunking as well.

However, this issue is not a blocker and can be dealt with when we are ready to exchange data with java side.

https://github.com/valeriy42/ml-cpp/blob/270b7aa6ba852f88dee80a16afac557876be0a86/include/api/CDataSummarizationJsonSerializer.h#L28-L28

tveasey:
Currently, we write large state documents out chunked but then only try and read the first chunk when restoring for incremental training. This means if our compressed state exceeds the maximum document size we fail to restore.

This change introduces a new `CSerializableFromCompressedChunkedJson` class which dechunks the compressed state into a temporary buffer. We expand this stream as we read it, so provided we read one document at a time (there is an enhancement for this), then the memory overhead is small if the compression factor is large, which it is.

I also took the opportunity to avoid all the temporary copies of compressed state by reading directly into a char array and then using pointers into this array to initialise the `rapidjson::String` objects during persist.

Fixes #1849.

### DIFF
--- a/lib/api/CSerializableToJson.cc
+++ b/lib/api/CSerializableToJson.cc
@@ -134,7 +134,7 @@ CSerializableFromCompressedChunkedJson::rawJsonStream(const std::string& compres
                 buffer.write(ifExists(payloadTag, getAsStringFrom, chunk),
                              ifExists(payloadTag, getStringLengthFrom, chunk));
                 done = chunk.HasMember(JSON_EOS_TAG);
-            } while (done == false && inputStream->eof() == false);
+            } while (done == false);
             consumeSpace(*inputStream);
             return decodeAndDecompress(buffer);
         } catch (const std::runtime_error& e) { LOG_ERROR(<< e.what()); }
--- a/lib/api/CSerializableToJson.cc
+++ b/lib/api/CSerializableToJson.cc
@@ -53,6 +53,12 @@ auto decodeAndDecompress(std::istream& inputStream) {
     return result;
 }
 
+void consumeSpace(std::istream& stream) {
+    while (std::isspace(stream.peek()) != 0) {
+        stream.get();
+    }
+}
+
 const std::string JSON_DOC_NUM_TAG{"doc_num"};
 const std::string JSON_EOS_TAG{"eos"};
 }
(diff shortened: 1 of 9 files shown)


lib/api/CSerializableToJson.cc
```diff
@@ -7,75 +7,139 @@
...
+                                                      const std::string& payloadTag,
+                                                      TIStreamPtr inputStream,
+                                                      std::iostream& buffer) {
+    if (inputStream != nullptr) {
+        rapidjson::IStreamWrapper isw{*inputStream};
+        try {
+            rapidjson::Document doc;
+            bool done{false};
+            do {
+                doc.ParseStream<rapidjson::kParseStopWhenDoneFlag>(isw);
```
droberts195, line 132: I think we need to check for a parse error after this as the RapidJSON way of doing things is to return a doc that contains the error information.  `ifExists` might prevent undefined behaviour in this case but it will lose the more useful diagnostic information of what was wrong.  There may well be other places in our code where we should be doing this.

```suggestion
                doc.ParseStream<rapidjson::kParseStopWhenDoneFlag>(isw);
                if (doc.HasParseError()) {
                    const char* error{rapidjson::GetParseError_En(doc.GetParseError())};
                    throw std::runtime_exception{"Error parsing JSON at offset " + std::to_string(doc.GetErrorOffset()) + ": " + ((error != nullptr) ? error : "No message")};
                }
```

droberts195 (commented):
Just one comment, about error reporting.

The other thing I realised is that we really ought to be making an allowance for the memory required for persistence and restoration in our memory estimates, otherwise we could cause the OOM killer to kill a process on memory-constrained machines, such as cloud containers.  But I will open a separate issue for that.

lib/api/CSerializableToJson.cc
tveasey: Good suggestion! This behaviour was just copied over from the original implementation. There are probably a couple of other places we should be doing the same.

tveasey: This was done in [65348db](https://github.com/elastic/ml-cpp/pull/1897/commits/65348dbe39335647335e2093a39f2fe0e8fdfca5).

tveasey:
> The other thing I realised is that we really ought to be making an allowance for the memory required for persistence and restoration in our memory estimates, otherwise we could cause the OOM killer to kill a process on memory-constrained machines, such as cloud containers.

I worried about this too. I think we may get away with this: training introduces other sources of memory overhead which aren't present at the time state is persisted and restored. This change also cuts down the number of simultaneous copies of state we hold. However, at the very least we may need to consider limit = max("peak during persist / restore", "during train"). #1898 is also important for reducing "peak during restore".

tveasey:
Thanks for the review @droberts195! I incorporated your suggestion.

droberts195 (approved):
LGTM

PR #1901, opened 2021-05-24
[ML] Incremental binomial logistic loss

tveasey:
This implements the incremental version of the binomial logistic loss. We add on a penalty which is the cross-entropy of the old and new tree predictions on the original training data.

### DIFF
--- a/lib/maths/CBoostedTreeLoss.cc
+++ b/lib/maths/CBoostedTreeLoss.cc
@@ -4,6 +4,7 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
+#include "maths/CBoostedTreeUtils.h"
 #include <maths/CBoostedTreeLoss.h>
 
 #include <core/CPersistUtils.h>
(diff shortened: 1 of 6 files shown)


lib/maths/CBoostedTreeLoss.cc
```diff
@@ -4,6 +4,7 @@
  * you may not use this file except in compliance with the Elastic License.
  */
 
+#include "maths/CBoostedTreeUtils.h"
```
droberts195, line 7:
```suggestion
#include <maths/CBoostedTreeUtils.h>
```

droberts195 (approved):
LGTM - just one nit about using angle brackets for library includes

lib/maths/CBoostedTreeLoss.cc
tveasey: Done in [c83ffe6](https://github.com/elastic/ml-cpp/pull/1901/commits/c83ffe6f36344efce60c1b32d68f0cb671b2e48a).

Issue #1724, opened 2021-02-04
[ML] Higher weighting for tokens at the beginning of messages during categorization


droberts195:
Many years ago we considered the idea of giving tokens near the beginning of messages higher weightings in categorization.  We rejected the idea, as we had a few data sets at the time that we got worse results for by doing this.

However, more recently we have seen data sets where giving higher weighting to tokens near the beginning of messages would give much better results.  Potentially the weight should taper to zero after a large number of tokens, so that subsequent tokens play no part whatsoever in categorization.

Some research is needed to see what weighting and tapering strategy would work best.  We don't want to change our categorization results _too_ dramatically from what they are today.

Also, for BWC this functionality would need to be enabled by an option that would default to false for jobs created in prior versions.

An example of a message that doesn't categorize well with no additional weighting for early tokens is:

```
a client request body is buffered to a temporary file /tmp/client-body/0000021894, client: 10.8.0.12, server: apm.35.205.226.121.ip.es.io, request: "POST /intake/v2/events HTTP/1.1", host: "apm.35.205.226.121.ip.es.io"
10.8.0.12 - - [29/Nov/2020:21:34:55 +0000] "POST /intake/v2/events HTTP/1.1" 202 0 "-" "elasticapm-dotnet/1.5.1 System.Net.Http/4.6.28208.02 .NET_Core/2.2.8" 27821 0.002 [default-apm-apm-server-8200] [] 10.8.1.19:8200 0 0.001 202 f961c776ff732f5c8337530aa22c7216
10.8.0.14 - - [29/Nov/2020:21:34:56 +0000] "POST /intake/v2/events HTTP/1.1" 202 0 "-" "elasticapm-python/5.10.0" 3594 0.002 [default-apm-apm-server-8200] [] 10.8.1.18:8200 0 0.001 202 61feb8fb9232b1ebe54b588b95771ce4
10.8.4.90 - - [29/Nov/2020:21:34:56 +0000] "OPTIONS /intake/v2/rum/events HTTP/2.0" 200 0 "http://opbeans-frontend:3000/dashboard" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Cypress/3.3.1 Chrome/61.0.3163.100 Electron/2.0.18 Safari/537.36" 292 0.001 [default-apm-apm-server-8200] [] 10.8.1.19:8200 0 0.000 200 5fbe8cd4d217b932def1c17ed381c66b
[...]

PR #1903, opened 2021-05-26
[ML] Boost weighting for multiple adjacent words.

edsavage:
In an effort to categorise the most important parts of a message, give a boost to the weighting when 3 or more dictionary words in the message are adjacent one another.

Relates to #1724

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -41,6 +41,11 @@
 
 == {es} version 7.14.0
 
+=== Enhancements
+
+* Favour multiple adjacent dictionary words when performing categorisation. (See
+  {ml-pull}1903[#1903].)
+
 === Bug Fixes
 
 * Make atomic operations safer for aarch64. (See {ml-pull}1893[#1893].)
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -43,7 +43,7 @@
 
 === Enhancements
 
-* Favour multiple adjacent dictionary words when performing categorisation. (See
+* Give higher weight to multiple adjacent dictionary words when performing categorization. (See
   {ml-pull}1903[#1903].)
 
 === Bug Fixes
--- a/include/core/CWordDictionary.h
+++ b/include/core/CWordDictionary.h
@@ -115,7 +115,8 @@ public:
             }
 
             std::size_t weight = (partOfSpeech == SPECIAL_PART1) ? EXTRA_WEIGHT1 : DEFAULT_EXTRA_WEIGHT;
-            weight *= (m_NumOfAdjacentDictionaryWords > 1 ? ADJACENT_PARTS_BOOST : 1);
+            std::size_t boost = (m_NumOfAdjacentDictionaryWords > 1 ? ADJACENT_PARTS_BOOST : 1);
+            weight *= boost;
 
             m_NumOfAdjacentDictionaryWords++;
 
--- a/include/core/CWordDictionary.h
+++ b/include/core/CWordDictionary.h
@@ -115,7 +115,8 @@ public:
             }
 
             std::size_t weight = (partOfSpeech == SPECIAL_PART1) ? EXTRA_WEIGHT1 : DEFAULT_EXTRA_WEIGHT;
-            std::size_t boost = (m_NumOfAdjacentDictionaryWords > 1 ? ADJACENT_PARTS_BOOST : 1);
+            std::size_t boost =
+                (m_NumOfAdjacentDictionaryWords > 1 ? ADJACENT_PARTS_BOOST : 1);
             weight *= boost;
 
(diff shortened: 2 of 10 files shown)


lib/core/unittest/CStringSimilarityTesterTest.cc
```diff
@@ -418,4 +348,56 @@ BOOST_AUTO_TEST_CASE(testWeightedEditDistance) {
...
+    BOOST_REQUIRE_EQUAL(size_t(13), sst.weightedEditDistance(serviceStart, sourceShutDown2));
+
+    BOOST_REQUIRE_EQUAL(size_t(21), sst.weightedEditDistance(noImageData, serviceStart));
+    BOOST_REQUIRE_EQUAL(size_t(21), sst.weightedEditDistance(serviceStart, noImageData));
+
+    BOOST_REQUIRE_EQUAL(size_t(21), sst.weightedEditDistance(noImageData, sourceShutDown1));
+    BOOST_REQUIRE_EQUAL(size_t(21), sst.weightedEditDistance(sourceShutDown1, noImageData));
+
+    BOOST_REQUIRE_EQUAL(size_t(23), sst.weightedEditDistance(serviceStart, empty));
+    BOOST_REQUIRE_EQUAL(size_t(23), sst.weightedEditDistance(empty, serviceStart));
```
droberts195, line 401: You don't need to cast the expected values with Boost.Test.  (The places where we do this are left over from the cppunit days.)

droberts195:
Some of the Java integration tests that have failed suggest there's a bug somewhere.

For example, `testStopOnWarn` failed with this:

```
Expected: <warn>
 but: was <ok>

at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:18)
at org.junit.Assert.assertThat(Assert.java:956)
at org.junit.Assert.assertThat(Assert.java:923)
at org.elasticsearch.xpack.ml.integration.CategorizationIT.lambda$testStopOnWarn$1(CategorizationIT.java:400)
at org.elasticsearch.xpack.ml.integration.CategorizationIT.testStopOnWarn(CategorizationIT.java:426)
```

Given that `testStopOnWarn` feeds a single message repeatedly for each partition it seems impossible that changes to weighting of tokens could change whether this test sees 100 of the same message in a row.  So it seems that an unintentional change to the warn/ok status has been introduced somewhere.

docs/CHANGELOG.asciidoc
```diff
@@ -41,6 +41,11 @@
 
 == {es} version 7.14.0
 
+=== Enhancements
+
+* Favour multiple adjacent dictionary words when performing categorisation. (See
```
droberts195, line 46: Text on our public website needs to use US spellings:

```suggestion
* Give higher weight to multiple adjacent dictionary words when performing categorization. (See
```

include/core/CWordDictionary.h
```diff
@@ -72,34 +72,71 @@ class CORE_EXPORT CWordDictionary : private CNonCopyable {
...
+            if (partOfSpeech == E_NotInDictionary) {
+                m_NumOfAdjacentDictionaryWords = 0;
+                return 0;
+            }
+
+            std::size_t weight = (partOfSpeech == SPECIAL_PART1) ? EXTRA_WEIGHT1 : DEFAULT_EXTRA_WEIGHT;
+            std::size_t boost = (m_NumOfAdjacentDictionaryWords > 1 ? ADJACENT_PARTS_BOOST : 1);
+            weight *= boost;
+
+            m_NumOfAdjacentDictionaryWords++;
```
droberts195, line 121: nit: throughout the C++ codebase we use prefix increment when the original value is not used.

```suggestion
            ++m_NumOfAdjacentDictionaryWords;
```

lib/model/CTokenListCategory.cc
```diff
@@ -309,13 +309,15 @@ bool CTokenListCategory::updateOrderedCommonTokenIds(const TSizeSizePrVec& newTo
...
     for (std::size_t tryOrderedCommonTokenBeginIndex = m_OrderedCommonTokenBeginIndex;
          tryOrderedCommonTokenBeginIndex < m_OrderedCommonTokenEndIndex;
          ++tryOrderedCommonTokenBeginIndex) {
 
         std::size_t newIndex{0};
-        for (std::size_t commonIndex = tryOrderedCommonTokenBeginIndex;
-             commonIndex < m_OrderedCommonTokenEndIndex; ++commonIndex) {
+        std::size_t tryWeight{0};
+        std::size_t commonIndex = tryOrderedCommonTokenBeginIndex;
+        for (; commonIndex < m_OrderedCommonTokenEndIndex; ++commonIndex) {
```
droberts195, line 320: I think `commonIndex` can still be initialized in the `for` loop?  It doesn't seem to be used outside the loop.

droberts195 (approved):
LGTM if it passes CI now

Issue #1866, opened 2021-04-22
[ML] Introduce accept randomize_seed configuration parameter


valeriy42:
https://github.com/valeriy42/ml-cpp/blob/21b549dba18ffad899e2105193c997deefc3852c/lib/maths/CBoostedTreeImpl.cc#L274-L274

PR #1905, opened 2021-06-02
[ML] More testing and addressing some TODOs for incremental training

tveasey:
This reenables and extends the unit test to check the incremental training API. Whilst working on this test I found some state which wasn't properly initialised by the factory for incremental training and some state which wasn't properly reset by the factory when preparing a model to retrain. We were also miscounting the number of hyperparameters to tune for incremental training. Since, for results to match, it is necessary to reset the seed at the start of incremental train, I also took the opportunity to address #1866 and support a rng seed in the JSON configuration for training.

### DIFF
--- a/include/api/CDataFrameAnalysisConfigReader.h
+++ b/include/api/CDataFrameAnalysisConfigReader.h
@@ -93,8 +93,6 @@ public:
         //! Get a boolean parameter.
         bool fallback(bool value) const;
         //! Get an unsigned integer parameter.
-        std::uint64_t fallback(std::uint64_t fallback) const;
-        //! Get an unsigned integer parameter.
         std::size_t fallback(std::size_t value) const;
         //! Get a signed integer parameter.
         std::ptrdiff_t fallback(std::ptrdiff_t value) const;
--- a/include/api/CDataFrameAnalysisConfigReader.h
+++ b/include/api/CDataFrameAnalysisConfigReader.h
@@ -93,6 +93,8 @@ public:
         //! Get a boolean parameter.
         bool fallback(bool value) const;
         //! Get an unsigned integer parameter.
+        std::uint64_t fallback(std::uint64_t fallback) const;
+        //! Get an unsigned integer parameter.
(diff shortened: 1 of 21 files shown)


include/api/CDataFrameAnalysisConfigReader.h
```diff
@@ -93,6 +93,8 @@ class API_EXPORT CDataFrameAnalysisConfigReader {
         //! Get a boolean parameter.
         bool fallback(bool value) const;
         //! Get an unsigned integer parameter.
+        std::uint64_t fallback(std::uint64_t fallback) const;
+        //! Get an unsigned integer parameter.
```
droberts195, line 97: You may find this is problematic on 64 bit Windows, where `long` is still 32 bits, so `size_t` is probably exactly the same underlying type as `uint64_t`, so then this overload will clash with the one immediately below.

I think you should either limit the possible fallback types to just a couple, say `size_t` and `ptrdiff_t`, or else if there are going to be masses of overloads then provide one for every raw type: `short`, `int`, `long`, `long long`, `unsigned short`, `unsigned int`, `unsigned long`, `unsigned long long`.  Then you can call the methods with `size_t`, `uintN_t`, whatever, and the overload mechanism will pick the right overload for the platform.  This is what we had to do in `CStringUtils::typeToString` in the end, and what the standard library equivalents will do.  But if we can get away with just `size_t` and `ptrdiff_t` then that's less verbose.

droberts195 (approved):
LGTM but you might need to fix an overload resolution problem on Windows

include/api/CDataFrameAnalysisConfigReader.h
tveasey: I can get away with `size_t`: I just need to add a cast where it is used to initialise the seed. (Windows CI did compile so I guess the type is different, but I don't actually have strong opinions on this so happy to change.)

tveasey: Actually, I was wrong. Windows didn't compile for this reason, I was looking at the wrong output.

PR #1906, opened 2021-06-02
[ML] Improve robustness of tensor result processing

davidkyle:
Tensors are read via `accessors` which have template parameters type (float, int, etc) and a dimension count. If the type or dimension is wrong the accessor will throw. Previously the only support type was a hard coded float, this change inspects the underlying tensor type and will either create a `double` or `float32` accessor. If the tensor is not one of those types an error is returned.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -11,6 +11,7 @@
 
 #include <seccomp/CSystemCallFilter.h>
 
+#include <torch/csrc/api/include/torch/types.h>
 #include <ver/CBuildInfo.h>
 
 #include <api/CIoManager.h>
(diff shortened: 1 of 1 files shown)


bin/pytorch_inference/Main.cc
```diff
@@ -78,7 +79,8 @@ void writeTensor(torch::TensorAccessor<float, 1> accessor,
     jsonWriter.EndArray();
 }
 
-void writeTensor(torch::TensorAccessor<float, 2> accessor,
+template<typename T>
+void writeTensor(torch::TensorAccessor<T, 2>& accessor,
```
tveasey, line 83: Unrelated to this change, but it feels slightly cleaner to have this implemented in terms of `writeTensor(torch::TensorAccessor<T, 1>& accessor,...`. `operator[]` on `torch::TensorAccessor<T, N>` returns ``torch::TensorAccessor<T, N-1>` so this should just work. (This could be made fully generic, but not sure it is worthwhile at the moment.)

tveasey (approved):
Changes LGTM. I left a comment regarding `writeTensor` which feels like a slight tidy up while you are in the area.

bin/pytorch_inference/Main.cc
davidkyle: This is much better. In the nested for loop I was using a new accessor `accessor[i]` was created for every access `accessor[i][j]`, that object is now reused.

PR #1907, opened 2021-06-03
[ML] Split data summarisation writing and retrainable model reading

tveasey:
This implements the delayed suggestion in #1894 to split up the file containing the data summarisation writing and the retrainable model reading.

### DIFF
--- a/lib/api/CDataSummarizationJsonWriter.cc
+++ b/lib/api/CDataSummarizationJsonWriter.cc
@@ -126,7 +126,9 @@ void CDataSummarizationJsonWriter::addToJsonStream(TGenericLineWriter& writer) c
     }
     rapidjson::ParseResult ok(doc.Parse(encodings.str()));
     if (static_cast<bool>(ok) == false) {
-        LOG_ERROR(<< "Failed parsing encoding json. Please report this error.");
+        LOG_ERROR(<< "Failed parsing encoding json: "
+                  << rapidjson::GetParseError_En(doc.GetParseError())
+                  << ". Please report this error.");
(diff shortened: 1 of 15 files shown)


lib/api/CDataSummarizationJsonWriter.cc
```diff
@@ -0,0 +1,170 @@
...
+
+    rapidjson::Document doc;
+    std::stringstream encodings;
+    {
+        core::CJsonStatePersistInserter inserter{encodings};
+        m_Encodings.acceptPersistInserter(inserter);
+    }
+    rapidjson::ParseResult ok(doc.Parse(encodings.str()));
+    if (static_cast<bool>(ok) == false) {
+        LOG_ERROR(<< "Failed parsing encoding json. Please report this error.");
```
droberts195, line 129: I know you've just moved this from a different file rather than writing it from scratch in this PR, but it will be frustrating if we ever see this message with no details.

```suggestion
        LOG_ERROR(<< "Failed parsing encoding json. Please report this error: "
                  << rapidjson::GetParseError_En(doc.GetParseError())));
```

droberts195 (approved):
LGTM, although it would be good to make one error a bit more informative

lib/api/CDataSummarizationJsonWriter.cc
tveasey: Done in [e13e88c](https://github.com/elastic/ml-cpp/pull/1907/commits/e13e88c0ffd540257f2b4cecd8ede25e1e25112c).

PR #1909, opened 2021-06-03
[ML] Compute bucket counts for 1s bucket spans

### LINKED ISSUE #1908: [ML] Bucket event counts are always 0 for 1 second bucket spans
Analysis of the results of simple count anomaly detection jobs where the bucket span has been configured to be 1s shows that the value for `event_count` for every bucket is 0. 

The log files contain errors similar to:
```
[2021-06-02T12:08:30,167][ERROR][o.e.x.m.p.l.CppLogMessageHandler] [node-0] [cloned] [autodetect/14211] [CCountingModel.cc@122] No statistics at 1622159474, current bucket = [1622159475,1622159476)
[2021-06-02T12:08:30,168][ERROR][o.e.x.m.p.l.CppLogMessageHandler] [node-0] [cloned] [autodetect/14211] [CJsonOutputWriter.cc@184] Simple count detector has no current rate
```

edsavage:
Fix a bug where calls to calculate the current bucket count for simple count
models always failed to return a value when the bucket span was one
second.

fixes #1908

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -44,7 +44,7 @@
 
 === Bug Fixes
 
-* Ensure bucket count calculation is performed for simple count jobs with 1s bucket spans.
+* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
   (See {ml-pull}1908[#1908].)
 
 == {es} version 7.14.0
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -42,6 +42,11 @@
   - the C++ front end to PyTorch and performs inference on models stored in the 
   TorchScript format. (See {ml-pull}1902[#1902].)
 
+=== Bug Fixes
+
+* Ensure bucket count calculation is performed for simple count jobs with 1s bucket spans.
```
droberts195, line 47:
```suggestion
* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
```

tveasey (approved):
Thanks for debugging this Ed! Change LGTM.

PR #1911, opened 2021-06-07
[ML] Improve incremental training test coverage

tveasey:
This builds out the unit testing of incremental training QoR and fixes the bugs in:
1. The initialisation of the loss derivatives,
2. The initialisation of the predictions and previous predictions,
3. The value of `eta` to use for the retrained tree.

We also need to ensure that the feature bag includes the features used in the original tree so we can make the same choice again if it is appropriate.

(The test is still WIP, but I think there are enough fixes in this PR to be worthwhile pushing, so it doesn't hold up evaluation on real data sets.)


droberts195 (approved):
LGTM

PR #1912, opened 2021-06-08
[ML] Finish up testing of incremental training with MSE loss

tveasey:
This finishes up testing incremental training for MSE. It also fixes bugs in
1. The calculation of the penalty for changing the tree
2. The calculation of the candidate splits (which need the loss derivatives to be calculated first)
3. The tree leaf values, which were being computed after removing the next tree to retrain predictions

We should also have been persisting the new training data row mask for failover and I've unified some train and incremental train code by always filling in the new training data row mask.

Finally, a failing unit test showed up some improvements to make to the approach to minimise the incremental logistic loss. In particular, the term due to the cross-entropy between the old and new tree predictions can be treated completely independently from the term due to the overall prediction errors so we can maintain a separate bucketing of the tree predictions and use this to approximate the contribution from this term to the loss much more accurately.

### DIFF
--- a/lib/maths/CBoostedTreeLoss.cc
+++ b/lib/maths/CBoostedTreeLoss.cc
@@ -393,7 +393,7 @@ CArgMinBinomialLogisticLossIncrementalImpl::objective() const {
         };
     }
 
-    // This is true if the forest predictions were identical.
+    // This is true if all the forest predictions were identical.
     if (bucketWidth(this->predictionMinMax()) == 0.0) {
         double prediction{mid(this->predictionMinMax())};
         return [prediction, this](double weight) {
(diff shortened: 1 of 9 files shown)


lib/maths/CBoostedTreeLoss.cc
```diff
@@ -368,38 +375,81 @@ void CArgMinBinomialLogisticLossIncrementalImpl::merge(const CArgMinLossImpl& ot
 CArgMinBinomialLogisticLossIncrementalImpl::TObjective
 CArgMinBinomialLogisticLossIncrementalImpl::objective() const {
 
-    // This is true if and only if all the predictions were identical. In this
-    // case we only need one pass over the data and can compute the optimal
-    // value from the counts of the two categories and the mean tree predictions.
-    if (this->bucketWidth() == 0.0) {
-        double prediction{this->midPrediction()};
-        return [prediction, this](double weight) {
+    // This is true if all the forst and tree predictions were identical.
```
valeriy42, line 378:
```suggestion
    // This is true if all the forest and tree predictions were identical.
```

valeriy42 (approved):
LGTM

PR #1932, opened 2021-06-18
[ML] Data summarization num rows in model metadata

valeriy42:
We write a number of rows in data summarization as a part of model metadata. This helps to estimate the correct number of rows for incremental learning.


tveasey (approved):
LGTM

PR #1934, opened 2021-06-18
[ML] Don't link libnsl on Linux when USE_NET is defined

### LINKED ISSUE #1933: [ML][PyTorch] Missing libnsl.so.1 on CentOS 8
The `pytorch_inference` app links to libnsl for the network to host order functions (ntohl) but libnsl is no longer included by default on CentOS 8 and [RHEL 8](https://www.ibm.com/docs/en/cloud-app-management/2019.4.0?topic=systems-preinstallation-linux)

davidkyle:
#1933 shows that libnsl is no longer present on newer Linux distributions. The library should not be required on Linux as we do not use the networking functions.

This change makes setting `USE_NET=1` in a makefile a no-op for Linux and macOS, on Windows `WS2_32.lib` is linked (for the network to host byte order functions in one case). 

Closes #1933

### DIFF
--- a/mk/linux.mk
+++ b/mk/linux.mk
@@ -48,7 +48,7 @@ SHELL_SCRIPT_EXT=.sh
 UT_TMP_DIR=/tmp/$(LOGNAME)
 RESOURCES_DIR=resources
 LOCALLIBS=-lm -lpthread -ldl -lrt
-NETLIBS=-lnsl
+NETLIBS=
 BOOSTVER=1_71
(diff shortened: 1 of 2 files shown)


davidkyle:
I added the `>non-issue` label as I not sure this is worth change logging. None of our current or previous code links `libnsl` as it was optimised out, this change fixes a problem in the unreleased `pytorch_inference` app

droberts195 (approved):
LGTM

PR #1937, opened 2021-06-24
[ML] Time inference calls in the pytorch app

davidkyle:
Adds a stopwatch timer and records the time in milliseconds for each inference call. The time is reported back in the result document. 

The total round trip time for an inference API call will also be of interest but this measure eliminates the vagaries of network overhead by being as close as possible to the libtorch call.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -31,9 +31,9 @@
 namespace {
 const std::string INFERENCE{"inference"};
 const std::string ERROR{"error"};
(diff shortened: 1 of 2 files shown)


dimitris-athanasiou (approved):
LGTM

benwtrent:
> this measure eliminates the vagaries of network overhead

It also eliminates the "time in queue". Named pipes are a single queue that could receive pressure depending on throughput demands. 

Having this "wall clock time on CPU" is nice :)

PR #1939, opened 2021-06-29
[ML] Adding Makefile and Python tests

valeriy42:
This PR restructures the content of the `jupyter` directory to improve modularization and testability. It splits the notebooks and python code:  notebooks should go in `jupyter/notebooks/project_name` and python code in `jupyter/src/project_name`. 

Currently, we have a single `project_name`, which is `incremental_learning`, but as we continue prototyping on C++ side, new projects will follow. 

I made `jupyter/src/incremental_learning` to a python package. It is versioned so we can easier identify what version of the code a user is referring to. I installed and configured `bump2version` to manage version using `make bump-minor` and `make bump-major`.

I added some unit tests in `tests/incremental_learning` directory. Also, we now have smoke tests for all Jupyter notebooks in `notebooks/incremental_learning` directory. Those can be executed using `make` as well.


tveasey (approved):
Nice tidy up!

PR #1940, opened 2021-07-01
[ML] Fix prediction task

valeriy42:
The prediction job will only predict new data and ignore data summarization.


tveasey (approved):
Changes and extra tests look good. It looks like an existing test needs updating to reflect the new behaviour, but otherwise LGTM.

PR #1944, opened 2021-07-07
[ML] Add validElasticLicenseKeyConfirmed command line arg to ml-cpp binaries

edsavage:
All ml-cpp binaries (except controller) now must have '--validElasticLicenseKeyConfirmed=true' present in the command line arguments in order to run. Failure to do so will result in the binary to exit with an appropriate error message.


droberts195 (approved):
LGTM

Note that when you backport to 7.x you will need to remove the `pytorch_inference` changes, as that binary doesn't exist in 7.x.

PR #1946, opened 2021-07-13
[ML][PYTORCH] Add options for the number of inter & intra threads

davidkyle:
Adds 2 options for controlling the number of threads used in inference.
- numThreads
- numInterOpThreads

The first option is the most important for our use case and the models we are evaluating. 
See https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html

In the first instance these settings will be used for benchmarking and in future we may want to expose them via config or API. 

This PR also deletes the _simplest_ test as the superseded by the model used in the [Java side tests](https://github.com/elastic/elasticsearch/blob/b2ae13506d96500bdb202a0cb3a5d39bafce2484/x-pack/plugin/ml/qa/native-multi-node-tests/src/javaRestTest/java/org/elasticsearch/xpack/ml/integration/PyTorchModelIT.java#L28)

The `at::get_parallel_info()` log message looks like this on my machine
```
2021-07-13 14:36:07,244126 UTC [67451] DEBUG Main.cc@268 ATen/Parallel:
	at::get_num_threads() : 4
	at::get_num_interop_threads() : 1
OpenMP 201107 (a.k.a. OpenMP 3.1)
	omp_get_max_threads() : 4
Intel(R) Math Kernel Library Version 2019.0.4 Product Build 20190411 for Intel(R) 64 architecture applications
	mkl_get_max_threads() : 4
Intel(R) MKL-DNN v1.6.0 (Git Hash 5ef631a030a6f73131c77892041042805a06064f)
std::thread::hardware_concurrency() : 16
Environment variables:
	OMP_NUM_THREADS : [not set]
	MKL_NUM_THREADS : [not set]
ATen parallel backend: OpenMP
```

Non-issue from the change log perspective as it is part of the as-yet-unreleased 3rd party models project


bin/pytorch_inference/CCmdLineParser.cc
```diff
@@ -52,6 +54,10 @@ bool CCmdLineParser::parse(int argc,
             ("logPipe", boost::program_options::value<std::string>(),
                         "Named pipe to write log messages to")
             ("logProperties", "Optional logger properties file")
+            ("numThreads", boost::program_options::value<std::int32_t>(),
+                        "Optionaly set number of threads LibTorch can use for inference - not pressent means use the LibTorch defaults")
+            ("numInterOpThreads", boost::program_options::value<std::int32_t>(),
+                        "Optionaly set number of threads LibTorch can use for inter operation parrllelism - not pressent means use the LibTorch defaults")
```
droberts195, line 60:
```suggestion
                        "Optionaly set number of threads LibTorch can use for inter operation parallelism - not present means use the LibTorch defaults")
```

bin/pytorch_inference/CCmdLineParser.cc
```diff
@@ -52,6 +54,10 @@ bool CCmdLineParser::parse(int argc,
             ("logPipe", boost::program_options::value<std::string>(),
                         "Named pipe to write log messages to")
             ("logProperties", "Optional logger properties file")
+            ("numThreads", boost::program_options::value<std::int32_t>(),
+                        "Optionaly set number of threads LibTorch can use for inference - not pressent means use the LibTorch defaults")
```
droberts195, line 58:
```suggestion
                        "Optionaly set number of threads LibTorch can use for inference - not present means use the LibTorch defaults")
```

bin/pytorch_inference/Main.cc
```diff
@@ -21,10 +20,14 @@
...
 #include "CCommandParser.h"
 
+#include <ATen/Parallel.h>
 #include <rapidjson/ostreamwrapper.h>
+#include <torch/csrc/api/include/torch/types.h>
 #include <torch/script.h>
 
+#include <cstdint>
 #include <memory>
+#include <ostream>
```
droberts195, line 30: It doesn't look like any of the additions should need this.

davidkyle: Hmm VSCode must have added it

droberts195 (approved):
LGTM

PR #1947, opened 2021-07-13
[ML] Write out training fraction and number of train rows in model meta data

tveasey:
This finishes up the TODOs from #1941. In particular, I remove all reference to train fraction per fold from instrumentation since it doesn't change from round to round and so isn't really a useful addition to the model stats. I also start writing out the number of rows used in training in the model meta data because these are needed to handle scaling hyperparameters properly when rerunning training on a different amount of data. This will be important, for example, when running incremental training.

To do this I've introduced a new section `train_parameters` into the model metadata object. Ultimately this should include other parameters the user overrides which are needed to reproduce training. These can be read directly from the model config and so can be managed in Java.

Note this PR can't be merged until the Java is updated to read the new metadata result format.


tveasey:
This is primarily needed for incremental training so I'll just make the change on that feature branch (we need other stats there as well).

PR #1948, opened 2021-07-14
[ML][PYTORCH] Add benchmark mode to the evaluate.py script

davidkyle:
Inference results contain a time that can be used for benchmarking. The inline documentation should be sufficient explanation for how benchmark mode works.

The `numThreads` option added in #1946 is now configurable argument to the script. 

In this case only the average time is calculated but more complicated metrics could be derived in future using the [Python stats library](https://docs.python.org/3/library/statistics.html)


dimitris-athanasiou (approved):
LGTM

PR #1950, opened 2021-07-15
[ML][PYTORCH] Pass --validElasticLicenseKeyConfirmed argument from evaluate.py

davidkyle:
When launching the pytorch app from `evaluate.py` pass `--validElasticLicenseKeyConfirmed=true` as required by #1944


droberts195 (approved):
LGTM

PR #1951, opened 2021-07-15
[ML] Correct acceptance criterion for incremental training

tveasey:
When we decide whether to accept the results of incremental training we compare the loss calculated for the best candidate with the loss calculated for the original model. Since the data summary comprises a subset of the training data we are in effect comparing training error on old data + validation error on new training data with something closer to validation error on all data. If we don't have much new data, or the improvement we can make on it is small, this typically causes us to reject models which actually perform better in test.

This pull request records the gap between the train and validation loss on the old training data and adds it on to the threshold to accept. This is recorded in the model meta data and needs to be added to the configuration parameters when Java calls incremental training. Also it became clear that there is no value in storing the per fold training percentage in the instrumentation since it is fixed for the duration of train and is determined either by user override or from the training data size and it is only needed for the initial train so I just removed it.

### DIFF
--- a/include/maths/CBoostedTreeImpl.h
+++ b/include/maths/CBoostedTreeImpl.h
@@ -478,7 +478,7 @@ private:
 private:
     std::uint64_t m_Seed{0};
     mutable CPRNG::CXorOShiro128Plus m_Rng;
-    EInitializationStage m_InitializationStage = E_NotInitialized;
+    EInitializationStage m_InitializationStage{E_NotInitialized};
     std::size_t m_NumberThreads;
     std::size_t m_DependentVariable{std::numeric_limits<std::size_t>::max()};
     TOptionalSize m_PaddedExtraColumns;
--- a/include/maths/CBoostedTreeImpl.h
+++ b/include/maths/CBoostedTreeImpl.h
@@ -127,6 +127,9 @@ public:
     //! Get the model produced by training if it has been run.
     const TNodeVecVec& trainedModel() const;
 
+    //! Get the mean gap in the loss between test and train examples.
+    double lossGap() const;
+
     //! Get the training loss function.
     TLossFunction& loss() const;
 
@@ -180,9 +183,6 @@ public:
 
     //! \return The best hyperparameters for validation error found so far.
     const CBoostedTreeHyperparameters& bestHyperparameters() const;
-
-    //! \return The fraction of data we use for train per fold when tuning hyperparameters.
-    double trainFractionPerFold() const;
 
     //! \return The full training set data mask, i.e. all rows which aren't missing
     //! the dependent variable.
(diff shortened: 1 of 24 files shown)


include/maths/CBoostedTreeImpl.h
```diff
@@ -214,7 +214,8 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     using TDoubleVecVec = std::vector<TDoubleVec>;
     using TPackedBitVectorVec = std::vector<core::CPackedBitVector>;
     using TImmutableRadixSetVec = std::vector<core::CImmutableRadixSet<double>>;
-    using TNodeVecVecDoubleDoubleVecTr = std::tuple<TNodeVecVec, double, TDoubleVec>;
+    using TNodeVecVecDoubleDoubleDoubleVecTuple =
```
valeriy42, line 217: Same here, maybe introduce a new structure.

include/maths/CBoostedTreeImpl.h
```diff
@@ -181,9 +184,6 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     //! \return The best hyperparameters for validation error found so far.
     const CBoostedTreeHyperparameters& bestHyperparameters() const;
 
-    //! \return The fraction of data we use for train per fold when tuning hyperparameters.
-    double trainFractionPerFold() const;
```
valeriy42, line 185: Can you maybe mention in the PR description why you are removing these methods? This doesn't seem to be related to your other changes.

valeriy42 (approved):
LGTM. Good job.  Just a few comments regarding readability.

include/maths/CBoostedTreeImpl.h
tveasey: I ended up porting #1947 functionality to the feature branch because I need to write out the new section in the meta data for this change as well. In fact, I think it probably makes sense just to implement this on the `feature/incremental-learning` branch because it is primarily needed for this anyway. As part of that change I realised there was no real need to add this to the instrumentation since it is fixed for the duration of train. I'll add above as well.

include/maths/CBoostedTreeImpl.h
```diff
@@ -61,8 +61,8 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     using TVector = CDenseVector<double>;
     using TMeanAccumulator = CBasicStatistics::SSampleMean<double>::TAccumulator;
     using TMeanVarAccumulator = CBasicStatistics::SSampleMeanVar<double>::TAccumulator;
-    using TMeanVarAccumulatorSizeDoubleTuple =
-        std::tuple<TMeanVarAccumulator, std::size_t, double>;
+    using TMeanVarAccumulatorDoubleSizeDoubleTuple =
```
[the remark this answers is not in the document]
tveasey: See [3b63510](https://github.com/elastic/ml-cpp/pull/1951/commits/3b63510b482e7aa86538a7d5f2ffceadfaf04ac1).

include/maths/CBoostedTreeImpl.h
tveasey: See [3b63510](https://github.com/elastic/ml-cpp/pull/1951/commits/3b63510b482e7aa86538a7d5f2ffceadfaf04ac1).

PR #1952, opened 2021-07-19
[ML] Switch ml-cpp repo to Elastic License 2.0

droberts195:
Starting with version 7.15 all the ML C++ code will be
licensed under the Elastic License 2.0.

The intention is that modified versions may be used in
production but only when invoked by the Elasticsearch
JVM such that an Elastic license key that permits ML
use has been validated before ML functionality is
invoked.

(As always, modified code is not _supported_, just
permitted by the license.)


droberts195:
I just checked and there was no entry in the release notes when Elasticsearch changed its license.  I'm following that precedent here, hence `>non-issue`.

tveasey (approved):

sophiec20 (approved):
LGTM

PR #1958, opened 2021-07-23
[ML] Recompute class weights before prediction task

valeriy42:
Since we don't store class weights, we must recompute them explicitly before performing the prediction task.

### DIFF
--- a/lib/maths/CBoostedTreeFactory.cc
+++ b/lib/maths/CBoostedTreeFactory.cc
@@ -226,10 +226,13 @@ CBoostedTreeFactory::buildForPredict(core::CDataFrame& frame, std::size_t depend
         }
     });
     this->prepareDataFrameForTrain(frame);
-    m_TreeImpl->computeClassificationWeights(frame);
 
     skipIfAfter(CBoostedTreeImpl::E_NotInitialized,
                 [&] { this->determineFeatureDataTypes(frame); });
+    skipIfAfter(CBoostedTreeImpl::E_NotInitialized, [&] {
+        m_TreeImpl->predict(frame);
+        m_TreeImpl->computeClassificationWeights(frame);
+    });
 
     m_TreeImpl->m_Instrumentation->updateMemoryUsage(core::CMemory::dynamicSize(m_TreeImpl));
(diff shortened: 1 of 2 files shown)


lib/maths/CBoostedTreeFactory.cc
```diff
@@ -226,6 +226,7 @@ CBoostedTreeFactory::buildForPredict(core::CDataFrame& frame, std::size_t depend
         }
     });
     this->prepareDataFrameForTrain(frame);
+    m_TreeImpl->computeClassificationWeights(frame);
```
tveasey, line 229: Good catch, but I don't think just this can be done. It relies on the data frame having predicted values written into it. I think you'd need to change this so it actually calls `CBoostedTreeImpl::predict` first, wrap the two operations in a `skipIfAfter(CBoostedTreeImpl::E_NotInitialized` and move them after the call to `determineFeatureDataTypes`.

That said do we want to recompute the weights in this case? It seems to me one wouldn't want them to change on calls to predict.

valeriy42: We didn't initialize weights before (e.g., we don't restore them from the model definition). I think we are not wrong by doing it since data summarization preserves the distribution of target classes.

valeriy42:
Thank you for the review @tveasey . I adjusted the fix as you suggested.

lib/maths/CBoostedTreeFactory.cc
tveasey: I guess it doesn't matter since we don't really intend this to be used for prediction except in test, but it seems conceptually dubious to me. As this is just for test I'm happy to leave as is for the time being. As I understand it, this would only be a change on the python side anyway, since we could always extract and supply the original weights. We can always revisit if this proves to be a problem.

tveasey (approved):
LGTM.

tveasey:
One last thought...

`CDataFrameTrainBoostedTreeRunner::runImpl` calls predict again on the new data. This no longer needed and doubles the work for new examples. I think we should remove the call to predict in the switch statement and add a comment that this happens as part of `buildForPredict`.

PR #1959, opened 2021-07-26
[ML] Add logging so we know which round we exited hyperparameter tuning

tveasey:
We can exit hyperparameter tuning early if we don't think we're making progress up to the estimation error in the validation loss. Currently, we report that we do this but not which round we stopped. This decision is a candidate for causing variation in our QoR from run to run in our QA suite, but since we don't output which round we actually stopped we can't tell. Here I update the logging to report exactly which round we exit on.


dimitris-athanasiou (approved):
LGTM

PR #1961, opened 2021-07-27
[ML][PYTORCH] Batch inference requests

davidkyle:
Inference requests can be batched by adding more rows to the input tensor. These batch calls are more performant than making multiple calls to `forward()` with a single input. The expected input is now a 2D array of tokens and 2D arrays of supporting arguments, the output is a 3D array. The JSON test files best document the input & output formats. 

The C++ and evaluate.py changes are mostly handling this extra dimension with some tidying up. The parsed inputs are read into a 1D vector so the tensor can be created from contiguous data. 

Additionally this removes the `inputs` field from the request parser and document it was only used in a test that has now been superseded by an integration test in elasticsearch.


davidkyle:
CI failed running the Elasticsearch rest tests which is to be expected given that the Java side has not been updated yet.

Only the Linux build failed, Windows and Mac passed which is surprising given the test should definitely fail with bad input. Do the Elasticsearch tests only run against Linux?

tveasey:
> Do the Elasticsearch tests only run against Linux?

They run against linux and linux-aarch64. Interestingly, they did pass against linux-aarch64.

davidkyle:
Ah the test is muted on Aarch64 because we have seen failures there.

https://github.com/elastic/elasticsearch/issues/75685

I haven't looked into that one yet

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -141,11 +129,15 @@ bool CCommandParser::validateJson(const rapidjson::Document& doc,
             return false;
         }
 
-        if (checkArrayContainsUInts(value) == false) {
-            errorHandler(doc[REQUEST_ID].GetString(),
-                         "Invalid command: array [" + varArgName +
-                             "] contains values that are not unsigned integers");
-            return false;
+        for (auto itr = value.Begin(); itr != value.End(); ++itr) {
+            const rapidjson::Value::ConstArray& innerArray = itr->GetArray();
```
tveasey, line 133: as before

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -155,7 +147,7 @@ bool CCommandParser::validateJson(const rapidjson::Document& doc,
     return true;
 }
 
-bool CCommandParser::checkArrayContainsUInts(const rapidjson::Value& arr) const {
```
tveasey, line 158: nit: these methods are really one liners and would be ever so slightly more efficient since `find_if` returns false and stops traversing on failure
```
return std::find_if(arr.Begin(), arr.End(), [](const auto& i) { return itr->IsUint64() == false; }) == arr.End();
```

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -203,21 +194,25 @@ void CCommandParser::jsonToRequest(const rapidjson::Document& doc) {
...
-        for (auto itr = v.Begin(); itr != v.End(); ++itr) {
-            arg.push_back(itr->GetUint64());
-        }
 
-        m_Request.s_SecondaryArguments.push_back(arg);
-        arg.clear();
+        const rapidjson::Value& arr = doc[varArgName];
+        TUint64Vec arg;
+        arg.reserve(m_Request.h * m_Request.w);
+        for (auto itr = arr.Begin(); itr != arr.End(); ++itr) {
```
tveasey, line 202: nit: I believe GetArray supports range based for loops so this could just be
```
for (const auto& innerArray : arr.GetArray()) {
    for (const auto &token : innerArray.GetArray()) {
        ...
```
This comment applies to all iterations over rapidjson arrays in this PR.

bin/pytorch_inference/CCommandParser.h
```diff
@@ -50,21 +50,27 @@ class CCommandParser {
...
 
+    //! The incoming JSON requests contain a 2D array of tokens representing
+    //! a batch of inference calls. To avoid copying, the input tensor
+    //! should be created directly from contiguous data so the 2D token
+    //! array is read into a 1D vector of size w * h where w & h are the
+    //! dimensions of in the JSON input. The secondary arguments are
+    //! treated in the same manner.
     struct SRequest {
+        std::int64_t w;
+        std::int64_t h;
```
tveasey, line 68: nit: these names seem a bit brief to me and also violate the coding standards. How about `s_NumberInferences` and `s_FeatureVectorDimension`. I think it would also make what the code is doing clearer.

bin/pytorch_inference/Main.cc
```diff
@@ -93,6 +83,16 @@ void writeTensor(const torch::TensorAccessor<T, 1UL>& accessor,
...
+    jsonWriter.StartArray();
+    for (int i = 0; i < accessor.size(0); ++i) {
+        writeTensor(accessor[i], jsonWriter);
+    }
+    jsonWriter.EndArray();
+}
+
+template<typename T>
+void writeTensor(const torch::TensorAccessor<T, 3UL>& accessor,
+                 ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
```
tveasey, line 95: This seems surprising to me: if you `accessor.size(0) == 1` then I think this method should assert that; otherwise this needs a `jsonWriter.StartArray();` and `jsonWriter.EndArray();` surely or else the JSON is malformed. This code could be made fully general and clearer IMO:
```
template<typename T, std::size_t N>
struct SDoWriteTensor {
    static void dispatch(const torch::TensorAccessor<T, N>& accessor,
                     ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
        jsonWriter.StartArray();
        for (int i = 0; i < accessor.size(0); ++i) {
            writeTensor(accessor[i])
        }
        jsonWriter.EndArray();
    }
};
template<typename T>
struct SDoWriteTensor<T, 1> {
    static void dispatch(const torch::TensorAccessor<T, 1>& accessor,
                     ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
        jsonWriter.Double(static_cast<double>(accessor[i]));
    }
};
template<typename T, std::size_t N>
void writeTensor(const torch::TensorAccessor<T, N>& accessor,
                     ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
    SDoWriteTensor<T, N>::dispatch(accessor, jsonWriter);
}
```
(Note as well looking at `TensorAccessor.h` it seems that `torch::TensorAccessor<T, 1>` is a scalar.) You have to wrap in a struct because partial template specialisation isn't allowed for functions, but I don't think this hurts readability much.

bin/pytorch_inference/unittest/CCommandParserTest.cc
```diff
@@ -288,4 +242,58 @@ BOOST_AUTO_TEST_CASE(testRequestHandlerExitsLoop) {
     BOOST_REQUIRE_EQUAL(1, parsed.size());
 }
 
+BOOST_AUTO_TEST_CASE(testParsingBatch) {
```
tveasey, line 245: I think a test where you run inference for randomised strings of tokens one at a time and in batch and check you always get the same results would be good. This way you can test many more cases.

tveasey (commented):
I did a pass through. Made some tidy up suggestions and also a possible test. I think there may be a loophole in the validation as well.

bin/pytorch_inference/CCommandParser.h
davidkyle: I changed these to `s_NumberInferences` and `s_NumberInputTokens`

bin/pytorch_inference/unittest/CCommandParserTest.cc
davidkyle: The test suite is only testing the input parsing code. `evaluate.py` and the `test_run.json` files cover the case of asserting batch inputs give the same results as a single input. Unfortunately we can't randomise those inputs yet.

davidkyle:
I hit a complication for batching inputs here in the native process, namely the inputs must all be of the same length.
When the input text tokenises to different number of tokens padding is used to create rows of the same length.

I can think of a couple of strategies to handle this problem:

1. If the input size changes run inference on any queued inputs and start queuing again with the new input
2. Pad all inputs to the maximum token length (typically 512 tokens)
3. Move the queuing and batching logic to the Java side where tokenisation occurs and add the padding there

If `2.` was followed the padding should be done in the Java side as we do not want to split the logic of preparing the input between the C++ and Java. Also the C++ would need to be made aware of configuration in order to correctly construct the inputs. This would not only increase the size of the input but also the output as results are calculated for padding tokens. In the NER use case the padding tokens have a results classification of outside a named entity.

The increased result size may make padding inputs to a max length prohibitive. In the NER use case an input 512 tokens long with 9 possible output classes in a batch of 100 = `512 * 9 * 100 * sizeof(double) = 3,686,400 bytes`

bin/pytorch_inference/Main.cc
davidkyle: The hack I was working around here is that in all cases I want the output to be a 3D array but some models/tasks give a 2D output e.g. sentiment analysis. The missing start and end array are actually written elsewhere and this meant that a 2D output was embedded in a 3rd dimension whereas the 3D output was not. I've added comments and rewritten to make it clear why this is happening.

bin/pytorch_inference/unittest/CCommandParserTest.cc
tveasey: Perhaps something to revisit in a separate PR. I'm happy for this to go in without the extra testing.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -96,38 +95,34 @@ bool CCommandParser::validateJson(const rapidjson::Document& doc,
...
+    if (doc.HasMember(TOKENS) == false) {
+        errorHandler(doc[REQUEST_ID].GetString(),
+                     "Invalid command: missing field [" + TOKENS + "]");
+        return false;
+    }
+
+    const rapidjson::Value& tokens = doc[TOKENS];
+    if (tokens.IsArray() == false) {
+        errorHandler(doc[REQUEST_ID].GetString(),
+                     "Invalid command: expected an array [" + TOKENS + "]");
```
tveasey, line 107: Typo
```suggestion
                     "Invalid command: expected an array of [" + TOKENS + "]");
```

benwtrent:
@davidkyle what is preventing the native side from padding the inputs itself? Does the padding HAVE to occur before the tokenization?

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -203,22 +199,21 @@ void CCommandParser::jsonToRequest(const rapidjson::Document& doc) {
...
-            arg.push_back(itr->GetUint64());
-        }
 
-        m_Request.s_SecondaryArguments.push_back(arg);
-        arg.clear();
+        const rapidjson::Value::ConstArray& outer = doc[varArgName].GetArray();
+        TUint64Vec arg;
+        arg.reserve(m_Request.s_NumberInferences * m_Request.s_NumberInputTokens);
+        for (const auto& val : outer) {
+            const auto innerArray = val.GetArray();
```
tveasey, line 208:
```suggestion
            const auto& innerArray = val.GetArray();
```

bin/pytorch_inference/Main.cc
```diff
@@ -90,12 +80,33 @@ void writeTensor(const torch::TensorAccessor<T, 1UL>& accessor,
...
+void writeInferenceResults(const torch::TensorAccessor<T, 3UL>& accessor,
+                           ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
+
+    jsonWriter.Key(INFERENCE);
+    writeTensor(accessor, jsonWriter);
+}
+
+template<typename T>
+void writeInferenceResults(const torch::TensorAccessor<T, 2UL>& accessor,
+                           ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
```
tveasey, line 103: Nice tidy up this is much clearer and also better to keep tensor writing generic IMO. (I was just wrong and one can of course overload functions whose arguments are class partial specialisations so no need to have the class bit either.) One thing I find interesting is you still have
```
template<typename T>
void writeTensor(const torch::TensorAccessor<T, 1UL>& accessor,
                 ml::core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>& jsonWriter) {
    jsonWriter.StartArray();
    for (int i = 0; i < accessor.size(0); ++i) {
        jsonWriter.Double(static_cast<double>(accessor[i]));
    }
    jsonWriter.EndArray();
}
```
when you know `torch::TensorAccessor<T, 1UL>` only contains a single value. Is this intentional? It seems surprising to me.

tveasey (approved):
Just spotted a couple of typos I think. I have one question on the results format, but happy for this to go as is. Good stuff!

bin/pytorch_inference/Main.cc
davidkyle: The 2nd template parameter `1UL` is number of dimensions, it may well be a scalar but it could also be a vector. Note `operator[]` returns an accessor with one fewer dimensions `torch::TensorAccessor<T, N -1>` or when `N == 1` the underlying data.

I really like this pattern matching aspect of C++ where there is a base case for a single dimension and the overloaded method handles the N > 1 cases. This would have taken much more code to write in Java.

davidkyle:
@benwtrent 

> what is preventing the native side from padding the inputs itself?

@tveasey @dimitris-athanasiou  and I had a discussion that is worth documenting here

The padding logic is very simple and could be moved to the C++ but I am reluctant to because it violates the design principle that the C++ app would perform no function other than to call `forward()`. When reasoning about code it is much better to have the logic that constructs the input requests (tokenization, supporting arguments, padding etc) all in one place rather than split between 2 code bases, but my main  concern is that we would have to make this app aware of various configuration options and the extra work that would involve. 

None of that is insurmountable so if we decided the gains from queuing here are sufficient we would make that change move the padding logic. Currently that's undecided, I want to review the option of queueing in `DeploymentManager.java` first.

I argued above that when padding is used it increases the size of the result tensor that is written back but that need not be the case. There is enough information in the input to determine which values are padding and filter those out so only relevant non padded scores are written back. 

When dispatching to multiple models the load balancing strategy plays an important part. For example if we implemented a queue of 100 and round-robin'd the requests that would be the worst case for latency as we would not see any results until all those queues have filled up. 

Finally we plan to investigate making multiple concurrent calls to `forward()` to support high throughput use cases.

bin/pytorch_inference/Main.cc
tveasey: Oh sorry, my mistake, looking at the implementation again I realise I'd just misread the code.
> I really like this pattern matching aspect of C++

Yes and variadic templates make this sort of idiom even more powerful.

davidkyle:
This has been left open for a while but I am merging now as the change will be valuable for future use-cases such as zero-shot classification and potentially batch inference will be exposed to the REST infer API

PR #1962, opened 2021-07-27
[ML] Aggressively prune dead split fields

edsavage:
Prune models (excluding for functions rare, freq_rare, time_of_day, time_of_week, lat_long) that haven't seen data updates for a
given period of time.

The window of time over which to prune is supplied in the `model_prune_window` field, which is an optional field in the job's `analysis_config` configuration. If not present no aggressive pruning is performed, no default value is used. However, the `bucket_span` value is used as a minimum value for the `model_prune_window` i.e. if the provided `model_prune_window` value is smaller than the `bucket_span` then the value of the `bucket_span` is used instead.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -56,12 +56,12 @@
 
 * Give higher weight to multiple adjacent dictionary words when performing categorization. (See
   {ml-pull}1903[#1903].)
-* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
-(See {ml-pull}1908[#1908].)
 
 === Bug Fixes
 
 * Make atomic operations safer for aarch64. (See {ml-pull}1893[#1893].)
+* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
+(See {ml-pull}1908[#1908].)
 
 == {es} version 7.13.0
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -41,6 +41,8 @@
 * Added a new application for evaluating PyTorch models. The app depends on LibTorch
   - the C++ front end to PyTorch and performs inference on models stored in the 
   TorchScript format. (See {ml-pull}1902[#1902].)
+* Prune models for split fields (by, partition) that haven't seen data updates for
+  a given period of time. (See {ml-pull}1962[#1962].)
 
 === Bug Fixes
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -50,15 +50,14 @@
 
 === Bug Fixes
 
-* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
-  (See {ml-pull}1908[#1908].)
-
 == {es} version 7.14.0
 
 === Enhancements
 
 * Give higher weight to multiple adjacent dictionary words when performing categorization. (See
   {ml-pull}1903[#1903].)
+* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
+(See {ml-pull}1908[#1908].)
 
 === Bug Fixes
 
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -41,6 +41,10 @@
 * Added a new application for evaluating PyTorch models. The app depends on LibTorch
   - the C++ front end to PyTorch and performs inference on models stored in the 
   TorchScript format. (See {ml-pull}1902[#1902].)
+
+== {es} version 7.15.0
+
+=== Enhancements
 * Prune models for split fields (by, partition) that haven't seen data updates for
   a given period of time. (See {ml-pull}1962[#1962].)
 
--- a/include/api/CAnomalyJob.h
+++ b/include/api/CAnomalyJob.h
@@ -413,6 +413,8 @@ protected:
                                               model::CResourceMonitor& resourceMonitor);
 
     //! Prune all the models that exceed \p buckets in age
+    //! A value of 0 for \buckets indicates that only 'obsolete' models will
+    //! be pruned, i.e. those which are so old as to be effectively dead.
     void pruneAllModels(std::size_t buckets = 0);
 
 private:
--- a/include/api/CAnomalyJob.h
+++ b/include/api/CAnomalyJob.h
@@ -399,7 +399,7 @@ protected:
     //! Get all the detectors.
     void detectors(TAnomalyDetectorPtrVec& detectors) const;
 
-    //! Get the detectors by parition
+    //! Get the detectors by partition
     const TKeyAnomalyDetectorPtrUMap& detectorPartitionMap() const;
 
     //! Get all sorted references to the detectors.
@@ -412,8 +412,8 @@ protected:
                                               const std::string& partitionFieldValue,
                                               model::CResourceMonitor& resourceMonitor);
 
-    //! Prune all the models
-    void pruneAllModels();
+    //! Prune all the models that exceed \p buckets in age
+    void pruneAllModels(std::size_t buckets = 0);
 
 private:
     //! The job ID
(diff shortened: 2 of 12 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -41,6 +41,8 @@
 * Added a new application for evaluating PyTorch models. The app depends on LibTorch
   - the C++ front end to PyTorch and performs inference on models stored in the 
   TorchScript format. (See {ml-pull}1902[#1902].)
+* Prune models for split fields (by, partition) that haven't seen data updates for
+  a given period of time. (See {ml-pull}1962[#1962].)
```
droberts195, line 45: This should be in the 7.15 section of the release notes.

include/api/CAnomalyJobConfig.h
```diff
@@ -243,6 +246,8 @@ class API_EXPORT CAnomalyJobConfig {
 
         core_t::TTime bucketSpan() const { return m_BucketSpan; }
 
+        core_t::TTime modelPruneWindow() const { return m_ModelPruneWindow; }
```
droberts195, line 249: It will be quite confusing to future maintainers that sometimes "prune window" is expressed as a number of buckets and other times it's expressed as a period of time.  It would help those future maintainers if every related method and member variable had a Doxygen comment saying that sometimes "prune window" is expressed as a number of buckets and other times it's expressed as a period of time, and which of the two cases applies for the method/variable that the comment relates to.  (I can see that some already do, but not this one.)

include/model/CAnomalyDetector.h
```diff
@@ -244,6 +244,10 @@ class MODEL_EXPORT CAnomalyDetector : public CMonitoredResource {
     //! that may be deleted as a result of this call.
     virtual void pruneModels();
 
+    // Remove dead models - i.e. those that have not seen activity
+    // in the last \p pruneWindow buckets
```
droberts195, line 248:
```suggestion
    //! Remove dead models - i.e. those that have not seen activity
    //! in the last \p pruneWindow buckets
```

include/model/CAnomalyDetectorModelConfig.h
```diff
@@ -336,6 +336,12 @@ class MODEL_EXPORT CAnomalyDetectorModelConfig {
     //! Get the bucket length.
     core_t::TTime bucketLength() const;
 
+    // Get the period of time at which to perform a potential prune of the models
```
droberts195, line 339:
```suggestion
    //! Get the period of time at which to perform a potential prune of the models
```

include/model/CAnomalyDetectorModelConfig.h
```diff
@@ -336,6 +336,12 @@ class MODEL_EXPORT CAnomalyDetectorModelConfig {
...
+    core_t::TTime modelPruneWindow() const;
+
+    // Set the period of time at which to perform a potential prune of the models
```
droberts195, line 342:
```suggestion
    //! Set the period of time at which to perform a potential prune of the models
```

lib/api/CAnomalyJob.cc
```diff
@@ -688,6 +688,13 @@ void CAnomalyJob::outputResults(core_t::TTime bucketStartTime) {
     this->writeOutAnnotations(annotations);
     this->writeOutResults(false, results, bucketStartTime, processingTime);
 
+    if (m_ModelConfig.modelPruneWindow() > core_t::TTime{0}) {
```
droberts195, line 691:
```suggestion
    if (m_ModelConfig.modelPruneWindow() > 0) {
```

It seems unnecessarily verbose to case 0 in this line, and it shouldn't make any difference.

lib/api/CAnomalyJobConfig.cc
```diff
@@ -677,6 +682,18 @@ void CAnomalyJobConfig::CAnalysisConfig::parse(const rapidjson::Value& analysisC
...
+    const std::string& modelPruneWindowString{
+        parameters[MODEL_PRUNE_WINDOW].fallback(EMPTY_STRING)};
+
+    if (modelPruneWindowString.empty() == false) {
+        m_ModelPruneWindow = CAnomalyJobConfig::CAnalysisConfig::durationSeconds(
+            modelPruneWindowString, core_t::TTime{0});
+
+        // Ensure that the model prune window is never smaller than the bucket span.
+        const core_t::TTime minModelPruneWindow{m_BucketSpan};
+        m_ModelPruneWindow = std::max(m_ModelPruneWindow, minModelPruneWindow);
```
droberts195, line 694: It's probably best to log a warning here instead of silently changing the user's choice.

lib/api/CAnomalyJob.cc
```diff
@@ -1574,8 +1581,12 @@ CAnomalyJob::detectorForKey(bool isRestoring,
     return itr->second;
 }
 
-void CAnomalyJob::pruneAllModels() {
-    LOG_INFO(<< "Pruning all models");
+void CAnomalyJob::pruneAllModels(std::size_t buckets) {
+    if (buckets == 0) {
+        LOG_INFO(<< "Pruning all models");
```
droberts195, line 1586: This message is now out-of-sync with the one two lines below, as the meaning of "all" has been implicitly changed.  According to the method that gets called it "prunes ancient models that are effectively dead".  We probably shouldn't put that in a log message, but maybe "Pruning obsolete models"?  If people see this message and the one two lines below then we need to make clear that this is generally less aggressive than the other type of pruning.

There's also an implication for the docs: we need to make clear that models get pruned when we expect them to contain no information at all _even when pruning is not enabled in the config_.

include/api/CAnomalyJob.h
```diff
@@ -412,8 +412,8 @@ class API_EXPORT CAnomalyJob : public CDataProcessor {
                                               const std::string& partitionFieldValue,
                                               model::CResourceMonitor& resourceMonitor);
 
-    //! Prune all the models
-    void pruneAllModels();
+    //! Prune all the models that exceed \p buckets in age
```
droberts195, line 415: There should be an extra comment to say that 0 has a special value, meaning "so old they're effectively dead".

(It's worth noting there's a discontinuity here, as 1 is more aggressive than 2, 2 is more aggressive than 3, 3 is more aggressive than 4, 0 less aggressive than 1000.)

docs/CHANGELOG.asciidoc
```diff
@@ -42,17 +42,22 @@
...
+=== Bug Fixes
 
 == {es} version 7.14.0
 
 === Enhancements
 
 * Give higher weight to multiple adjacent dictionary words when performing categorization. (See
   {ml-pull}1903[#1903].)
+* Ensure bucket `event_count` is calculated for jobs with 1 second bucket spans.
+(See {ml-pull}1908[#1908].)
```
droberts195, line 60: This should be in the bug fixes section for 7.14.

lib/api/CAnomalyJob.cc
```diff
@@ -688,6 +688,18 @@ void CAnomalyJob::outputResults(core_t::TTime bucketStartTime) {
...
     this->writeOutResults(false, results, bucketStartTime, processingTime);
 
+    if (m_ModelConfig.modelPruneWindow() > 0) {
+        // Ensure that bucketPruneWindow is always rounded _up_
+        // to the next whole number of buckets (this doesn't really matter if we enforce
+        // that the model prune window always be an exact multiple of bucket span in the
+        // corresponding Java code)
+        core_t::TTime bucketPruneWindow{
+            m_ModelConfig.modelPruneWindow() / m_ModelConfig.bucketLength() +
+            (m_ModelConfig.modelPruneWindow() % m_ModelConfig.bucketLength() != 0)};
```
droberts195, line 698: There's an implicit `bool` to `long` cast here.  I suspect that will be causing a compiler warning on some platforms - you could check the console log of the Windows build of the PR CI.

If it is then an alternative pattern we use elsewhere would be:

```suggestion
        core_t::TTime bucketPruneWindow{
            (m_ModelConfig.modelPruneWindow() + m_ModelConfig.bucketLength() - 1) / m_ModelConfig.bucketLength()};
```

droberts195 (approved):
I saw a couple of nits but feel free to merge without further review.

edsavage:
retest

edsavage:
I think CI has got itself confused regarding the status of the changelog. Ignoring the complaints that it does not require modification and merging changes.

Issue #1963, opened 2021-07-27
[ML] Gradle fails Elasticsearch tests when run with the ml-cpp side-loaded


davidkyle:
When using the `elasticsearch-extra` trick to run against a local build of `ml-cpp` (i.e. create the `elasticsearch-extra` directory as a sibling of the `elasticsearch` repo and softlink `ml-cpp` inside it) the integration tests in the Elasticsearch repo fail even if the individual tests pass. The error is:

```
FAILURE: Build failed with an exception.

* What went wrong:
Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1m 23s
```

Elasticsearch is currently on Gradle version 7.1.1 while this repo is using Gradle 6.8.3, perhaps upgrading would do the trick?

## Reproduce
Set up the elasticsearch-extra repo and run a test, any test will do it.
```
 ./gradlew ':x-pack:plugin:ml:test' --tests "org.elasticsearch.xpack.ml.inference.nlp.BertRequestBuilderTests.testBuildRequest"
```

davidkyle:
> perhaps upgrading would do the trick?
 
No updating `gradle-wrapper.properties` to be the same as the in the Elasticsearch repo did not work. 

While upgrading I got this error

```
> Configure project :buildSrc
The RepositoryHandler.jcenter() method has been deprecated. This is scheduled to be removed in Gradle 8.0. JFrog announced JCenter's shutdown in February 2021. Use mavenCentral() instead. Consult the upgrading guide for further information: https://docs.gradle.org/7.1.1/userguide/upgrading_version_6.html#jcenter_deprecation
        at build_9wlfif6ooejmas8bcbgtnbwqi$_run_closure1.doCall(/Users/davidkyle/source/ml-cpp/buildSrc/build.gradle:8)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

So the deprecation must be tackled.

droberts195:
The `jCenter()` problem is trivial to fix.  (We already have `mavenCentral()` listed too, so just need to delete one line of code.)

Unfortunately another deprecation looks more tricky:

```
[...]

PR #1965, opened 2021-07-28
[ML] Remove jCenter() from buildSrc and just use mavenCentral()

droberts195:
jCenter() is going away, so we shouldn't rely on it any more.

Additionally, the internal Gradle progress logger isn't
available from Maven Central.  We shouldn't really be using
this internal class and progress logging is not essential,
so I have removed this from our S3 upload code.

Fixes one of the problems of #1963


dimitris-athanasiou (approved):
LGTM

PR #1970, opened 2021-07-29
[ML] Adding upload tasks to build.gradle

droberts195:
Gradle has deprecated the -b option, which means that we
cannot continue to split functionality between build.gradle
and upload.gradle in the long term.

This PR starts the combination process by adding the tasks
from upload.gradle into build.gradle.

The PR does _not_ remove upload.gradle, as it is referenced
by CI jobs.  The overall change needs to be done as a
multi-stage process, first adding functionality to
build.gradle, then updating CI jobs, and finally removing
the upload.gradle file.

Fixes one more deprecation identified in #1963

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,12 @@
 description = 'Builds the Machine Learning native binaries'
 
+import org.elastic.gradle.UploadS3Task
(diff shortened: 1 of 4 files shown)


edsavage (approved):
LGTM

PR #1974, opened 2021-07-30
[ML] Include PyTorch libraries in cross-compiled builds

droberts195:
Previously PyTorch libraries were only included in
native builds.  This was a mistake.

Fixes elastic/elasticsearch#75685


davidkyle (approved):
LGTM

PR #1975, opened 2021-07-30
[ML] S3 upload task isn't available when embedded in ES build

droberts195:
After the changes of #1970 we get an instant failure when
building ml-cpp as part of the ES build by putting it under
elasticsearch-extra.

This change makes importing UploadS3Task and the tasks of
that type conditional on ml-cpp being standalone.


dimitris-athanasiou (approved):
LGTM

PR #1985, opened 2021-08-10
[ML] Split out 8.0.0-alpha1 in the changelog

droberts195:
8.0.0-alpha1 will be released soon. This PR moves the changelog
items that went into alpha1 into a separate section so that it's
clear which items (if any) need to be release noted for alpha2
and beyond.


edsavage (approved):
LGTM

PR #1986, opened 2021-08-11
[ML] Do not persist detectors with only empty models

edsavage:
Pruning may result in a situation where all the feature models of a detector are effectively empty. This presents an opportunity to reclaim memory by way of excluding such a detector from the persisted state, as on restoration the detector will no longer be present.

Labelling as `<non-issue` as the only observable change of this PR is a potential reduction in model memory size following a persist/restore cycle.

Relates #1962

### DIFF
--- a/lib/model/CIndividualModel.cc
+++ b/lib/model/CIndividualModel.cc
@@ -339,6 +339,20 @@ void CIndividualModel::doPersistModelsState(core::CStatePersistInserter& inserte
     }
 }
 
+bool CIndividualModel::shouldPersistModel() const {
+    // If a feature model's memory usage is zero, it indicates that the model
+    // has been pruned and replaced with a stub. If all feature models are
+    // such stubs then we choose to not persist the model
+    // in order to reclaim memory after a persist/restore cycle.
+    std::size_t modelMemory{0};
+    for (const auto& feature : m_FeatureModels) {
+        for (const auto& model : feature.s_Models) {
+            modelMemory += model->memoryUsage();
+        }
+    }
+    return modelMemory > 0 ? true : false;
+}
+
 void CIndividualModel::doAcceptPersistInserter(core::CStatePersistInserter& inserter) const {
     inserter.insertValue(WINDOW_BUCKET_COUNT_TAG, this->windowBucketCount(),
                          core::CIEEE754::E_SinglePrecision);
--- a/lib/model/CIndividualModel.cc
+++ b/lib/model/CIndividualModel.cc
@@ -339,19 +339,9 @@ void CIndividualModel::doPersistModelsState(core::CStatePersistInserter& inserte
     }
 }
 
-bool CIndividualModel::shouldPersistModel() const {
-    // If a feature model's memory usage is zero, it indicates that the model
-    // has been pruned and replaced with a stub. If all feature models are
-    // such stubs then we choose to not persist the model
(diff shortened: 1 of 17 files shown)


lib/model/CIndividualModel.cc
```diff
@@ -339,6 +339,20 @@ void CIndividualModel::doPersistModelsState(core::CStatePersistInserter& inserte
...
+    // has been pruned and replaced with a stub. If all feature models are
+    // such stubs then we choose to not persist the model
+    // in order to reclaim memory after a persist/restore cycle.
+    std::size_t modelMemory{0};
+    for (const auto& feature : m_FeatureModels) {
+        for (const auto& model : feature.s_Models) {
+            modelMemory += model->memoryUsage();
+        }
+    }
+    return modelMemory > 0 ? true : false;
```
droberts195, line 353: Since the final condition is just `> 0` you can make this shortcircuit instead of checking all models.  This could be quite beneficial when there are huge numbers of models that _haven't_ been pruned.

```suggestion
    for (const auto& feature : m_FeatureModels) {
        for (const auto& model : feature.s_Models) {
            if (model->memoryUsage() > 0) {
                return true;
            }
        }
    }
    return false;
```

droberts195 (approved):
LGTM apart from one minor performance thing.

Maybe there is something equivalent we could do for population models - do you know @tveasey?  But if it's complicated then that can be left to a followup PR.

lib/model/CIndividualModel.cc
```diff
@@ -339,6 +339,21 @@ void CIndividualModel::doPersistModelsState(core::CStatePersistInserter& inserte
     }
 }
 
+bool CIndividualModel::shouldPersistModel() const {
+    // If a feature model's memory usage is zero, it indicates that the model
+    // has been pruned and replaced with a stub. If all feature models are
+    // such stubs then we choose to not persist the model
+    // in order to reclaim memory after a persist/restore cycle.
+    for (const auto& feature : m_FeatureModels) {
+        for (const auto& model : feature.s_Models) {
```
tveasey, line 348: It feels a bit fragile to check memory usage here. I think it would make the code self documenting if you introduced a method `maths::CModel::shouldPersist` then this could be:
```
if (std::any_of(feature.s_Models.begin(), feature.s_Models.end(), [](const auto& model) { return model.shouldPersist(); }) {
  return true;
}
```

tveasey: I would also be inclined to move this inner loop into a method `SFeatureModels::shouldPersist()` which can then be shared by all implementations. With this, the implementation of this method just becomes:
```
  return std::any_of(m_FeatureModels.begin(), m_FeatureModels.end(), [](const auto& model) { return model.shouldPersist()});
```

Exactly the same check can be applied in `CMetricPopulationModel` and `CEventRatePopulationModel`. So for example:
```
bool CMetricPopulationModel::shouldPersist() const {
  return std::any_of(m_FeatureModels.begin(), m_FeatureModels.end(), [](const auto& model) { return model.shoulPersist(); });
}
```

tveasey (commented):
Overall looks fine. However, my inclination would be to push `shouldPersist` methods to some other classes to improve robustness and readability. I also think you can and should apply exactly the same logic to both `CEventRatePopulationModel` and `CMetricPopulationModel` as described in the comment.

tveasey (approved):
Thanks for iterating! LGTM.

PR #1993, opened 2021-08-16
[ML] Improve test robustness

tveasey:
This test has failed for me in a back port and also a couple of times in pull requests builds. It depends on timing and so I make it more robust and incidentally test assertions in more cases, which should improve coverage slightly.

(Note that since a version of this is already committed to 7.x there is no need to back port.)


droberts195 (approved):
LGTM

PR #1994, opened 2021-08-17
[ML] Scripting to generate full incremental training test matrix

tveasey:
This adds a script to generate random test cases from the full test matrix which comprises our target data sets and transforms. It also makes configuration of various directories more robust to different layouts and usage (for example it didn't work properly after installing the `incremental_learning` package using `setup.py` previously). I also changed transform to always copy the data set before transforming (this fits in better with `experiment_driver.py`) and made variable naming more consistent. Finally, this finishes up the implementation of `experiment_driver.py` to support all transform types.

The full experiments.json configuration file will added be a separate PR which migrates the data sets in this repo to our cloud bucket.

### DIFF
--- a/jupyter/scripts/incremental_learning/experiment_driver.py
+++ b/jupyter/scripts/incremental_learning/experiment_driver.py
@@ -146,73 +146,48 @@ def transform_dataset(dataset: pd.DataFrame,
         update_dataset, test2_dataset = train_test_split(update_dataset, test_size=test_fraction)
         return train_dataset, update_dataset, test1_dataset, test2_dataset
 
+    transform = None
+
     if transform_name == 'resample_metric_features':
-        train_dataset, test1_dataset = train_test_split(dataset, test_size=test_fraction)
-        update_dataset = resample_metric_features(
-            dataset=train_dataset,
+        transform = lambda dataset : resample_metric_features(
+            dataset=dataset,
             seed=_seed, 
             fraction=transform_parameters['fraction'], 
             magnitude=transform_parameters['magnitude'],
             metric_features=transform_parameters['metric_features'])
-        test2_dataset = resample_metric_features(
-            dataset=test1_dataset,
-            seed=_seed, 
-            fraction=transform_parameters['fraction'], 
-            magnitude=transform_parameters['magnitude'],
-            metric_features=transform_parameters['metric_features'])
-        return train_dataset, update_dataset, test1_dataset, test2_dataset
 
     if transform_name == 'shift_metric_features':
-        train_dataset, test1_dataset = train_test_split(dataset, test_size=test_fraction)
-        update_dataset = shift_metric_features(
-            dataset=train_dataset,
+        transform = lambda dataset : shift_metric_features(
+            dataset=dataset,
             seed=_seed,
             fraction=transform_parameters['fraction'],
             magnitude=transform_parameters['magnitude'],
             categorical_features=transform_parameters['categorical_features'])
-        test2_dataset = shift_metric_features(
-            dataset=test1_dataset,
-            seed=_seed,
-            fraction=transform_parameters['fraction'],
-            magnitude=transform_parameters['magnitude'],
-            categorical_features=transform_parameters['categorical_features'])
-        return train_dataset, update_dataset, test1_dataset, test2_dataset
 
     if transform_name == 'rotate_metric_features':
-        train_dataset, test1_dataset = train_test_split(dataset, test_size=test_fraction)
-        update_dataset = rotate_metric_features(
-            dataset=train_dataset,
+        transform = lambda dataset : rotate_metric_features(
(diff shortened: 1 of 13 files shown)


tveasey:
retest

jupyter/scripts/incremental_learning/experiment_driver.py
```diff
@@ -128,42 +132,109 @@ def read_dataset(_run, dataset_name):
 
 
 @ex.capture
-def transform_dataset(dataset: pd.DataFrame, transform_name: str, transform_parameters, _seed: int) -> pd.DataFrame:
+def transform_dataset(dataset: pd.DataFrame,
```
Winterflower, line 135: Not a big deal, but it would be nice to add types for the `transform_parameters` variable. It appears to be a dict based on later in the code. Given that the rest of the function params have type info, why doesn't this one?

jupyter/scripts/incremental_learning/experiment_driver.py
```diff
@@ -128,42 +132,109 @@ def read_dataset(_run, dataset_name):
...
+
+    if transform_name == 'shift_metric_features':
+        train_dataset, test1_dataset = train_test_split(dataset, test_size=test_fraction)
+        update_dataset = shift_metric_features(
+            dataset=train_dataset,
+            seed=_seed,
+            fraction=transform_parameters['fraction'],
+            magnitude=transform_parameters['magnitude'],
+            categorical_features=transform_parameters['categorical_features'])
+        test2_dataset = shift_metric_features(
```
Winterflower, line 173: There is a lot of code repetition here due to the "switch" statement like functionality. It's possible to use runtime reflection to do this more concisely, but this can be a future refactoring.

jupyter/scripts/incremental_learning/generate_experiment_matrix.py
```diff
@@ -0,0 +1,117 @@
...
+    result = {}
+    for dataset_name in args.classification_datasets:
+        if download_dataset(dataset_name):
+            result[dataset_name] = feature_fields(dataset_name)
+    for dataset_name in args.regression_datasets:
+        if download_dataset(dataset_name):
+            result[dataset_name] = feature_fields(dataset_name)
+    return result
+
+def needs_metric_features(transform_name: str):
```
Winterflower, line 53: Not sure what is going on in this method - are you checking that the supplied `transform_name` is one of the approved metric transforms?

jupyter/src/incremental_learning/config.py
```diff
@@ -12,21 +12,33 @@
 import platform
 from sys import exit
 
-from pathlib2 import Path
+from pathlib import Path
 
-root_dir = Path(__file__).parent.parent.parent.absolute()
+# Check for resource called in ancestor directories of path.
+def find_ancestor_dir(path: Path, resource: str):
```
Winterflower, line 18: It would be nice to add a bit more explanation here in form of comments or docstrings. I don't fully understand what is happening here.

Winterflower (approved):
There are a few things that could be improved imho, but nothing that is a major blocker.

jupyter/scripts/incremental_learning/experiment_driver.py
tveasey: I agree in principle, but I inherited this. I think @valeriy42 said he ran into issues defining the type of this with sacred. Since he's away at the moment and it isn't really part of this PR (I just added some extra parameters) I think I'll leave it for the time being and revisit when I have everything working end-to-end.

jupyter/scripts/incremental_learning/generate_experiment_matrix.py
tveasey: This lists out all transforms which absolutely require that there is at least one metric feature (we don't generate the experiment against datasets which only have categorical features). I added documentation in [72c2a29](https://github.com/elastic/ml-cpp/pull/1994/commits/72c2a290fe948e6e8341da565c561adae57ffbb2).

jupyter/scripts/incremental_learning/experiment_driver.py
tveasey: I think streamlining this is a good idea. I'm going to stop short of generating the code, but I made it a lot less verbose in [e0b90ec](https://github.com/elastic/ml-cpp/pull/1994/commits/e0b90ec26570996b34ecf45218ead741a974e5cd).

jupyter/src/incremental_learning/config.py
tveasey: I added fuller documentation in [d4bdd1b](https://github.com/elastic/ml-cpp/pull/1994/commits/d4bdd1bf982a28328c1b8ba5d37b706226ec8ec5).

tveasey:
The one failure is just an intermittent failure due to timing in CI. I'm going to go ahead and merge.

PR #1995, opened 2021-08-18
[ML] Improve initialisation of search for the best tree topology change penalty

tveasey:
Previously, we were basing the initialisation of this parameter on the gain of the selected splits in the original tree (which is trained on a different data set). This is not the critical value which controls whether it will have an effect in incremental training. What we care about is the variation we see in the gain between different choices of split for the incrementally trained trees, since the parameter is only important if it is large(ish) on this scale.

This PR measures that quantity directly and chooses the range to search to include values where the penalty will have little effect and values where it will have significant effect.

(There is still a question mark in my mind as to whether this penalty should stop us making a change altogether if the gain is negative for the tree's currently selected feature. We could for example refuse to change the sign of gain with this penalty. This is something I would like to explore with our new experiment framework and is left as a follow on change if it is beneficial.)


valeriy42 (approved):
LGTM

PR #2001, opened 2021-08-24
[ML] Incremental training evaluation experiments

tveasey:
This generates the full experiments file we plan to use for evaluating incremental training and fixes some outstanding issues  in this area. I also took the opportunity to migrate all test datasets out of the repo since the infrastructure downloads them when they are used.


valeriy42 (approved):
LGTM. Good stuff :+1:

PR #2005, opened 2021-08-31
[ML] Ensure initialiser list ctor is used

davidkyle:
From the debug logging added in #2003 it is clear the input size is not being initialised correctly

`input size: [4354725592, 140222092279815]`

This change ensures the initialiser list ctor is called


droberts195 (approved):
LGTM

PR #2007, opened 2021-09-01
[ML] Fix invalid reference when initialising input sizes

davidkyle:
Use a local var with same lifespan.

I've left the debug logging in place while this is tested and will clean it up once the tests have passed.


droberts195 (approved):
LGTM

PR #2009, opened 2021-09-03
[ML] Store and restore training jobs with  models

valeriy42:
This PR adds functionality to store and restore training jobs. It mainly requires storing the model blob object, but other metadata like `input_filename`, `dependent_variable`, `analysis_name` are also stored for completeness.  Jobs are stored as gzipped JSON files.

The job can be stored with 
```python
job.store(destination=local_path)
```
and restored with 
```python
job = Job.fromFile(source=local_path)
```
See unit tests in `test_store_restore.py` for more examples.

Furthermore, I implemented functionality to upload, download and delete job files in the Google storage bucket.

### DIFF
--- a/jupyter/src/incremental_learning/job.py
+++ b/jupyter/src/incremental_learning/job.py
@@ -16,7 +16,8 @@ import string
 import tempfile
 import time
 from typing import Union
-
+from pathlib import Path
+import sys
 import libtmux
 import numpy as np
 import pandas
@@ -44,8 +45,8 @@ class Job:
     """Job class .
     """
 
-    def __init__(self, input: Union[str, tempfile._TemporaryFileWrapper],
-                 config: Union[str, tempfile._TemporaryFileWrapper],
+    def __init__(self, input: Union[str, tempfile._TemporaryFileWrapper] = None,
+                 config: Union[str, tempfile._TemporaryFileWrapper] = None,
                  persist: Union[None, str, tempfile._TemporaryFileWrapper] = None,
                  restore: Union[None, str, tempfile._TemporaryFileWrapper] = None,
                  verbose: bool = True, run=None):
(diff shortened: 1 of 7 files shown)


jupyter/src/incremental_learning/job.py
```diff
@@ -272,6 +280,101 @@ def get_model_update_data(self) -> str:
...
+            try:
+                json.dump(obj=self, fp=fp, cls=JobJSONEncoder)
+            except:
+                ex = sys.exc_info()
+                logger.error("Failed storing job {}: {}".format(self.name, ex))
+                success = False
+        return success
+
+    @classmethod
+    def asJob(cls, state: dict):
```
tveasey, line 304: nit: stick with snake case conventions used in this file
```suggestion
    def as_job(cls, state: dict):
```

jupyter/src/incremental_learning/job.py
```diff
@@ -272,6 +280,101 @@ def get_model_update_data(self) -> str:
...
+            job.restore_filename = state['restore_filename']
+            job.name = state['name']
+            job.model = state['model']
+            if 'results' in state:
+                job.results =  state['results']
+            job.initialized = True
+        return job
+
+    @classmethod
+    def fromFile(cls, source: Path):
```
tveasey, line 322: As before
```suggestion
    def from_file(cls, source: Path):
```

tveasey (approved):
Nice! Functionality looks good; left a couple of very minor nits on naming, otherwise LGTM.

jupyter/src/incremental_learning/job.py
valeriy42: renamed

jupyter/src/incremental_learning/job.py
valeriy42: renamed

PR #2011, opened 2021-09-03
[ML] Obtain prediction probabilities

valeriy42:
This PR extends Job functionality with an ability to return prediction probabilities for label "true" for binary classification. This is important for the computation of ROC AUC.

### DIFF
--- a/jupyter/src/incremental_learning/job.py
+++ b/jupyter/src/incremental_learning/job.py
@@ -98,6 +98,7 @@ class Job:
             self._set_analysis_name()
             if self.is_classification():
                 self._set_num_classes()
+                self._configure_num_top_classes()
             self.initialized = True
         else:
             self.initialized = False
@@ -118,6 +119,13 @@ class Job:
             with open(self.config_filename) as fp:
                 config = json.load(fp)
             self.num_classes = config['analysis']['parameters']['num_classes']
+
+    def _configure_num_top_classes(self):
+        with open(self.config_filename) as fc:
+            config = json.load(fc)
+        config['analysis']['parameters']['num_top_classes'] = self.num_classes
+        with open(self.config_filename, 'wt') as fc:
+            json.dump(config, fc)
 
     def clean(self):
         if is_temp(self.output):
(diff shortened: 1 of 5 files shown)


jupyter/src/incremental_learning/job.py
```diff
@@ -200,6 +200,30 @@ def get_predictions(self) -> np.array:
...
 
+    def get_probabilities(self) -> Union[None, np.array]:
+        """Return prediction probabilities for the class "true".
+
+        Note: Only works for binary classification.
+
+        Returns:
+            Union[None, np.array]: prediction probabilities. 
+        """
+        if self.analysis_name == 'regression':
```
tveasey, line 211: Could we cache the number of classes in the `self` and fail as well if > 2. It means it won't fail silently and just output nonsense. Alternatively, we could add output top classes to the job configs and get from there (this will work in all cases). This can be done automatically by the framework, i.e. setting from `num_classes` in the config file.

jupyter/src/incremental_learning/job.py
```diff
@@ -200,6 +200,30 @@ def get_predictions(self) -> np.array:
...
+                "Failed to obtain prediction probabilities for the regression job {}."
+                .format(self.name))
+            return None
+        probabilities = []
+        for item in self.results:
+            if 'row_results' in item:
+                target_prediction = item['row_results']['results']['ml']['{}_prediction'.format(
+                    self.dependent_variable)]
+                prediction_probability = item['row_results']['results']['ml']['prediction_probability']
+                if target_prediction == 'false':
```
tveasey, line 222: Does this work? `target_prediction` should be the class label in the input data. I would be inclined to use `{class_name: probability}`. Again we could cache the classes in `self` so we can get at it. Alternatively, this isn't an issue if we read from `top_classes`.

tveasey (commented):
Just a couple of things. The main one being that I don't think we can relying on `target_prediction` being true/false.

jupyter/src/incremental_learning/job.py
valeriy42: Good point, now I raise exceptions if it's `regerssion` or `num_classes` !=2 .

jupyter/src/incremental_learning/job.py
valeriy42: I rewrote the code to augment `num_top_classes` into configuration and then use it for working with prediction probability.

valeriy42:
Thank you for the pointer, @tveasey. I am using `top_classes` now.

tveasey (approved):
LGTM

PR #2013, opened 2021-09-06
[ML] Cleanly handle mismatch in the classification config num_classes and the number classes in the data

tveasey:
An error in the transforms was causing us to generate spurious classes in the transformed data set. This showed up we weren't properly validating the input and in fact failed by SIGSEGV in this case. I've update the validate method to check the data is consistent with the config and also put in explicit protection at the point where we got the undefined behaviour.


valeriy42 (approved):
LGTM

PR #2016, opened 2021-09-07
[ML] Raise runtime error when C++ process fails

valeriy42:
When the training job fails, we want the failure to propagate through the Python framework to mark the experiment as failed.


tveasey (approved):
LGTM

PR #2017, opened 2021-09-07
[ML] Add an option to force update the model

tveasey:
Currently, we assess whether we think that incremental training will reduce loss vs the original model before committing the change. This relies on us being able to accurately estimate the test error for the original model. I've run into cases where we seem to significantly underestimate this on the transformed data we simulate in our test framework. It isn't clear why this is happening: I'm still investigating. However, it is also convenient to simply force accept the best change we found, both for test and also as a user (if one can assess model generalisation error independently). This adds the option and a parameter to the test framework.


tveasey:
retest

valeriy42 (approved):
LGTM

PR #2020, opened 2021-09-08
[ML][PYTORCH] Update macOS build to libtorch 1.9

davidkyle:
Changes the BLAS library to be explicitly set to `vecLib` which is part of the `Accelerate` framework.

non-issue as the change is to unreleased code

### DIFF
--- a/build-setup/macos.md
+++ b/build-setup/macos.md
@@ -155,7 +155,7 @@ Download the graphical installer for Python 3.7.9 from <https://www.python.org/f
 
 Install using all the default options.  When the installer completes a Finder window pops up.  Double click the `Install Certificates.command` file in this folder to install the SSL certificates Python needs.
 
-### PyTorch 1.8.0
+### PyTorch 1.9.0
 
 PyTorch requires that certain Python modules are installed.  To install them:
 
@@ -166,7 +166,7 @@ sudo /Library/Frameworks/Python.framework/Versions/3.7/bin/pip3.7 install instal
 Then obtain the PyTorch code:
 
 ```
-git clone --depth=1 --branch=v1.8.0 https://github.com/pytorch/pytorch.git
+git clone --depth=1 --branch=v1.9.0 https://github.com/pytorch/pytorch.git
 cd pytorch
 git submodule sync
 git submodule update --init --recursive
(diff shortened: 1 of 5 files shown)


dev-tools/download_macos_deps.sh
```diff
@@ -25,7 +25,7 @@ DEST=/usr
 case `uname -m` in
 
     arm64)
-        ARCHIVE=local-arm64-apple-macosx11.1-3.tar.bz2
+        ARCHIVE=local-arm64-apple-macosx11.1-4.tar.bz2
```
droberts195, line 28: This file is in place now.

build-setup/macos.md
```diff
@@ -183,18 +183,18 @@ external processes.
...
 export BUILD_CAFFE2=OFF
 export USE_NUMPY=OFF
 export USE_DISTRIBUTED=OFF
-export USE_MKLDNN=OFF
+export USE_MKLDNN=ON
 export USE_QNNPACK=OFF
 export USE_PYTORCH_QNNPACK=OFF
 [ $(uname -m) = x86_64 ] && export USE_XNNPACK=OFF
 # TODO: recheck if this is still necessary next time we upgrade
 [ $(uname -m) != x86_64 ] && export CMAKE_OSX_ARCHITECTURES=`uname -m`
```
droberts195, line 196: These two lines can be deleted now - the workaround is no longer necessary.

droberts195 (approved):
LGTM

There is one more workaround that can be removed for ARM.

build-setup/macos.md
```diff
@@ -183,18 +183,18 @@ external processes.
...
 
 ```
-# TODO: add BLAS=vecLib when we upgrade to 1.9
+export BLAS=vecLib
 export BUILD_TEST=OFF
 export BUILD_CAFFE2=OFF
 export USE_NUMPY=OFF
 export USE_DISTRIBUTED=OFF
-export USE_MKLDNN=OFF
+export USE_MKLDNN=ON
```
droberts195, line 191: Turns out this doesn't work in PyTorch 1.9 when built for ARM macOS.

The fixes that make `oneDNN` work on Apple silicon are in versions 2.1.3 and 2.2 according to https://github.com/oneapi-src/oneDNN/issues/984#issuecomment-811952961.  And PyTorch 1.9 is using version 2.1.2 😿 .

So I'll make this setting conditional on the CPU type for the time being.

droberts195: It should be possible to build MKLDNN on M1 in PyTorch 1.10 - see https://github.com/pytorch/pytorch/pull/57928.  We'll have to see if there are further problems downstream of simply building it though.

dev-tools/download_macos_deps.sh
droberts195: This file is now updated to a version containing libraries for the correct architecture. 🤦

davidkyle:
retest

PR #2021, opened 2021-09-08
[ML] Reinitialise predictions before estimating initial loss

tveasey:
Since we incrementally train to get the range to use for the tree topology change penalty, we need to reinitialise the predictions before computing the loss for the original model.

### DIFF
--- a/jupyter/src/incremental_learning/transforms.py
+++ b/jupyter/src/incremental_learning/transforms.py
@@ -16,7 +16,7 @@ from typing import List
 from typing import Set
 import random
 
-import pandas
+from .config import logger
 
 def metric_features(categorical_features : List[str],
                     dataset : DataFrame) -> List[str]:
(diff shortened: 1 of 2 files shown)


jupyter/src/incremental_learning/transforms.py
```diff
@@ -65,12 +65,13 @@ def partition_on_metric_ranges(seed : int,
     '''
 
     if matches_columns(metric_features, dataset):
-        print(list(dataset.columns), 'does not contain', metric_features)
+        print('ERROR:', list(dataset.columns), 'does not contain', metric_features)
```
valeriy42, line 68: It would be better if you use the `logger` from `config` instead of printing the messages:
```python
from .config import logger
...
logger.error("{} does not contain {}".format(dataset.columns, metric_features))
```

Same below.

valeriy42 (approved):
LGTM. Just a minor comment regarding Python code.

jupyter/src/incremental_learning/transforms.py
tveasey: Corrected in [1b78fa7](https://github.com/elastic/ml-cpp/pull/2021/commits/1b78fa701cd44cc288e22992683f762b40718c3f).

PR #2022, opened 2021-09-09
[ML] Random seed support for experiment tracking

valeriy42:
This change allows passing the `seed` parameter to the configuration of the `data_frame_analytics`. It also adds a unit test for reproducible runs with sacred.


tveasey (approved):
LGTM

PR #2023, opened 2021-09-09
[ML] Parallelize calls to forward on pytorch models

dimitris-athanasiou:
Parallelizes model forwarding. A new parameter called `numParallelForwardingThreads`
specifies the size of the thread pool used to do model forwarding in parallel. The default
value is `1` for now.

### DIFF
--- a/bin/pytorch_inference/CCmdLineParser.cc
+++ b/bin/pytorch_inference/CCmdLineParser.cc
@@ -34,8 +34,8 @@ bool CCmdLineParser::parse(int argc,
                            bool& isRestoreFileNamedPipe,
                            std::string& loggingFileName,
                            std::string& logProperties,
-                           std::int32_t& numThreads,
-                           std::int32_t& numInterOpThreads,
+                           std::int32_t& numLibTorchThreads,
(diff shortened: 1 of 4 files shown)


dimitris-athanasiou:
This PR is doing the work needed to parallelize model forwarding for pytorch models. However, initial testing revealed that using more than 1 thread results to slowdown. Looking a bit into it revealed that pytorch already uses multithreading for the maths library. Thus, adding parallel forwarding probably leads to thread contention.

My plan is to hold on until we have merged the library optimization work and revisit this.

dimitris-athanasiou:
After fixing a problem with the benchmark script, it is clear that parallel forwarding can indeed provide significant speedups.

I have now exposed a new parameter called `numParallelForwardingThreads` so that we deploy this on more machines and benchmark it. This should provide information as to what would be an appropriate configuration for all the various threading parameters at play.

Also, I profiled this change to see if copying the request each time should be optimized. The result was that it is insignificant compared to `module.forward`.

tveasey (approved):
Changes look good. Just a small suggestion on naming.

PR #2024, opened 2021-09-13
[ML] Performance optimisation for classification and regression model training

tveasey:
Caching example splits proves to be a good runtime optimisation. Testing using our experiment driver on around 250 small data sets the mean runtime dropped from 380s to 320s. However, the runtime improvements were larger on the larger data sets. I'll attach some benchmarks for large datasets as well. I also removed `CImmutableRadixSet` which was no longer used since looking up splits is not on the hot path.

Note that this does have potential to change results slightly. The source of the difference is that we now store the candidate splits in float precision rather than double precision.


include/maths/CBoostedTreeFactory.h
```diff
@@ -139,8 +139,10 @@ class MATHS_EXPORT CBoostedTreeFactory final {
     //! Estimate the maximum booking memory that training the boosted tree on a
     //! data frame with \p numberRows row and \p numberColumns columns will use.
     std::size_t estimateMemoryUsage(std::size_t numberRows, std::size_t numberColumns) const;
-    //! Get the number of columns training the model will add to the data frame.
-    std::size_t numberExtraColumnsForTrain() const;
+    //! Estimate the number of columns training the model will add to the data frame.
+    static std::size_t estimatedExtraColumns(std::size_t numberColumns,
```
valeriy42, line 143: Don't we know the number of extra columns for sure? "Estimate" sounds like something that may deviate in the future, e.g. `estimateMemoryUsage`.

include/core/CImmutableRadixSet.h
```diff
@@ -1,146 +0,0 @@
-/*
```
valeriy42, line 1: Would you please mention in the PR description why this file gets deleted? It is not obvious.

include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -46,6 +45,30 @@ class CBoostedTreeNode;
 class CDataFrameCategoryEncoder;
 class CEncodedDataFrameRowRef;
 
+//! \brief Manages accessing the bytes of CFloatStorage.
+//!
+//! DESCRIPTION:\n
```
valeriy42, line 50: nice! :rocket:

include/maths/CBoostedTreeLeafNodeStatistics.h
```diff
@@ -236,7 +257,6 @@ class MATHS_EXPORT CBoostedTreeLeafNodeStatistics final {
         void swap(CSplitsDerivatives& other) {
             std::swap(m_NumberLossParameters, other.m_NumberLossParameters);
             m_Derivatives.swap(other.m_Derivatives);
-            m_MissingDerivatives.swap(other.m_MissingDerivatives);
```
valeriy42, line 239: Don't you need to handle missing derivative extra anymore?

valeriy42 (approved):
Good work and great results! I have just a few comments regarding readability. Other than those, LGTM! :+1:

include/maths/CBoostedTreeFactory.h
tveasey: We don't know exactly how many encoded features we will choose upfront (for example it'll vary with categorical fields present and how much information they carry). Since we now have additional some storage for each such feature this is only approximate.

include/maths/CBoostedTreeLeafNodeStatistics.h
tveasey: I've just put them at the end of `m_Derivatives`. This was convenient because I need to assign an integer for "this is missing" so I just use # splits + 1 (conveniently it takes all branching out of the hot method `addRowDerivatives`)

include/core/CImmutableRadixSet.h
tveasey: This was to provide optimisation for `addRowDerivatives` but since computing splits is no longer on a hot path I simplified and just use `std::upper_bound`.

tveasey:
The failure was timing related `CStringStoreTest/testStringStore` so I'm going to go ahead and merge this.

tveasey:
Training on 80% Higgs 1M runtime before 9057s after 7641s.

PR #2025, opened 2021-09-14
[ML] Only start thread pool for parallel forwarding if multiple threads

dimitris-athanasiou:
... are used. This is a follow up to #2023. Now that we start the
thread pool for parallel forwarding to LibTorch models, if the setting
is `1` we start an additional thread that isn't necessary. This commit
removes the use of that extra thread if `numParallelForwardingThreads`
is set to `1`.


davidkyle (approved):
LGTM

PR #2027, opened 2021-09-14
[ML] Performance optimisation for classification and regression model training

tveasey:
Caching example splits proves to be a good runtime optimisation. Testing using our experiment driver on around 250 small data sets the mean runtime dropped from 380s to 320s. However, the runtime improvements were larger on the larger data sets. I'll attach some benchmarks for large datasets as well. I also removed CImmutableRadixSet which was no longer used since looking up splits is not on the hot path.

Note that this does have potential to change results slightly. The source of the difference is that we now store the candidate splits in float precision rather than double precision.

This ports #2024 to the incremental training feature branch. This collides with a certain amount of code change so I've essentially reapplied the change to this branch.


tveasey:
Training on 80% Higgs 1M runtime before 9057s after 7641s.

PR #2030, opened 2021-09-15
[ML] Upgrade to PyTorch 1.9 on Windows

droberts195:
macOS and Linux have already been upgraded.

Moving to version 1.9 allows a build workaround
to be removed.

We also enable oneDNN, but do not yet use MKL as
the BLAS. (Potentially we can try that in the next
upgrade, after we've had a chance to see the Linux
benefit in production.)


davidkyle (approved):
LGTM

PR #2028, opened 2021-09-15
[ML] Upgrade to gcc 10.3 for Linux compilation

droberts195:
gcc 10.3 contains the fix for the bug that hinders
compilation of PyTorch on aarch64.

The binutils package is also upgraded, from version
2.34 to version 2.37, and patchelf is upgraded from
version 0.10 to version 0.13.

PR #2032, opened 2021-09-20
[ML] Use gcc 10.3 Docker image for Jupyter Dockerfile

droberts195:
The base image needs to match what the source code
expects, which is version 19 since #2028 was merged.

### DIFF
--- a/jupyter/docker/Dockerfile
+++ b/jupyter/docker/Dockerfile
@@ -7,7 +7,7 @@
 # compliance with the Elastic License 2.0 and the foregoing additional
 # limitation.
 
-FROM docker.elastic.co/ml-dev/ml-linux-build:17 as builder 
+FROM docker.elastic.co/ml-dev/ml-linux-build:19 as builder 
 
 RUN yum install -y epel-release && \
     yum install -y ccache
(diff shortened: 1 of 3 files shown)


dimitris-athanasiou (approved):
LGTM

droberts195:
The test failure that led to the test threshold change in https://github.com/elastic/ml-cpp/pull/2032/commits/3d3db61849bba9876bbc706d36dda4998a37b1d0 was:

```
[Exception] - difference{0.0119332} between minRecalls[0][0]{0.85760517799352753} and minRecalls[1][0]{0.84749190938511332} exceeds 1%
 == [File] - CDataFrameUtilsTest.cc
 == [Line] -1493
```

@tveasey and @valeriy42 the threshold increase is pretty small so I don't think it's a problem but just wanted to flag it for you.  I imagine there's some performance improvement in gcc 10.3 that leads to single-threaded and multi-threaded diverging slightly more.

droberts195:
The latest test failure is detailed in #2033. I don't think that's a problem that's been introduced by this PR - it seems to have wider implications for test reproducibility - so I'll merge as-is and leave #2033 to be investigated separately.

PR #2036, opened 2021-09-22
[ML] Fix unit tests for skip_model_update rule

edsavage:
A number of minor fixes for the `testIgnoreSamplingGivenDetectionRules`
tests that exercise the skip_model_update rule.

In some test cases the rule was improperly configured leading to it
never being executed.

In the case of
`CMetricPopulationModelTest/testIgnoreSamplingGivenDetectionRules`
messages were not being sampled by the models, leading to the rule never
firing as no data was being passed to it. This was resolved by
generating a more complex set of messages to pass to the models.


droberts195 (approved):
LGTM as far as it goes

But in a followup it would be good to uncomment the code below `// TODO These checks fail see elastic/machine-learning-cpp/issues/485` in `lib/model/unittest/CMetricModelTest.cc` and `lib/model/unittest/CMetricPopulationModelTest.cc`, because it seems we still don't have proper test coverage for the metric case.  (And presumably uncommenting will require changing something else in the test or production code.)

Issue #2033, opened 2021-09-20
[ML] Failure in CBoostedTreeTest/testMseIncrementalForTargetDrift


droberts195:
The following error was seen on the `feature/incremental-training` branch for a PR build of #2032:

```
[Exception] - critical check errorDecreaseOnNew > 2.0 * errorIncreaseOnOld has failed [0 <= 0]
 == [File] - CBoostedTreeTest.cc
 == [Line] -918
```

The test could be trivially fixed by changing the assertion on line 918 from `>` to `>=`.  However, this comment makes me think this might be a sign of a worse problem with the test:

```
    // By construction, the prediction error for the old training data must
    // increase because the old and new data distributions overlap and their
    // target values disagree. However, we should see proportionally a much
    // larger reduction in the new training data predcition error.
```

If the test is supposed to be constructed such that the prediction errors for the old and new data distributions cannot both be 0 then something is wrong in the test setup code.

Another thing to note is that this failure only occurred once out of several CI runs. So this test is not deterministic for a given platform/architecture. At one time we were always using the same random number generator seed for ml-cpp unit tests so that behaviour was consistent. If we are no longer doing that then we need to get the test code to print the random number generator seed for each run so that problems like this can be debugged.

tveasey:
This means that the result of incremental training was not accepted, i.e. the validation error was not reduced relative to the original model. I'm presuming it is triggered by the change in compiler perturbing the behaviour of both training and incremental training. We could no doubt update the test data slightly to fix this, say make the drift larger, but I think this warrants investigation.

PR #2039, opened 2021-09-23
[ML] Fix Docker build caching

valeriy42:
This PR fixes caching when building Docker containers. New `ccache` sloppiness settings help to improve cache hits and accelerate re-building. 

I also pre-build the part of the Jupyter container that does not rely on `requirements.txt`. Now, it can be downloaded from the docker registry and does not has to be built locally.


tveasey (approved):
Changes LGTM. The failing unit test is unrelated and tracked in #2033.

PR #2042, opened 2021-09-23
[ML] Remove all trace of elasticsearch-extra

droberts195:
Building ml-cpp and elasticsearch side-by-side by
putting the ml-cpp directory under elasticsearch-extra
is now deprecated by Gradle, and will be removed in
Gradle 8.

With the recent release manager updates we no longer
have any need to use this functionality anywhere in
the main and 7.x branches. This allows the build.gradle
file to be simplified.


dimitris-athanasiou (approved):
LGTM

PR #2046, opened 2021-09-27
[ML] Remove obsolete warning

droberts195:
CStateMachine.h contained a warning about a problem that was
fixed four years ago. When it was fixed the unit test was
reenabled (and hasn't failed since), but the warning comment
was left in. This PR removes the comment.


dimitris-athanasiou (approved):
LGTM

PR #2050, opened 2021-09-28
[ML] Add threading benchmark mode to LibTorch evaluate.py

dimitris-athanasiou:
Adds a mode for benchmarking the different threading parameters
of the libtorch process. The new flag is `--threading_benchmark`.

### DIFF
--- a/bin/pytorch_inference/evaluate.py
+++ b/bin/pytorch_inference/evaluate.py
@@ -40,6 +40,17 @@ LibTorch. If not set LibTorch will choose the defaults. Second, we have
 `--numParallelForwardingThreads` which controls how many threads are
 calling LibTorch's forwarding. If not set it defaults to 1.
 
+THREADING_BENCHMARK MODE
+-----------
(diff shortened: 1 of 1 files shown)


bin/pytorch_inference/evaluate.py
```diff
@@ -297,13 +317,32 @@ def test_evaluation(args):
             print('SUCCESS: inference results match expected', flush=True)
             print()
 
+def threading_benchmark(args):
+    threading_options = [1, 2, 3, 4, 8, 12, 16]
+    results = []
+    for libtorch_threads in threading_options:
+        for mkl_threads in threading_options:
```
davidkyle, line 324: I can't see where `mkl_threads` is used apart from the `print` statements

dimitris-athanasiou: ah yes, forgot to remove that

davidkyle (approved):
LGTM

PR #2051, opened 2021-09-28
[ML] Handle chunked state in Python Job object

tveasey:
If either the data summary or model is large enough to require chunking we need to ensure we pass all the chunks to the `data_frame_analyzer` from the Python `Job` object. It doesn't matter how we join documents because we read the stream using `kParseStopWhenDoneFlag`.


valeriy42 (approved):
Great catch :rocket:

PR #2052, opened 2021-09-28
[ML] Handle field mismatch in incremental training

tveasey:
The data which was prepared for incremental training of the DGA model from sunburst domains used a different field order and renamed one of the fields. Whilst this wouldn't typically happen in Java, which sorts the field values it supplies by name, it is fragile to rely on this. It is also conceivable that a subset of features aren't available for the update data. This checks that there are no additional fields supplied (likely to indicate a user error) and handles the case that the fields are supplied in a different order or are missing altogether.

### DIFF
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -260,8 +260,8 @@ void CDataFrameAnalyzer::captureFieldNames(const TStrVec& fieldNames) {
         if (m_DataFrame->columnNames().empty() == false &&
             m_DataFrame->columnNames() != columnNames) {
 
-            // We take the view that missing features are not fatal, since they may
-            // not be available for the new set, but extra features are likely to
+            // We take the view that missing fields are not fatal, since they may
+            // not be available for the new set, but extra fields are likely to
             // indicate user error.
 
             TSizeVec positions(columnNames.size());
--- a/lib/api/CDataFrameAnalyzer.cc
+++ b/lib/api/CDataFrameAnalyzer.cc
@@ -11,6 +11,7 @@
 
 #include <api/CDataFrameAnalyzer.h>
 
+#include <boost/unordered/unordered_map_fwd.hpp>
 #include <core/CContainerPrinter.h>
 #include <core/CDataFrame.h>
 #include <core/CFloatStorage.h>
@@ -128,7 +129,7 @@ void CDataFrameAnalyzer::run() {
     // We create the writer in run so that when it is finished destructors
     // get called and the wrapped stream does its job to close the array.
 
-    auto analysisRunner = m_AnalysisSpecification->runner();
+    auto* analysisRunner = m_AnalysisSpecification->runner();
     if (analysisRunner != nullptr) {
         // We currently use a stream factory because the results are wrapped in
         // an array. This is managed by the CJsonOutputStreamWrapper constructor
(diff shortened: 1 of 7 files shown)


lib/api/CDataFrameAnalyzer.cc
```diff
@@ -250,10 +255,49 @@ void CDataFrameAnalyzer::captureFieldNames(const TStrVec& fieldNames) {
...
+            if (extraColumnNames.empty() == false) {
+                HANDLE_FATAL(<< "Input error: supplying additional columns "
+                             << core::CContainerPrinter::print(extraColumnNames) << ".");
+            }
+
+            for (const auto& name : m_DataFrame->columnNames()) {
+                auto i = std::lower_bound(columnNames.begin(), columnNames.end(), name);
+                if (i == columnNames.end() || *i != name) {
+                    LOG_WARN(<< "Missing column '" << name << "'");
+                    m_ColumnMap->push_back(columnNames.size());
```
dimitris-athanasiou, line 291: This seems to denote missing values with `columnNames.size()`. However, in a comment in `CDataFrameAnalyzer.h` we have ```// This has values: -2 (unset), -1 (missing), >= 0 (control field index).```. Should we not be denoting missing values with `-1`? Or this refers to something else?

lib/maths/CBoostedTreeImpl.cc
```diff
@@ -231,7 +231,7 @@ void CBoostedTreeImpl::train(core::CDataFrame& frame,
 
     m_TrainingProgress.progressCallback(m_Instrumentation->progressCallback());
 
-    std::int64_t lastMemoryUsage(this->memoryUsage());
+    std::int64_t lastMemoryUsage{static_cast<std::int64_t>(this->memoryUsage())};
```
dimitris-athanasiou, line 234: The changes in this file seem unrelated. Are they intentional? (just checking they didn't pop here from another piece of work).

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -1523,6 +1519,255 @@ BOOST_AUTO_TEST_CASE(testClassificationIncrementalTraining) {
             row->copyTo(newTrainingData.back().begin());
         }
     });
+
+    frame->resizeRows(0);
```
dimitris-athanasiou, line 1523: This seems to complete this test but it's not strictly related to the field order stuff, right?

tveasey: This is just the diff getting confused because there is similarity to test I added. If you scroll down you'll see that this was essentially as it was before.

lib/maths/CBoostedTreeImpl.cc
tveasey: I noticed some compile warnings from these statements when I added some debug. While I could factor this out into a separate change, since it is all on a branch which we'll squash when we merge I'm not sure it is worthwhile.

lib/api/CDataFrameAnalyzer.cc
tveasey: I agree it would be more consistent in `CDataFrameAnalyzer`, but I also think it is more natural where it's used in `CDataFrame::parseAndWriteRow` to make this a vector of unsigned ints and denote missing as I have done. I guess this is a tradeoff and I don't have strong opinions, if you find the suggested version clearer I'm happy to change: see [e7be3e6](https://github.com/elastic/ml-cpp/pull/2052/commits/e7be3e6d2ddfdd751392090d3f6bed2585b11fce).

dimitris-athanasiou: I find it easier to read and understand without e7be3e6. Let it be inconsistent :-)

dimitris-athanasiou (approved):
LGTM

Issue #1272, opened 2020-05-20
[ML] Improve the usability of `skip_model_update` rules


tveasey:
`skip_model_update` is potentially a powerful technique for excluding updates to the model by known bad events, but it has some usability issues:
1. If you apply it from the start of the modelling it can stop the model learning anything at all.
2. It can stop the model ever adapting to some change in data characteristics.
3. It doesn't stop score quantiles being updated. This means anomaly severities from excluded values can diminish over time. However, if you are told something should be excluded because it is a known bad event you shouldn't decrease its severity based on frequency.
4. You can only apply it based on absolute thresholds, such as the magnitude of the difference from prediction. A more natural approach would be to apply it based on the anomaly score or some related measure.

Note that 3 mitigates problem 1 to some extent we could also have a fade in. Problem 2 could be addressed if we made the model mainly (but not completely) ignore the skipped data. This would probably give an all round nicer experience.

PR #2053, opened 2021-09-28
[ML] Improve aspects of skip_model_update rule

edsavage:
Move away from completely ignoring a sample when the skip_model_update
rule applies to instead adding the sample with a small multiplicative
weight applied. This has the benefit of reducing the impact of the
matched samples on the model while still allowing the model to react to
changes in data characteristics.

Related to elastic/ml-cpp#1272

### DIFF
--- a/lib/model/CEventRateModel.cc
+++ b/lib/model/CEventRateModel.cc
@@ -285,9 +285,21 @@ void CEventRateModel::sample(core_t::TTime startTime,
                     continue;
                 }
 
+                // initialCountWeight returns a weight value as double:
+                // 0.0 if checkScheduledEvents is true
+                // 1,0 if both checkScheduledEvents and checkRules are false
+                // A small weight - 0.005 - if checkRules is true.
+                // This weight is applied to countWeight (and therefore scaledCountWeight) as multiplier.
+                // This reduces the impact of the values affected by the skip_model_update rule
+                // on the model while not completely ignoring them. This still allows the model to
+                // learn from the affected values - addressing point 1. and 2. in
+                // https://github.com/elastic/ml-cpp/issues/1272, Namely
+                // 1. If you apply it from the start of the modelling it can stop the model learning anything at all.
+                // 2. It can stop the model ever adapting to some change in data characteristics
                 core_t::TTime sampleTime = model_t::sampleTime(feature, time, bucketLength);
-                if (this->shouldIgnoreSample(feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID,
-                                             sampleTime)) {
+                double initialCountWeight = this->initialCountWeight(
+                    feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID, sampleTime);
+                if (initialCountWeight == 0.0) {
                     model->skipTime(sampleTime - lastBucketTimesMap[pid]);
                     continue;
                 }
@@ -304,7 +316,7 @@ void CEventRateModel::sample(core_t::TTime startTime,
                     feature, static_cast<double>(data_.second.s_Count));
                 TDouble2Vec value{count};
                 double winsorisationDerate = this->derate(pid, sampleTime);
-                double countWeight = this->learnRate(feature);
+                double countWeight = initialCountWeight * this->learnRate(feature);
                 // Note we need to scale the amount of data we'll "age out" of the residual
                 // model in one bucket by the empty bucket weight so the posterior doesn't
                 // end up too flat.
@@ -631,8 +643,8 @@ bool CEventRateModel::fill(model_t::EFeature feature,
         model->seasonalWeight(maths::DEFAULT_SEASONAL_CONFIDENCE_INTERVAL, time, result);
         return maths_t::seasonalVarianceScaleWeight(result);
     }()};
-    bool skipAnomalyModelUpdate = this->shouldIgnoreSample(
-        feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID, time);
+    double initialCountWeight{this->initialCountWeight(
+        feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID, time)};
 
     params.s_Feature = feature;
     params.s_Model = model;
@@ -650,7 +662,7 @@ bool CEventRateModel::fill(model_t::EFeature feature,
     params.s_ComputeProbabilityParams
         .addCalculation(model_t::probabilityCalculation(feature))
         .addWeights(weights)
-        .skipAnomalyModelUpdate(skipAnomalyModelUpdate);
+        .initialCountWeight(initialCountWeight);
 
     return true;
 }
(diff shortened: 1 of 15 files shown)


lib/model/CEventRateModel.cc
```diff
@@ -285,9 +285,21 @@ void CEventRateModel::sample(core_t::TTime startTime,
                     continue;
                 }
 
+                // initialCountWeight returns a weight value as double:
+                // 0.0 if checkScheduledEvents is true
+                // 1,0 if both checkScheduledEvents and checkRules are false
```
droberts195, line 290:
```suggestion
                // 1.0 if both checkScheduledEvents and checkRules are false
```

lib/model/CMetricModel.cc
```diff
@@ -244,9 +244,22 @@ void CMetricModel::sample(core_t::TTime startTime,
                     continue;
                 }
 
+                // initialCountWeight returns a weight value as double:
+                // 0.0 if checkScheduledEvents is true
+                // 1,0 if both checkScheduledEvents and checkRules are false
```
droberts195, line 249:
```suggestion
                // 1.0 if both checkScheduledEvents and checkRules are false
```

lib/model/CMetricPopulationModel.cc
```diff
@@ -385,8 +385,21 @@ void CMetricPopulationModel::sample(core_t::TTime startTime,
                     LOG_ERROR(<< "Missing model for " << this->attributeName(cid));
                     continue;
                 }
+                // initialCountWeight returns a weight value as double:
+                // 0.0 if checkScheduledEvents is true
+                // 1,0 if both checkScheduledEvents and checkRules are false
```
droberts195, line 390:
```suggestion
                // 1.0 if both checkScheduledEvents and checkRules are false
```

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -2329,7 +2335,7 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
     // Check the last value times of the underlying models are the same
     // const maths::CUnivariateTimeSeriesModel *timeSeriesModel =
     //     dynamic_cast<const maths::CUnivariateTimeSeriesModel*>(modelWithSkipView->model(model_t::E_IndividualMeanByPerson, 0));
-    // BOOST_TEST_REQUIRE(timeSeriesModel != 0);
+    // BOOST_TEST_REQUIRE(timeSeriesModel != nullptr);
```
droberts195, line 2338: This same change can be made on line 1319 of `lib/model/unittest/CMetricPopulationModelTest.cc`

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -2292,26 +2292,32 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
...
     }
     modelNoSkip->sample(startTime, endTime, m_ResourceMonitor);
     modelWithSkip->sample(startTime, endTime, m_ResourceMonitor);
 
-    // Checksums will be different due to the data gatherers
+    // Checksums will be different due to the small weighting applied to the sample
+    // added to the model with the detector rule.
     BOOST_TEST_REQUIRE(modelWithSkip->checksum() != modelNoSkip->checksum());
 
     // TODO this test fails due a different checksums for the decay rate and prior
```
droberts195, line 2323: It would be good to replace the commented out code with some checks on the "underlying models" which are supposed to be the same.  If it's non-trivial then feel free to do this in a followup PR.

lib/model/unittest/CMetricPopulationModelTest.cc
```diff
@@ -1226,48 +1223,94 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
...
-    withSkipChecksum = modelWithSkipView
-                           ->model(model_t::E_PopulationMeanByPersonAndAttribute, 1)
-                           ->checksum();
-    noSkipChecksum =
-        modelNoSkipView->model(model_t::E_PopulationMeanByPersonAndAttribute, 1)->checksum();
-    BOOST_REQUIRE_EQUAL(withSkipChecksum, noSkipChecksum);
+    BOOST_TEST_REQUIRE(withSkipChecksum != noSkipChecksum);
 
     // TODO These checks fail see elastic/ml-cpp/issues/2043
     // Check the last value times of all the underlying models are the same
```
droberts195, line 1316: As above, please replace the commented out checks with checks on the underlying models (in a followup PR if it's non-trivial).

droberts195 (commented):
Looks good.  Most of my review comments are about old code comments that need changing.

lib/model/unittest/CMetricModelTest.cc
edsavage: The comment was out of date and misleading.  We now do expect the underlying models to differ (due to the application of different sample weights). I've updated the comment and added an explicit check for the expected differing checksums.

lib/model/unittest/CMetricModelTest.cc
```diff
@@ -2292,44 +2292,54 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
...
+        modelWithSkipView->model(model_t::E_IndividualMeanByPerson, 0);
+    BOOST_TEST_REQUIRE(mathsModelWithSkip != nullptr);
+    uint64_t withSkipChecksum = mathsModelWithSkip->checksum();
+    const maths::CModel* mathsModelNoSkip =
+        modelNoSkipView->model(model_t::E_IndividualMeanByPerson, 0);
+    BOOST_TEST_REQUIRE(mathsModelNoSkip != nullptr);
+    uint64_t noSkipChecksum = mathsModelNoSkip->checksum();
+    BOOST_TEST_REQUIRE(withSkipChecksum != noSkipChecksum);
 
     // TODO These checks fail see elastic/ml-cpp/issues/2043
```
droberts195, line 2338: This issue doesn't seem to tie up with the code that's commented out.  The issue says the problem is with checksums, but the code below is comparing last times.

It seems like we shouldn't have this issue reference here and either the last times _should_ be the same, or, if they are not then it's because we've skipped data from the most recent bucket and the commented out assertions should just be deleted as they're obviously never going to work.

lib/model/unittest/CMetricPopulationModelTest.cc
```diff
@@ -1226,54 +1223,100 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
...
-
-    withSkipChecksum = modelWithSkipView
-                           ->model(model_t::E_PopulationMeanByPersonAndAttribute, 1)
-                           ->checksum();
-    noSkipChecksum =
-        modelNoSkipView->model(model_t::E_PopulationMeanByPersonAndAttribute, 1)->checksum();
-    BOOST_REQUIRE_EQUAL(withSkipChecksum, noSkipChecksum);
+    BOOST_TEST_REQUIRE(withSkipChecksum != noSkipChecksum);
 
     // TODO These checks fail see elastic/ml-cpp/issues/2043
```
droberts195, line 1315: Again, it seems like the issue doesn't match what's commented out, because the issue is for checksums while the commented out code is for last value times.  Or am I missing something?

droberts195 (approved):
LGTM assuming the new tests pass

PR #2054, opened 2021-09-29
[ML] Rename LibTorch threading parameters

dimitris-athanasiou:
Renames `--numLibTorchThreads` to `--inferenceThreads`.
Renames `--numParallelForwardingThreads` to `--modelThreads`.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -16,6 +16,7 @@
 #include <core/CRapidJsonConcurrentLineWriter.h>
 #include <core/CSetEnv.h>
 #include <core/CStopWatch.h>
+#include <core/CStringUtils.h>
 #include <core/Concurrency.h>
 
 #include <seccomp/CSystemCallFilter.h>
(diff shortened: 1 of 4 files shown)


bin/pytorch_inference/Main.cc
```diff
@@ -290,14 +293,14 @@ int main(int argc, char** argv) {
         return EXIT_FAILURE;
     }
 
-    if (numLibTorchThreads != -1) {
-        at::set_num_threads(numLibTorchThreads);
-    }
-    if (numLibTorchInterOpThreads != -1) {
-        at::set_num_interop_threads(numLibTorchInterOpThreads);
-    }
+    at::set_num_threads(inferenceThreads);
```
droberts195, line 296: This can currently be set to 0 or a huge number.  Probably we should constrain it into the range [1, `std::thread::hardware_concurrency`].

bin/pytorch_inference/Main.cc
```diff
@@ -290,14 +293,14 @@ int main(int argc, char** argv) {
...
+
+    // This is not used as we don't call at::launch anywhere.
+    // Setting it to 1 to ensure there is no thread pool sitting around.
+    at::set_num_interop_threads(1);
+
     LOG_DEBUG(<< at::get_parallel_info());
-    LOG_DEBUG(<< "Number of parallel forwarding threads: " << numParallelForwardingThreads);
+    LOG_DEBUG(<< "Model threads: " << modelThreads);
```
droberts195, line 303: Similarly, maybe we should constrain this to be at most `std::thread::hardware_concurrency`.  (0 is already equivalent to 1 for this one.)

droberts195 (approved):
LGTM, but left a suggestion to constrain the supplied values

bin/pytorch_inference/Main.cc
dimitris-athanasiou: Ah yes, capping. Will add this.

dimitris-athanasiou:
@droberts195 I have added validation to the threading parameters to keep them within `[1, std::thread::hardware_concurrency]`.

droberts195 (approved):
LGTM

PR #2057, opened 2021-10-04
[ML] Capture best hyperparameters after restoring for final train

tveasey:
We were missing a capture of the best hyperparameters by instrumentation after restoring for final train. This is most noticeable when training with all parameters overridden in which case we see uninitialised values in the analysis stats for the job, i.e. before
[
Screenshot 2021-10-04 at 09 46 18
](url)
after
Screenshot 2021-10-04 at 09 46 45

As above I've confirmed it works manually and I don't think the cost benefit means it is worthwhile writing a unit test since  the test is not straightforward to write and it is unlikely that this behaviour will get broken.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -55,6 +55,8 @@
 
 * Speed up training of regression and classification models (See {ml-pull}2024[#2024].)
 * Improve aspects of implementation of skip_model_update rule (See {ml-pull}2053[#2053].)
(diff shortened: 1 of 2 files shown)


tveasey:
retest

tveasey:
Awaits fix for https://gitlab.com/libeigen/eigen/-/issues/2336.

valeriy42 (approved):
LGTM :+1: Good catch!

tveasey:
retest

tveasey:
retest

PR #2058, opened 2021-10-04
[ML] Switch to backup repo for Eigen

droberts195:
The main Eigen repo on GitLab has been accidentally
deleted. Hopefully it will be recovered by tomorrow,
but if not we can merge this PR, which instead clones
a temporary repo that's been created as an interim
workaround.


droberts195:
I would prefer not to have to merge this, as it will be extra work to revert it again.  But if the GitLab problem is still not fixed by tomorrow then I'll merge it first thing.

tveasey (approved):
LGTM

droberts195:
This is not needed - the Eigen repo has been restored.

PR #2060, opened 2021-10-04
[7.15] [ML] Switch to backup repo for Eigen

droberts195:
The main Eigen repo on GitLab has been accidentally
deleted. Hopefully it will be recovered by tomorrow,
but if not we can merge this PR, which instead clones
a temporary repo that's been created as an interim
workaround.

Backport of #2058


droberts195:
This is not needed - the Eigen repo has been restored.

PR #2068, opened 2021-10-11
[ML] Maths library restructure

edsavage:
Restructure the maths library into three smaller shared libraries:
 * libMlMathsTimeSeries - functions specific to time series analysis
 * libMlMathsAnalytics - functions specific to data frame analytics
 * libMlMathsCommon - general purpose functions used by the other two
 libraries and elsewhere.


lib/api/dump_state/Makefile
```diff
@@ -12,7 +12,7 @@ include $(CPP_SRC_HOME)/mk/defines.mk
 
 TARGET=dump_state$(EXE_EXT)
 
-ML_LIBS=$(LIB_ML_CORE) $(LIB_ML_MATHS) $(LIB_ML_MODEL) $(LIB_ML_API)
+ML_LIBS=$(LIB_ML_CORE) $(LIB_ML_MODEL) $(LIB_ML_API)
```
droberts195, line 15: I suspect this will work on macOS, but not Linux.  On Linux if the `model` library references a maths library function that has a default value that uses a static constant then the `.so` will need to be linked - something to look out for in the PR CI.

lib/maths/time_series/Makefile
```diff
@@ -0,0 +1,51 @@
...
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+include $(CPP_SRC_HOME)/mk/defines.mk
+
+TARGET=$(OBJS_DIR)/libMlMathsTimeSeries$(DYNAMIC_LIB_EXT)
+
+USE_BOOST=1
+USE_BOOST_FILESYSTEM_LIBS=1
```
droberts195, line 16: Are you sure this is needed for all 3 split libraries?  It seems like it might not be widely used within the maths library.

lib/maths/analytics/Makefile
```diff
@@ -0,0 +1,41 @@
...
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+include $(CPP_SRC_HOME)/mk/defines.mk
+
+TARGET=$(OBJS_DIR)/libMlMathsAnalytics$(DYNAMIC_LIB_EXT)
+
+USE_BOOST=1
+USE_BOOST_FILESYSTEM_LIBS=1
```
droberts195, line 16: Are you sure this is needed for all 3 split libraries?  It seems like it might not be widely used within the maths library.

lib/maths/common/Makefile
```diff
@@ -0,0 +1,83 @@
...
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+include $(CPP_SRC_HOME)/mk/defines.mk
+
+TARGET=$(OBJS_DIR)/libMlMathsCommon$(DYNAMIC_LIB_EXT)
+
+USE_BOOST=1
+USE_BOOST_FILESYSTEM_LIBS=1
```
droberts195, line 16: Are you sure this is needed for all 3 split libraries?  It seems like it might not be widely used within the maths library.

droberts195 (commented):
I only reviewed the `Makefile`s.

I am assuming all the `.h` and `.cc` changes are mechanical renames or deleting entire classes. If you made non-trivial changes to any of these files please flag them up - we have had bugs introduced in the past where people sneaked unrelated refactors into huge unreviewable change sets like this one.

I also think you might have missed the `Makefile`s under `bin`.  Or maybe they're in here but the browser search function couldn't find them because the the volume to search was too great 😆 .

include/maths/analytics/ImportExport.h
```diff
@@ -0,0 +1,47 @@
...
+#endif
+
+//! See https://devblogs.microsoft.com/cppblog/optimizing-the-layout-of-empty-base-classes-in-vs2015-update-2-3/
+#define EMPTY_BASE_OPT __declspec(empty_bases)
+
+#else
+
+// Empty string on Unix
+#define MATHS_ANALYTICS_EXPORT
+#define EMPTY_BASE_OPT
```
droberts195, line 43: Now this is in 3 files, does it cause warnings about defining a macro without undefining it first?  This could be causing build noise on some platforms even if it doesn't happen with clang.

It's also not best practice to be defining the same thing multiple times, so maybe now this is used in multiple libraries it should be moved to `core/WindowsSafe.h`.  Then the classes that need `EMPTY_BASE_OPT` will have to include `core/WindowsSafe.h`, but at least they'll all get the same definition if the way this is done in the future changes.  `STDCALL` is already defined in `core/WindowsSafe.h`, so there's already a precedent.

PR #2070, opened 2021-10-13
[ML] Maths library refactor

edsavage:
Merge the maths library refactor changes to the incremental-learning feature branch.

This splits the maths library into three shared libraries:
 * libMlMathsTimeSeries - functions specific to time series analysis
 * libMlMathsAnalytics - functions specific to data frame analytics
 * libMlMathsCommon - general purpose functions used by the other two libraries and elsewhere.

Oddly I notice a unit test failure in `CBoostedTreeTest/testFeatureBag` at line 1424:
`fatal error: in "CBoostedTreeTest/testFeatureBags": critical check distanceToSorted(selectedForTree) < 0.007 has failed [0.010800000000000001 >= 0.0070000000000000001]` and also at line 1425
`fatal error: in "CBoostedTreeTest/testFeatureBags": critical check distanceToSorted(selectedForNode) < 0.01 has failed [0.0114 >= 0.01]` when testing locally on my Mac. As the test fails in exactly the same way prior to the merge I don't think it can be related to the code changes. I've left the test as-is to see how it behaves under CI


edsavage:
Merging this now with a view to a further investigation of possible memory errors afterwards.

PR #2071, opened 2021-10-13
[ML] Refactor boosted tree hyperparameters and improve test coverage

tveasey:
The management of boosted tree hyperparameters neatly factors out of `CBoostedTreeImpl` which contains too much functionality. This PR makes that change. It would also be good to pull some code related to the hyperparameter bounding box initialisation out of `CBoostedTreeFactory`, but since this is already a large change we can do this as a follow up. This also means we can better unit test hyperparameter optimisation.

Whilst making this change it highlighted some issues around overrides:
1. We weren't properly respecting the maximum number of trees parameter
2. Regularisation hyperparameter scaling wasn't working correctly. In particular, we would double scale parameters failing over in incremental training. Not account for scaling when initialising the tree topology change penalty and also misapply scaling if values were overridden by copying from the job stats and launching a new process.
3. We weren't properly overriding downsample factor.

### DIFF
--- a/lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
+++ b/lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
@@ -1621,7 +1621,7 @@ BOOST_AUTO_TEST_CASE(testIncrementalTrainingFieldMismatch) {
             .predictionDownsampleFactor(downsampleFactor)
             .predictionFeatureBagFraction(featureBagFraction)
             .previousTrainLossGap(lossGap)
-            .previousTrainNumberRows(4 * numberExamples / 5)
+            .previousTrainNumberRows(4 * numberExamples / 5) // This needs to count for a train fold.
             .predictionPersisterSupplier(persisterSupplier)
             .predictionRestoreSearcherSupplier(restorerSupplier)
             .regressionLossFunction(TLossFunctionType::E_BinaryClassification)


include/maths/CBoostedTreeImpl.h
```diff
@@ -472,85 +424,79 @@ class MATHS_EXPORT CBoostedTreeImpl final {
     void startProgressMonitoringTrainIncremental();
 
     //! Record the training state using the \p recordTrainState callback function
-    void recordState(const TTrainingStateCallback& recordTrainState) const;
+    void outputState(const TTrainingStateCallback& recordTrainState) const;
```
valeriy42, line 427: "output" is out of sync with the documentation and the parameter name.

include/maths/CBoostedTreeImpl.h
```diff
@@ -472,85 +424,79 @@ class MATHS_EXPORT CBoostedTreeImpl final {
...
 
     //! Record hyperparameters for instrumentation.
-    void recordHyperparameters();
-
-    //! Populate the list of tunable hyperparameters.
-    void initializeTunableHyperparameters();
-
-    //! Use Sobol sampler for for random hyperparamers.
-    void initializeHyperparameterSamples();
+    void outputHyperparameters();
```
valeriy42, line 430: Same here, why don't you like "record"?

valeriy42:
Really nice work altogether! I like how easy it is to test new code. I left a few comments regarding readability.

valeriy42:
I also think we could safely factor out all code related to storing/restoring and checking invariants of `CBoostedTreeImpl`, which would further reduce complexity of the class.

include/maths/CBoostedTreeImpl.h
tveasey: For me it helps readability to make it clearer that this is sending stuff externally. We also for example capture the best hyperparameters and to me recording could just as easily be internally capturing the stuff rather than writing it elsewhere. I agree that this is inconsistent as is and I don't feel too strongly about the output terminology. It might be too much to make the knock on changes for consistency, I'll see, in which case I'll revert.

PR #2073, opened 2021-10-18
[ML] Address some issues with ANOVA for GP

tveasey:
Because we scale the evaluation point in `CBayesianOptimisation::add` we weren't correctly getting values in the range `[0,1]` when computing the various ANOVA terms. It also isn't necessary to scale the kernel since this is already handled by scaling the inputs. At the same time I addressed the early stopping criterion which is affected by this change. In particular, what we really care about is are we going to have a significant impact on QoR if we continue which relates to the relative change in the mean loss we expect across the search bounding box. This measures this quantity directly and requires we expect at least a 0.1% improvement by continuing.


valeriy42 (approved):
LGTM.

PR #2088, opened 2021-10-27
[ML] Fix multi-step learning

valeriy42:
- Updated the logger format to output time
- Metrics and configurations from the update steps are stored in the -metrics index
- Source file of the experiment_driver is stored in the index.


tveasey (approved):
LGTM

PR #2096, opened 2021-11-01
[ML] Improve skip_model_update behaviour.

edsavage:
Improve the outcome of using the skip_model_update rule by not updating
the quantiles whenever the rule applies to a value. This has the effect
of ensuring that the severities of the affected anomalies are not
diminished over time, i.e. the expected behaviour of this code change is to
skew the distribution of anomaly severities towards being more severe.

Relates to issue #1272


droberts195 (commented):
The code looks good, but please can you add a test to prove it works?

lib/model/CEventRateModel.cc
```diff
@@ -422,6 +422,13 @@ bool CEventRateModel::computeProbability(std::size_t pid,
             continue;
         }
 
+        if (this->initialCountWeight(
```
tveasey, line 425: I think the intention would be much clearer if we had a new method called something `shouldSkipUpdate`. This could be implemented on `CModel` and behind the scene just do 
```
return this->initialCountWeight(
                feature, pid, model_t::INDIVIDUAL_ANALYSIS_ATTRIBUTE_ID,
                model_t::sampleTime(feature, startTime, bucketLength)) != 1.0
```

lib/model/CAnnotatedProbability.cc
```diff
@@ -166,6 +166,7 @@ void SAnnotatedProbability::swap(SAnnotatedProbability& other) noexcept {
     s_DescriptiveData.swap(other.s_DescriptiveData);
     std::swap(s_CurrentBucketCount, other.s_CurrentBucketCount);
     std::swap(s_BaselineBucketCount, other.s_BaselineBucketCount);
+    std::swap(s_ShouldUpdateQuantiles, other.s_ShouldUpdateQuantiles);
```
tveasey, line 169: I'm pretty sure we should add this to persist and restore as well since we have persistence for this object.

lib/model/CEventRateModel.cc
tveasey: I also wonder whether it is strictly more correct to do this before checking `shouldIgnoreResult` and continuing because we probably want to not update quantiles even if we are ignoring the result.

tveasey (commented):
I think it would be nice to have a method which wraps up the logic to test the initial weight to make the intention clearer and also for consistency with ignoring the result. Regarding the test. I think we can test this qualitatively as follows: 1. create a rule you know will apply to some test data, 2. run AD with and without the rule against the data, 3. check that the anomaly scores are higher for the values to which the rule condition matches.

edsavage:
With respect to testing, following Tom's suggestion I created an AD job using the `nyc_taxi` dataset:
 * Indices: nyc_taxi
 * bucket_span: 3h
 * detector: low_sum(value)
 
With no rule in effect the anomalies had the following distribution

Severity | Anomalies
-- | --
Warning | 42
Minor | 3
Major | 2
Critical | 4
Total | 51

Adding the rule `skip_model_update when actual lt 40000` and running the job with the code changes included in this PR the anomalies were as follows:

Severity | Anomalies
-- | --
Warning | 31
Minor | 14
Major | 9
Critical | 4
Total | 58

i.e. there are slightly more anomalies and their severities are skewed away from `warning` to consisting of more `minor` and `major`  severities.

tveasey (approved):
LGTM

PR #2097, opened 2021-11-01
[ML] Support supplying hyperparameter ranges

tveasey:
For train by query it is useful to be able to control hyperparameter ranges for each iteration from the coordinating process.

### DIFF
--- a/lib/api/CDataFrameTrainBoostedTreeRunner.cc
+++ b/lib/api/CDataFrameTrainBoostedTreeRunner.cc
@@ -196,7 +196,7 @@ CDataFrameTrainBoostedTreeRunner::CDataFrameTrainBoostedTreeRunner(
                     [](double x) { return x <= 0.0; })) {
         HANDLE_FATAL(<< "Input error: '" << ETA_GROWTH_RATE_PER_TREE << "' should be positive.");
     }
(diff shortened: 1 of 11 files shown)


lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -173,43 +177,53 @@ CDataFrameTrainBoostedTreeRunner::CDataFrameTrainBoostedTreeRunner(
         m_CustomProcessors.CopyFrom(*parameters[FEATURE_PROCESSORS].jsonObject(),
                                     m_CustomProcessors.GetAllocator());
     }
-    if (alpha != -1.0 && alpha < 0.0) {
+
+    if (std::any_of(alpha.begin(), alpha.end(), [](double x) { return x < 0.0; })) {
```
valeriy42, line 181: This would be a good candidate for a function to avoid redundancy and improve readability.

valeriy42 (approved):
LGTM. Just a single minor remark concerning readability.

lib/api/CDataFrameTrainBoostedTreeRunner.cc
tveasey: I thought about this but many of the conditions are slightly different so I'd have to have different named functions for each possible condition we have on the variable. In the end I decided it was best to just go with the lambda approach.

PR #2101, opened 2021-11-02
[7.15] [ML] Lock the clang version used to build Boost in ARM CI

droberts195:
If we upgrade Xcode on one of our macOS ARM build servers
then we won't rebuild Boost for old branches that are past
feature freeze, but Boost adds the compiler version into
the names of its libraries. Therefore when running in CI
we need to hardcode the version of clang that was used to
build Boost for each branch.

We used to avoid this problem by always using a macOS
build platform for which Xcode was no longer being updated.
For example, the last version of Xcode that works on
Mojave is 11.3, so by advancing the build platform to
Mojave after Xcode 11.4 was released we could be sure that
Xcode would never be upgraded on a build server where we
build for release. (Plus this meant the code was proven by
unit tests to run on old versions of macOS.) With Big Sur
on ARM this approach wasn't possible, as Big Sur was the
first version of macOS to work on ARM. As the years go by
we should revert to the old approach of building on an
older version of macOS for releases.

Backport of #2098

### DIFF
--- a/dev-tools/jenkins_ci.sh
+++ b/dev-tools/jenkins_ci.sh
@@ -99,6 +99,15 @@ case `uname` in
         else
             TASKS="clean buildZip check"
         fi
+        # For macOS we usually only use a particular version as our build platform
+        # once Xcode has stopped receiving updates for it. However, with Big Sur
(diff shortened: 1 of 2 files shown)


droberts195:
retest

PR #2102, opened 2021-11-03
[ML] Fix blocking call cancellation on recent glibc

droberts195:
In recent versions of glibc the pthread_kill function
requires the __NR_getpid system call. This is the case
in Ubuntu 18.04. It is not the case in RHEL/CentOS 7.
Since __NR_getpid was not permitted by our system call
filter pthread_kill did not work on recent versions of
glibc with bad consequences for cleaning up after
connection timeouts.

Additionally there was a much more minor bug where the
named pipe did not get removed if cancellation of the
named pipe connection occurred at one specific place.


edsavage (approved):
LGTM

PR #2106, opened 2021-11-03
[ML] Upgrading RapidJSON

droberts195:
RapidJSON hasn't had a new release for 5 years, but lots of
improvements have been committed to its master branch, so we
should pick up those improvements. (The commit we were
previously using wasn't a numbered release either, so it's
not like we're doing anything dodgier than we were before.)

https://github.com/Tencent/rapidjson/commit/0d4517f15a8d7167ba9ae67f3f22a559ca841e3b
is the commit of RapidJSON that this PR incorporates into the
ml-cpp repo.


dimitris-athanasiou (approved):
LGTM

PR #2118, opened 2021-11-08
[ML] Adding an unbuffered istream wrapper for RapidJSON

droberts195:
In #2106 it became apparent that the behaviour of RapidJSON's
own IStreamWrapper class had changed between the 2017 version
we used previously and the latest 2021 version. In the older
version it was unbuffered and in the newer version it was
buffered.

An attempt was made to work around this in #2106 by specifying
a single character buffer, however, in debug builds this then
fell foul of an assertion that the buffer size was at least 4
characters.

This PR adds a separate unbuffered istream wrapper class that
we can use with RapidJSON in cases where we don't want the
wrapper to consume extra characters beyond the end of the
document that's being parsed. Having a separate class should
also reduce the inefficiency of taking one character at a
time, as there's no need to call a stream method to refill
the buffer on every read.

### DIFF
--- a/bin/pytorch_inference/CCommandParser.cc
+++ b/bin/pytorch_inference/CCommandParser.cc
@@ -12,9 +12,9 @@
 #include "CCommandParser.h"
 
 #include <core/CLogger.h>
+#include <core/CRapidJsonUnbufferedIStreamWrapper.h>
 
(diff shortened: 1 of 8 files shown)


valeriy42 (approved):
LGTM

PR #2119, opened 2021-11-08
[ML] Fix warnings due to -Wsuggest-override compiler flag

edsavage:
When upgrading to macOS `Big Sur` and `Xcode 13.1` a large number of compiler warnings similar to
```
include/maths/common/CNormalMeanPrecConjugate.h:288:25: warning: 'memoryUsage' overrides a member function but is not marked 'override' [-Wsuggest-override]
    virtual std::size_t memoryUsage() const;
```
are generated.

This PR adds the `override` keyword where necessary to subdue such compiler warnings.

### DIFF
--- a/include/api/CAnomalyJobConfigReader.h
+++ b/include/api/CAnomalyJobConfigReader.h
@@ -65,7 +65,7 @@ public:
     public:
         explicit CParseError(const std::string& what)
             : std::runtime_error{what} {}
-        virtual ~CParseError() noexcept override = default;
+        ~CParseError() noexcept override = default;
     };
 
 public:
--- a/include/api/CAnomalyJobConfigReader.h
+++ b/include/api/CAnomalyJobConfigReader.h
@@ -65,7 +65,7 @@ public:
     public:
         explicit CParseError(const std::string& what)
             : std::runtime_error{what} {}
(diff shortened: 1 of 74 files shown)


droberts195 (commented):
I think historically when we've used `override` we've removed `virtual` from the functions it's been applied to.  It would be nice to do this for all these others too, and I don't even think it's very hard.  On a Mac, starting from what's currently in this PR and with a working directory at the top of the ml-cpp repo, the following should do it:

```
echo 'ColumnLimit: 9999' >> .clang-format
dev-tools/clang-format.sh
find . -name '*.cc' -o -name '*.h' | xargs -n 100 sed -i '' 's/virtual \(.*\) override/\1 override/'
git checkout .clang-format
dev-tools/clang-format.sh
```

(The `sed` syntax is slightly different on Linux, but I think the above should work on a Mac.)

droberts195 (approved):
LGTM if CI passes

droberts195:
If this cherry-picks cleanly to 7.16 I'm thinking it would be best to backport to that branch too.  We'll be building that branch for a long time, and it's going to get annoying if the build logs are cluttered with masses of warnings.  But hold off merging any backport PR to 7.16 until next week, just in case there's some weird side effect of the change.

Issue #2120, opened 2021-11-08
[ML] Avoiding C++17 warnings for Windows builds


droberts195:
Our Windows build logs are currently warning of a number of C++ features we are using that are deprecated in C++17 and removed in C++20.  We should try to stop using these before we upgrade to C++20 (although if we cannot then it may be possible to continue to use the deprecated features by defining certain macros).

In our own code we have:

```
C:\Users\jenkins\workspace\elastic+machine-learning+main+windows+debug\include\core/Concurrency.h(139,18): warning C4996: 'std::result_of_t': warning STL4014: std::result_of and std::result_of_t are deprecated in C++17. They are superseded by std::invoke_result and std::invoke_result_t. You can define _SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
C:\Users\jenkins\workspace\elastic+machine-learning+main+windows+debug\include\core/Concurrency.h(141,20): warning C4996: 'std::result_of_t': warning STL4014: std::result_of and std::result_of_t are deprecated in C++17. They are superseded by std::invoke_result and std::invoke_result_t. You can define _SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
CMemoryUsageTest.cc(244,29): warning C4996: 'std::allocator<void>::const_pointer': warning STL4010: Various members of std::allocator are deprecated in C++17. Use std::allocator_traits instead of accessing these members directly. You can define _SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
```

These should be easy to avoid by changing our code to comply with C++20.

In Boost we have:

```
[...]

PR #2123, opened 2021-11-09
[ML] Fix C++17 deprecation warnings

edsavage:
Fixes or workarounds for C++17 deprecation warnings emanating directly
from the ml-cpp codebase (not 3rd-party)

Relates #2120


droberts195 (approved):
LGTM if it passes CI on all platforms

Issue #2072, opened 2021-10-18
[ML] Uninitialised values in `CBoostedTreeTest/testMseIncrementalForTargetDrift`


edsavage:
When running the `CBoostedTreeTest/testMseIncrementalForTargetDrift` test under `valgrind` the following error is reported:

```
2021-10-14 10:20:54,084638 UTC [23487] DEBUG CTestObserver.cc@28 +-----------------------------------------------------+
2021-10-14 10:20:54,153600 UTC [23487] DEBUG CTestObserver.cc@29 |  CBoostedTreeTest/testMseIncrementalForTargetDrift  |
2021-10-14 10:20:54,161977 UTC [23487] DEBUG CTestObserver.cc@30 +-----------------------------------------------------+
==23487== Conditional jump or move depends on uninitialised value(s)
==23487==    at 0x56EC792: equal_pointees<boost::optional<double> > (compare_pointees.hpp:32)
==23487==    by 0x56EC792: operator==<double> (optional_relops.hpp:63)
==23487==    by 0x56EC792: ml::maths::analytics::CBoostedTreeFactory::initializeUnsetRegularizationHyperparameters(ml::core::CDataFrame&) (CBoostedTreeFactory.cc:984)
==23487==    by 0x56EE2EA: ml::maths::analytics::CBoostedTreeFactory::initializeHyperparameters(ml::core::CDataFrame&) (CBoostedTreeFactory.cc:681)
==23487==    by 0x56EF03A: ml::maths::analytics::CBoostedTreeFactory::buildForTrain(ml::core::CDataFrame&, unsigned long) (CBoostedTreeFactory.cc:167)
==23487==    by 0x1D9E45: CBoostedTreeTest::testMseIncrementalForTargetDrift::test_method() (CBoostedTreeTest.cc:839)
==23487==    by 0x1DBEEF: CBoostedTreeTest::testMseIncrementalForTargetDrift_invoker() (CBoostedTreeTest.cc:781)
==23487==    by 0x4172B6D: boost::detail::function::function_obj_invoker0<boost::detail::forward, int>::invoke(boost::detail::function::function_buffer&) (in /usr/local/gcc93/lib/libboost_unit_test_framework-gcc9-mt-x64-1_71.so.1.71.0)
[...]

PR #2125, opened 2021-11-09
[ML] Refactor parameter search initialisation

tveasey:
This moves handling of initialisation of parameter search ranges largely into `CBoostedTreeHyperparameters`. Previously it was split between there and `CBoostedTreeFactory`. It also improves handling of user overrides of parameter search ranges and properly encapsulates the logic for searching logged values or not in the parameter definition.

### DIFF
--- a/include/maths/analytics/CBoostedTreeHyperparameters.h
+++ b/include/maths/analytics/CBoostedTreeHyperparameters.h
@@ -325,7 +325,7 @@ public:
     using THyperparameterImportanceVec =
         std::vector<boosted_tree_detail::SHyperparameterImportance>;
 
-    //! \brief The arguments to initialize the
+    //! \brief The arguments to the initial search we perform for each parameter.
     class MATHS_ANALYTICS_EXPORT CInitializeFineTuneArguments {
     public:
         using TUpdateParameter = std::function<bool(CBoostedTreeImpl&, double)>;
@@ -556,6 +556,13 @@ public:
     //! At the end we use a smooth curve fit through all test loss values (using
     //! LOWESS regression) and use this to get a best estimate of where the true
     //! minimum occurs.
+    //!
+    //! \return A vector comprising
+    //! <pre>
+    //!   | minimum value for fine tune |
+    //!   |       initial value         |
+    //!   | maximum value for fine tune |
+    //! </pre>
     TOptionalVector3x1 testLossLineSearch(const CInitializeFineTuneArguments& args,
                                           double intervalLeftEnd,
                                           double intervalRightEnd) const;
--- a/include/maths/analytics/CBoostedTreeHyperparameters.h
+++ b/include/maths/analytics/CBoostedTreeHyperparameters.h
@@ -66,7 +66,7 @@ public:
     }
 
     //! Get the value.
-    T value() const { return m_Value; }
+    T value() const { return m_Scale * m_Value; }
 
     //! Set to \p value.
     //!
@@ -79,13 +79,25 @@ public:
         return *this;
     }
 
-    //! Multiply by \p scale.
+    //! Get the multiplier which is applied when the parameter is read.
+    T scale() const { return m_Scale; }
+
+    //! Set the multiplier which is applied when the parameter is read to \p scale.
+    CBoostedTreeParameter& scale(T scale) {
+        m_Scale = scale;
+        return *this;
+    }
+
+    //! Convert the scale to a multiplier of the parameter.
     //!
-    //! \note Has no effect if the parameter is fixed.
-    CBoostedTreeParameter& forceScale(T scale) {
-        m_Value *= scale;
-        m_MinValue *= scale;
-        m_MaxValue *= scale;
+    //! \warning Handle with care since this also applies the scaling to the range
+    //! limits. This is consistent with the behaviour of scaling which can move
+    //! values outside this interval.
+    CBoostedTreeParameter& captureScale() {
+        m_Value *= m_Scale;
+        m_MinValue *= m_Scale;
+        m_MaxValue *= m_Scale;
+        m_Scale = T{1};
         return *this;
     }
 
@@ -102,6 +114,7 @@ public:
         m_Value = value;
         m_MinValue = value;
         m_MaxValue = value;
+        m_Scale = T{1};
         m_FixedToRange = true;
     }
 
@@ -112,6 +125,7 @@ public:
         m_Value = common::CTools::truncate(m_Value, minValue, maxValue);
         m_MinValue = minValue;
         m_MaxValue = maxValue;
+        m_Scale = T{1};
         m_FixedToRange = true;
     }
 
@@ -144,6 +158,9 @@ public:
         return m_LogSearch == false || value > 0.0;
     }
 
+    //! Get the unscaled value converted to a search value.
+    double toSearchValue() const { return this->toSearchValue(m_Value); }
+
     //! Convert \p value to the value used by BO for fine tuning.
     double toSearchValue(T value) const {
         return m_LogSearch ? common::CTools::stableLog(static_cast<double>(value))
@@ -165,6 +182,7 @@ public:
         m_SavedValue = m_Value;
         m_SavedMinValue = m_MinValue;
         m_SavedMaxValue = m_MaxValue;
+        m_SavedScale = m_Scale;
     }
 
     //! Load the saved value.
@@ -172,6 +190,7 @@ public:
         m_Value = m_SavedValue;
         m_MinValue = m_SavedMinValue;
         m_MaxValue = m_SavedMaxValue;
+        m_Scale = m_SavedScale;
         return *this;
     }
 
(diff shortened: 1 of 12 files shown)


lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -1135,138 +912,151 @@ void CBoostedTreeFactory::initializeUnsetEta(core::CDataFrame& frame) {
...
 
 void CBoostedTreeFactory::initializeUnsetTreeTopologyPenalty(core::CDataFrame& frame) {
 
-    auto& hyperparameters = m_TreeImpl->m_Hyperparameters;
+    if (m_TreeImpl->m_Hyperparameters.incrementalTraining() == false) {
+        return;
+    }
 
-    if (hyperparameters.treeTopologyChangePenalty().rangeFixed() == false) {
+    skipIfAfter(CBoostedTreeImpl::E_NotInitialized, [&] {
```
valeriy42, line 993: Can we maybe get away without using these large unnamed lambda functions? I think it impairs the readability.

lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -94,14 +107,214 @@ void CBoostedTreeHyperparameters::resetSearch() {
...
+    //   2. Fine tune, via Bayesian Optimisation targeting expected improvement,
+    //      and stop if the expected improvement small compared to the current
+    //      minimum test loss,
+    //   3. Calculate the parameter interval which gives the lowest test losses,
+    //   4. Fit an OLS quadratic approximation to the test losses in the interval
+    //      from step 3 and use it to estimate the best parameter value,
+    //   5. Compare the size of the residual errors w.r.t. to the OLS curve from
+    //      step 4 with its variation over the interval from step 3 and truncate
+    //      the returned interval if we can determine there is a low chance of
+    //      missing the best solution by doing so.
```
valeriy42, line 155: I think, it would improve readability if you either split this long function into individual steps or relate the individual steps in the comment to the places in the code where it happens.

valeriy42 (commented):
Good work on refactoring the code :rocket:  I left some comments to improve readability.

lib/maths/analytics/CBoostedTreeFactory.cc
tveasey: I don't agree here. It is essentially no extra code other than just putting this inside a code block `{}`. This is the mechanism by which we ensure we reinitialise properly if we failover halfway through building and I think the mechanism is effective. Since it is really unrelated I think I will leave as is. Happy to discuss alternative approaches offline.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -747,202 +653,105 @@ void CBoostedTreeFactory::initializeUnsetRegularizationHyperparameters(core::CDa
...
-                }
+                hyperparameters.initializeFineTuneSearchInterval(
+                    CBoostedTreeHyperparameters::CInitializeFineTuneArguments{
+                        frame, *m_TreeImpl, m_GainPerNode90thPercentile,
+                        2.0 * m_GainPerNode90thPercentile / m_GainPerNode1stPercentile,
+                        [](CBoostedTreeImpl& tree, double logDepthPenalty) {
+                            auto& parameter = tree.m_Hyperparameters.depthPenaltyMultiplier();
+                            parameter.set(parameter.fromSearchValue(logDepthPenalty));
+                            return true;
+                        }},
```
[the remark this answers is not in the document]
tveasey: Ok, I though about this. The one downside is that none of these parameters have a sensible default and the only way of enforcing they are supplied is via a constructor. Let me see if there might be something better we can do here.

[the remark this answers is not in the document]
tveasey: I think on balance I'd prefer to leave as is. The other option would be to have a `validate` method which is called inside `initializeFineTuneSearchInterval`, but I prefer the fact that the correct behaviour is enforced at compile time.

include/maths/analytics/CBoostedTreeHyperparameters.h
```diff
@@ -262,13 +293,70 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeHyperparameters {
...
-    using TAddInitialRangeFunc =
-        std::function<void(boosted_tree_detail::EHyperparameter, TDoubleDoublePrVec&)>;
+    using TVector3x1 = common::CVectorNx1<double, 3>;
+    using TOptionalVector3x1 = boost::optional<TVector3x1>;
     using TMeanAccumulator = common::CBasicStatistics::SSampleMean<double>::TAccumulator;
     using TMeanVarAccumulator = common::CBasicStatistics::SSampleMeanVar<double>::TAccumulator;
     using THyperparameterImportanceVec =
         std::vector<boosted_tree_detail::SHyperparameterImportance>;
 
+    //! \brief The arguments to initialize the
```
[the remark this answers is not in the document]
tveasey: See [b7ec94a](https://github.com/elastic/ml-cpp/pull/2125/commits/b7ec94a0cd76dba12f2412a66e61c6382414bfe4).

include/maths/analytics/CBoostedTreeHyperparameters.h
```diff
@@ -426,14 +514,29 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeHyperparameters {
...
     void resetSearch();
 
+    //! Compute the fine tune search interval for \p parameter.
+    void initializeFineTuneSearchInterval(const CInitializeFineTuneArguments& args,
+                                          TDoubleParameter& parameter) const;
+
+    //! Perform a line search for the test loss w.r.t. a single hyperparameter.
+    //! At the end we use a smooth curve fit through all test loss values (using
+    //! LOWESS regression) and use this to get a best estimate of where the true
+    //! minimum occurs.
```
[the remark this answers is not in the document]
tveasey: See [b7ec94a](https://github.com/elastic/ml-cpp/pull/2125/commits/b7ec94a0cd76dba12f2412a66e61c6382414bfe4).

lib/maths/analytics/CBoostedTreeHyperparameters.cc
tveasey: See [e5858c7](https://github.com/elastic/ml-cpp/pull/2125/commits/e5858c70df6ff8999fed13f67e7901d6665049f2).

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -601,6 +625,7 @@ void CBoostedTreeFactory::initializeHyperparametersSetup(core::CDataFrame& frame
     // This needs to be tied to the learn rate to avoid bias.
     hyperparameters.maximumNumberTrees().setToRangeMidpointOr(
         computeMaximumNumberTrees(hyperparameters.eta().value()));
+    hyperparameters.retrainedTreeEta().setToRangeMidpointOr(1.0);
```
valeriy42, line 628: I would set it to `0.5` instead of `1.0` as a fallback value.

valeriy42 (approved):
LGTM.

lib/maths/analytics/CBoostedTreeFactory.cc
tveasey: The only thing is that this interacts with whether or not you have `predictionChangeCost` non-zero. I was tending to find 1 was good in conjunction with this non-zero. In the multistep train from scratch I agree all your results suggests 1 is a bad choice. How about I set it to 1 if `predictionChangeCost` > 0 and `eta` otherwise?

include/maths/analytics/CBoostedTreeUtils.h
```diff
@@ -63,14 +63,15 @@ enum EHyperparameter {
...
     E_FeatureBagFraction,
-    E_PredictionChangeCost,
     // Incremental train parameters.
+    E_PredictionChangeCost,
     E_RetrainedTreeEta,
     E_TreeTopologyChangePenalty
 };
 
-constexpr std::size_t NUMBER_HYPERPARAMETERS = E_TreeTopologyChangePenalty + 1; // This must be last hyperparameter
+constexpr std::size_t NUMBER_HYPERPARAMETERS(E_TreeTopologyChangePenalty + 1); // This must be last hyperparameter
```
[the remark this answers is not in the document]
tveasey: Yes. I was thinking this needs to cast, but it compiles with your suggestion.

lib/maths/analytics/CBoostedTreeFactory.cc
tveasey: [32a64c3](https://github.com/elastic/ml-cpp/pull/2125/commits/32a64c37b81ec05b0fc5cba2befaff9ca7026849)

include/maths/analytics/CBoostedTreeUtils.h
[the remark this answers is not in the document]
tveasey: [63a2e8b](https://github.com/elastic/ml-cpp/pull/2125/commits/63a2e8b1470b8098c4df516100c5c699260411f0)

tveasey:
The platform specific failures are likely related to #2072, which I plan to investigate shortly. We need this for other testing, so I'm going to go ahead and merge now.

PR #2127, opened 2021-11-10
[ML] Re-run seccomp unit tests on more Linux distributions

droberts195:
Our system call filtering is particularly sensitive to the
exact glibc and kernel versions being used. At present our
CI builds only run the C++ unit tests inside the Docker
container. This means the seccomp unit tests will only run
in CI on glibc 2.17 (with our current CentOS 7 build Docker
images). We can improve coverage by copying the test program
out of the Docker container and running it again on the host
operating system. Then it will run against the glibc version
on the host operating system, which is likely to be much
newer than the version in the Docker container. (The Docker
container deliberately uses the oldest glibc version that we
need to be able to run on.)

Potentially we could re-run more unit tests in the future
using the same approach, but some would dramatically increase
CI runtimes. For example, the maths library unit tests take
16 minutes to run. This PR is starting small by just re-running
the seccomp library tests, which are very fast. We can iterate
in the future.

### DIFF
--- a/3rd_party/3rd_party.sh
+++ b/3rd_party/3rd_party.sh
@@ -48,7 +48,7 @@ case `uname` in
             BOOST_ARCH=a64
         fi
         BOOST_EXTENSION=mt-${BOOST_ARCH}-1_77.dylib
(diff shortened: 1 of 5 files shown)


dimitris-athanasiou (approved):
LGTM

PR #2128, opened 2021-11-10
[ML] Fix a bug in hyperparameter scaling for classification and regression model training

tveasey:
When we adjust the downsample factor we scale the parameters which are added to terms which are proportional to the amount of data so that the degree of regularisation remains constant in expectation. We do this correctly when we fine tune, but using the wrong scaling when doing the initial search for downsample factor. This is liable to lead to us choosing smaller downsample factors than we otherwise might.


dimitris-athanasiou (approved):
LGTM

PR #2135, opened 2021-11-22
[ML] Use unbufferred RapidJSON istream wrapper everywhere in the code

valeriy42:
After #2121, there were still some places in the code where the old rapidjson istream wrapper was used. This led to errors, and this PR fixes this.

PR #2136, opened 2021-11-22
[ML] Replace rapidjson::istreamwrapper in CJsonStateRestoreTraverser

droberts195:
One of the changes of #2135 applies to the main branch as well
as the incremental learning branch. Although it doesn't cause
any test failures it should be replaced before 8.1 feature
freeze just in case there's a place where buffering makes a
difference in a scenario not covered by our testing.


dimitris-athanasiou (approved):
LGTM

PR #2138, opened 2021-11-24
[ML] Removing null pointer output workaround for Boost.Test

droberts195:
Removing Boost.Test output function for nullptr now
this is provided by a standard operator<< overload.


edsavage (approved):
LGTM

PR #2139, opened 2021-11-25
[ML] Set model state compatibility version to 8.0.0

droberts195:
Setting this for the new major version gives us a nice
clean break for model state compatibility.  Model snapshots
generated by 8.0 will not be loadable by 7.x nodes, and
in the version where we eventually stop supporting upgrades
from 7.x to the latest version we'll be able to remove all
the model state backwards compatibility code related to
version 7.x.

Proactively avoiding a repeat of elastic/elasticsearch#81011

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -45,7 +45,7 @@
 
 === Bug Fixes
 
-* ...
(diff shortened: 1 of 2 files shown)


dimitris-athanasiou (approved):
LGTM

PR #2141, opened 2021-11-29
[PYTORCH] Limit total number of model & inference threads <= max threads

davidkyle:
The validation of the `inferenceThreads` and `modelThreads` parameters should not allow the sum to exceed `maxThreads` (std::thread::hardware_concurrency()).

The validation logic is a little complicated as when `modelThreads ==1` no thread pool is setup which effectively means `modelThreads == 0` and inference can use the rest of the threads. I pulled the validation method out into a new file so it can be unit tested.

### DIFF
--- a/bin/pytorch_inference/SettingsValidator.cc
+++ b/bin/pytorch_inference/SettingsValidator.cc
@@ -1,0 +1,62 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the following additional limitation. Functionality enabled by the
+ * files subject to the Elastic License 2.0 may only be used in production when
+ * invoked by an Elasticsearch process with a license key installed that permits
+ * use of machine learning features. You may not use this file except in
+ * compliance with the Elastic License 2.0 and the foregoing additional
+ * limitation.
+ */
+
+#include "SettingsValidator.h"
+
+#include <core/CLogger.h>
+
+#include <algorithm>
+#include <thread>
+
+namespace ml {
+namespace torch {
+
+void validateThreadingParameters(std::int32_t maxThreads,
+                                 std::int32_t& inferenceThreads,
+                                 std::int32_t& modelThreads) {
+    if (maxThreads == 0) {
+        LOG_WARN(<< "Could not determine hardware concurrency; setting max threads to 1");
+        maxThreads = 1;
+    }
+    if (inferenceThreads < 1) {
+        LOG_WARN(<< "Setting inference threads to minimum value of 1; value was "
+                 << inferenceThreads);
+        inferenceThreads = 1;
+    } else if (inferenceThreads > maxThreads) {
+        LOG_WARN(<< "Setting inference threads to maximum value of "
+                 << maxThreads << "; value was " << inferenceThreads);
+        inferenceThreads = maxThreads;
+    }
+    if (modelThreads < 1) {
+        LOG_WARN(<< "Setting model threads to minimum value of 1; value was " << modelThreads);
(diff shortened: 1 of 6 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@
 * Improve skip_model_update rule behaviour (See {ml-pull}2096[#2096].)
 * Upgrade Boost libraries to version 1.77. (See {ml-pull}2095[#2095].)
 * Upgrade RapidJSON to 31st October 2021 version. (See {ml-pull}2106[#2106].)
+* Prevent over-subscription of threads in pytorch_inference (See {ml-pull}2106[#2141].)
```
droberts195, line 38:
```suggestion
* Prevent over-subscription of threads in pytorch_inference. (See {ml-pull}2141[#2141].)
```

dimitris-athanasiou (approved):
LGTM

bin/pytorch_inference/SettingsValidator.cc
```diff
@@ -0,0 +1,59 @@
...
+        modelThreads = std::max(1, maxThreads - 1);
+    }
+
+    if (modelThreads + inferenceThreads > maxThreads) {
+        std::int32_t oldInferenceThreadCount{inferenceThreads};
+        inferenceThreads = std::max(1, maxThreads - modelThreads);
+        LOG_WARN(<< "Sum of model threads [" << modelThreads << "] and inference threads ["
+                 << oldInferenceThreadCount << "] is greater than max threads ["
+                 << maxThreads << "]. Setting model threads to " << modelThreads
+                 << " and inference threads to " << inferenceThreads);
```
droberts195, line 55: Nobody looks in the logs unless there's a fatal problem, so in reality what will happen here is that the user configures, say, 1000 threads of each type and receives no feedback.  Silently we'll decide for them to favour model threads over inference threads.  This is not really in the spirit of Elasticsearch's "no leniency" policy.

So I think it's OK to merge this PR as-is as a first step, but it would be good to also add functionality to validate the thread settings in some way on the Java side too, or at least report how many actually got used for each process.

davidkyle: I agree it is sneaky to change the value here, we are logging the change but that information is not clearly manifested. That said, it is desirable to prevent a value of 1000 threads.

The thread settings apply to the entire deployment which is spread over heterogeneous compute nodes, a setting that works for 1 node may not work well for another. I will look to see if there is an equivalent of `std:thread:hardware_concurrency()` in Java and do the validation in the server and somehow report that in the deployment stats.

droberts195: Elasticsearch has the number of processors it's allowed to use in the node stats - search for `os.allocated_processors` in https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html.  That could be used for some sort of sanity check on the magnitude of the configured numbers.  However, just thinking about it we _don't_ want the current number of allocated processors to be a hard limit on the configuration.  Maybe in the future what we want is that you configure how many threads you want across all nodes and the cluster autoscales to a size where it can supply that number of threads.

Maybe it's best to document that the configured numbers are maximums and have each C++ process pass back the number it actually decided to use in the first message it sends back to the Java side.  Then we could report actual threads used in our stats instead of just configured threads.

But all this is a lot more work.

davidkyle: https://github.com/elastic/elasticsearch/issues/81149

davidkyle: > document that the configured numbers are maximums

👍 I added this to the backlog

davidkyle:
run elasticsearch-ci

davidkyle:
retest

PR #2144, opened 2021-11-30
[ML] Improve training stability for regression and classification models

tveasey:
We need to check that the objective actually decreases after snapping to the constraint bounding box for L-BFGS with constraint. Also, addresses some cross platform variation due to variation `std::exp` and  `std::log` when computing ANOVA for hyperparameter losses.


valeriy42 (approved):
LGTM. Good work catching this! :rocket:

PR #2145, opened 2021-12-01
[ML] Experiment multistep learning with hyperparameters ranges

valeriy42:
This PR adds the code for running multistep learning experiments with the search of hyperparameters within specific ranges.

Note: the Jupyter notebooks in the `prototypes/` directory are preliminary and don't need to be reviewed.

The experiment driver makes use of the `comment` parameter of sacred instead of `tags` and makes its use obligatory. Also, I added `submit` config to control whether or not to submit results to the index.


jupyter/src/incremental_learning/job.py
```diff
@@ -577,8 +577,27 @@ def update(dataset_name: str,
         config['analysis']['parameters']['force_accept_incremental_training'] = True
 
     for name, value in original_job.get_hyperparameters().items():
-        if name not in ['retrained_tree_eta', 'tree_topology_change_penalty']:
+        if name not in ['retrained_tree_eta', 'tree_topology_change_penalty', 'soft_tree_depth_limit', 'gamma', 'alpha'] \
+            and name not in config['analysis']['parameters'].keys():
             config['analysis']['parameters'][name] = value
+        if name in ['gamma', 'alpha']:
```
tveasey, line 583: I'm not sure this apply to single step case where you have a small summary of the original data set and update with some new data. I think it would be good to factor out this logic into something like `multistep_overrides` and have some parameter to opt in to this behaviour.

jupyter/scripts/multi_step_learning/parameter_ranges/experiment_driver.py
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env python
```
tveasey, line 1: As before I wonder if there is common functionality with `incremental_training/experiment_driver.py`. Would be nice to factor into `src` if we do have some.

tveasey (commented):
Just some minor suggestions. Overall looks good. The one thing is I would like to keep hyperparameter overrides optional and tied multistep training since I think the right behaviour will end up being different.

jupyter/scripts/multi_step_learning/parameter_ranges/job_submitter.py
```diff
@@ -0,0 +1,87 @@
...
+import argparse
+import json
+import multiprocessing
+import os
+import subprocess
+
+from jinja2 import Environment, FileSystemLoader
+from pathlib import Path
+
+def init_task_spooler():
```
tveasey, line 21: This comment seems to have got lost: Maybe factor out common code with `incremental_training/job_submitter`.

valeriy42: Moved to `incremental_training/job_submitter.py`.

jupyter/src/incremental_learning/job.py
valeriy42: Indeed, I was also thinking along those lines. Added `hyperparameter_overrides` as a optional parameter.

jupyter/scripts/multi_step_learning/parameter_ranges/experiment_driver.py
valeriy42: Thank you for the review @tveasey . I factored out the common functionality and introduced `hyperparameter_overrides` in the `jop.py/udpate()` function to keep it general.

tveasey (approved):
Good stuff! LGTM

PR #2147, opened 2021-12-01
[ML] Some improvements to Bayesian Optimisation implementation

tveasey:
These were found debugging sources of cross-platform variation.
1. `1 + erf` underflows when `erf` is within double epsilon of 1. `boost::normal` distribution has a better cdf implementation which computes values accurate to double precision as small as minimum (positive) double. This means we get better estimates of EI and its gradient.
2. Depending on the kernel parameters EI can be very small. In the case when it underflows we broke ties by dimension lexicograpical order for the points. From an optimisation standpoint this is a silly choice. I've changed it to prefer the we tried point which is most dissimilar to points we've already tried (i.e. pure exploration).

Both of these incidentally mean the exact choice of points BO will propose is more sensitive to numerics which can vary slightly between platforms, because, for example, there are differences in standard library implementations.

### DIFF
--- a/lib/maths/common/CBayesianOptimisation.cc
+++ b/lib/maths/common/CBayesianOptimisation.cc
@@ -58,11 +58,6 @@ const std::string RNG_TAG{"rng"};
 // narrow deep valley that the Gaussian Process hasn't sampled.
 const double MINIMUM_KERNEL_SCALE_FOR_EXPECTATION_MAXIMISATION{1e-8};
 
-// For values less than this the EI is essentially zero and we should fallback to
-// dissimilarity. Note that the supplied function values are scaled so that their
-// variance is one and so this is a relative constant.
-double NEGLIGIBLE_EI{1e-12};
-
 //! A version of the normal c.d.f. which is stable across our target platforms.
 double stableNormCdf(double z) {
     return CTools::stable(CTools::safeCdf(boost::math::normal{0.0, 1.0}, z));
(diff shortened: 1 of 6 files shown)


lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -710,6 +702,28 @@ double CBayesianOptimisation::kernel(const TVector& a, const TVector& x, const T
...
+    return x - m_MinBoundary.cwiseQuotient(m_MaxBoundary - m_MinBoundary);
+}
+
+double CBayesianOptimisation::dissimilarity(const TVector& x) const {
+    double sum{0.0};
+    double min{0.0};
+    for (const auto& y : m_FunctionMeanValues) {
+        double dxy{las::distance(x, y.first)};
+        sum += dxy;
+        min += std::min(min, -dxy);
```
valeriy42, line 722: Can you maybe leave a comment on why do you need this `min` term?

valeriy42 (approved):
Good work on catching this! I am wondering how this switch to dissimilarity exploration co-works with early stopping. From what I understand, if the surface of the GP is flat, then early stopping will kick in and prevent the exploration process, won't it?

lib/maths/common/CBayesianOptimisation.cc
tveasey: I added a comment in [2f20ec1](https://github.com/elastic/ml-cpp/pull/2147/commits/2f20ec1ed7e62fe55350232e4874f20b0ee2c476).

tveasey:
> From what I understand, if the surface of the GP is flat, then early stopping will kick in and prevent the exploration process, won't it?

There are other places than fine tune where we use BO: for example in the per parameter line searches. However, interestingly EI can be very small even if the surface is not flat. For example, if the predicted variance is very small then it can also wind up very small and certainly anywhere away from the immediate vicinity of the minimum loss parameter choice we've seen to date.

PR #2148, opened 2021-12-02
[ML] Test hyperparameter change for incremental learning

### LINKED ISSUE #2146: [ML] Reproducibility of incremental training results
After the update step, we get different results depending on whether we are searching for hyperparameters or explicitly specifying the found values in the configuration. This difference makes it impossible to analyze the impact of hyperparameters on the system in the multistep learning regime. 

Hence, we need to introduce comprehensive testing of this behavior and find the root cause of the difference.

valeriy42:
This PR introduces integration tests for incremental learning to verify the following behavior:
- We obtain the same results whether we are searching for an optimal hyperparameter value or we are specifying this value explicitly
- If a user specifies a search range for a hyperparameter, the identified optimal hyperparameter is indeed within this range. 
- We can achieve monotonically increasing or decreasing hyperparameter values by specifying the feasible search interval accordingly

I also integrated the changes from #2153 to disable hyperparameter scaling using a config parameter since this change simplified the test.

Closes #2146


valeriy42:
pull

tveasey (approved):
Test LGTM. I know you grabbed the hyperparameter disabling in advance, but happy for you to commit once this is merged into your branch.

PR #2137, opened 2021-11-24
[ML] Upgrade to Eigen 3.4.0

edsavage:
Eigen 3.4.0 has better support for more modern C++ language features and
as such quietens a few C++17 deprecation warnings encountered with
version 3.3.7.

Relates to #2120

PR #2149, opened 2021-12-02
[ML] Fix deprecation warning for macOS build

edsavage:
Deprecation warnings were enabled for a build on macOS 11.6.1 using clang
13.0.0 by way of adding the `-Wdeprecated` flag to clang's arguments.
The majority of the warnings related to implicit constructors of one
sort or another, with a notable exception being `OSSpinLock` being
deprecated in favour of `os_unfair_lock` since macOS 10.12.

labelling as `>non-issue` as no functional change has been made.

### DIFF
--- a/include/api/CHierarchicalResultsWriter.h
+++ b/include/api/CHierarchicalResultsWriter.h
@@ -153,7 +153,7 @@ public:
                                const TResultWriterFunc& resultWriter,
                                const TPivotWriterFunc& pivotsWriterFunc);
 
-    virtual ~CHierarchicalResultsWriter() override = default;
+   ~CHierarchicalResultsWriter() override = default;
 
     //! Write \p node.
     void visit(const model::CHierarchicalResults& results, const TNode& node, bool pivot) override;
--- a/include/api/CHierarchicalResultsWriter.h
+++ b/include/api/CHierarchicalResultsWriter.h
@@ -153,7 +153,7 @@ public:
                                const TResultWriterFunc& resultWriter,
                                const TPivotWriterFunc& pivotsWriterFunc);
 
-   ~CHierarchicalResultsWriter() override = default;
+    ~CHierarchicalResultsWriter() override = default;
 
     //! Write \p node.
     void visit(const model::CHierarchicalResults& results, const TNode& node, bool pivot) override;
--- a/include/api/CHierarchicalResultsWriter.h
+++ b/include/api/CHierarchicalResultsWriter.h
@@ -153,6 +153,8 @@ public:
                                const TResultWriterFunc& resultWriter,
                                const TPivotWriterFunc& pivotsWriterFunc);
 
+    virtual ~CHierarchicalResultsWriter() override = default;
+
     //! Write \p node.
     void visit(const model::CHierarchicalResults& results, const TNode& node, bool pivot) override;
 
--- a/include/core/CFastMutex.h
+++ b/include/core/CFastMutex.h
@@ -17,7 +17,7 @@
 
 #ifndef Windows
 #ifdef MacOSX
-#include <libkern/OSAtomic.h>
+#include <os/lock.h>
 #else
 #include <pthread.h>
 #endif
@@ -76,7 +76,7 @@ private:
 #ifdef Windows
     SRWLOCK m_Mutex;
 #elif defined(MacOSX)
-    OSSpinLock m_Mutex;
+    os_unfair_lock m_Mutex;
 #else
     pthread_mutex_t m_Mutex;
 #endif
--- a/include/core/CFastMutex.h
+++ b/include/core/CFastMutex.h
@@ -45,16 +45,10 @@ namespace core {
 //!
 //! All errors are just warnings - no action taken.
 //!
-//! On Mac OS X, spin locks are also used, which might seem strange
-//! given that Mac OS X will have relatively few available CPU cores
-//! (say 2-4).  However, Mac OS X spinlocks are actually quite clever
-//! about yielding if they don't quickly get a lock, and are hence
-//! more akin to adaptive mutexes than pure spinlocks.  The source
-//! code at confirms this:
-//! http://www.opensource.apple.com/source/Libc/Libc-825.25/x86_64/sys/spinlocks_asm.s
-//! The other reason is that pthread mutexes are appallingly slow on
-//! Mac OS X so it's easy to massively outperform them - e.g. see:
-//! http://www.mr-edd.co.uk/blog/sad_state_of_osx_pthread_mutex_t
+//! On Mac OS X, spin locks used to be used, but OSSpinLock has now
+//! been tagged as deprecated as of macOS 10.12. Instead, Apple have
+//! provided os_unfair_lock as a replacement (see the documentation here
+//! https://developer.apple.com/documentation/os/1646466-os_unfair_lock_lock)
 //!
 //! On Linux, standard non-recursive mutexes are used, on the
 //! assumption that Linux may have relatively few available CPU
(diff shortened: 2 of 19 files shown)


include/api/CHierarchicalResultsWriter.h
```diff
@@ -153,6 +153,8 @@ class API_EXPORT CHierarchicalResultsWriter : public model::CHierarchicalResults
                                const TResultWriterFunc& resultWriter,
                                const TPivotWriterFunc& pivotsWriterFunc);
 
+    virtual ~CHierarchicalResultsWriter() override = default;
```
droberts195, line 156: For all other overridden virtual functions we're not specifying virtual as well, so:

```suggestion
   ~CHierarchicalResultsWriter() override = default;
```

include/core/CFastMutex.h
```diff
@@ -76,7 +76,7 @@ class CORE_EXPORT CFastMutex : private CNonCopyable {
 #ifdef Windows
     SRWLOCK m_Mutex;
 #elif defined(MacOSX)
-    OSSpinLock m_Mutex;
+    os_unfair_lock m_Mutex;
```
droberts195, line 79: How does the performance compare?  The [docs](https://developer.apple.com/documentation/os/1646466-os_unfair_lock_lock) say:

> This is a replacement for the deprecated OSSpinLock. This function doesn't spin on contention, but instead waits in the kernel to be awoken by an unlock.

To me this sounds like there's a chance of no benefit over `pthread_mutex_t`, in which case the implementation could be shared with Linux.

There are tests in `CReadWriteLockTest` that test the performance of all the different types of lock, not just read/write.  If you dig out some old CI logs and look at the timestamps on `Starting fast mutex lock throughput test at` and `Finished fast mutex lock throughput test at` for a macOS build then you should be able to compare the performance of the old and new types.  You can also do this on your laptop.

include/model/CHierarchicalResultsAggregator.h
```diff
@@ -52,6 +52,8 @@ class MODEL_EXPORT CHierarchicalResultsAggregator
 public:
     CHierarchicalResultsAggregator(const CAnomalyDetectorModelConfig& modelConfig);
 
+    virtual ~CHierarchicalResultsAggregator() override = default;
```
droberts195, line 55:
```suggestion
    ~CHierarchicalResultsAggregator() override = default;
```

include/model/CHierarchicalResultsLevelSet.h
```diff
@@ -59,6 +59,8 @@ class CHierarchicalResultsLevelSet : public CHierarchicalResultsVisitor {
     explicit CHierarchicalResultsLevelSet(const T& bucketElement)
         : m_BucketElement(bucketElement) {}
 
+    virtual ~CHierarchicalResultsLevelSet() override = default;
```
droberts195, line 62:
```suggestion
   ~CHierarchicalResultsLevelSet() override = default;
```

include/model/CHierarchicalResultsNormalizer.h
```diff
@@ -108,6 +108,8 @@ class MODEL_EXPORT CHierarchicalResultsNormalizer
 public:
     CHierarchicalResultsNormalizer(const CAnomalyDetectorModelConfig& modelConfig);
 
+    virtual ~CHierarchicalResultsNormalizer() override = default;
```
droberts195, line 111:
```suggestion
    ~CHierarchicalResultsNormalizer() override = default;
```

include/model/CHierarchicalResultsPopulator.h
```diff
@@ -28,6 +28,8 @@ class MODEL_EXPORT CHierarchicalResultsPopulator : public CHierarchicalResultsVi
     //! Constructor
     CHierarchicalResultsPopulator(const CLimits& limits);
 
+    virtual ~CHierarchicalResultsPopulator() override = default;
```
droberts195, line 31:
```suggestion
    ~CHierarchicalResultsPopulator() override = default;
```

include/model/CHierarchicalResultsProbabilityFinalizer.h
```diff
@@ -31,6 +31,8 @@ namespace model {
 //! breadth first pass over the results.
 class MODEL_EXPORT CHierarchicalResultsProbabilityFinalizer : public CHierarchicalResultsVisitor {
 public:
+    virtual ~CHierarchicalResultsProbabilityFinalizer() override = default;
```
droberts195, line 34:
```suggestion
    ~CHierarchicalResultsProbabilityFinalizer() override = default;
```

lib/model/CHierarchicalResults.cc
```diff
@@ -167,6 +167,8 @@ void aggregateLayer(ITR beginLayer,
 //! that it is either the person or partition field of that node.
 class CCommonInfluencePropagator : public CHierarchicalResultsVisitor {
 public:
+    virtual ~CCommonInfluencePropagator() override = default;
```
droberts195, line 170:
```suggestion
    ~CCommonInfluencePropagator() override = default;
```

include/core/CFastMutex.h
edsavage: With respect to performance. I tested locally on my laptop (MacBook Pro 2017, 2.9GHz Quad-Core Intel Core i7) and increased the number of iterations from 1000000 to 100000000. The test `CReadWriteLockTest/testPerformanceVersusMutex` was then run in a loop 10 times and the reported run times averaged.

Timings for `CFastMutex`:
* using `OSSpinLock` - average time taken = 2.9s
* using `os_unfair_lock` - average time taken = 9.1s
* using `pthread_mutex_t` (same as Linux)- average time take = 20.6s

So while the move to `os_unfair_lock` is appreciably slower than the deprecated `OSSpinLock` it still out strips `pthread_mutex_t`.

droberts195: I also had a look at what happened on macOS on ARM.

Before (from the PR CI build of #2137):

```
2021-12-01 09:26:10,389468 UTC [94129] INFO  CReadWriteLockTest.cc@357 Starting fast mutex lock throughput test at 01:26:10
2021-12-01 09:26:10,462001 UTC [94129] INFO  CReadWriteLockTest.cc@373 Finished fast mutex lock throughput test at 01:26:10
```

462001 - 389468 = 72533

After (from this PR's CI):

```
2021-12-03 10:13:57,598510 UTC [27626] INFO  CReadWriteLockTest.cc@357 Starting fast mutex lock throughput test at 02:13:57
2021-12-03 10:13:57,684028 UTC [27626] INFO  CReadWriteLockTest.cc@373 Finished fast mutex lock throughput test at 02:13:57
```

684028 - 598510 = 85518

It's ideal because there are too few iterations to dominate OS process scheduling effects, but those numbers are very close, so maybe the new mutex is more efficient on ARM.  Since ARM is the future for macOS we might as well get rid of the deprecated function usage.

droberts195 (approved):
LGTM

include/core/CFastMutex.h
edsavage: Agreed. 

I've just noticed that I've omitted to update the comments in `CFastMutex.h`. I'll do so now.

include/core/CFastMutex.h
```diff
@@ -45,16 +45,10 @@ namespace core {
...
-//! On Mac OS X, spin locks are also used, which might seem strange
-//! given that Mac OS X will have relatively few available CPU cores
-//! (say 2-4).  However, Mac OS X spinlocks are actually quite clever
-//! about yielding if they don't quickly get a lock, and are hence
-//! more akin to adaptive mutexes than pure spinlocks.  The source
-//! code at confirms this:
-//! http://www.opensource.apple.com/source/Libc/Libc-825.25/x86_64/sys/spinlocks_asm.s
-//! The other reason is that pthread mutexes are appallingly slow on
-//! Mac OS X so it's easy to massively outperform them - e.g. see:
-//! http://www.mr-edd.co.uk/blog/sad_state_of_osx_pthread_mutex_t
```
droberts195, line 57: I think it's worth keeping this bit about why we don't just use pthread_mutex_t on macOS

edsavage: Sure, sadly that blog post no longer exists though :-/

PR #2150, opened 2021-12-02
[ML] Improve regression and classification training stability

tveasey:
Currently, if we make any different choice in any round of training it affects the random number generator state which will affect the numbers we choose for all subsequent rounds. This tends to amplify differences in results especially those related to say slightly different numerics on different platforms.

With this change I reset the rng and then jump (by 2^64 values) after each forest we train. This means different choices for a particular round don't affect random numbers drawn in subsequent rounds. The jump is large enough that sequences don't overlap.


valeriy42 (approved):
LGTM. Good idea with the jumps!

PR #2152, opened 2021-12-06
[ML] Silence warnings stemming from PyTorch headers

edsavage:
On Windows a number of deprecation warnings are seen emanating from
PyTorch header files (See #2120). An issue has been opened on the
PyTorch repo requesting that these deprecations be addressed
(https://github.com/pytorch/pytorch/issues/69290). While the issue has
been triaged it is suspected that it will be some time before fixes are
forthcoming. 

The changes in this PR silence such deprecation warnings for
`pytorch_inference` only.

Labelling as `>non-issue` as this introduces no functional change.


bin/pytorch_inference/Makefile
```diff
@@ -22,6 +22,10 @@ USE_NET=1
 USE_RAPIDJSON=1
 USE_TORCH=1
 
+ifeq ($(OS),Windows)
+    CPPFLAGS+=-D_HAS_FEATURES_REMOVED_IN_CXX20 -D_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
+endif
```
droberts195, line 27: Please can you add a comment above these lines that links to the PyTorch issue you opened.  It can say that these lines can be removed if the PyTorch code is changed to avoid using features that are removed in C++20.

Also, it's important to note that while these lines would allow us to upgrade to C++20 on Windows, if clang or gcc unconditionally removes the stuff that's supposed to be removed in C++20 then we still won't be able to switch to C++20 even though their compilers operating in C++17 mode don't warn us.

droberts195 (approved):
LGTM

PR #2153, opened 2021-12-07
[ML] Add the ability to optionally disable hyperparameter scaling

tveasey:


lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -168,6 +168,8 @@ CDataFrameTrainBoostedTreeRunner::CDataFrameTrainBoostedTreeRunner(
```
valeriy42, line 168: You are missing `theReader.addParameter(...)` above.

valeriy42 (approved):
LGTM. A single comment.

lib/api/CDataFrameTrainBoostedTreeRunner.cc
tveasey: Done in [e294c87](https://github.com/elastic/ml-cpp/pull/2153/commits/e294c872d42482e2b16ebee3e9a3866e002d72d1).

PR #2156, opened 2021-12-09
[ML] Verify generalization error estimation for multi-step learning

### LINKED ISSUE #2155: [ML] Verify generalization error estimation for multi-step learning
Cross-validation error for hyperparameter estimation should estimate the generalization error (aka test error on the holdout dataset). This works as long as the test and training datasets have identical distributions. For instance, the training set distribution will become skewed when we use disagreement sampling or residual-based sampling for finding update samples.

Implement a test that verifies that the test error does not increase when we do multi-step learning with a random sampling of the update data (distributions of the training data and test data are the same).

## Why is it important?

We optimize hyperparameters, assuming that the validation error matches the test error. 

Suppose the assumption doesn't hold, and we observe the multi-step learning to overfit. In that case, we don't know if the problem is in the algorithm or the representation of the generalization error during training.

valeriy42:
This PR introduces a test that ensures that for multi-step learning, the generalization error is always non-increasing. This means we are avoiding overfitting.

Closes #2155


valeriy42:
I will keep this PR as a draft until #2157 is implemented. Once we can use a hold-out dataset for validation, I can 

1. pass the test data as a validation dataset and make sure that the test error indeed does not decrease from one step to another
2. sample test data and validation data from a (largish) dataset and verify that the test error is not increasing for more than a difference between the test data and validation data distributions.

valeriy42:
@tveasey it would be great if you could take a brief look at this PR.

tveasey (approved):
LGTM

PR #2160, opened 2021-12-15
[ML] Fixing the way we set up the ES JVM on linux-aarch64

droberts195:
PR CI tests run some of the Elasticsearch integration tests.
We've always had a slightly hacky way of doing this on
linux-aarch64.  For JDK 17 things are a bit different because
Java has fixed the page size issues that affected earlier
versions.  Each Elasticsearch CI worker now has a symlink
from "jdk17" to the actual JDK directory on the machine, so
we should use that.

This PR should fix errors like "ERROR: JAVA_HOME is set to an
invalid directory: /var/lib/jenkins/.java/adoptopenjdk17" in
https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+linux-aarch64/749/console

### DIFF
--- a/dev-tools/run_es_tests.sh
+++ b/dev-tools/run_es_tests.sh
@@ -83,27 +83,10 @@ fi
 # distribution that it's running on, which is:
 # - 4KB for Ubuntu, Debian and SLES
 # - 64KB for RHEL and CentOS
-# We still disable the CDS archive as an extra measure to avoid warnings related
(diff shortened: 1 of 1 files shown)


droberts195:
retest

droberts195:
retest

dimitris-athanasiou (approved):
LGTM

PR #2163, opened 2021-12-21
[ML] Publish third party library dependencies in separate artifact

### LINKED ISSUE #2158: Publish third party library dependencies in separate artifact
I'm sure you all hear the gripes about the ML JAR being a hefty download. Right now Elasticsearch developers incur a 350MB download at least once a day (or whenever the latest SNAPSHOT changes). Some folks survive with less than ideal internet access and when S3 is acting wonky this slows everyone down.

The _vast_ majority of this artifact is dependent libraries which rarely change. One thought is to publish separate artifacts, one with just the ML code (ml-nodeps) and another with just the dependencies (ml-deps), the current distribution could remain (or become ml-full). The idea here being that the dependencies artifact would change infrequently so the locally cached artifact could be reused more often than not, with only the `-nodeps` artifact needing to be redownloaded. Gradle handles this for us by doing HEAD requests on artifacts to see if they have changed, if not, use the locally cached one. This saves us downloading a few hundred megs of native binaries because of a single change in the `ml-cpp` repo. Thoughts?

droberts195:
In addition to the artifacts we currently produce we'll separately
bundle:

1. The binaries and other resources based on our own code
2. 3rd party dependencies

We expect the 3rd party dependencies to change very rarely, so
this will reduce the number of times this second artifact needs
to be downloaded onto Elasticsearch developers' laptops, helping
developers with slow internet connections.

This change makes no difference whatsoever to what gets released
to end users.

Closes #2158


dimitris-athanasiou (approved):
LGTM

build.gradle
```diff
@@ -197,6 +197,7 @@ task buildUberZip(type: Zip) {
   with uberZipSpec
   destinationDirectory = file("${buildDir}/distributions")
   archiveVersion = project.version
+  reproducibleFileOrder = true
```
dimitris-athanasiou, line 200: Is this variable read anywhere?

droberts195: Yes, internally Gradle's `Zip` task will read it via [this](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/bundling/AbstractArchiveTask.html#isReproducibleFileOrder--).

I think it basically means, "add the files in alphabetical order instead of the order you encounter them in the file system".

mark-vieira: I think we may need `preserveFileTimestamps = false` as well here to have truly reproducible artifacts.

dev-tools/jenkins_ci.sh
```diff
@@ -146,6 +146,10 @@ case `uname` in
             IVY_REPO="${GIT_TOPLEVEL}/../ivy"
             mkdir -p "${IVY_REPO}/maven/org/elasticsearch/ml/ml-cpp/$VERSION"
             cp "../build/distributions/ml-cpp-$VERSION-linux-$HARDWARE_ARCH.zip" "${IVY_REPO}/maven/org/elasticsearch/ml/ml-cpp/$VERSION/ml-cpp-$VERSION.zip"
+            # Since this is all local, for simplicity, cheat with the dependencies/no-dependencies split
```
mark-vieira, line 149: Not sure I follow this bit.

droberts195: This is for the step of testing ml-cpp PRs where we run a subset of Elasticsearch integration tests on Linux using the ml-cpp code that's changed by the PR.

We were already cheating here, because we renamed the zip file containing just the code for Linux on the current architecture to the name of the zip file that's supposed to contain the code for all platforms.

Since the test is taking place on the local machine and the Elasticsearch build will "download" the newly built code from an Ivy repo on the local disk, we might as well just put all the files in one zip and have an empty zip for the other one - the combination of the two zip files will still contain all the necessary files.

Partly this depends on how the Elasticsearch side works.  I've opened https://github.com/elastic/elasticsearch/pull/82268 with my first attempt at this and it doesn't work at all - if I point what's currently in that PR at a local Ivy repo containing two zip files then it doesn't currently incorporate either of them into the build.  Please let me know if you have any ideas for how to make it work.

droberts195:
retest

dev-tools/jenkins_ci.sh
mark-vieira: Yeah, the project substitution stuff is going to be tricky since you're saying "replace both these things with the same project artifact". There are a few different things we could do here, to solve this problem which we can discuss separate from this PR. I think for now the simplest thing is just to use the "full" artifact when integrating with the Elasticsearch build.

droberts195:
retest

droberts195:
For the time being this will have no effect, but merging it now allows the release manager changes to be made, and then the Elasticsearch side will follow after that.

PR #2164, opened 2021-12-21
[ML] Tighten up the pattern used to copy Torch dependencies

droberts195:
The old pattern was too lax, copying all files with "c10" in
their names, which also matches "gcc10". This was causing us
to copy the symlinked versions of the Boost libraries and hence
ship each Boost library twice on Linux.


dimitris-athanasiou (approved):
LGTM

PR #2167, opened 2022-01-06
[ML] Ensure we have sufficient resolution when we reinitialise component models 

### LINKED ISSUE #2166: [ML] Better handling of edge cases when reinitialising models after detecting new seasonal components
We test for seasonality over a range of time window lengths. When we test longer windows we bucket data over longer intervals to reduce the memory we consume. If there are are both slow and fast periodicities in the data, say 30 min and 1 day repeats, then when we detect daily periodicity we may not have enough points to properly characterise the 30 min seasonal pattern if the job itself uses a short bucket interval.

In general, when we detect new seasonality we typically choose to reinitialise all other seasonal models where we deem we have enough resolution since seasonal components we're not modelling will typically introduce bias in other component models. Normally we handle the edge case by excluding short seasonalities from consideration when testing long windows. We achieve this by thresholding the ratio between the window bucket length and the period. We currently use the same setting for this to decide test for new seasonal components and whether to reinitialise ones we're already modelling. This decision should probably be decoupled and we should require higher resolution to reinitialise seasonal components when we adding new ones.

tveasey:
When we retest if a modelled seasonal component is still appropriate or may contain bias we now require higher resolution if shorter time windows are available. This is important because if the time series has a mixture of fast and slow seasonality we will detect the fast seasonality first and then potentially reinitialise it when we detect the slow seasonality. We should only do this if the window provides us with sufficient resolution to initialise it properly.

Closes #2166.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -72,6 +72,17 @@
   - the C++ front end to PyTorch - and performs inference on models stored in the
   TorchScript format. (See {ml-pull}1902[#1902].)
 
+
+== {es} version 7.17.0
+
+=== Bug Fixes
+
+* Avoid transient poor time series modelling after detecting new seasonal components.
+  This can affect cases where we have fast and slow repeats in the data, for example
+  30 mins and 1 day, and the job uses a short bucket length. The outcome can be transient
+  poor predictions and model bounds, and sometimes false positive anomalies. (See
+  {ml-pull}2167[#2167].)
+
 == {es} version 7.16.0
 
 === Enhancements
--- a/include/maths/time_series/CExpandingWindow.h
+++ b/include/maths/time_series/CExpandingWindow.h
@@ -88,6 +88,9 @@ public:
     //! Get the number of bucket values.
     std::size_t size() const;
 
+    //! Check if there are shorter windows.
+    bool haveShorterWindows() const;
+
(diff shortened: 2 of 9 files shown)


droberts195:
In the case of 30 minute periodicity being detected first, then 24 hour periodicity being detected later, is the effect of this PR that we'll opt not to model the 24 hour periodicity at all if doing so would ruin the modelling of the 30 minute periodic pattern?

tveasey:
> In the case of 30 minute periodicity being detected first, then 24 hour periodicity being detected later, is the effect of this PR that we'll opt not to model the 24 hour periodicity at all if doing so would ruin the modelling of the 30 minute periodic pattern?

No. We simply are more careful about which windows will trigger us to reinitialise components. In particular, we will only consider reinitialising a component if we have sufficient resolution. There is a tradeoff here since the current models will contain bias due to averaging over the unmodelled components. Either way it'll rectify, but I think it probably makes sense to keep the current model if we can't do a good job of reinitialising it on the principle of least surprise.

Here are the results in a prototypical example, from the unit test I added (note the second one is more zoomed but they both cover the detection of the daily periodicity it just isn't as clear in the second case because there is no reinitialisation issue).

before

Screenshot 2022-01-06 at 12 13 26

after

Screenshot 2022-01-06 at 12 10 26

droberts195 (approved):
LGTM

PR #2172, opened 2022-01-13
[ML] Add new tree during incremental learning

valeriy42:
When considering the changes in the forest during the incremental learning step, we now allow the system to add up to `max_num_new_trees` to the forest. This only happens if adding new trees doesn't lead to overfitting. 

The parameter `max_num_new_trees` has to be explicitly specified by the user since its default value `0` leads to the old behavior.


tveasey (approved):
This is a really nice implementation. Great work! (Note the test failure looks to be down to a small difference in numerics arm.)

PR #2173, opened 2022-01-14
[ML] Upload combined artifacts to GCS as well as S3

droberts195:
The upload to S3 is done in the same way as before, except
that the call to Gradle is now made from a script rather
than being part of the Jenkins job config. This will make
it easier to evolve in the future without making infra
changes.

The upload to GCS is done by a Docker image owned by the
release team that generates checksums and metadata such
that release manager can generate a release from a specific
ml-cpp commit rather than always using the latest commit on
a given branch.


Conky5 (approved):
LGTM!

jmlrt (approved):
LGTM⛴

PR #2177, opened 2022-01-17
[ML] Download debug symbols as well as binaries in combine step

droberts195:
The "combine artifacts" step of the ML CI workflow previously
used to download platform-specific binaries from S3 and upload
an artifact built by combining them to S3. It didn't need to
do anything with the platform-specific debug symbols that were
already on S3.

Now that the "combine artifacts" step also uploads artifacts to
GCS it also needs to upload the zips containing the debug symbols.
Therefore, this commit changes the step that downloads
platform-specific artifacts to get the debug symbols too.

Followup to #2173

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -302,6 +302,13 @@ class DownloadPlatformSpecific extends DefaultTask {
         }
       }
       zip.close()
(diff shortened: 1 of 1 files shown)


edsavage (approved):
LGTM

PR #2180, opened 2022-01-17
[ML] Include version in dependency report file name

droberts195:
Currently the dependency report is always simply "dependencies.csv".

For integration into the new release manager workflow the file name
needs to also contain the qualified version.

PR #2181, opened 2022-01-18
[8.0] [ML] Include version in dependency report file name

droberts195:
Currently the dependency report is always simply "dependencies.csv".

For integration into the new release manager workflow the file name
needs to also contain the qualified version.

Backport of #2180


jmlrt (approved):
LGTM⛴

PR #2183, opened 2022-01-18
[ML] Better handle almost periodic data in time series anomaly detection

tveasey:
Currently, we have can issues with our time series model when:
1. The data repeat is not a multiple of the bucket length (or at least not very close to a multiple),
2. The data are periodic but the pattern shifts from time to time.

In the first case we get slow precession of the periodic pattern and generally we don't adapt to this fast enough. The effect is we lose prediction accuracy and sensitivity over time. In the second case we get similar effects, but also often generate bursts of anomalies when the pattern shifts slightly.

This change introduces a latent time shift for each seasonal component which is modelled and solves for the shift which minimises the difference between its predictions and the observed values. Shifts are accumulated over each period which is observed, which allows us to handle slow precession.

### DIFF
--- a/include/maths/common/CLbfgs.h
+++ b/include/maths/common/CLbfgs.h
@@ -301,7 +301,8 @@ private:
 
         double smin;
         double fmin;
-        CSolvers::globalMinimize(probes, f_, smin, fmin);
+        double fsd;
+        CSolvers::globalMinimize(probes, f_, smin, fmin, fsd);
         xs = m_X - smin * step;
         fs = f(xs);
     }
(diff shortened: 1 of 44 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -37,6 +37,8 @@
 * Upgrade RapidJSON to 31st October 2021 version. (See {ml-pull}2106[#2106].)
 * Upgrade Eigen library to version 3.4.0. (See {ml-pull}2137[#2137].)
 * Prevent over-subscription of threads in pytorch_inference. (See {ml-pull}2141[#2141].)
+* Better handle small shifts of the seasonal patterns in time series data.
+  (See {ml-pull}2183[#2183].)
```
droberts195, line 41: Please can you move this into the 8.2.0 section.

tveasey:
This evolved a bit while I tested it and I think there is a simpler smaller PR inside this one which captures the current behaviour. I'm going to have a second go at implementing this.

PR #2184, opened 2022-01-18
[ML] Specify main instead of master when using infra Docker image

droberts195:
The infra repo is still using a master branch instead of main.
However, the infra Docker image has been changed so that if the
branch to use is specified as main then it gets mapped to master.
Therefore it's more future-proof to specify main rather than
master. Specifying main means that when the infra repo is changed
from master to main nothing will need changing in the ml-cpp
repo. The change at that time will be in the infra repo, which is
easier to coordinate with a branch rename in that same repo.


edsavage (approved):
LGTM

PR #2189, opened 2022-01-20
[ML] Fix regression in incremental learning performance

valeriy42:
I investigated the recent performance regression in the results of incremental learning benchmarks. It came down to the change I made in b196d084, where I introduced explicit override of the hyperparameters. 

I added this override to the experiment driver of incremental learning.

I reran the benchmark and documented the results [here](https://github.com/elastic/ml-cpp/files/7903293/Investigate.performance.regression.pdf). Performance regression disappeared.


valeriy42:
@tveasey this is a trivial PR. More FYI than anything else.

PR #2190, opened 2022-01-24
[ML] Support testing on a holdout set rather than cross-validation when training

tveasey:
When training from scratch on very large data sets using multiple rounds of incremental training, we have both plenty of data and we don't want the examples selected to include in the training set in later rounds to be used for validation. They will depend on the selection criterion and so will not be a uniform sampling of the full data set. Also this process is likely to leak information about the model.

In this case, it is better to use a fixed evaluation set which is held out from training and can be sampled at random upfront. This change introduces that holdout set. Since we maintain row masks for the test and train sets for cross-validation it is easy to hijack the cross-validation mechanism for this purpose by simply having one fold.

### DIFF
--- a/include/maths/analytics/CBoostedTreeHyperparameters.h
+++ b/include/maths/analytics/CBoostedTreeHyperparameters.h
@@ -639,7 +639,9 @@ public:
     bool selectNext(const TMeanVarAccumulator& testLossMoments,
                     double explainedVariance = 0.0);
 
-    //! Capture the current hyperparameters if they're the best we've seen so far.
+    //! Capture the current hyperparameters if they're the best we've seen.
+    //!
+    //! \return True if we they are the best hyperparameters.
     bool captureBest(const TMeanVarAccumulator& testLossMoments,
                      double meanLossGap,
                      double numberKeptNodes,
--- a/include/maths/analytics/CBoostedTreeHyperparameters.h
+++ b/include/maths/analytics/CBoostedTreeHyperparameters.h
@@ -640,7 +640,7 @@ public:
                     double explainedVariance = 0.0);
 
     //! Capture the current hyperparameters if they're the best we've seen so far.
-    void captureBest(const TMeanVarAccumulator& testLossMoments,
+    bool captureBest(const TMeanVarAccumulator& testLossMoments,
                      double meanLossGap,
                      double numberKeptNodes,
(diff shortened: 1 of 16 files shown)


lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -792,6 +790,182 @@ BOOST_AUTO_TEST_CASE(testLowTrainFractionPerFold) {
...
+                double actual{(*row)[cols - 1]};
+                double prediction{regression->prediction(*row)[0]};
+                double previousPrediction{regression->previousPrediction(*row)[0]};
+                expectedMse.add(maths::common::CTools::pow2(actual - prediction));
+                expectedMsd.add(maths::common::CTools::pow2(prediction - previousPrediction));
+            }
+        },
+        &holdoutRowMask);
+    LOG_DEBUG(<< expectedMse << " " << expectedMsd);
+    LOG_DEBUG(<< maths::common::CBasicStatistics::mean(expectedMse) +
```
valeriy42, line 959: What does this number mean?

valeriy42 (approved):
LGTM. :tada:  Good change. I have only some minor comments.

lib/maths/analytics/unittest/CBoostedTreeTest.cc
tveasey: This was left over from debugging an issue. See [512fcde](https://github.com/elastic/ml-cpp/pull/2190/commits/512fcdee1660348dfdbda0f9ac35ac20fec042b2).

include/maths/analytics/CBoostedTreeHyperparameters.h
```diff
@@ -640,7 +640,7 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeHyperparameters {
                     double explainedVariance = 0.0);
 
     //! Capture the current hyperparameters if they're the best we've seen so far.
-    void captureBest(const TMeanVarAccumulator& testLossMoments,
+    bool captureBest(const TMeanVarAccumulator& testLossMoments,
```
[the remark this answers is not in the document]
tveasey: Added in [e774df4](https://github.com/elastic/ml-cpp/pull/2190/commits/e774df4a3db39639eab0aabc7db05c67ec4d42f1).

PR #2192, opened 2022-01-31
[ML] Fix data race in classification weight calculation

tveasey:
Whilst evaluating incremental training we found an issue with the classification weights calculation for maximising minimum recall. We were setting the weights to [0, 1], which effectively means we'll predict a constant class. Clearly this is a bug since the class we never predict has zero recall.

I noticed three things: 1) we have a data race reading and writing to a probability vector when setting up the ROC curve as a precursor to the calculation (this is just a clear bug), 2) we allow the optimisation routine to select 0/1 weights which is never a good idea, 3) we aren't robust to infinite and NaN values in the input (although this shouldn't happen they would cause optimisation to bail and use one of the interval initial points, which is a bad idea). I haven't yet been able to confirm that any of these caused our problem, and it seems like something else must be at play. However, it worthwhile fixing these issues.


valeriy42 (approved):
LGTM. Good catch!

PR #2194, opened 2022-02-01
[ML] Improve classification weights robustness

tveasey:
This addresses two potential issues with the maximise minimum recall classification weight calculation:
  1. We allow the optimisation routine to select 0/1 weights which would never be a good idea,
  2. We aren't robust to infinite and NaN values in the input. Although this shouldn't happen they would cause optimisation to bail and use one of the interval end points, which is a bad idea.

Note this also removes a duplicate header (CMathsFuncsDetail.h == CMathsFuncsForMatrixAndVectorTypes.h).


valeriy42 (approved):
LGTM

PR #2196, opened 2022-02-04
[ML] Address cause of "[CStatisticalTests.cc@102] Test statistic is nan"  log errors

tveasey:
This unifies the handling of edge cases for calculating the F-test statistic which needs to handle 0 / 0 and x / 0. This was spread over various places and not always applied consistently. There were also some uses which did not properly scale the statistic by the degrees of freedom. This second fix has an impact (albeit v.small) on our results as tweaks to the test thresholds signify.

This addresses some sources of errors observed in our cloud logging clusters.


valeriy42 (approved):
LGTM

PR #2197, opened 2022-02-07
[ML] Improve handling of non-finite values

tveasey:
This makes a number of changes to improve numeric robustness around handling of NaN and infinite values:
1. Trap NaN and inf inputs to distribution probability of less likely samples calculations,
2. Avoid using x * y, which can underflow, when testing for root bracketing,
3. Avoid v * w / x * y which can lead to NaN when the products underflow,
4. Trap the case `maxIterations` is supplied as 0 to various numeric routines, which would lead looping over all positive integers,
5. Tidy up and improve the logging of non finite values in various places (avoid duplicate logging and add extra useful information).

These were found trying to reproduce "x = NaN, distribution = N5boost4math23students_t_distribution..." log errors by injecting NaNs at various places.

### DIFF
--- a/include/maths/common/CSolvers.h
+++ b/include/maths/common/CSolvers.h
@@ -54,15 +54,18 @@ private:
                                                      const double fa,
                                                      const double fb,
                                                      const double fc) {
-        return a * fb * fc / (fa - fb) / (fa - fc) +
-               b * fa * fc / (fb - fa) / (fb - fc) + c * fa * fb / (fc - fa) / (fc - fb);
+        // The bracketing is important here: if f(x) < (min double)^(1/2)
+        // the products will underflow.
+        return a * (fb / (fa - fb)) * (fc / (fa - fc)) +
+               b * (fa / (fb - fa)) * (fc / (fb - fc)) +
+               c * (fa / (fc - fa)) * (fb / (fc - fb));
     }
 
     //! A secant interpolation of two distinct function values.
     //! WARNING the caller must ensure that \p fa != \p fb.
     static inline double
     secantInterpolate(const double a, const double b, const double fa, const double fb) {
-        return b - fb * (b - a) / (fb - fa);
+        return b - (b - a) * (fb / (fb - fa));
     }
 
     //! Bisect the interval [\p a, \p b].
@@ -126,11 +129,14 @@ private:
             fx = fb;
         }
 
-        double w = x, v = x;
-        double fw = fx, fv = fx;
-        double s = 0.0, sLast = 0.0;
+        double w = x;
+        double v = x;
+        double fw = fx;
+        double fv = fx;
+        double s = 0.0;
+        double sLast = 0.0;
 
-        std::size_t n = maxIterations;
+        std::size_t n = std::max(maxIterations, std::size_t{1});
 
         do {
             double xm = bisect(a, b);
@@ -230,7 +236,7 @@ private:
 
         std::size_t n = maxIterations;
         for (/**/; n > 0; --n) {
-            if (fa * fb <= 0.0) {
+            if (std::signbit(fa) != std::signbit(fb) || fa == 0.0 || fb == 0.0) {
                 break;
             }
 
@@ -263,7 +269,7 @@ private:
         }
 
         maxIterations = maxIterations - n;
-        return fa * fb <= 0.0;
+        return std::signbit(fa) != std::signbit(fb) || fa == 0.0 || fb == 0.0;
     }
 
 public:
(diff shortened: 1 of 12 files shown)


include/maths/common/CSolvers.h
```diff
@@ -570,7 +583,7 @@ class MATHS_COMMON_EXPORT CSolvers {
                 return true;
             }
 
-            if (fa * fs > 0.0) {
+            if (std::signbit(fa) == std::signbit(fs)) {
```
droberts195, line 586: Should this also confirm both variables are not equal to zero?  The new condition is slightly different in the case of one or both variables being zero with the same sign.

include/maths/common/CSolvers.h
```diff
@@ -685,7 +698,7 @@ class MATHS_COMMON_EXPORT CSolvers {
             maxIterations -= n;
             return true;
         }
-        if (fa * fb > 0.0) {
+        if (std::signbit(fa) == std::signbit(fb)) {
```
droberts195, line 701: As above, this is slightly different to before if one or both variables is zero.

include/maths/common/CSolvers.h
```diff
@@ -701,7 +714,7 @@ class MATHS_COMMON_EXPORT CSolvers {
                 return true;
             }
 
-            if (fa * fc > 0.0) {
+            if (std::signbit(fa) == std::signbit(fc)) {
```
droberts195, line 717: As above, this is slightly different to before if one or both variables is zero.

lib/maths/common/COneOfNPrior.cc
```diff
@@ -320,7 +320,7 @@ void COneOfNPrior::addSamples(const TDouble1Vec& samples,
     for (std::size_t i = 0; i < samples.size(); ++i) {
         double xi = samples[i];
         double ni = maths_t::countForUpdate(weights[i]);
```
droberts195, line 322: nit: calculation of `ni` could move inside the `if`

lib/maths/common/CTools.cc
```diff
@@ -774,7 +774,7 @@ operator()(const CLogTDistribution& logt, double x, maths_t::ETail& tail) const
     double step = std::max(b2, std::exp(l) - b2);
     double growthFactor = 1.0;
     for (;;) {
-        if (maxIterations == 0 || f2 <= fx) {
+        if (maxIterations == 2 || f2 <= fx) {
```
droberts195, line 777: It's a bit weird to stop at 2.  I guess it relates to the way `step` is handled differently when it's a certain proportion of `MAX_ITERATIONS`, but please could you add a comment here to confirm it's not a mistake.

lib/maths/common/unittest/CSolversTest.cc
```diff
@@ -13,6 +13,7 @@
 #include <core/CLogger.h>
 #include <core/CStringUtils.h>
 
+#include <limits>
```
droberts195, line 16: This is already present in the correct place below, so can be removed from here.

include/maths/common/CSolvers.h
tveasey: Those conditions are actually handled in the code immediately above.

lib/maths/common/CTools.cc
tveasey: Actually, this was accidentally committed now I added protection for maxIterations being 0 in the solver code.

include/maths/common/CSolvers.h
tveasey: As above neither variable can be 0 here.

lib/maths/common/CTools.cc
tveasey: See [33a3ef2](https://github.com/elastic/ml-cpp/pull/2197/commits/33a3ef2464aab0cb830532f243d7a5e7bcb263b5).

lib/maths/common/COneOfNPrior.cc
tveasey: Done in [bc63e9f](https://github.com/elastic/ml-cpp/pull/2197/commits/bc63e9fa078fe98f7386f8e5c3b995a59dd50f0b).

lib/maths/common/unittest/CSolversTest.cc
tveasey: See [783369e](https://github.com/elastic/ml-cpp/pull/2197/commits/783369ed44a8dc9b4b7be54fb434c723a3f10558).

droberts195 (approved):
LGTM

PR #2202, opened 2022-02-08
[ML] Handle precession and time shifts of seasonality in anomaly detection

tveasey:
Currently, we have can issues with our time series modelling when:

1. The data are seasonal and the period is not a multiple of the bucket length (or at least not very close to a multiple),
2. The data have seasonality, but the pattern shifts from time to time.

In both cases we get precession of the periodic pattern and generally we don't adapt to this fast enough. The effect is we lose prediction accuracy and sensitivity over time.

This change introduces a latent time shift for each seasonal component and solves for the shift which minimises the difference between the predictions and the observed values. Shifts are accumulated over each period which is observed, which allows us to handle slow precession.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -33,7 +33,7 @@
 === Enhancements
 
 * Better handle small shifts of the seasonal patterns in time series data.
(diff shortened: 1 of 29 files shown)


valeriy42 (approved):
LGTM

PR #2204, opened 2022-02-08
[ML] Correct loss gap calculation when evaluating using a holdout set

tveasey:
When we are assessing generalisation error using a holdout set there is no need to account for the train/test loss gap comparing the output of  incremental training with the original model.


valeriy42 (approved):
LGTM

PR #2208, opened 2022-02-16
[ML] Pass ml-cpp version to release manager integration code

droberts195:
Previously the release manager integration code used its own
version number. Generally this didn't matter, but it meant
there was a period when builds would fail as the version was
bumped in all the repos. Passing in the version from the ml-cpp
repo will mean that builds don't fail if the version bumps in
ml-cpp and infra are out of sync.


edsavage (approved):
LGTM

PR #2213, opened 2022-02-17
[ML] Fix bug restoring data gatherer state for time of day/week and anomaly detection

tveasey:
We were not properly restoring the data gatherer state for time of day and week anomaly detection functions. This could lead to "No queue item for time <x>" and "Time <x> is out of range. Returning earliest bucket index" log errors and us potentially missing data for them.


droberts195 (approved):
LGTM

PR #2216, opened 2022-02-21
[ML] Remove some unused (and buggy) functionality in anomaly detection data gathering

tveasey:
This PR removes some functionality for updating time which was not being used and in fact wasn't properly maintaining class invariants. It also adds some extra useful information in various log messages, in case we still have problems in this area, after fixing #2213.

Finally, it addresses assorted warnings in this code generated by clang-tidy. Principally, we were using bind, because the code predated lambdas, however calling via lambdas often produces better inlining.


tveasey:
> My only question would be whether it's now possible to remove #include <functional> from any of the modified source files.

I'll double check.

PR #2219, opened 2022-02-22
[ML] Strengthen checks on state restoration

edsavage:
Perform additional checks during restoration of anomaly detector job
state in order to fail quickly and cleanly when errors in the state are
encountered.

Labelling as `>non-issue` as the changes have no functional impact on users


droberts195 (approved):
LGTM

PR #2220, opened 2022-02-23
[ML] Refactor state traverser checks for bad state

edsavage:
Centralise checks for bad state in the public methods of the state
traverser itself.

Relates #2219


tveasey (approved):
LGTM

PR #2221, opened 2022-02-23
[ML] Assorted tidy ups

tveasey:
This addresses a variety of sensible clang-tidy suggestions.

Principally:
1. Migrating from `std::bind` to lambdas,
2. Addressing implicit conversions of ints and pointers to bools,
3. Matching function declaration and definition parameter names,
4. Not using lower case explicit integer type qualifiers,
5. Using `empty()` rather than 'size() == 0`,
6. Using explicit constructors,
7. Preferring range based for loops,
8. Preferring auto when initialising from a cast.

These are not fully addressed in the code base. I focused mainly on the `maths` library. We can gradually address more cases in other PR. Whilst making these changes I noticed in various places we were also silently swallowing errors restoring types in constructors. In these, cases we now set the traverser state as bad.

Finally, we can now remove pre 6.3 state restoration from main. Code and associated tests for these upgrade paths have been removed.


lib/maths/time_series/CTimeSeriesDecomposition.cc
```diff
@@ -127,59 +121,38 @@ bool CTimeSeriesDecomposition::acceptRestoreTraverser(const common::SDistributio
...
-                        &CComponents::acceptRestoreTraverser, &m_Components,
-                        std::cref(params), m_LastValueTime, std::placeholders::_1)))
+                    traverser.traverseSubLevel([this](auto& traverser_) {
+                        return m_CalendarCyclicTest.acceptRestoreTraverser(traverser_);
+                    }))
+            RESTORE(COMPONENTS_6_3_TAG, traverser.traverseSubLevel([&](auto& traverser_) {
+                return m_Components.acceptRestoreTraverser(params, traverser_);
+            }))
         }
-    } else {
```
droberts195, line 165: Maybe return false if neither supported version tag is found?

lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -1268,59 +1261,35 @@ bool CUnivariateTimeSeriesModel::acceptRestoreTraverser(const common::SModelRest
...
                 m_AnomalyModel = std::make_unique<CTimeSeriesAnomalyModel>(),
-                traverser.traverseSubLevel(std::bind(
-                    &CTimeSeriesAnomalyModel::acceptRestoreTraverser,
-                    m_AnomalyModel.get(), std::cref(params), std::placeholders::_1)),
+                traverser.traverseSubLevel([&](auto& traverser_) {
+                    return m_AnomalyModel->acceptRestoreTraverser(params, traverser_);
+                }),
                 /**/)
         }
-    } else {
```
droberts195, line 1295: Else return false?

lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -2696,59 +2662,38 @@ bool CMultivariateTimeSeriesModel::acceptRestoreTraverser(const common::SModelRe
...
```
droberts195, line 2724: Else return false?

droberts195:
Also, in case you didn't see it:

> A format error has been detected within the following files:
> `lib/api/unittest/CRestorePreviousStateTest.cc`

lib/maths/time_series/CTimeSeriesDecomposition.cc
tveasey: Done in [ca10699](https://github.com/elastic/ml-cpp/pull/2221/commits/ca106994958c0af9ea88d7b2bb9d008293d1281c).

droberts195 (approved):
LGTM

PR #2222, opened 2022-03-01
[ML] Support example weights in classification and regression model training

tveasey:
This allows the user to specify a column to hold example weights. In this case it is excluded from candidate features and weights that example's contribution to the loss. As before by default no weighting is applied. A nice side effect is we actually reduce memory usage since there was an unused placeholder for example weights added to the data frame.


lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -367,30 +367,30 @@ void CBoostedTreeFactory::initializeNumberFolds(core::CDataFrame& frame) const {
 
 void CBoostedTreeFactory::resizeDataFrame(core::CDataFrame& frame) const {
 
+    if (m_RowWeightColumn != UNIT_ROW_WEIGHT_COLUMN &&
+        (m_RowWeightColumn >= frame.numberColumns() ||
+         m_RowWeightColumn == m_TreeImpl->columnHoldingDependentVariable())) {
```
valeriy42, line 372: Do you also need to check that `m_RowWeightColumn` is not a categorical column?

PR #2224, opened 2022-03-02
Use deps and nodeps archives for project artifacts

mark-vieira:
This relates to https://github.com/elastic/elasticsearch/pull/84536. For composite build integration to work using the separate "deps" and "nodeps" artifacts, we need to tweak the build to expose _both_ as project artifacts so they can be properly resolved by the Elasticsearch build. This PR includes a few things to make that happen:

1. Apply the Gradle "base" plugin to inherit the necessary configuration so that the `default` configuration is properly consumed by the ES build.
2. Introduce additional tasks to create the "deps" and "nodeps" artifacts from the locally built zip (isntead of from a downloaded snapshot)
3. Remove superfluous configuration which is provided by base plugin conventions.
4. Modify lifecycle tasks (ie. clean, assemble, check, etc) to work with base plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -327,7 +327,6 @@ def dependenciesSpec(source) {
 
 task buildDependenciesZip(type: Zip) {
   description = 'Create a zip of dependencies only from combined platform-specific C++ distributions'
-  dependsOn buildUberZip
 
   classifier = 'deps'
   reproducibleFileOrder = true
@@ -335,7 +334,7 @@ task buildDependenciesZip(type: Zip) {
   // timestamp diffs between otherwise identical archives
   preserveFileTimestamps = false
 
-  with dependenciesSpec(zipTree(buildUberZip))
+  with dependenciesSpec(zipTree(buildUberZip.archiveFile))
 }
 
 task buildDependenciesZipFromDownloads(type: Zip) {
@@ -373,7 +372,6 @@ def noDependenciesSpec(source) {
 
 task buildNoDependenciesZip(type: Zip) {
   description = 'Create a zip excluding dependencies from combined platform-specific C++ distributions'
-  dependsOn buildUberZip
 
   classifier = 'nodeps'
   reproducibleFileOrder = true
@@ -381,7 +379,7 @@ task buildNoDependenciesZip(type: Zip) {
   // timestamp diffs between otherwise identical archives
   preserveFileTimestamps = false
 
-  with noDependenciesSpec(zipTree(buildUberZip))
+  with noDependenciesSpec(zipTree(buildUberZip.archiveFile))
 }
 
 task buildNoDependenciesZipFromDownloads(type: Zip) {
--- a/build.gradle
+++ b/build.gradle
@@ -137,6 +137,7 @@ task format(type: Exec) {
 
 task buildZip(type: Zip) {
   dependsOn strip
+  archiveClassifier = artifactClassifier
   from("${buildDir}/distribution") {
     // Don't copy Windows import libraries
     exclude "**/*.lib"
@@ -155,7 +156,7 @@ task buildZip(type: Zip) {
 
 task buildZipSymbols(type: Zip) {
   dependsOn strip
-  classifier = "debug"
+  archiveClassifier = "${artifactClassifier}-debug"
   from("${buildDir}/distribution") {
     // only take debug files
     include "**/*-debug"
@@ -201,7 +202,7 @@ check {
 }
 
 assemble {
-  dependsOn = ['buildUberZip', 'buildZipSymbols']
+  dependsOn = ['buildUberZip', 'buildZip', 'buildZipSymbols']
   description = 'Assemble the C++ part of Machine Learning'
 }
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -146,16 +153,9 @@ def zipSpec = copySpec {
...
 }
 
-task buildZip(type: Zip) {
+task buildZipSymbols(type: Zip) {
   dependsOn strip
-  archiveBaseName = artifactName
-  with zipSpec
-  destinationDirectory = file("${buildDir}/distributions")
-  archiveVersion = project.version
-  archiveClassifier = artifactClassifier
```
droberts195, line 155: There's a complication with changing what the classifier is used for and combining `buildZip` and `buildUberZip` into one task.

Previously a `gradle assemble` in the `ml-cpp` repo (completely independent of Elasticsearch) would result in these files being generated:

```
bash-3.2$ gradle assemble
bash-3.2$ ls -l build/distributions 
total 371512
-rw-rw-r--  1 dave  dev  44297817 Mar  3 10:38 ml-cpp-8.2.0-SNAPSHOT-darwin-x86_64.zip
-rw-rw-r--  1 dave  dev  44297817 Mar  3 10:38 ml-cpp-8.2.0-SNAPSHOT.zip
-rw-rw-r--  1 dave  dev  96359137 Mar  3 10:39 ml-cpp-debug-8.2.0-SNAPSHOT-darwin-x86_64.zip
```

On the PR branch it results in this:

```
bash-3.2$ gradle assemble
bash-3.2$ ls -l build/distributions 
total 284488
-rw-rw-r--  1 dave  dev  96359137 Mar  3 10:33 ml-cpp-8.2.0-SNAPSHOT-debug.zip
-rw-rw-r--  1 dave  dev  44297817 Mar  3 10:32 ml-cpp-8.2.0-SNAPSHOT.zip
```

So we no longer get the platform specific artifacts.  We still need those so that in the full workflow the combine step can download all the platform specific files and combine them.

I don't mind moving where `debug`/`deps`/`nodeps` appears in the file names, so would it be possible to keep the approach of using classifiers for these but still generate files with the platform in their classifiers too, for example, `ml-cpp-8.2.0-SNAPSHOT-darwin-x86_64-debug.zip` where the classifier is `darwin-x86_64-debug`?

build.gradle
```diff
@@ -432,3 +439,9 @@ wrapper {
     println "Added checksum to wrapper properties"
   }
 }
+
+
+artifacts {
+  'default' buildDependenciesZip
+  'default' buildNoDependenciesZip
```
droberts195, line 446: For some reason only the `deps` file gets generated.  This is the same even if the order of these is swapped.

In the `elasticsearch` repo:

```
bash-3.2$ ./gradlew :x-pack:plugin:ml:assemble --include-build ../ml-cpp
.
.
.
> Task :x-pack:plugin:ml:extractNativeLicenses FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':x-pack:plugin:ml:extractNativeLicenses'.
> Cannot expand ZIP '/Users/dave/ml-cpp/build/distributions/ml-cpp-8.2.0-SNAPSHOT-nodeps.zip' as it does not exist.
```

And looking in the `ml-cpp` repo after that:

```
ls -l build/distributions
total 449712
-rw-rw-r--  1 dave  dev  44010914 Mar  3 11:09 ml-cpp-8.2.0-SNAPSHOT-deps.zip
-rw-rw-r--  1 dave  dev  44297817 Mar  3 11:09 ml-cpp-8.2.0-SNAPSHOT.zip
```

droberts195 (commented):
I tested this today and I think there are still a couple of problems.

build.gradle
mark-vieira: Ah yeah, I didn't realize the classifier included the current architecture. I've added this back.

droberts195:
After the latest change the pure ml-cpp build output for a single platform looks better:

```
ls -l build/distributions
total 569648
-rw-rw-r--  1 dave  dev   96359137 Mar  3 18:31 ml-cpp-8.2.0-SNAPSHOT-darwin-x86_64-debug.zip
-rw-rw-r--  1 dave  dev   44297817 Mar  3 18:30 ml-cpp-8.2.0-SNAPSHOT-darwin-x86_64.zip
-rw-rw-r--  1 dave  dev  140656022 Mar  3 18:32 ml-cpp-8.2.0-SNAPSHOT.zip
```

But the side by side build from Elasticsearch using `./gradlew :x-pack:plugin:ml:assemble --include-build ../ml-cpp` now fails like this:

```
> Task :ml-cpp:buildNoDependenciesZip FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ml-cpp:buildNoDependenciesZip'.
> Cannot fingerprint input file property 'rootSpec$1$1$1': Cannot convert the provided notation to a File or URI: task ':ml-cpp:buildUberZip'.
  The following types/formats are supported:
    - A String or CharSequence path, for example 'src/main/java' or '/usr/include'.
    - A String or CharSequence URI, for example 'file:/usr/include'.
    - A File instance.
    - A Path instance.
    - A Directory instance.
    - A RegularFile instance.
    - A URI or URL instance.
    - A TextResource instance.
```

mark-vieira:
That should now be addressed.

droberts195 (approved):
LGTM

Thanks for doing the refactoring.  I just need to adapt to the move of `-debug` to the end of some of the artifact names in a couple more places before merging it.

PR #2225, opened 2022-03-07
[ML] Speedup finding change points in the time series trend

tveasey:
We check for change points in the trend both when performing change point detection and detecting trend and seasonality in our time series modelling. We do this by testing the significance of the change in residual variance with and without the change point. The current approach has O(n^2) complexity with respect to the time window length n. However, least squares regression allows one to compute residual moments in an online fashion, using largely the same set of statistics we currently maintain for estimating parameters. This allows one to drop the complexity to O(n). The slight downside is that we compute variance using the numerically lossy E[X^2] - E[X]^2 formulation. However, we can centre the data before computing variance which means in practice this isn't a problem.

For the sort of window lengths we currently test I got around a 6X performance improvement using this new approach. The impact on worst case bucket processing latency is stronger than average case because the tests are only performed intermittently. This change also adds a couple of constexpr if statements, since we're now on C++17, to ensure unneeded code really is removed altogether.

### DIFF
--- a/include/maths/common/CLeastSquaresOnlineRegression.h
+++ b/include/maths/common/CLeastSquaresOnlineRegression.h
@@ -45,10 +45,8 @@ namespace least_squares_online_regression_detail {
 //! when computing parameters.
 template<typename T>
 struct CMaxCondition {
-    static const double VALUE;
+    static constexpr double VALUE{1e15};
 };
-template<typename T>
-const double CMaxCondition<T>::VALUE = 1e15;
 
 //! Used for getting the default maximum condition number to use
 //! when computing parameters.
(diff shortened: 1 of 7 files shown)


include/maths/common/CLeastSquaresOnlineRegression.h
```diff
@@ -97,25 +100,28 @@ struct MATHS_COMMON_EXPORT CMaxCondition<CFloatStorage> {
 //!
 //! \tparam N_ The degree of the polynomial.
 // clang-format off
-template<std::size_t N_, typename T = CFloatStorage>
+template<std::size_t N_, typename T = CFloatStorage, bool R_2 = false>
```
droberts195, line 103: Please add a Doxygen comment for `R_2`.

include/maths/common/CLeastSquaresOnlineRegression.h
```diff
@@ -271,6 +283,46 @@ class CLeastSquaresOnlineRegression : boost::addable<CLeastSquaresOnlineRegressi
...
+    //! be computed from the statistics maintained.
+    //! \param[in] maxCondition The maximum condition number for
+    //! the Gramian this will consider solving. If the condition
+    //! is worse than this it'll fit a lower order polynomial.
+    //! \param[out] result Filled in with the regression model R^2.
+    bool r2(double& result, double maxCondition = MAX_CONDITION) const {
+
+        result = 0.0;
+
+        if (R_2 == false) {
```
droberts195, line 298: Could be `if constexpr`

droberts195 (approved):
LGTM

include/maths/common/CLeastSquaresOnlineRegression.h
tveasey: See [2ccc477](https://github.com/elastic/ml-cpp/pull/2225/commits/2ccc477a760824de25c3d84cc22649d05c7d4c04).

PR #2229, opened 2022-03-14
[ML] Better handling of low cardinality features for training classification and regression models

tveasey:
If the number of distinct values of a feature is less than the number of splits we test, we can compute an exhaustive set of splits once upfront and cache the rows' split for the duration of training. This speed up training by 30% on a data set with many low cardinality features.


valeriy42 (approved):
LGTM overall. Can you maybe leave a comment about what cardinality is considered a low cardinality? 2 or 5 or 10?

PR #2230, opened 2022-03-15
[ML] Better handling of extreme outliers in time series modelling

tveasey:
We currently handle outliers when decomposing time series on a window by reducing their weight. This is usually sufficient, however extreme outliers can still create problems given the current approach. Ignoring values altogether is risky; for example, periodic spikes are large outliers for many measures. In the sorts of data we deal with extreme values are often localised in time, for example a very large spike. This PR makes a couple of changes to try and deal better with extreme values:
1. Exclude a single short interval within the window being decomposed which is very different from all other values in that window.
2. Discard any component scaling or trend piece we fit to a short interval when computing values with which to initialise the model (since such pieces are often used to "fit" outliers).

Here is a prototypical example of the sort of data which creates problems and the change in behaviour.
**before**
Screenshot 2022-03-15 at 19 37 47
**after**
Screenshot 2022-03-15 at 19 37 29

In this case the spike creating problems is around 1500X larger than typical.
Screenshot 2022-03-15 at 19 38 19

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -42,6 +42,8 @@
 * Reduce worst case bucket processing time for anomaly detection. (See {ml-pull}2225[#2225].)
 * Improve handling of low cardinality features for training classification
   and regression models. (See {ml-pull}2229[#2229].)
+* Improve handling of extremely large outliers in time series modelling.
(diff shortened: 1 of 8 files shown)


droberts195 (approved):
LGTM

PR #2232, opened 2022-03-17
[ML] Allow on-the-fly adjustment of the number of threads in use

tveasey:
For PyTorch model inference we want to be able to control the number of threads we assign to parallel calls to forward on-the-fly to adjust the number of cores each model consumes on a node as cluster wide tasks are added or removed. This lays the groundwork by adding a dynamic setting for the number of threads the pool will use. The idea is we always start the process thread pool with the hardware concurrency (divided by the number of threads we give libtorch), but then adjust the threads actually used by control message. Unused threads will simply be idle waiting to pop an empty queue so are effectively free.

cc @dimitris-athanasiou.

### DIFF
--- a/lib/core/CStaticThreadPool.cc
+++ b/lib/core/CStaticThreadPool.cc
@@ -42,7 +42,7 @@ CStaticThreadPool::~CStaticThreadPool() {
 }
 
 void CStaticThreadPool::numberThreadsInUse(std::size_t threads) {
-    threads = std::max(std::min(threads, m_Pool.size()), 1UL);
+    threads = std::max(std::min(threads, m_Pool.size()), std::size_t{1});
     m_NumberThreadsInUse.store(threads);
 }
 
--- a/lib/core/CStaticThreadPool.cc
+++ b/lib/core/CStaticThreadPool.cc
@@ -25,6 +25,7 @@ std::size_t computeSize(std::size_t hint) {
 
 CStaticThreadPool::CStaticThreadPool(std::size_t size)
     : m_Busy{false}, m_Cursor{0}, m_TaskQueues{computeSize(size)} {
+    m_NumberThreadsInUse.store(m_TaskQueues.size());
     m_Pool.reserve(m_TaskQueues.size());
     for (std::size_t id = 0; id < m_TaskQueues.size(); ++id) {
         try {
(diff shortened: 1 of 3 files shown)


lib/core/CStaticThreadPool.cc
```diff
@@ -40,9 +41,14 @@ CStaticThreadPool::~CStaticThreadPool() {
     this->shutdown();
 }
 
+void CStaticThreadPool::numberThreadsInUse(std::size_t threads) {
+    threads = std::max(std::min(threads, m_Pool.size()), 1UL);
```
droberts195, line 45:
```suggestion
    threads = std::max(std::min(threads, m_Pool.size()), std::size_t{1});
```

I think `size_t` is `unsigned long long` on some platforms, so the `UL` suffix might not be portable.

lib/core/unittest/CStaticThreadPoolTest.cc
```diff
@@ -18,7 +18,9 @@
 
 #include <atomic>
 #include <chrono>
+#include <mutex>
 #include <thread>
+#include <unordered_set>
```
droberts195, line 23: We're still using `boost::unordered_set` everywhere else.  I think it's best we stick with one across the codebase.  Using `std` here will just encourage mixing the two.

tveasey: See [f047acd](https://github.com/elastic/ml-cpp/pull/2232/commits/f047acdece21675beec11357850a4b73d3defdd6).

lib/core/CStaticThreadPool.cc
tveasey: See [f047acd](https://github.com/elastic/ml-cpp/pull/2232/commits/f047acdece21675beec11357850a4b73d3defdd6).

droberts195 (approved):
LGTM

lib/core/unittest/CStaticThreadPoolTest.cc
```diff
@@ -202,7 +204,46 @@ BOOST_AUTO_TEST_CASE(testWithExceptions) {
...
+                executionThreads.insert(std::this_thread::get_id());
+            });
+        }
+
+        for (;;) {
+            std::scoped_lock<std::mutex> lock{mutex};
+            if (numberProcessedTasks == 200) {
+                LOG_DEBUG(<< "# threads used = " << executionThreads.size());
+                // A subset of threads can steal all the work.
+                BOOST_REQUIRE(numberThreadsInUse <= executionThreads.size());
```
dimitris-athanasiou, line 240: Shouldn't we be asserting that `executionThreads.size() <= numberThreadsInUse` instead?

tveasey: We should indeed!

dimitris-athanasiou (commented):
LGTM

PR #2233, opened 2022-03-18
[ML] Upgrade to PyTorch 1.11 on Windows

droberts195:
There are no changes to the build process since 1.9.

We still don't ship with MKL on Windows. The size of MKL proved
controversial on Linux, so it's best not to increase distribution
size even further by adding in the Windows version. (We can still
revisit this in the future if there's sufficient demand for NLP
in production on Windows.)


davidkyle (approved):
LGTM

PR #2234, opened 2022-03-21
[ML] Generate encoding without training

### LINKED ISSUE #2231: [ML] Split encoding and training stages
We need to separate the computation of MIC values and encodings from the training itself. This way we can perform the first step on the large dataset to ensure that encoding is correct, and the second step on the small dataset to ensure we can finish within a reasonable amount of time.

We can use the persist/restore mechanism to recover the state of the training process with the new dataframe.

valeriy42:
This PR enables the separation of encoding from training. This way we can perform (fast) encoding step on a large dataset, while subsequent (slow) training is performed on a smaller dataset.

Closes #2231


valeriy42:
retest

include/maths/analytics/CBoostedTreeImpl.h
```diff
@@ -237,14 +237,15 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeImpl final {
     //! Tag progress through initialization.
     enum EInitializationStage {
         E_NotInitialized = 0,
-        E_SoftTreeDepthLimitInitialized = 1,
```
tveasey, line 240: This will invalidate persisted state. One option would be to bump the state when we restore. Alternatively, we could increment the initialisation stage. Since this is still on a feature branch, I'm not sure this should be done now, but maybe we should capture this in an issue so it isn't forgotten.

lib/api/CDataFrameTrainBoostedTreeRunner.cc
```diff
@@ -436,6 +439,9 @@ void CDataFrameTrainBoostedTreeRunner::runImpl(core::CDataFrame& frame) {
     this->validate(frame, dependentVariableColumn);
 
     switch (m_Task) {
+    case E_Encode:
+        m_BoostedTree = m_BoostedTreeFactory->buildForEncode(frame, dependentVariableColumn);
+        break;
```
tveasey, line 444: We probably need separate code for memory estimation, but this is still outstanding for incremental training so not necessarily part of this change.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -38,6 +38,7 @@
 
 #include <rapidjson/prettywriter.h>
 
+#include <boost/make_shared.hpp>
```
tveasey, line 41: Why not `std::shared_ptr`?

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
```diff
@@ -2217,4 +2218,83 @@ BOOST_AUTO_TEST_CASE(testProgressMonitoringFromRestart) {
...
+    test::CDataFrameAnalyzerTrainingFactory::addPredictionTestData(
+        TLossFunctionType::E_MseRegression, fieldNames, fieldValues,
+        restoredAnalyzer, rowsTrain);
+    restoredAnalyzer.handleRecord(fieldNames, {"", "", "", "", "", "", "", "$"});
+
+    // reset logger for next tests
+    logger.reset();
+
+    // Check that no error messages were logged
+    BOOST_TEST_REQUIRE(logDataSPtr->rdbuf()->in_avail() == 0);
```
tveasey, line 2297: Should we perhaps assert something about the predictions, i.e. at least that we have reasonable predictions written?

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -139,11 +172,11 @@ CBoostedTreeFactory::buildForTrain(core::CDataFrame& frame, std::size_t dependen
 
     this->prepareDataFrameForTrain(frame);
 
-    m_TreeImpl->m_InitializationStage != CBoostedTreeImpl::E_NotInitialized
+    m_TreeImpl->m_InitializationStage != CBoostedTreeImpl::E_EncodingInitialized
         ? this->skipProgressMonitoringFeatureSelection()
         : this->startProgressMonitoringFeatureSelection();
 
-    skipIfAfter(CBoostedTreeImpl::E_NotInitialized,
+    skipIfAfter(CBoostedTreeImpl::E_EncodingInitialized,
```
tveasey, line 179: The checkpointing logic doesn't seem quite correct. You need to create a checkpoint for encoding by persisting `skipCheckpointIfAtOrAfter`. Also you should never be checking earlier checkpoints after later ones as in the following code  to `selectFeaturesAndEncodeCategories`, etc. Let's discuss this offline.

tveasey (commented):
I did a pass through. I think the logic around managing initialisation stages isn't quite correct, but probably easiest to discuss.

lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -1456,7 +1456,8 @@ BOOST_AUTO_TEST_CASE(testMseIncrementalAddNewTrees) {
     };
 
     core::CPackedBitVector holdoutRowMask(numberHoldoutRows, true);
-    holdoutRowMask.extend(false, batch2Size - numberHoldoutRows);
+    holdoutRowMask.extend(false, batch1Size + batch2Size - numberHoldoutRows);
+    BOOST_TEST_REQUIRE(batch2->numberRows() = holdoutRowMask.size());
```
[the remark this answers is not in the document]
valeriy42: :+1:

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: `CLogger::reconfigure` expects `boost::shared_ptr`. I wouldn't change the interfaces there so we won't unnecessarily diverge further from `main`.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
valeriy42: I test the predictions in `CBoostedTreeTest`. At the api level, I simply check if the process works successfully when triggered from the outside.

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
droberts195: We have to keep using `boost::shared_ptr` for passing arguments to Boost::Log, because Boost::Log still uses them.

tveasey: Thanks!

lib/api/unittest/CDataFrameAnalyzerTrainingTest.cc
tveasey: Ok, that seems reasonable.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -158,7 +178,8 @@ CBoostedTreeFactory::buildForTrain(core::CDataFrame& frame, std::size_t dependen
 
     this->startProgressMonitoringInitializeHyperparameters(frame);
 
-    if (this->initializeFeatureSampleDistribution()) {
+    if (m_TreeImpl->m_InitializationStage == CBoostedTreeImpl::E_EncodingInitialized ||
+        this->initializeFeatureSampleDistribution()) {
```
tveasey, line 182: On reflection, I don't think you can have this test here because we can failover at a later stage as part of `this->initializeHyperparameters(frame)`
```suggestion
    if (this->initializeFeatureSampleDistribution()) {
```

tveasey: Or perhaps better inside `CBoostedTreeFactory::initializeFeatureSampleDistribution()`:
```
    if (m_TreeImpl->m_FeatureSampleProbabilities.size() > 0) {
        return true;
    }
```

tveasey (approved):
LGTM

valeriy42:
retest

PR #2235, opened 2022-03-21
[ML] Upgrade to PyTorch 1.11 on macOS

droberts195:
It's now possible to use the MKLDNN library on macOS on ARM.

There are no other changes to the build process since 1.9.

I've added a comment to say we should upgrade the version of
Python we use to build PyTorch before the next upgrade, as
this seems to be the root cause of the difficulties with
correct architecture detection on ARM.


dimitris-athanasiou (approved):
LGTM

davidkyle (approved):
LGTM

PR #2238, opened 2022-03-22
[ML] Upgrade to PyTorch 1.11 on Linux

### LINKED ISSUE #2126: [ML] Change the list of MKL libraries we redistribute
In our build setup instructions we specify the MKL libraries that we will redistribute, for example on Linux: https://github.com/elastic/ml-cpp/blob/e262592493c6b4433319bed523e8085c0601a0ab/build-setup/linux.md#L280-L285

Based on the information in the "Computational Layer" section of the table in http://portal.nacad.ufrj.br/online/intel/mkl/common/mkl_userguide/GUID-C823F752-DDA3-4EFB-B673-222C2720FAFA.htm, the subset of libraries we are shipping is not quite correct.

We claim that our code will run on processors that support at minimum SSE4.2 instructions.  This means that we should be shipping `libmkl_mc3.so` in addition to the libraries we are currently shipping.  Additionally, since we _won't_ run on processors that support a maximum of SSSE3 nor Hi-k Core 2 processors (from 2009), we _don't_ need to ship `libmkl_vml_mc.so` nor `libmkl_vml_mc2.so`.

We should correct this next time we upgrade PyTorch and have to revisit the PyTorch dependencies.

droberts195:
Together with #2233 and #2235 this completes the upgrade to
PyTorch 1.11 for all platforms.

A small change to the way things are done is that we now copy
all MKL libraries into the Docker images, then copy the required
subset into our release distribution. This means we can change
the subset of MKL libraries that we redistribute without having
to rebuild our Docker images.

Fixes #2126

### DIFF
--- a/3rd_party/3rd_party.sh
+++ b/3rd_party/3rd_party.sh
@@ -70,7 +70,7 @@ case `uname` in
                 MKL_LOCATION=/usr/local/gcc103/lib
                 MKL_EXTENSION=.so
                 MKL_PREFIX=libmkl_
-                MKL_LIBRARIES=`cd "$MKL_LOCATION" && ls $MKL_PREFIX*$MKL_EXTENSION`
+                MKL_LIBRARIES='avx avx2 avx512 avx512_mic core def gnu_thread intel_lp64 mc3 vml_avx vml_avx2 vml_avx512 vml_avx512_mic vml_cmpt vml_def vml_mc3'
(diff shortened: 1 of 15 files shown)


dimitris-athanasiou (approved):
LGTM

davidkyle (approved):
LGTM

PR #2239, opened 2022-03-22
[8.1] [ML] Fix the PyTorch version reported in the dependency info

droberts195:
We upgraded to PyTorch 1.9 before 8.0.0 was released, but the
dependency info was still saying 1.8.

Backport of #2237


droberts195:
The build failure was because there was an Elasticsearch version bump during the build.  Since this is such a simple change there's no point re-running the tests.

PR #2245, opened 2022-03-29
[ML] Boosted tree hyperparameter refactor

tveasey:
This migrates the refactoring we did for hyperparameters onto the main branch from the incremental-learning feature branch. The main idea is better separation of concerns managing optimisation of hyperparameter optimisation and the rest of training.

This is a useful precursor to stopping after coarse parameter tuning and I would like to have that available on main sooner than we will have incremental training. It should reduce the overhead and scope for introducing bugs maintaining both branches.


valeriy42 (approved):
LGTM

PR #2247, opened 2022-03-30
[ML] Fix error message for failure to create reverse search

droberts195:
The reverse search for categorization is limited to a certain
length (currently 10000 characters).

It's possible that we cannot create a reverse search for a
category if the cost of including any token in that reverse
search would be higher than this maximum cost.

This PR corrects that logic for logging that edge case.
Previously there was a chance we'd erroneously log an error
reporting an inconsistency between two variables in the code.
Now we should correctly always log a message reporting the
actual problem. Additionally, the message that reports the
problem is downgraded from error to warning to fit better
with the overall Elasticsearch log level policy.


droberts195:
retest

PR #2248, opened 2022-03-30
[ML] Don't use moved from variable

tveasey:
#2245 introduced a using variable after moving it mistake. This corrects that. Since it hasn't been released, I've marked as a non-issue.


droberts195 (approved):
LGTM

tveasey:
retest

PR #2249, opened 2022-03-30
[ML] Numerical robustness improvements for Bayesian Optimisation

tveasey:
This improves handling of numerical issues computing EI.

In particular, we weren't checking the condition of kernel covariance matrix. We don't need a lot of accuracy, but we need to handle edge cases, like singular matrices, NaNs, etc. In this case, zeroing the EI and its gradient is fine, the solver will fallback to taking random step.

Note that this is also a potential source of `[CTools.cc@152] x = NaN, distribution = class...` log errors.


docs/CHANGELOG.asciidoc
```diff
@@ -58,6 +58,8 @@
 
 * Fix possible source of "Discarding sample = -nan(ind), weight = 1, variance scale = 1"
   log errors training classification and regression models. (See {ml-pull}2226[#2226].)
+* Fix possible source of "x = NaN, distribution = class boost::math::normal_distribution<..."
+  log errors training classification and regression models. (See {ml-pull}2249[#2249].)
```
droberts195, line 62: Since Easter is coming up and this still isn't merged it's likely to miss 8.2.0.  If it does slip to a subsequent release then please remember to move the release note.

valeriy42 (approved):
LGTM

PR #2251, opened 2022-03-31
[ML] Correct restart from failure when fine tuning classification and regression models

tveasey:
Previously, we would enter fine tuning loop even if we exited early in the event of restart from failover. This adds state to record early exit so we pick up correctly. Also, to monitor progress correctly we need to always start progress monitoring for final train, independent of whether we're training using a hold set or cross-validation.


valeriy42 (approved):
LGTM. Good catch.

PR #2252, opened 2022-04-01
[ML] Compress quantiles for persistence

droberts195:
Historically we have always persisted the quantiles state
uncompressed, unlike the model state which is compressed.

Originally the quantiles state was small and this didn't
matter. However, since we started storing per-partition
scores in version 6.5 the quantiles state can get pretty
big, and we can significantly reduce the memory usage of
the JVM that has to handle the quantiles documents by
compressing the quantiles state.

This PR does exactly that: the quantiles state is now
compressed during persistence. No Java-side changes are
needed, as we can easily tell during restoration whether
we're restoring compressed or uncompressed data.

The slight downside is that if jobs move nodes in a mixed
version cluster from a newer node to an older node then
the new quantiles will be unreadable on the older node.
This is a pretty unlikely scenario, as we expect each node
to get restarted once in a rolling upgrade, so jobs should
not need to move away from a newer node unless there's an
unexpected failure in addition to the controlled restarts.


droberts195:
retest

edsavage (approved):
LGTM

PR #2256, opened 2022-04-25
[ML] Improvements to the outlier influence calculation

tveasey:
Currently, we compute influence of features on the outlier score by searching for the minimum value of the outlier score allowing one coordinate to vary at a time. This works well if the data aren't projected, but when the data are high dimensional we randomly project them. We then have to work out how changes in projected coordinates should be attributed to the original features. This is only possible approximately and can lead to confusing results.

The normal definition of influence is the (Gateaux) derivative of the outlier score with respect to the feature values. This has the advantage that we can simply compute the change in the projection space for small changes in a single feature value. This changes switches to use this definition. For the unprojected case the results are broadly similar: the unit tests pass with minor tweaks. However, this fixes confusing results in the other case.


lib/maths/analytics/COutliers.cc
```diff
@@ -600,79 +601,68 @@ void CEnsemble<POINT>::CScorer::add(const TMeanVarAccumulator2Vec& logScoreMomen
...
+            result.interpolate(LOG_KNOTS, KNOTS_P_OUTLIER,
+                               common::CSplineTypes::E_ParabolicRunout);
+            return result;
+        }()};
+
+        return [&](double cdfComplement) {
+            double logCdfComplement{common::CTools::fastLog(
+                std::max(cdfComplement, std::numeric_limits<double>::min()))};
+            return P_OUTLIER.value(logCdfComplement);
+        };
```
tveasey, line 630: Note that this represents a small change to how we compute outlier scores for extreme outliers. Before things would saturate which means we couldn't reliably compute the derivative of the overall score w.r.t. changes in feature values. This extends the range to `std::numeric_limits<double>::min()`. It also avoids making the function as flat around P = 0.5 for the same reason. The resulting scores are still broadly similar.

valeriy42 (approved):
LGTM

PR #2257, opened 2022-04-27
[ML] Fix AnovaTotalCoefficientOfVariation formula for Bayesian optimization

valeriy42:
This PR fixes the formula for scaling the ANOVA coefficient of variation for the Gaussian process. It also adds a unit test to ensure that the computed coefficient is similar to the empirical for different sampling intervals and scales.


tveasey (approved):
Good catch. LGTM.

valeriy42:
The fix is superseded by #2259

PR #2258, opened 2022-04-27
[ML][PYTORCH] Add control message parsing to update number of model threads

davidkyle:
Allows dynamically updating the number of allocations the process will work with in parallel.

The request message is of the form:
```
{"request_id": "1", "control": 0, "num_allocations": 4}
```

The output is confirmation of the change

```
{"thread_settings": {"request_id": "1", "num_threads_per_allocation": 1, "num_allocations": 4}}
```


bin/pytorch_inference/CThreadSettings.cc
```diff
@@ -19,9 +18,25 @@
...
+
+std::int32_t CThreadSettings::numAllocations() const {
+    return m_NumAllocations;
+}
+
+void CThreadSettings::numAllocations(std::int32_t numAllocations) {
+    m_NumAllocations = numAllocations;
+}
+
+void CThreadSettings::validateThreadingParameters(std::int32_t maxThreads,
```
davidkyle, line 37: This logic needs reviewing

bin/pytorch_inference/Main.cc
```diff
@@ -120,7 +120,7 @@ void writeError(const std::string& requestId,
                 const std::string& message,
                 ml::core::CRapidJsonConcurrentLineWriter& jsonWriter) {
     jsonWriter.StartObject();
-    jsonWriter.Key(RESULT);
+    jsonWriter.Key(ERROR);
```
davidkyle, line 123: Now errors are written as errors not a special type of results with an error message. On the Java parsing side this change makes the code clearer.

docs/CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@
 * Upgrade PyTorch to version 1.11. (See {ml-pull}2233[#2233], {ml-pull}2235[#2235]
   and {ml-pull}2238[#2238].)
 * Upgrade zlib to version 1.2.12 on Windows. (See {ml-pull}2253[#2253].)
+* Add control message to dynamically change the number of model allocations. (See {ml-pull}2258[#2258].)
```
droberts195, line 38: We should think carefully about how we want this to look in the release notes. We don't want both the C++ and Java PRs to appear as separate items. If the Java PR is going to make it clearer to the user how they would use it then maybe just that one should appear and this PR should be `>non-issue`. Or otherwise the Java PR should be a `>non-issue`.

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -81,60 +85,104 @@ bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler,
...
+    return validateInferenceRequestJson(doc, errorHandler);
+}
+
+CCommandParser::EMessageType
+CCommandParser::validateControlMessageJson(const rapidjson::Document& doc,
+                                           const TErrorHandlerFunc& errorHandler) const {
+
+    const rapidjson::Value& control = doc[CONTROL];
+    if (control.IsInt() == false ||
+        control.GetInt() >= SControlMessage::EControlMessageType::E_Unknown) {
```
droberts195, line 135: Negative numbers should also be rejected here.

droberts195 (commented):
Looks good apart from some small nits.

My main comment is that to get this integrated without breaking CI for everyone 3 PRs will be needed:

1. A Java PR that mutes every Java test that uses this process (there's a list in https://gradle-enterprise.elastic.co/s/3dwvcwrhuryhk/tests/overview?outcome=failed)
2. This PR
3. A Java PR that unmutes the tests and also adds the Java side of the change

In between 2 and 3 you'll need to wait for the corresponding ml-cpp builds (both snapshot and non-snapshot) to get uploaded to S3 so that elasticsearch-ci can see them.

droberts195:
> I think the CI mute dance is worth it in this case so we can consistently adapt to the new names

Yes, agreed.  This PR won't pass CI until those mutes are merged to the ES repo.

Also, it seems that there are some formatting violations in this PR after the latest changes.

bin/pytorch_inference/CThreadSettings.cc
```diff
@@ -0,0 +1,75 @@
...
+        // leave one thread for inference
+        LOG_WARN(<< "Setting number of allocations to maximum value of "
+                 << std::max(1, maxThreads - 1) << "; value was " << numAllocations);
+        numAllocations = std::max(1, maxThreads - 1);
+    }
+
+    if (numAllocations + numThreadsPerAllocation > maxThreads) {
+        std::int32_t oldInferenceThreadCount{numThreadsPerAllocation};
+        numThreadsPerAllocation = std::max(1, maxThreads - numAllocations);
+        LOG_WARN(<< "Sum of allocation cound [" << numAllocations << "] and inference threads ["
```
droberts195, line 68: `cound` is a typo here

bin/pytorch_inference/CThreadSettings.cc
droberts195: It's less important what the exact logic is if the Java side is going to protect us from exceeding these limits in the normal case, and these checks are only for the unfortunate case of ML nodes being lost.

docs/CHANGELOG.asciidoc
davidkyle: I've changed my mind on this because it is an internal change to enable another feature. In fact even the Java side of this might not be release noted as that is also part of the change to improve model allocations

bin/pytorch_inference/CThreadSettings.cc
davidkyle: @dimitris-athanasiou what do you think? Should we delete this code now or keep a simplified version of the logic that rejects thread settings > hardware concurrency

bin/pytorch_inference/CCommandParser.cc
```diff
@@ -81,60 +85,104 @@ bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler,
...
         }
     }
 
     return true;
 }
 
-bool CCommandParser::validateJson(const rapidjson::Document& doc,
-                                  const TErrorHandlerFunc& errorHandler) const {
+CCommandParser::EMessageType
+CCommandParser::validateJson(const rapidjson::Document& doc,
```
tveasey, line 110: I suspect that all these methods can be static. Probably slightly cleaner if they don't use any state.

bin/pytorch_inference/Main.cc
```diff
@@ -355,14 +370,18 @@ int main(int argc, char** argv) {
 
     // Starting the executor with 1 thread will use an extra thread that isn't necessary
     // so we only start it when more than 1 threads are set.
-    if (modelThreads > 1) {
-        ml::core::startDefaultAsyncExecutor(modelThreads);
+    if (numAllocations > 1) {
+        ml::core::startDefaultAsyncExecutor(numAllocations);
```
tveasey, line 374: We should probably be able to
```
else {
    ml::core::stopDefaultAsyncExecutor();
}
```

tveasey (approved):
LGTM. Just a couple of minor suggestions.

PR #2259, opened 2022-04-27
[ML] Some corrections to ANOVA for Bayesian Optimisation

tveasey:
This corrects the handling of the shifting and scaling of the function values when computing various quantities related to the GP. In particular, we were not properly undoing the transform in `evaluate`, `evaluate1D` and various ANOVA related functionality. We could end up computing square roots of negative numbers in `evaluate1D` and `anovaTotalVariance`. At the same time I took the opportunity to ensure GP domain is [0, 1]^n rather than just scaling. This leads to more predictable numerics for ML and also simplifies the code.

I've added tests that the kernel parameters are invariant under operations to the function range which should preserve its value: changing level and scale. I also do the same for `evaluate`, `evaluate1D` and various ANOVA related functionality.

### DIFF
--- a/lib/maths/common/CBayesianOptimisation.cc
+++ b/lib/maths/common/CBayesianOptimisation.cc
@@ -375,11 +375,11 @@ double CBayesianOptimisation::anovaMainEffect(const TVector& Kinvf, int dimensio
                 integrate1dKernel(m_KernelParameters(dimension + 1), xi(dimension)) *
                 prodXt(xi, dimension);
     }
-    double shift2{CTools::pow2(m_RangeShift)};
+    double scale2{CTools::pow2(m_RangeScale)};
     double theta02{CTools::pow2(m_KernelParameters(0))};
     double f0{this->anovaConstantFactor()};
     double f02{CTools::pow2(f0)};
-    return (theta02 * (theta02 * sum1 - 2.0 * f0 * sum2) + f02) / shift2;
+    return (theta02 * (theta02 * sum1 - 2.0 * f0 * sum2) + f02) / scale2;
 }
 
 double CBayesianOptimisation::anovaMainEffect(int dimension) const {
--- a/lib/maths/common/CBayesianOptimisation.cc
+++ b/lib/maths/common/CBayesianOptimisation.cc
@@ -123,9 +123,8 @@ CBayesianOptimisation::CBayesianOptimisation(core::CStateRestoreTraverser& trave
 }
 
 void CBayesianOptimisation::add(TVector x, double fx, double vx) {
-    m_FunctionMeanValues.emplace_back(
-        (x - m_MinBoundary).cwiseQuotient(m_MaxBoundary - m_MinBoundary),
-        toScaled(m_RangeShift, m_RangeScale, fx));
+    m_FunctionMeanValues.emplace_back(this->to01(std::move(x)),
+                                      toScaled(m_RangeShift, m_RangeScale, fx));
     m_ErrorVariances.push_back(CTools::pow2(m_RangeScale) * vx);
 }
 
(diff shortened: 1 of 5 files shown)


lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -61,6 +63,16 @@ const std::string RNG_TAG{"rng"};
 // narrow deep valley that the Gaussian Process hasn't sampled.
 const double MINIMUM_KERNEL_SCALE_FOR_EXPECTATION_MAXIMISATION{1e-8};
 
+//! Affine transform \p scale * (\p fx - \p shift).
+double toScaled(double shift, double scale, double fx) {
+    return scale * (fx - shift);
+}
+
+//! Affine transform \p shift + \p scale * \p fx.
```
valeriy42, line 71:
```suggestion
//! Affine transform \p shift +  \p fx / \p scale.
```

lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -114,9 +126,32 @@ CBayesianOptimisation::CBayesianOptimisation(core::CStateRestoreTraverser& trave
...
+    if (CMathsFuncs::isFinite(fx) == false || CMathsFuncs::isFinite(vx) == false) {
+        LOG_ERROR(<< "Discarding point (" << x.transpose() << "," << fx << "," << vx << ")");
+        return;
+    }
+
+    x = this->to01(std::move(x));
+    fx = toScaled(m_RangeShift, m_RangeScale, fx);
+    vx = CTools::pow2(m_RangeScale) * vx;
+
+    std::size_t duplicate(std::find_if(m_FunctionMeanValues.begin(),
```
valeriy42, line 138: This is interesting. Why do you expect there to be duplicates?

lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -250,20 +284,26 @@ double CBayesianOptimisation::evaluate1D(const TVector& Kinvf, double input, int
...
-
-    // We rewrite theta_0^2 sum - f_0 as (theta + f) * (theta - f) where
-    // theta = theta_0 sum^(1/2) and f = f_0^(1/2) because it has better
-    // numerics.
-    double theta{m_KernelParameters(0) * std::sqrt(sum)};
-    double f{std::sqrt(this->anovaConstantFactor(Kinvf))};
-    return (theta + f) * (theta - f);
+    double f2{this->anovaConstantFactor(Kinvf)};
+
+    if (std::signbit(sum) == std::signbit(f2)) {
```
valeriy42, line 296: Can you please add a comment, what otherwise can go wrong here?

lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -399,38 +445,50 @@ CBayesianOptimisation::minusLikelihoodAndGradient() const {
...
+    // We need to be careful when we compute the kernel decomposition. Basically,
+    // if the kernel matrix is singular to working precision then if the function
+    // value vector projection onto the null-space has non-zero length the likelihood
+    // function is effectively -infinity. This follow from the fact that although
+    // log(1 / lambda_i) -> +infinity, -1/2 sum_i{ ||f_i||^2 / lambda_i } -> -infinity
+    // faster for all ||f_i|| > 0 and lambda_i sufficiently small. Here {lambda_i}
+    // denote the Eigenvalues of the nullspace. We use a rank revealing decomposition
+    // and compute the likelihood on the complement of the row space and add on a
+    // large penalty to discourage choosing kernel parameters for which the kernel
+    // is singular.
```
valeriy42: Nice catch!

valeriy42 (approved):
Good work on catching the numerical issue and writing unit tests to verify the correct behaviour! :rocket:

lib/maths/common/CBayesianOptimisation.cc
tveasey: I don't in current usage but I think this code should handle them, i.e. we shouldn't be fragile, and there is a natural way to handle this.

lib/maths/common/CBayesianOptimisation.cc
tveasey: See [726ea78](https://github.com/elastic/ml-cpp/pull/2259/commits/726ea7879d39c20ab374e5b5dda35525747d39b0).

lib/maths/common/CBayesianOptimisation.cc
tveasey: See [726ea78](https://github.com/elastic/ml-cpp/pull/2259/commits/726ea7879d39c20ab374e5b5dda35525747d39b0).

PR #2260, opened 2022-05-02
[ML] Coarse parameter tuning early stopping

### LINKED ISSUE #2240: [ML] Early stopping for coarse parameter tuning
To deal with large datasets effectively, we need to accelerate the coarse parameter tuning. To this end, we can use GP with an early stopping criterion similar to what we do for fine parameter tuning. Furthermore, if coarse parameter tuning shows no significant dependency of results from the hyperparameter values, we can skip the fine parameter tuning (for this hyperparameter) altogether.

Since this performance improvement is mainly intended to accelerate training on large datasets, we can only use it for specified holdout datasets. This limitation allows avoiding the complexity of dealing with cross-fold variation and the requirement that the stopping criterion would have to work similarly well for large and small datasets.

valeriy42:
This PR enables the early stopping of hyperparameter tuning after the coarse parameter tuning. This should improve training time at the cost of potentially slightly lower accuracy.

Closes #2240


include/maths/analytics/CBoostedTreeHyperparameters.h
```diff
@@ -65,6 +65,10 @@ class CBoostedTreeParameter final {
         : m_Value{value}, m_MinValue{value}, m_MaxValue{value}, m_LogSearch{logSearch == E_LogSearch} {
     }
 
+    CBoostedTreeParameter(const CBoostedTreeParameter<T>& other) = default;
+    CBoostedTreeParameter& operator=(const CBoostedTreeParameter<T>& other) = default;
+    ~CBoostedTreeParameter() = default;
```
tveasey, line 70: Was there a reason to add these?

lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -46,6 +46,32 @@ CBoostedTreeHyperparameters::CBoostedTreeHyperparameters() {
     this->initializeTunableHyperparameters();
 }
 
+CBoostedTreeHyperparameters::CBoostedTreeHyperparameters(const CBoostedTreeHyperparameters& other) {
+    *this = other;
+}
+
+CBoostedTreeHyperparameters& CBoostedTreeHyperparameters::
+operator=(const CBoostedTreeHyperparameters& other) {
```
tveasey, line 54: I'm not keen on having actual copying only copying part of the object. I think it would be better to have a named method to make it clear this isn't a real copy. Maybe something like:
```
void copyCurrentValues(const CBoostedTreeHyperparameters& other);
```

The same comment applies to the copy constructor.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -1710,6 +1756,7 @@ bool CBoostedTreeFactory::skipCheckpointIfAtOrAfter(int stage, const F& f) {
         f();
         m_TreeImpl->m_InitializationStage =
             static_cast<CBoostedTreeImpl::EInitializationStage>(stage);
+        // LOG_DEBUG(<<"Checkpoints stage " << stage);
```
tveasey, line 1759: Remove

include/maths/analytics/CBoostedTreeHyperparameters.h
```diff
@@ -749,8 +791,8 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeHyperparameters {
 
     //@ \name Hyperparameter Optimisation
     //@{
-    bool m_StopHyperparameterOptimizationEarly{true};
-    bool m_StoppedHyperparameterOptimizationEarly{false};
+    bool m_EarlyHyperparameterOptimizationStoppingEnabled{true};
+    mutable bool m_StopHyperparameterOptimizationEarly{false};
```
tveasey, line 795: This name seems worse to me. This tells you if we did in fact stop hyperparameter optimisation early. `m_StopHyperparameterOptimizationEarly` is an imperative so suggests we're telling the class to "stop hyperparameter optimization early". If you don't like `m_StoppedHyperparameterOptimizationEarly` maybe `m_DidStopHyperparameterOptimizationEarly` or some such.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -119,6 +119,15 @@ auto validInputStream(core::CDataSearcher& restoreSearcher) {
...
 }
+
+double minBoundary(const maths::analytics::CBoostedTreeHyperparameters::TDoubleParameter parameter,
+                   double maxBoundary,
+                   double interval) {
+    maxBoundary = parameter.toSearchValue(maxBoundary);
+    interval = parameter.toSearchValue(interval);
+    double minBoundary{maxBoundary - interval};
+    return parameter.fromSearchValue(minBoundary);
+}
```
tveasey, line 130: My inclination is to make this generic even if you only use for TDoubleParameter at present:
```suggestion
template<typename T>
double minBoundary(const CBoostedTreeParameter<T>& parameter,
                   T maxBoundary,
                   T interval) {
    maxBoundary = parameter.toSearchValue(maxBoundary);
    interval = parameter.toSearchValue(interval);
    T minBoundary{maxBoundary - interval};
    return parameter.fromSearchValue(minBoundary);
}
```

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -187,6 +196,19 @@ CBoostedTreeFactory::buildForTrain(core::CDataFrame& frame, std::size_t dependen
...
+                auto parameters = hyperparameterLoss.first.selectParametersVector(
+                    m_TreeImpl->m_Hyperparameters.tunableHyperparameters());
+                m_TreeImpl->m_Hyperparameters.addObservation(
+                    parameters, hyperparameterLoss.second, 0.0, true);
+            }
+            if (m_TreeImpl->m_Hyperparameters.stopEarly() == false) {
+                m_TreeImpl->m_Hyperparameters.resetSearch();
+            }
+            m_TreeImpl->m_Hyperparameters.resetBayesianOptimization();
+        }
```
tveasey, line 211: This just seems better placed in `CBoostedTreeHyperparameters::initializeSearch()`. This doesn't really seem part of the factory responsibilities. You then also don't need to add anything to `CBoostedTreeHyperparameters` public interface related to this change.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -756,6 +780,21 @@ void CBoostedTreeFactory::initializeUnsetRegularizationHyperparameters(core::CDa
                   << core::CContainerPrinter::print(gainAndTotalCurvaturePerNode));
     });
 
+    // initialize unset regularization hyperparameters with meaningful values
+    if (hyperparameters.earlyStoppingEnabled()) {
```
tveasey, line 784: I'm not sure about this being conditional on enabling early stopping. It seems weird to me to change this behaviour based on this flag. My feeling is it is probably better to use the true minimum we consider rather than zero anyway.

include/maths/analytics/CBoostedTreeFactory.h
```diff
@@ -387,6 +394,7 @@ class MATHS_ANALYTICS_EXPORT CBoostedTreeFactory final {
     TBoostedTreeImplUPtr m_TreeImpl;
     mutable std::size_t m_PaddedExtraColumns{0};
     TTrainingStateCallback m_RecordTrainingState{noopRecordTrainingState};
+    THyperparametersDoublePrVecSPtr m_HyperparametersLosses;
```
tveasey, line 397: This feels misplaced to me here. They belong in `CBoostedTreeHyperparameters` IMO since it is responsible for maintaining all other state related to hyperparameter optimisation.

lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -1440,10 +1442,11 @@ BOOST_AUTO_TEST_CASE(testMseIncrementalForOutOfDomain) {
 
     LOG_DEBUG(<< "increase on old = " << errorIncreaseOnOld);
     LOG_DEBUG(<< "decrease on new = " << errorDecreaseOnNew);
-    BOOST_TEST_REQUIRE(errorDecreaseOnNew > 100.0 * errorIncreaseOnOld);
+    BOOST_TEST_REQUIRE(errorDecreaseOnNew > 57.0 * errorIncreaseOnOld);
 }
 
-BOOST_AUTO_TEST_CASE(testMseIncrementalAddNewTrees) {
+// TODO #2271 Fix flacky test and re-enabled
```
tveasey, line 1448: Needs resolving before commit.

tveasey (commented):
I did a pass through. My main comment is it feels like the implementation got split between `CBoostedTreeFactory` and `CBoostedTreeHyperparameters`. I think it would be cleaner inside `CBoostedTreeHyperparameters`. Another thing I notice is this has had a slightly larger impact on the unit tests than I would expect. I'd like to spend a little time digging into these changes as well, before it gets merged.

include/maths/analytics/CBoostedTreeHyperparameters.h
valeriy42: You are right, these default definitions are not necessary here.

lib/maths/analytics/CBoostedTreeFactory.cc
valeriy42: I want this logic to be executed for initial training and not for incremental training (since parameter tuning for incremental training is negligible). By moving it to `CBoostedTreeHyperparameters::initializeSearch()` IMO it becomes a side-effect, hidden in the implementation. I can see your point since the logic operators on `m_Hyperparameters` exclusively, and so it would reduce coupling to refactor the logic into `CBoostedTreeHyperparameters`.

include/maths/analytics/CBoostedTreeHyperparameters.h
valeriy42: Let's discuss the appropriate names offline.

lib/maths/analytics/CBoostedTreeHyperparameters.cc
valeriy42: Alright, I introduced the private method `copyRegularizationParameters()` and removed the public constructor and assignment operators.

include/maths/analytics/CBoostedTreeFactory.h
valeriy42: Thank you for the suggestion. I moved the member variable into `CBoostedTreeHyperparameters` :+1:

tveasey (approved):

lib/maths/analytics/CBoostedTreeFactory.cc
tveasey: Thanks!

lib/maths/analytics/unittest/CBoostedTreeTest.cc
tveasey: This was handled in a separate change.

lib/maths/analytics/CBoostedTreeFactory.cc
tveasey: This turned out to be due to handling of max number trees.

PR #2261, opened 2022-05-03
[ML] Correct initialisation of the trend model after detecting seasonality for edge case

tveasey:
There are various reasons we can decide not to remove the seasonal component from a portion of the window we test when we think the seasonality is piecewise linearly scaled. In this case the residual values there are unreliable and should be ignored when reinitialising the trend model. This issue could lead to us to very significantly over estimate the prediction errors and inflate the model bounds after detecting new seasonal components in certain edge cases.


droberts195 (approved):
LGTM

As this affects results, to leave time for due diligence by QA, please _don't_ backport to 8.2 or 7.17 branches until after the 8.2.1 and 7.17.4 releases, and please adjust the release notes to say this is in 8.2.2 and 7.17.5.

PR #2264, opened 2022-05-04
[ML] Add support for the CMake build system

edsavage:
Add the ability to build the `ml-cpp` repository using the `CMake` build system. This has both the advantage of being somewhat more modern (and hence accessible for future maintainers) but also being fully supported by IDEs such as `CLion`.

To build all targets manually:

```
. ./set_env.sh
cmake -B cmake_build $CMAKE_FLAGS
cmake --build cmake_build -j`nproc`
```

To build a particular target e.g. `install-distributed`:
```
cmake --build cmake_build -j`nproc` -t install-distributed
```

To obtain a list of all possible build targets
```
cmake --build cmake_build -j`nproc` -t help
```

Labelling as a WIP as there's still work to be done formalising the procedure to add `cmake` to the `Docker` build images and tidying the steps for the Windows build.

Labelling as a non-issue as these changes pertain solely to the build system.


droberts195 (commented):
Thanks for doing this.  It will be great if we can get properly switched over to a modern build system.

My high level comments are:

- It would be best if this replaced the existing make system instead of duplicating it.  Otherwise it will be painful and error prone trying to maintain both in the future.
- It seems like an enormous amount is lumped into the top level `CMakeLists.txt`.  Would it be possible for some of the library-specific stuff to be delegated to a file in the `lib` directory and some of the program-specific stuff to be delegated to a file in the `bin` directory?  Then maybe we could also preserve the ability to build `devbin` if we ever want to in the future.
- The commands in `build.gradle` should be switched to CMake.

I think this is going to need several iterations to get right.

set_env.sh
```diff
@@ -113,7 +123,7 @@ case $SIMPLE_PLATFORM in
         ;;
 
     windows)
-        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows
+        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows:$ROOT/PROGRA~1/Python37/Scripts/
```
droberts195, line 126: I am not sure the Windows VMs in elasticsearch-ci will have this version of Python, so if this is now required then it's an extra complication.  Previously we only needed Python to build PyTorch, which we don't do in CI.  Why do we need it now to build our own C++ code?

dev-tools/docker/linux_aarch64_cross_image/Dockerfile
```diff
@@ -30,8 +33,8 @@ RUN \
   ln -s usr/lib64 lib64
 
 # For compiling with hardening and optimisation
-ENV CFLAGS "-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2 -msse4.2 -mfpmath=sse"
-ENV CXXFLAGS "-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2 -msse4.2 -mfpmath=sse"
+ENV CFLAGS "-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2 -march=armv8-a+crc+crypto"
+ENV CXXFLAGS "-g -O3 -fstack-protector -D_FORTIFY_SOURCE=2 -march=armv8-a+crc+crypto"
```
droberts195, line 37: I think this is wrong.  In this image we're building tools that will run on `x86_64` in order to subsequently do cross compilation to `aarch64`.  So for the build of those tools that will run on `x86_64` we need options that will make the tools run fast on `x86_64`.

Arguably the options here for building `gcc` and `binutils` should be `CFLAGS_FOR_BUILD` and `CXXFLAGS_FOR_BUILD`.  It seems that `configure.sh` will copy `CFLAGS` to both `CFLAGS_FOR_BUILD` _and_ also `CFLAGS_FOR_TARGET` and `CXXFLAGS_FOR_TARGET` if those more specialised variables aren't specified.  But then `CFLAGS_FOR_TARGET` and `CXXFLAGS_FOR_TARGET` don't seem to actually affect the default options that the resulting cross compiler will apply when building for the target (or at least not the ones we're specifying here).

Arguably it would be more correct to specify `CFLAGS_FOR_BUILD` and `CXXFLAGS_FOR_BUILD` here, then replace them with `CFLAGS` and `CXXFLAGS` when building `patchelf` lower down (as that doesn't know about `CFLAGS_FOR_BUILD` and `CXXFLAGS_FOR_BUILD`).  But as far as I can see the way it's being done at the moment doesn't actually cause a problem.  We specify `-march=armv8-a+crc+crypto` in the compiler options when we use it to build our code for `aarch64`.

So I would just put this back how it was.

set_env.sh
edsavage: Weirdly, this is the install location of cmake on the windows system I've been testing on. I've no idea why or how it's ended up under a python scripts folder. I'll need to investigate whether its possible to install it somewhere more sensible.

set_env.sh
```diff
@@ -95,15 +116,15 @@ fi
...
+        PATH=/usr/local/gcc103/bin:/usr/bin:/bin:/usr/local/gcc103/sbin:/usr/sbin:/sbin:/usr/local/bin:/usr/local/cmake/bin
         ;;
 
     macos)
         PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin
         ;;
 
     windows)
-        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows
+        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows:$ROOT/PROGRA~2/MICROS~2/2019/PROFES~1/Common7/IDE/COMMON~1/MICROS~1/CMake/CMake/bin
```
droberts195, line 127: If CMake is installed according to the [build setup instructions](https://github.com/elastic/ml-cpp/blob/main/build-setup/windows.md#cmake) then it ends up being version 3.19 and goes in `C:\Program Files\CMake\bin`:

Screenshot 2022-05-16 at 14 06 39

If CMake is installed using Visual Studio then it goes in the directory you've specified here, and currently is version 3.20:

Screenshot 2022-05-16 at 14 05 56

We _could_ change the build instructions to tell people to install CMake with Visual Studio.  That might affect the scripts that build the CI workers as we might need to specify an extra option when doing the unattended Visual Studio install.  If we go this way we should also change the build setup instructions to tell people to always use that CMake on Windows.

However, on the other hand we may benefit in the future from having CMake decoupled from Visual Studio.  If we install our own then we can potentially upgrade CMake without having to upgrade Visual Studio.  I think having CMake decoupled is better _unless_ Microsoft has fixed something in their build of CMake that makes it work better with Visual Studio than vanilla CMake.  In that case we _should_ use Microsoft's build of CMake, accepting that Microsoft now dictates our CMake upgrade policy and should have comments to say that wherever we tell people to install CMake on other platforms, e.g. "On Windows we are constrained to use CMake 3.20, so no platform can use a CMake feature that was introduced after this."

CMakeLists.txt
```diff
@@ -0,0 +1,285 @@
...
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
+
+  message(STATUS "VCVER $ENV{VCVER}")
+  STRING(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" VCVER_REGEX_MATCH  $ENV{VCVER})
+  set(VCVER_MAJOR ${CMAKE_MATCH_1})
+  set(VCVER_MINOR ${CMAKE_MATCH_2})
+
+  string(CONCAT BOOST_VCVER "vc" ${VCVER_MAJOR} ${VCVER_MINOR} )
+  string(SUBSTRING ${BOOST_VCVER} 0 5 BOOST_VCVER)
```
droberts195, line 194: Why is the substring needed here?

If major is 14 and minor is 2 then surely the concat alone produces the right result.

If minor is not just 2 then I am wondering if there's some subtle future problem lurking here.

droberts195 (commented):
This seems to be making good progress but it looks like there are still quite a few tricky details to sort out.

CMakeLists.txt
edsavage: The minor version is 20 in this case. However the Boost naming scheme only appears to use the first digit of it - see [version-numbers-for-visual-studio-2017-boost-and-cmake ](https://stackoverflow.com/questions/42730478/version-numbers-for-visual-studio-2017-boost-and-cmake#:~:text=And%20here%20it%20seems%20that,talk%20about%20boost's%20compiler%20naming.)

set_env.sh
```diff
@@ -32,22 +32,37 @@ case `uname` in
...
+            export CMAKE_INSTALL_PREFIX="$CPP_SRC_HOME/build/distribution/platform/$BUNDLE_PLATFORM"
+            export CMAKE_TOOLCHAIN_FILE="tc-linux-aarch64.cmake"
+            export CMAKE_FLAGS="-DCMAKE_TOOLCHAIN_FILE=$CMAKE_TOOLCHAIN_FILE"
         fi
         ;;
 
     MINGW*)
         SIMPLE_PLATFORM=windows
         BUNDLE_PLATFORM=windows-x86_64
+        export CMAKE_GENERATOR='MinGW Makefiles'
```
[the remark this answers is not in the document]
edsavage: I chose to support MinGW Makefiles as that was what had gone before and would have a minimal impact on the build infrastructure, but it has proved a bit messy to implement in places (such as the issue with the LOCAL_DRIVE variable). I'll investigate supporting Visual Studio project files instead.

set_env.sh
```diff
@@ -76,13 +91,19 @@ unset COMPILER_PATH
...
     #     is followed by the 32 bit SDK bin directory
     COMPILER_PATH=$ROOT/$PFX86_DIR/$MSVC_DIR/2019/Professional/VC/Tools/MSVC/$VCVER/bin/Hostx64/x64:$ROOT/$PFX86_DIR/$MSVC_DIR/2019/Professional/Common7/IDE:$ROOT/$PFX86_DIR/$WIN_KITS_DIR/8.0/bin/x64:$ROOT/$PFX86_DIR/$WIN_KITS_DIR/8.0/bin/x86
+    export INCLUDE=$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/atlmfc/include:$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/include:$ROOT/$WINSDKBASE/10/Include/$UCRTVER/ucrt:$ROOT/$WINSDKBASE/10/Include/$UCRTVER/shared:$ROOT/$WINSDKBASE/10/Include/$UCRTVER/um:$ROOT/$WINSDKBASE/10/Include/$UCRTVER/winrt:$ROOT/$WINSDKBASE/10/Include/$UCRTVER/cppwinrt
+    export LIB=$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/atlmfc/lib/x64:$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/lib/x64:$ROOT/$WINSDKBASE/10/Lib/$UCRTVER/ucrt/x64:$ROOT/$WINSDKBASE/10/Lib/$UCRTVER/um/x64
+    export LIBPATH=$PATH:$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/atlmfc/lib/x64:$ROOT/$VCBASE/VC/Tools/MSVC/$VCVER/lib/x64:$ROOT/$VCBASE/VC/Tools/MSVC/VCVER/lib/x86/store/REFERE~1:$ROOT/$WINSDKBASE/10/UNIONM~1/$UCRTVER:$ROOT/$WINSDKBASE/10/REFERE~1/$UCRTVER:$ROOT/Windows/MICROS~1.NET/FRAMEW~2/V40~1.303
```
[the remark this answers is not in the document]
edsavage: Agreed. These may not even be needed if we move to Visual Studio project files instead.

set_env.sh
edsavage: Agreed, I think unless there's a very good reason not to then we should use our own build of CMake on all platforms.

set_env.sh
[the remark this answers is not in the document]
droberts195: I am not clear on why it makes a difference.  My naive assumption was that CMake would generate the next level of build system files and then that next level of build system files should be expected to just work.  Or is the problem that currently we have some Makefile fragments embedded in the CMake config that we're telling CMake to just pass through?  If so I guess the solution is to replace those fragments that are passed through directly with higher level CMake directives.

CMakeLists.txt
```diff
@@ -0,0 +1,285 @@
...
+# Set these here to avoid any confusion. See the documentation
+# in build-setup for more details
+if (WIN32)
+  set(CMAKE_MAKE_PROGRAM "gnumake.exe" CACHE STRING "" FORCE)
+  set(CMAKE_BUILD_TYPE "Release" CACHE STRING "" FORCE)
+  set(CMAKE_C_COMPILER "cl.exe" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_COMPILER "cl.exe" CACHE STRING "" FORCE)
+elseif(UNIX AND NOT APPLE)
+  set(CMAKE_C_COMPILER   "gcc" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_COMPILER "g++" CACHE STRING "" FORCE)
```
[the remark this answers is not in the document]
edsavage: CMake correctly determines the compiler to use on Mac and Windows targets (`clang` and `cl.exe` respectively). 

However for Linux targets the compiler defaults to `/usr/bin/cc`. We need to point CMake towards `/usr/local/gcc103/bin/gcc` instead. 

I'll update the comment to make this clearer.

dev-tools/docker/linux_aarch64_cross_image/Dockerfile
```diff
@@ -83,3 +83,12 @@ RUN \
   cd .. && \
   rm -rf patchelf-0.13.20210805.a949ff2
 
+# Install CMake
+# v3.5 minimum is required
```
droberts195, line 87: This comment is wrong: `CMakeLists.txt` says `cmake_minimum_required(VERSION 3.9)`.

Please also update the same comment in the other files where it appears.

set_env.sh
```diff
@@ -103,7 +103,7 @@ case $SIMPLE_PLATFORM in
         ;;
 
     windows)
-        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows
+        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows::$ROOT/PROGRA~1/CMake/bin
```
droberts195, line 106:
```suggestion
        PATH=/mingw64/bin:/usr/bin:/bin:$ROOT/Windows/System32:$ROOT/Windows:$ROOT/PROGRA~1/CMake/bin
```

cmake/compiler/msvc.cmake
```diff
@@ -0,0 +1,59 @@
...
+
+list(APPEND ML_SYSTEM_INCLUDE_DIRECTORIES ${CMAKE_SOURCE_DIR}/3rd_party/include ${ROOT}/usr/local/include ${VCINCLUDES} ${WINSDKINCLUDES})
+
+list(APPEND ML_COMPILE_DEFINITIONS
+	_CRT_SECURE_NO_WARNINGS
+	_CRT_NONSTDC_NO_DEPRECATE
+	WIN32_LEAN_AND_MEAN
+    NOMINMAX
+	NTDDI_VERSION=0x06010000
+	_WIN32_WINNT=0x0601 Windows)
```
droberts195, line 28: This is indented with a mixture of tabs and spaces.

Is there any hard requirement to use tabs in CMake files?  If not then I think it would be best to replace tabs with spaces in all the CMake build files.

cmake/compiler/vs2019.cmake
```diff
@@ -0,0 +1,36 @@
...
+  set(ROOT $ENV{ROOT})
+endif()
+  
+execute_process(COMMAND bash -c "cygpath -m -s \"${ROOT}/Program Files (x86)/Microsoft Visual Studio/2019/Professional\"" OUTPUT_VARIABLE VCBASE OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "cygpath -m -s \"${ROOT}/Program Files (x86)/Windows Kits\"" OUTPUT_VARIABLE WINSDKBASE OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "cygpath -m -s \"${ROOT}/Program Files (x86)\"" OUTPUT_VARIABLE PFX86_DIR OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "cd ${PFX86_DIR} && cygpath -m -s \"Microsoft Visual Studio\"" OUTPUT_VARIABLE MSVC_DIR OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "cd ${PFX86_DIR} && cygpath -m -s \"Windows Kits\"" OUTPUT_VARIABLE WIN_KITS_DIR OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "/bin/ls -1 ${PFX86_DIR}/${MSVC_DIR}/2019/Professional/VC/Tools/MSVC" COMMAND bash -c "tail -1" OUTPUT_VARIABLE VCVER OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND bash -c "/bin/ls -1 ${WINSDKBASE}/10/Include" COMMAND bash -c "tail -1" OUTPUT_VARIABLE UCRTVER OUTPUT_STRIP_TRAILING_WHITESPACE)
```
droberts195, line 29: One thing I am interested in is whether the build still needs to run in Git bash.  If CMake is a native program and is generating build files for Visual Studio which is also a native program, maybe the build can now run natively and avoid all this path translation stuff?

Maybe there's still too much that secretly relies on bash functionality on Windows and it's not practical to do this yet.  If it's difficult then don't attempt it in this PR.  But it would be interesting to make a list of reasons why Git bash still needs to be in the mix for Windows builds.

dev-tools/docker/linux_aarch64_cross_image/Dockerfile
```diff
@@ -83,3 +83,12 @@ RUN \
   cd .. && \
   rm -rf patchelf-0.13.20210805.a949ff2
 
+# Install CMake
+# v3.9 minimum is required
```
droberts195, line 87:
```suggestion
# v3.19.2 minimum is required
```

dev-tools/docker/linux_aarch64_native_image/Dockerfile
```diff
@@ -111,18 +111,14 @@ RUN \
   /usr/local/bin/pip3.7 install numpy pyyaml setuptools cffi typing_extensions future six requests dataclasses
 
 # Install CMake
-# v3.5 minimum is required
+# v3.9 minimum is required
```
droberts195, line 114:
```suggestion
# v3.19.2 minimum is required
```

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -112,18 +112,14 @@ RUN \
   /usr/local/bin/pip3.7 install numpy ninja pyyaml setuptools cffi typing_extensions future six requests dataclasses
 
 # Install CMake
-# v3.5 minimum is required
+# v3.9 minimum is required
```
droberts195, line 115:
```suggestion
# v3.19.2 minimum is required
```

build-setup/macos.md
```diff
@@ -19,12 +19,25 @@ umask 0002
...
+cmake -B cmake_build -DCMAKE_TOOLCHAIN_FILE=cmake/darwin-aarch64.cmake
+cmake --build cmake_build -v -j`nproc`
+```
+
+or, more simply, use Gradle
+```
+./gradlew :compile
+```
+
+Note that we configure the build to be of type `Release`, and specify the compiler flag `-g` in the cmake configuration files in order to obtain a fully optimized build along with debugging symbols. This is used in preference to `RelWithDebInfo` as `Release` generally gives a higher optimization level than `RelWithDebInfo` (`O3` vs `O2` respectively)
```
droberts195, line 39: If this is moved to in https://github.com/elastic/ml-cpp/blob/main/CONTRIBUTING.md#working-with-the-ml-cpp-codebase then there's no need to duplicate it for the different platforms.  Also it makes more sense in a "Working with the codebase" section because setting up a build machine is a one-off task per machine but people might want to refer back to how to build more frequently.

build-setup/windows.md
```diff
@@ -4,10 +4,10 @@ These instructions are for setting up a Windows build machine from scratch.
 
 For subsequent machines, the setup can be simplified by copying components from the first machine - see [windows_simple.md](windows_simple.md).
 
-The build system assumes Windows itself is installed on the `C:` drive. If this is not the case on your machine, the Makefiles will not work.
-
 On Windows, the build system runs within Git for Windows. This provides a bash shell that enables our Unix build system to work on Windows without alteration.
 
+The build system assumes Windows itself is installed on the `/C` drive. If this is not the case on your machine, set the `ROOT` environment variable to the appropriate drive.
```
droberts195, line 9:
```suggestion
The build system assumes Windows itself is installed on the `C:` drive. If this is not the case on your machine, set the `ROOT` environment variable to the appropriate drive.
```

Also, it would be nice to remove this requirement.  CMake should be less sensitive than MinGW to drive letters.  But this can be left for a future PR.

build-setup/windows.md
```diff
@@ -4,10 +4,10 @@ These instructions are for setting up a Windows build machine from scratch.
...
+
 If you use Gradle to build Machine Learning you do not need to work in Git Bash, as Gradle tasks will run it when necessary.  In order to develop interactively on Windows, you must work in Git Bash.  You will need the following environment variables to be defined:
```
droberts195, line 11: > you must work in Git Bash

Like being tethered to the `C:` drive, it would be nice to avoid the need to work in Git Bash.  CMake generating Visual Studio project files _should_ make it possible to work outside of Git Bash.

It would be nice to do an exercise to list all the reasons why Git Bash is still needed as an integral part of the build system and then figure out how each reason could be avoided by using CMake functionality instead.

But this can be left for a future PR.

build-setup/windows.md
```diff
@@ -21,21 +21,32 @@ You will also need to add several directories to your `PATH` environment variabl
...
+cmake -B cmake_build -DCMAKE_TOOLCHAIN_FILE=cmake/windows-x86_64.cmake
+cmake --build cmake_build --config Release -v -j`nproc`
+```
+
+or, more simply, use Gradle
+```
+./gradlew :compile
+```
+
+Note that we configure the build to be of type `Release`, and specify the compiler flag `/Zi` in the cmake configuration files in order to obtain a fully optimized build along with debugging symbols. This is preferable to say, using the `RelWithDebInfo` configuration as `RelWithDebInfo` has some undesirable side effects (such as disabling inlining)
```
droberts195, line 49: As for the other platforms, consolidate this in https://github.com/elastic/ml-cpp/blob/main/CONTRIBUTING.md#working-with-the-ml-cpp-codebase

cmake/doxygen.cmake
```diff
@@ -0,0 +1,43 @@
...
+  find_package(Doxygen)
+  if (NOT DOXYGEN_FOUND)
+    add_custom_target(doxygen COMMAND false
+                              COMMENT "Doxygen not found")
+    return()
+  endif()
+
+  set(DOXYGEN_GENERATE_HTML          YES)
+  set(DOXYGEN_HTML_OUTPUT            ${output})
+  set(DOXYGEN_PROJECT_NAME           "Ml C++")
```
droberts195, line 12:
```suggestion
  set(DOXYGEN_PROJECT_NAME           "ML C++")
```

edsavage:
retest

droberts195 (approved):
LGTM

I'm happy to merge this to the feature branch and continue to iterate in followup PRs.

PR #2265, opened 2022-05-09
[ML] Correct initialisation of the trend model after detecting seasonality for edge case

tveasey:
Backport #2261.


docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,13 @@
 
 //=== Regressions
 
+== {es} version 8.2.2
```
droberts195, line 31: If you merge this PR soon it will go into 8.2.1.

PR #2267, opened 2022-05-10
[ML] Improve time series model stability and anomaly scoring consistency for sparse data

tveasey:
Our tests to find change points and seasonality in time series data have lower precision for data which are sparse (i.e. when most bucket are empty). False detection tends to lead to us switch between making changes then undoing them or modelling seasonality and the dropping it, since we test from time-to-time and will subsequently undo our mistake. Changing the model also tends to produce less consistent scoring of anomalous events than for non-sparse data.

To address these shortcomings, this PR parameterises these tests so they are less sensitive when the data are sparse. Clearly there is a tradeoff here with recall, but provided the behaviour is clear we still detect both changes and seasonality with high recall for sparse data. It also adds additional testing for sparse synthetic data.

This affects anomaly detection results, but only significantly in the case that the data are sparse.


lib/model/CIndividualModel.cc
```diff
@@ -70,10 +70,6 @@ const std::string FIRST_BUCKET_TIME_TAG("c");
 const std::string LAST_BUCKET_TIME_TAG("d");
 const std::string FEATURE_MODELS_TAG("e");
 const std::string FEATURE_CORRELATE_MODELS_TAG("f");
-// Extra data tag deprecated at model version 34
-// TODO remove on next version bump
-//const std::string EXTRA_DATA_TAG("g");
-//const std::string INTERIM_BUCKET_CORRECTOR_TAG("h");
```
droberts195, line 76: `INTERIM_BUCKET_CORRECTOR_TAG` was only removed in 6.4 (in #97), so we might still see states that contain it (and 6.4+ will silently ignore it).  I think it would be best to leave it commented out until that possibility has gone.

`EXTRA_DATA_TAG` is ancient and can go.  (The comment above it just relates to `EXTRA_DATA_TAG`.)

tveasey: See [c2bef08](https://github.com/elastic/ml-cpp/pull/2267/commits/c2bef089579b57eb3dd64d7d0b8ca2557c43ad7b ).

PR #2268, opened 2022-05-16
[ML] Fix potential use of unintialised value when minimising lowess regression function

tveasey:
If for any reason the extrapolation range for lowess regression returns zero then `globalMinimize`, to find the function minimum, would be supplied an empty collection of points and fail. This would then lead to `minimum` returning a uninitialised value for the minimum. This shouldn't normally happen in our usage of it, but we have seen the error message associated with supplying an empty collection to `globalMinimize` in our QA suite.


valeriy42 (approved):
LGTM

PR #2270, opened 2022-05-16
[ML] Address the root cause for "actual equals typical equals zero" anomalies

tveasey:
We've had multiple issues reported in the past where we report high anomaly scores when the actual is the same as the typical and both are equal to zero. This has been exacerbated in the past because we've also had some instabilities in the modelling when we stop receiving data for a partition, which was an edge case when this could occur.

In all problematic cases, the underlying reason is the if we know the data are non-negative we truncate the prediction when we report the typical value. However, up until now we haven't truncated when computing probabilities.

This changes switches to always truncating our prediction. (This is a halfway house to a more rigorous approach which would condition the predicted distribution to be non-negative.) This should be good enough to ensure we never create non-zero anomaly scores for cases where actual equals typical equals zero except when the result is a multi-bucket anomaly.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -35,6 +35,7 @@
 * Upgrade PyTorch to version 1.11. (See {ml-pull}2233[#2233], {ml-pull}2235[#2235]
   and {ml-pull}2238[#2238].)
 * Upgrade zlib to version 1.2.12 on Windows. (See {ml-pull}2253[#2253].)
+* Address root cause for actuals equals typical equals zero anomalies. (See {ml-pull}2270[#2270].)
 
 === Bug Fixes
 
--- a/lib/maths/time_series/CTimeSeriesModel.cc
+++ b/lib/maths/time_series/CTimeSeriesModel.cc
@@ -593,16 +593,15 @@ CUnivariateTimeSeriesModel::CUnivariateTimeSeriesModel(
     const TDecayRateController2Ary* controllers,
     const TMultibucketFeature* multibucketFeature,
     bool modelAnomalies)
-    : common::CModel(params), m_Id(id), m_IsNonNegative(false), m_IsForecastable(true),
-      m_TrendModel(trendModel.clone()), m_ResidualModel(residualModel.clone()),
+    : common::CModel(params), m_Id(id), m_TrendModel(trendModel.clone()),
+      m_ResidualModel(residualModel.clone()),
       m_MultibucketFeature(multibucketFeature != nullptr ? multibucketFeature->clone()
                                                          : nullptr),
       m_MultibucketFeatureModel(multibucketFeature != nullptr ? residualModel.clone() : nullptr),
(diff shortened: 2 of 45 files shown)


lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -127,16 +123,10 @@ const std::string VERSION_7_11_TAG("7.11");
...
-//const std::string RNG_6_3_TAG{"d"}; Removed in 6.5
 const std::string CONTROLLER_6_3_TAG{"e"};
 const core::TPersistenceTag TREND_MODEL_6_3_TAG{"f", "trend_model"};
 const core::TPersistenceTag RESIDUAL_MODEL_6_3_TAG{"g", "residual_model"};
 const std::string ANOMALY_MODEL_6_3_TAG{"h"};
-
-//const std::string RECENT_SAMPLES_6_3_TAG{"i"}; Removed in 6.5
-//const std::string CANDIDATE_CHANGE_POINT_6_3_TAG{"j"}; Removed in 7.11
-//const std::string CURRENT_CHANGE_INTERVAL_6_3_TAG{"k"}; Removed in 7.11
-//const std::string CHANGE_DETECTOR_6_3_TAG{"l"}; Removed in 7.11
```
droberts195, line 139: I don't think these should be deleted yet.  8.x still has to be able to load state going back to 6.3, so it's good to know what tags were used in that.  Even though they're obviously ignored because they're commented out they still serve a useful purpose as a reminder that a state might get restored that discards these.

docs/CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@
 * Upgrade PyTorch to version 1.11. (See {ml-pull}2233[#2233], {ml-pull}2235[#2235]
   and {ml-pull}2238[#2238].)
 * Upgrade zlib to version 1.2.12 on Windows. (See {ml-pull}2253[#2253].)
+* Address root cause for actuals equals typical equals zero anomalies. (See {ml-pull}2270[#2270].)
 
 === Bug Fixes
```
droberts195, line 40: Since you're editing this file anyway you could move the release note for PR 2261 (line 48) into the 8.2.1 section.  We changed our minds about which version to release it in but only updated the release notes on the 8.2 branch.

droberts195 (approved):
LGTM

The overall approach looks good. I don't know the code well enough to say whether the truncation has been applied everywhere where it needs to be. Probably the biggest chance of a mistake in this PR is a line that _hasn't_ been changed rather than one that has, but at least that would be no worse than now.

docs/CHANGELOG.asciidoc
droberts195: Actually don't worry about this - I've done it in #2272

lib/maths/time_series/CTimeSeriesModel.cc
tveasey: See [44ddc14](https://github.com/elastic/ml-cpp/pull/2270/commits/44ddc14320809228500abdbccb5c82a3c3eecdf5).

PR #2272, opened 2022-05-18
[ML] Permit new system calls used by Ubuntu 22.04

droberts195:
Ubuntu 22.04 uses glibc 2.35 and the implementation of
pthread_create has been changed to use more system calls
than it used to.

In order for pthread_create to work on glibc 2.35 we need
to allow the rt_sigprocmask, rseq and clone3 system calls.

Fixes elastic/elasticsearch#86877


przemekwitek (approved):
LGTM

Issue #2276, opened 2022-05-23
[ML] Revisit correction for bucket length when computing probability quantiles normalising results


tveasey:
Currently, we try and account for the different number of results per time interval for different bucket lengths when computing probability quantiles as part of rate limiting normalised scores.

The assumption we make should result in roughly the same score distribution when modelling pure noise (I need to recheck this). However, it will tend to pull down scores when there are long lasting anomalies in the data for short bucket lengths. This has probably been exacerbated by multi-bucket anomaly detection. In particular, we see cases that the correction has the reverse impact to the one we'd like: producing less consistent scoring for different bucket lengths.

We should revisit whether any sort of bucket length correction rate limiting normalised scores is a good idea.

PR #2280, opened 2022-05-24
[ML] Improve robustness to outliers after detecting changes in time series

tveasey:
We have some code which affects the weighting we apply to outliers immediately after detecting a change in a time series. This is to deal with time series which flip-flop between values: we don't want to flag every change as anomalous, but rather learn the "envelope" of the changes. However, we need to be a bit more careful than we are with allowing the model to incorporate unusual values just after a change. If we get unlucky and there is a large outlier around this time it pollutes the model.

To address the flip-flop case we simply need to increase the weight of values whose prediction error is on the same scale as the change we detected. Values whose prediction error is significantly larger can be reweighted as normal. I noticed this while debugging #2276.

**Before**
Screenshot 2022-05-24 at 12 28 55

**After**
Screenshot 2022-05-24 at 12 28 42

### DIFF
--- a/lib/maths/time_series/CTimeSeriesDecompositionDetail.cc
+++ b/lib/maths/time_series/CTimeSeriesDecompositionDetail.cc
@@ -317,6 +317,8 @@ const core::TPersistenceTag LAST_CHANGE_POINT_TIME_7_11_TAG{"i", "last_change_po
 const core::TPersistenceTag LAST_CANDIDATE_CHANGE_POINT_TIME_7_11_TAG{
     "j", "last_candidate_change_point_time"};
 const core::TPersistenceTag LAST_CHANGE_POINT_7_11_TAG{"k", "last_change_point"};
+// Version 8.3
+const core::TPersistenceTag WINSORIZATION_DERATE_8_3_TAG{"l", "winsorization_derate"};
 
 // Seasonality Test Tags
 // Version 7.9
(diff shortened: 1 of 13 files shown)


lib/maths/time_series/CTimeSeriesDecompositionDetail.cc
```diff
@@ -317,6 +317,8 @@ const core::TPersistenceTag LAST_CHANGE_POINT_TIME_7_11_TAG{"i", "last_change_po
 const core::TPersistenceTag LAST_CANDIDATE_CHANGE_POINT_TIME_7_11_TAG{
     "j", "last_candidate_change_point_time"};
 const core::TPersistenceTag LAST_CHANGE_POINT_7_11_TAG{"k", "last_change_point"};
+// Version 8.3
+const core::TPersistenceTag WINSORIZATION_DERATE_8_3_TAG{"l", "winsorization_derate"};
```
droberts195, line 321: It's probably best to update `MODEL_SNAPSHOT_MIN_VERSION` in `CAnomalyJob.cc` to `8.3.0` to match this.

Maybe it's not that critical if it's acceptable for the data behind this tag to get lost if a job moves around in a mixed version cluster.

But bumping the version that the Java side understands will help in the long run if we ever want to remove this.

lib/maths/time_series/unittest/CTimeSeriesModelTest.cc
```diff
@@ -294,7 +294,7 @@ void reinitializeResidualModel(double learnRate,
 
 class CDebug {
 public:
-    static const bool ENABLED{false};
+    static const bool ENABLED{true};
```
droberts195, line 297: Did you mean to leave this on for checkin?

If so, please add `results.py` to the `.gitignore` file for this directory so nobody accidentally checks it in.

lib/maths/time_series/CTimeSeriesDecompositionDetail.cc
tveasey: See [3adb43b](https://github.com/elastic/ml-cpp/pull/2280/commits/3adb43bb1eed624e6e2736c4f56b54e9c5302285).

lib/maths/time_series/unittest/CTimeSeriesModelTest.cc
tveasey: See [d0bdf29](https://github.com/elastic/ml-cpp/pull/2280/commits/d0bdf2951aebdecbd2ebc7a399cf034a5ab2bfba).

PR #2283, opened 2022-05-25
[ML] Target distribution-preserving sampling

valeriy42:
For incremental learning on one-sided feedback, we need an ability to extend data summarization sample without "forgetting" that samples of underrepresented categories. For instance, if we receive the new batches of data containing only class "0", we want our data summarization to include still all the instances of the class "1" it had initially. This PR implements this capability. For regression, it would still behave as initially, creating a sample from the quantiles of the complete training dataset. This way, we can handle situations like target drift, where the target domain can extend orbitrary.


valeriy42:
retest

valeriy42:
retest

lib/maths/analytics/unittest/CDataFrameUtilsTest.cc
```diff
@@ -802,6 +802,133 @@ BOOST_AUTO_TEST_CASE(testStratifiedSamplingRowMasks) {
...
+
+        TDoubleDoubleUMap actualCategoryCounts;
+        frame->readRows(1, 0, frame->numberRows(),
+                        [&](const core::CDataFrame::TRowItr& beginRows,
+                            const core::CDataFrame::TRowItr& endRows) {
+                            for (auto row = beginRows; row != endRows; ++row) {
+                                actualCategoryCounts[(*row)[0]] += 1.0;
+                            }
+                        },
+                        &samplingRowsMask);
```
tveasey, line 860: Maybe factor this out into a lambda and use for computing actual and expected counts. Alternatively, I guess this could use CDataFrameUtils::categoryCounts, but maybe testing via an alternative implementation is better.

tveasey (commented):
Overall a nice clean implementation!

I know we discussed preserving the class counts on the original training data, but my main thought is would a better criterion be that we chose the more balanced distribution of old vs old + new?

Also, I realised that this is going to interact badly with multi-step learning for train by query, where we really do want to retain all the data. We can revisit that later, but it might be worthwhile leaving a TODO in `dataSummarization`.

include/maths/analytics/CDataFrameUtils.h
```diff
@@ -320,13 +320,40 @@ class MATHS_ANALYTICS_EXPORT CDataFrameUtils : private core::CNonInstantiatable
...
+                              std::size_t desiredNumberSamples,
+                              std::size_t numberBuckets,
+                              const core::CPackedBitVector& allTrainingRowsMask);
+
+    //! Compute sampling on the \p frame that preserves the distribution on \p distributionSourceRowsMask.
+    //!
+    //! For categorical \p targetColumn, the items of
+    //! different categories of the \p distributionSourceRowsMask and of the result sample will match.
+    //! For numerical \p targetColumn, the quantiles distribution of \p allTrainingRowsMask with be
+    //! preserved.
```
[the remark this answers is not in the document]
valeriy42: Thank you. I'll use your version.

lib/maths/analytics/unittest/CDataFrameUtilsTest.cc
valeriy42: I refactored it to use `CDataFrameUtils::categoryCounts` for `actualCategoryCounts` and I use explicit calculation for `expectedCategoryCounts`. This way I can test the `categoryCounts` implementation and keep the test code clean.

valeriy42:
Thank you for the review @tveasey . I integrated your suggestions. It would be great if you could have another pass.

tveasey:
> I am not sure what you mean here. For multi-step learning, we would use the complete training dataset as data summarization so that there won't be any sampling.

I was thinking we preserved the size of original training data set, but reading again I see you actually take a fixed fraction of the total training set. I wonder about this: it allows the sample size to grow without bounds over many retrains. But let's not worry about that at this point.

lib/maths/analytics/CBoostedTreeImpl.cc
```diff
@@ -2527,12 +2527,41 @@ core::CPackedBitVector CBoostedTreeImpl::dataSummarization(const core::CDataFram
...
+    auto allTrainingData{dataSummarizationRowMask | newDataSampleRowMask};
+
+    if (m_Loss->isRegression()) {
+        // For regression we preserve the quantile distribution of the complete training sample.
+        core::CPackedBitVector rowMask{CDataFrameUtils::distributionPreservingSamplingRowMask(
+            m_NumberThreads, frame, m_DependentVariable, m_Rng, sampleSize, 10,
+            allTrainingData, allTrainingData)};
+        return rowMask;
+    }
+    // For classification we preserve the class distribution of the data summarization sample.
```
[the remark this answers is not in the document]
tveasey: Maybe update the comment now too.

lib/maths/analytics/CBoostedTreeImpl.cc
```diff
@@ -2527,12 +2527,56 @@ core::CPackedBitVector CBoostedTreeImpl::dataSummarization(const core::CDataFram
...
+    auto classImbalance = [&](const core::CPackedBitVector& rowMask) {
+        auto categoryCounts = CDataFrameUtils::categoryCounts(
+            m_NumberThreads, frame, rowMask, {m_DependentVariable})[m_DependentVariable];
+        double maxClassCount{
+            *(std::max_element(categoryCounts.begin(), categoryCounts.end()))};
+        double minClassCount{
+            *(std::min_element(categoryCounts.begin(), categoryCounts.end()))};
```
tveasey, line 2568: nit: seems like the perfect opportunity for `minmax_element`.

tveasey (approved):
LGTM aside from a couple of very minor points. Also looks like formatting is off.

PR #2284, opened 2022-05-25
[ML] Improve naming

tveasey:
We were using Winsorization to refer to the weight we apply to outlying data when updating time series models. Morally, what we do is much closer to iterative reweighting of outliers. It is therefore less confusing to refer to this as an outlier weight. We were also misspelling Winsorization as Winsorisation. Since we do actually do Winsorization in some of our clustering code I've also corrected this typo.


droberts195 (approved):
LGTM

PR #2285, opened 2022-05-26
[ML] Improve normalisation of anomaly detection results for short bucket lengths

### LINKED ISSUE #2276: [ML] Revisit correction for bucket length when computing probability quantiles normalising results

tveasey:
This changes the way we normalise for short bucket lengths.

Previously, we tried to apply consistent rate limiting of alerts across different bucket lengths by accounting for the different number of results we receive per time interval. The approach we took proved to interact badly with uncertainty in the estimates of extreme quantiles. The upshot is we actually could get bias where our scores would be too low short bucket lengths. The problem was particularly severe if the time series had extended runs of unusual values.

This takes a different approach: we simply only update the quantiles with the high water mark in each fixed time interval. This achieves the same thing, but works reliably for shorter bucket lengths. I've added a test that rate limiting is working as expected across multiple bucket lengths.

This will potentially cause a non-trivial results change. One thing to consider is if we need special handling for upgrading existing jobs. First off we need to see differences on our QA suite.

Fixes #2276.

### DIFF
--- a/include/model/CAnomalyScore.h
+++ b/include/model/CAnomalyScore.h
@@ -23,14 +23,13 @@
 #include <boost/optional.hpp>
 #include <boost/unordered_map.hpp>
 
+#include <cstdint>
 #include <functional>
 #include <iosfwd>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
-
-#include <stdint.h>
 
 namespace CAnomalyScoreTest {
 struct testNormalizerGetMaxScore;
@@ -138,7 +137,7 @@ public:
         };
 
     public:
-        CNormalizer(const CAnomalyDetectorModelConfig& config);
+        explicit CNormalizer(const CAnomalyDetectorModelConfig& config);
 
         //! Does this normalizer have enough information to normalize
         //! anomaly scores?
@@ -196,6 +195,7 @@ public:
     private:
         using TDoubleDoublePr = std::pair<double, double>;
         using TDoubleDoublePrVec = std::vector<TDoubleDoublePr>;
+
         //! \brief Wraps a maximum score.
         class CMaxScore {
         public:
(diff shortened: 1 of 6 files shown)


include/model/CAnomalyScore.h
```diff
@@ -268,13 +268,13 @@ class MODEL_EXPORT CAnomalyScore {
...
         //! The maximum possible normalized score.
-        double m_MaximumNormalizedScore = 100.0;
+        double m_MaximumNormalizedScore{100.0};
 
         //! The approximate HIGH_PERCENTILE percentile raw score.
         uint32_t m_HighPercentileScore;
         //! The number of scores less than the approximate
         //! HIGH_PERCENTILE percentile raw score.
-        uint64_t m_HighPercentileCount = 0;
+        uint64_t m_HighPercentileCount{0};
```
droberts195, line 277:
```suggestion
        std::uint64_t m_HighPercentileCount{0};
```

include/model/CAnomalyScore.h
```diff
@@ -268,13 +268,13 @@ class MODEL_EXPORT CAnomalyScore {
         //! The normalized anomaly score knot points.
         TDoubleDoublePrVec m_NormalizedScoreKnotPoints;
         //! The maximum possible normalized score.
-        double m_MaximumNormalizedScore = 100.0;
+        double m_MaximumNormalizedScore{100.0};
 
         //! The approximate HIGH_PERCENTILE percentile raw score.
         uint32_t m_HighPercentileScore;
```
droberts195, line 274:
```suggestion
        std::uint32_t m_HighPercentileScore;
```

lib/api/unittest/CResultNormalizerTest.cc
```diff
@@ -60,7 +60,7 @@ BOOST_AUTO_TEST_CASE(testInitNormalizerPartitioned) {
 
     // The maximum bucket_time influencer probability in the farequote data used to initialise
     // the normaliser is 2.1305076675260463E-24, so this should map to the highest normalised
-    // score which is 90.69091
+    // score which is 90.69091.
```
droberts195, line 63: This comment no longer matches the value on line 81.

lib/model/CAnomalyScore.cc
```diff
@@ -271,23 +279,19 @@ bool CAnomalyScore::CNormalizer::canNormalize() const {
...
         // Nothing to do.
         return true;
     }
-
     if (m_RawScoreQuantileSummary.n() == 0) {
-        LOG_ERROR(<< "No scores have been added to the quantile summary");
-        return false;
+        score = 0.0;
+        return true;
     }
```
droberts195, line 289: Are we intentionally going through this code branch now?  I would have thought it's a coding error to ask to normalize a score when we've never generated any score.

tveasey: We don't update the quantiles with every score any more for short bucket lengths. So at the very beginning we hit the situation we haven't yet added any scores. I think this is harmless: it is only the first 30 mins of modelling, in practice scores always work out as zero at this point anyway. I don't think it is worthwhile trying to detect that we're not in start up, so I just don't bother to error in this case anymore.

include/model/CAnomalyScore.h
tveasey: My plan is to try and address some fo the clang tidy suggestions we're getting now we run it over the codebase. As part of this work I'm going to do one big change to switch all our uses of `stdint.h` to `cstdint` and migrate to the std namespace versions. I think I can do this with regex and I reckon it is better than making piecemeal changes for this.

include/model/CAnomalyScore.h
tveasey: See below.

droberts195 (approved):
LGTM

PR #2286, opened 2022-05-27
[ML] Fix a source of "Discarding sample = nan, weights = ..." log errors

tveasey:
Recent changes to anomaly detection code exposed a pre-existing bug in sampling code for one of our distribution models. In particular, we were computing `E[X 1{[a, b]}] / E[1{[a, b]}]` for log-normal and gamma distributions and not properly handling the case `a -> 0`. As `a` gets close to 0 one can simply snap it to 0 and use an alternative form for `E[X 1{[0, b]}]` which is more numerically pleasant. Even then we can run into `E[X 1{[a, b]}] / E[1{[a, b]}] = 0/0` for finite precision. In this case, we need to compute an asymptotic approximation.

The upshot is previously we would generate NaN samples. These were subsequently discarded, so as far as I can tell the actual impact isn't too severe. However, it would generate log errors when we trapped and discarded the bad values.

### DIFF
--- a/lib/maths/common/CTools.cc
+++ b/lib/maths/common/CTools.cc
@@ -598,10 +598,10 @@ operator()(const lognormal& logNormal, double x, maths_t::ETail& tail) const {
         //                        + 2 * s^2 * (log(x) - m))^(1/2))  if x > mode
 
         double logx = std::log(x);
-        double squareScale = CTools::pow2(logNormal.scale());
-        double discriminant = std::sqrt(CTools::pow2(squareScale) +
-                                        (logx - logNormal.location() + 2.0 * squareScale) *
-                                            (logx - logNormal.location()));
+        double squareScale = pow2(logNormal.scale());
+        double discriminant = std::sqrt(
+            pow2(squareScale) + (logx - logNormal.location() + 2.0 * squareScale) *
+                                    (logx - logNormal.location()));
         double m = boost::math::mode(logNormal);
         this->tail(x, m, tail);
         double y = m * std::exp(x > m ? -discriminant : discriminant);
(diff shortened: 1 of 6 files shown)


lib/maths/common/CTools.cc
```diff
@@ -1393,63 +1399,88 @@ operator()(const lognormal& logNormal, double a, double b) const {
...
         return expa == expb ? (2.0 * a / (a + b)) * b
                             : (expa + expb) / (expa / a + expb / b);
     }
 
-    double erfa_ = a_ == NEG_INF ? -1.0 : std::erf(a_);
-    double erfb_ = b_ == POS_INF ? 1.0 : std::erf(b_);
-    return mean * (erfb - erfa) / (erfb_ - erfa_);
+    double erfa_ = std::erf(a_);
+    double erfb_ = std::erf(b_);
+    return (erfb - erfa) == (erfb_ - erfa_) ? mean : mean * (erfb - erfa) / (erfb_ - erfa_);
```
droberts195, line 1445: Is it possible that `erfb_ - erfa_` could be zero when `erfb - erfa` is not zero?

lib/maths/common/CTools.cc
```diff
@@ -1393,63 +1399,88 @@ operator()(const lognormal& logNormal, double a, double b) const {
...
+        double expb = std::exp((shape - 1.0) * std::log(b) - rate * b);
+        return expa == expb ? (a + b) / 2.0 : (a * expa + b * expb) / (expa + expb);
     }
 
-    double gama_ = a <= 0.0 ? 0.0 : boost::math::gamma_p(shape, rate * a);
-    double gamb_ = b == POS_INF ? 1.0 : boost::math::gamma_p(shape, rate * b);
-    return mean * (gamb - gama) / (gamb_ - gama_);
+    double gama_ = boost::math::gamma_p(shape, rate * a);
+    double gamb_ = boost::math::gamma_p(shape, rate * b);
+    return (gamb - gama) == (gamb_ - gama_) ? mean : mean * (gamb - gama) / (gamb_ - gama_);
```
droberts195, line 1483: Is it possible that `gamb_ - gama_ ` could be zero when `gamb - gama ` is not zero?

lib/maths/common/CTools.cc
tveasey: In this case the answer should be inf (at least to working precision). But maybe though I should actually trap to max double because inf probably creates downstream issues. It shouldn't really happen for realistic data but better safe than sorry.

droberts195 (approved):
LGTM

lib/maths/common/CTools.cc
tveasey: See [cbaa29e](https://github.com/elastic/ml-cpp/pull/2286/commits/cbaa29e7862fec6da97f6fb5233edb3582b5c4ed).

lib/maths/common/CTools.cc
tveasey: See [cbaa29e](https://github.com/elastic/ml-cpp/pull/2286/commits/cbaa29e7862fec6da97f6fb5233edb3582b5c4ed).

tveasey:
retest

PR #2291, opened 2022-05-31
[ML] Apply consistent size penalty selecting best classification and regression model

tveasey:
We currently apply a small penalty to prefer selecting small models when the validation loss is similar. This is based on the model size, but is also parameterised by the mean size of all models trained up to each point where we test. The mean size changes through the optimisation loop and means we don't apply a completely consistent penalty when comparing the candidate model with the current best model (whose penalty was calculated using earlier parameters). This change stores the best model size as well so we compute penalties using the same parameters.


lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -468,10 +492,13 @@ bool CBoostedTreeHyperparameters::captureBest(const TMeanVarAccumulator& testLos
     // We capture the parameters with the lowest error at one standard
     // deviation above the mean. If the mean error improvement is marginal
     // we prefer the solution with the least variation across the folds.
-    double testLoss{lossAtNSigma(1.0, testLossMoments) + this->modelSizePenalty(numberNodes)};
+    double testLoss{lossAtNSigma(1.0, testLossMoments)};
+    double penalizedTestLoss{testLoss + this->modelSizePenalty(numberNodes)};
+    double bestPenalizedTestLoss{m_BestForestTestLoss + this->modelSizePenalty(numberNodes)};
```
valeriy42, line 497:
```suggestion
    double bestPenalizedTestLoss{m_BestForestTestLoss + this->modelSizePenalty(m_BestForestNumberNodes)};
```
Shouldn't this be `m_BestForestNumberNodes` now?

PR #2292, opened 2022-06-01
[ML] Migrate to cstdint and std namespace versions of integer types

tveasey:
Now we're running clang tidy on the codebase we plan to gradually address its suggestions. One is to migrate our uses of `stdint.h` to `cstdint`, which is more idiomatic C++. In conjunction with this change we are only guaranteed that types are defined in the `std::` namespace.


droberts195 (approved):
LGTM

PR #2294, opened 2022-06-02
[ML] Reactivate Java  integration tests

valeriy42:
This PR ensures backward compatibility with formats and definitions on the main branch. To this end, the configuration parameter `data_summarization_fraction` now defaults to 0.0, which leads to the same behavior as before introducing incremental learning. If `data_summarization_fraction` is set to a non-zero value, we produce additional definitions and information required for incremental updates.

It also enables Java integration tests to ensure interoperability.


include/api/CDataFrameAnalysisInstrumentation.h
```diff
@@ -225,6 +228,7 @@ class API_EXPORT CDataFrameTrainBoostedTreeInstrumentation final
     bool m_AnalysisStatsInitialized{false};
     std::string m_LossType;
     TLossVec m_LossValues;
+    api_t::EDataFrameTrainBoostedTreeTask m_Task{api_t::EDataFrameTrainBoostedTreeTask::E_Train};
```
tveasey, line 231: I guess we have a choice about whether to make the enum a enum class or not. Since this is the only thing in api_t my inclination is to keep as is but then I would be consistent with qualifying:
```suggestion
    api_t::EDataFrameTrainBoostedTreeTask m_Task{api_t::E_Train};
```

lib/api/CInferenceModelMetadata.cc
```diff
@@ -24,8 +25,12 @@ void CInferenceModelMetadata::write(TRapidJsonWriter& writer) const {
...
     this->writeFeatureImportanceBaseline(writer);
     this->writeHyperparameterImportance(writer);
-    this->writeTrainProperties(writer);
-    this->writeDataSummarization(writer);
+    LOG_DEBUG(<< "Nuber data summarization rows " << m_NumDataSummarizationRows);
+    if (m_NumDataSummarizationRows > 0) {
+        // only output if data summarization fraction was specified.
+        this->writeTrainProperties(writer);
+        this->writeDataSummarization(writer);
+    }
```
tveasey, line 33:
```suggestion
    LOG_DEBUG(<< "Number data summarization rows " << m_NumDataSummarizationRows);
    if (m_NumDataSummarizationRows > 0) {
        // Only output if data summarization fraction was specified.
        this->writeTrainProperties(writer);
        this->writeDataSummarization(writer);
    }
```

lib/api/CInferenceModelMetadata.cc
```diff
@@ -271,20 +276,30 @@ void CInferenceModelMetadata::hyperparameterImportance(
...
+        // Not tuned directly.
+        case maths::analytics::boosted_tree_detail::E_MaximumNumberTrees:
+            hyperparameterName = CDataFrameTrainBoostedTreeRunner::MAX_TREES;
+            break;
+
+        // Incremental train hyperparameters.
         case maths::analytics::boosted_tree_detail::E_PredictionChangeCost:
+            if (m_Task != api_t::E_Update) {
+                continue;
+            }
```
tveasey, line 288: It seems surprising to me that this is in `hyperparameterImportance` if the task is not E_Update. Is this definitely needed?

lib/maths/analytics/CBoostedTreeImpl.cc
```diff
@@ -2523,6 +2523,9 @@ core::CPackedBitVector CBoostedTreeImpl::dataSummarization(const core::CDataFram
 
     core::CPackedBitVector allTrainingRowsMask{this->allTrainingRowsMask()};
 
+    if (m_DataSummarizationFraction == 0.0) {
+        return ~allTrainingRowsMask;
```
tveasey, line 2527: I think this shouldn't negate all training row mask because this will be true for test row masks. We probably get away with this because we never pass such rows now inference for them happens in Java, but I think this should be
```suggestion
        return core::CPackedBitVector{frame.numberRows(), false);
```
Also I would move to before `core::CPackedBitVector allTrainingRowsMask{this->allTrainingRowsMask()};`.

tveasey (commented):
Good stuff! This was easier than I anticipated. Just a couple of small things.

tveasey (approved):
LGTM

lib/api/CInferenceModelMetadata.cc
tveasey: It turns out that this can be addressed in `CBoostedTreeHyperparameters` which should only return importances for relevant hyperparameters.

PR #2297, opened 2022-06-07
[ML] Memory management for incremental training

### LINKED ISSUE #1790: [ML] Boosted tree incremental training memory estimation
The memory requirements are different to regular train and we need custom upfront memory estimation code.

tveasey:
We were missing memory management for the new incremental training, prediction and compute encodings tasks. This change adds it in. Note that we also need to write out the trained model size to meta data and supply this when we call tasks which load a pre-existing model to do proper accounting.

Closes #1790.

### DIFF
--- a/lib/api/unittest/testfiles/model_meta_data/model_meta_data.schema.json
+++ b/lib/api/unittest/testfiles/model_meta_data/model_meta_data.schema.json
@@ -112,6 +112,9 @@
                 },
                 "loss_gap" : {
                     "type" : "number"
+                },
+                "trained_model_memory_usage": {
+                    "type" : "number"
                 }
             },
             "required" : [
--- a/lib/maths/analytics/CBoostedTreeFactory.cc
+++ b/lib/maths/analytics/CBoostedTreeFactory.cc
@@ -133,12 +133,8 @@ CBoostedTreeFactory::buildForEncode(core::CDataFrame& frame, std::size_t depende
 
     skipCheckpointIfAtOrAfter(CBoostedTreeImpl::E_EncodingInitialized, [&] {
         this->initializeMissingFeatureMasks(frame);
-        this->initializeNumberFolds(frame);
-        // There are only "old" training examples for the initial train.
-        if (frame.numberRows() > m_TreeImpl->m_NewTrainingRowMask.size()) {
-            m_TreeImpl->m_NewTrainingRowMask.extend(
-                false, frame.numberRows() - m_TreeImpl->m_NewTrainingRowMask.size());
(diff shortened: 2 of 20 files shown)


lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -1631,14 +1642,31 @@ std::size_t CBoostedTreeFactory::estimateMemoryUsageForTrain(std::size_t numberR
...
 CBoostedTreeFactory::estimateMemoryUsageForTrainIncremental(std::size_t numberRows,
                                                             std::size_t numberColumns) const {
-    return m_TreeImpl->estimateMemoryUsageTrainIncremental(numberRows, numberColumns);
+    std::size_t maximumNumberTrees{this->mainLoopMaximumNumberTrees(
+        m_TreeImpl->m_Hyperparameters.eta().fixed()
+            ? m_TreeImpl->m_Hyperparameters.eta().value()
+            : computeEta(numberColumns))};
+    CScopeBoostedTreeParameterOverrides<std::size_t> overrides;
+    overrides.apply(m_TreeImpl->m_Hyperparameters.maximumNumberTrees(), maximumNumberTrees);
+    return m_TreeImpl->estimateMemoryUsageForTrainIncremental(numberRows, numberColumns);
```
valeriy42, line 1657: Should this code be moved to m_TreeImpl?

tveasey: The tricky part is that `mainLoopMaximumNumberTrees` logic relates to `CBoostedTreeFactory`. This sort of configuration seems to naturally belong on the factory, so I'm not sure I'd want to move that. `estimateMemoryUsageForTrainIncremental` does implement the bulk of memory estimation so I feel like this is probably ok.

tveasey:
> I wonder if we need another test for memory estimate

Yes good point. I'll add some more testing in this area.

tveasey:
I added some more testing (and it actually found a pre-existing bug :)). Could you take another look @valeriy42?

lib/api/unittest/testfiles/model_meta_data/model_meta_data.schema.json
```diff
@@ -112,6 +112,9 @@
                 },
                 "loss_gap" : {
                     "type" : "number"
+                },
+                "trained_model_memory_usage": {
+                    "type" : "number"
```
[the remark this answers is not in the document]
tveasey: Agreed. I'll add an issue so we capture this, but will make the change when we actually merge to main.

PR #2298, opened 2022-06-09
[ML] Early stopping for coarse parameter tuning

valeriy42:
This PR enables the early stopping of hyperparameter tuning after the coarse parameter tuning. Skipping fine parameter tuning should improve training time at the cost of potentially slightly lower accuracy.


tveasey (approved):
Port LGTM

PR #2306, opened 2022-06-15
[ML] Fix flaky test

### LINKED ISSUE #2263: [ML] CStaticThreadPoolTest.testNumberThreadsInUse failed on Linux
`CStaticThreadPoolTest.testNumberThreadsInUse` failed in an 8.2 Linux build: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+8.2+linux/5/testReport/junit/lib.core.CStaticThreadPoolTest/CStaticThreadPoolTest/testNumberThreadsInUse/

This doesn't happen every time.  It's unclear what was different about the build where it failed.

#2262 might make this easier to investigate if it happens again.

tveasey:
We had a race in our test of limiting the number of threads in use. It was possible for a worker thread to have loaded `m_NumberThreadsInUse` before it got reset in the main thread and so steal one task. This was causing occasional test failures. 

We don't care about this in practice. We only expect that the thread pool soon adapts to only scheduling tasks on a limited number of threads after making a change to `m_NumberThreadsInUse` (thread specific queues need to be drained anyway). Therefore, we simply update the test logic to reflect that this can happen.

Fixes #2263

### DIFF
--- a/include/core/CStaticThreadPool.h
+++ b/include/core/CStaticThreadPool.h
@@ -46,6 +46,9 @@ public:
     CStaticThreadPool& operator=(const CStaticThreadPool&) = delete;
     CStaticThreadPool& operator=(CStaticThreadPool&&) = delete;
 
+    //! Get the number of threads in use.
+    std::size_t numberThreadsInUse() const;
+
     //! Adjust the number of threads which are being used by the pool.
     //!
     //! \note \p threads should be in the range [1, pool size].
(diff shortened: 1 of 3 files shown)


droberts195:
Does this fix #2263?

tveasey:
Yes it does.

droberts195 (approved):
LGTM

PR #2307, opened 2022-06-16
[ML] Updating CMake Docker config to build new official images

droberts195:
The official images need simple numeric versions rather than
"cmake2", etc.

Also installing the latest version of CMake on the Docker
images so we don't have to rebuild them if we want to use
slightly more recent CMake features.


edsavage (approved):
LGTM

Thanks Dave

dev-tools/docker/macosx_image/Dockerfile
```diff
@@ -52,9 +52,8 @@ RUN \
...
 # v3.19.2 minimum is required
 RUN \
-  wget --quiet https://github.com/Kitware/CMake/releases/download/v3.19.3/cmake-3.19.3-Linux-x86_64.sh && \
-  mkdir -p /usr/local/gcc103 && \
-  chmod +x cmake-3.19.3-Linux-x86_64.sh && \
-  ./cmake-3.19.3-Linux-x86_64.sh --skip-license --prefix=/usr/local/gcc103 && \
-  rm -f cmake-3.19.3-Linux-x86_64.sh
+  wget --quiet https://github.com/Kitware/CMake/releases/download/v3.23.2/cmake-3.23.2-Linux-x86_64.sh && \
+  chmod +x cmake-3.23.2-Linux-x86_64.sh && \
+  ./cmake-3.23.2-Linux-x86_64.sh --skip-license --prefix=/usr/local && \
```
droberts195, line 57: I changed this because on Linux it's convenient if all the tools for building a given branch of ml-cpp are in the same subdirectory that relates to the compiler we're using for that branch.  It makes it easier to keep the right stuff when trimming down a Docker image.  But on macOS XCode is installed separately, g++ isn't used, and it's silly to create a `gcc103` directory just for CMake.

droberts195:
The CI test failure was fixed by #2306, which is not on this feature branch.

I'm going to merge this anyway, as the only change since the last successful CI run was in the contributing guide.

PR #2312, opened 2022-06-17
[ML] Write test output to files

edsavage:
To aid diagnosis of unit test failures duplicate stdout and stderr to
log files. This functionality is not fully portable as it makes use of
the `tee` command, as such it is not supported on Windows.


cmake/functions.cmake
```diff
@@ -194,12 +194,14 @@ function(ml_add_test_executable _target)
         ${CMAKE_CURRENT_BINARY_DIR}/$<IF:$<CONFIG:Release>,Release,Debug>/ml_test_${_target}.exe
         ${CMAKE_CURRENT_BINARY_DIR}/ml_test_${_target}.exe
       COMMAND ${CMAKE_CURRENT_BINARY_DIR}/ml_test_${_target}
```
droberts195, line 196: I think we should redirect to a file on Windows too.

cmake/functions.cmake
```diff
@@ -194,12 +194,14 @@ function(ml_add_test_executable _target)
...
+      COMMENT "Running test: ml_test_${_target}"
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
       )
   else()
     add_custom_target(test_${_target}
       DEPENDS ml_test_${_target}
-      COMMAND ${CMAKE_CURRENT_BINARY_DIR}/ml_test_${_target}
+      COMMAND ${CMAKE_CURRENT_BINARY_DIR}/ml_test_${_target} 2>&1 | tee ${CMAKE_CURRENT_BINARY_DIR}/ml_test_${_target}.log
```
droberts195, line 203: I think it's best not to use `tee`, as this will still cause the CI job output to be unreadable, with output from multiple processes interleaved.

What would be nice is to do something similar to Gradle, where it hides the output in a file _unless_ there's a failure, in which case it prints it all at the end.

So we'd need to do something like:

1. Run `ml_test` and redirect the output to a file
2. Remember the exit code of `ml_test`
3. If the exit code is not zero, `cat` the file
4. Make the success of the whole step depend not on the exit code of `cat` but the one we remembered from `ml_test`

Hopefully by doing a full `cat` at the end there'll be less interleaving than there is if `ml_test` writes to `stderr` as it runs.

droberts195 (commented):
I think it would be a good idea to introduce a test failure into the next commit on this PR so that we can see what the CI output looks like for each platform when a test failure is reported.

Then in the final commit the test can be fixed and we can look at the (hopefully clean) CI output for all platforms in that final CI run.

droberts195 (approved):
LGTM if you remove the failing test before merging

Issue #2311, opened 2022-06-16
[ML] Investigate replacing uses of bash and other Unix tools in CMake build system with platform independent code


edsavage:
The initial approach to migrating to a `CMake` build system has continued to assume the presence of a Unix like build environment. On Windows this requires the presence of `git bash` and the minimal GNU toolset - `MinGW`.  It is desirable to lessen the dependency on such tools on Windows as it lowers the bar for developers on that platform and simplifies the setting up of the build environment.

What follows is a list of places in the `CMake` build system that make calls to `bash` or other Unix commands.

 ### `lib/ver/CMakeLists.txt`
```
lib/ver/CMakeLists.txt:execute_process(COMMAND date "+%Y" OUTPUT_VARIABLE BUILD_YEAR OUTPUT_STRIP_TRAILING_WHITESPACE)
lib/ver/CMakeLists.txt:execute_process(COMMAND id COMMAND awk -F ")" "{ print $1 }" COMMAND awk -F "(" "{ print $2 }" OUTPUT_VARIABLE USER_NAME
lib/ver/CMakeLists.txt:execute_process(COMMAND awk -F= "/elasticsearchVersion/ {gsub(/-.*/,\"\"); print $2}" $ENV{CPP_SRC_HOME}/gradle.properties OUTPUT_VARIABLE PRODUCT_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
```

Suggestions:
* id: Use native “whoami” command that gives response "vm-win2016-64-1\eds” and parse using cmake string commands
* date: use native “date /T” (requires extensions to be installed and parse out the year
* awk: use CMake file READ command to read in gradle.properties and then search and parse out elasticsearchVersion value

###  `3rd_party/CMakeLists.txt:`
```
add_custom_target(
    eigen
    DEPENDS 3rd_party.sh pull-eigen.sh
    COMMAND bash -c "./3rd_party.sh --add ${INSTALL_DIR}"
    COMMAND bash -c "./pull-eigen.sh"
    WORKING_DIRECTORY cmake${CMAKE_CURRENT_SOURCE_DIR}
)
```

###  `cmake/compiler/vs2019.cmake`
```
execute_process(COMMAND bash -c "cygpath -m -s \"${ROOT}/Program Files (x86)/Microsoft Visual Studio/2019/Professional\"" OUTPUT_VARIABLE VCBASE OUTPUT_STRIP_TRAILING_WHITESPACE)
[...]

PR #2313, opened 2022-06-17
[ML] Portability improvements

edsavage:
Replace calls to unix like commands with either CMake commands or the
appropriate command for the native platform.

Relates #2311


3rd_party/pull-eigen.cmake
```diff
@@ -0,0 +1,46 @@
...
+# limitation.
+#
+
+# Script to get the appropriate version of Eigen, if not already present.
+#
+# If updating this script ensure the license information is correct in the
+# licenses sub-directory.
+
+# This cmake script is expected to be called from a target or custom command with WORKING_DIRECTORY set to this files location
+#cd `dirname "$BASH_SOURCE"`
```
droberts195, line 18:
```suggestion
```

3rd_party/pull-eigen.cmake
```diff
@@ -0,0 +1,46 @@
...
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+
+# Script to get the appropriate version of Eigen, if not already present.
+#
+# If updating this script ensure the license information is correct in the
+# licenses sub-directory.
+
+# This cmake script is expected to be called from a target or custom command with WORKING_DIRECTORY set to this files location
```
droberts195, line 17:
```suggestion
# This cmake script is expected to be called from a target or custom command with WORKING_DIRECTORY set to this file's location
```

cmake/compiler/vs2019.cmake
```diff
@@ -15,20 +15,35 @@ else()
   set (VS2019_VARIABLES_ 1)
 endif()
 
-set(ROOT "/c")
+set(ROOT "c:")
```
droberts195, line 18: Windows drive letters are traditionally capital letters

```suggestion
set(ROOT "C:")
```

cmake/compiler/vs2019.cmake
```diff
@@ -15,20 +15,35 @@ else()
   set (VS2019_VARIABLES_ 1)
 endif()
```
droberts195, line 17:
```suggestion

# TODO: when everything else that needs to be consistent is ready, set this to the current drive instead of hard coding C:
```

cmake/variables.cmake
```diff
@@ -18,8 +18,11 @@ set(SNAPSHOT yes)
 
 set(ML_APP_NAME controller)
 
-execute_process(COMMAND awk -F= "/elasticsearchVersion/ {gsub(/-.*/,\"\"); print $2}"
-  $ENV{CPP_SRC_HOME}/gradle.properties OUTPUT_VARIABLE ML_VERSION_NUM OUTPUT_STRIP_TRAILING_WHITESPACE)
+file(READ ${CMAKE_SOURCE_DIR}/gradle.properties GRADLE_PROPERTIES)
+if(${GRADLE_PROPERTIES} MATCHES "elasticsearchVersion=([0-9\.]+)\n")
```
droberts195, line 22: Does this need to explicitly match the newline character?  I am thinking it would be better if it didn't.  As it stands it will break if somebody introduces a trailing space into the Gradle properties file.

lib/ver/CMakeLists.txt
```diff
@@ -12,14 +12,21 @@
...
-  $ENV{CPP_SRC_HOME}/gradle.properties OUTPUT_VARIABLE PRODUCT_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-if(${VERSION_QUALIFIER})
-  set(PRODUCT_VERSION "${PRODUCT_VERSION}-${VERSION_QUALIFIER}")
+string(TIMESTAMP BUILD_YEAR "%Y")
+
+if(MSVC)
+  set(USER_NAME $ENV{USERNAME})
+else()
+  set(USER_NAME $ENV{USER})
```
droberts195, line 20: I don't think `$USER` is reliable on Linux, in particular in Docker containers or when using `su` to change user.  Maybe we should still use `id` on *nix.

lib/ver/CMakeLists.txt
```diff
@@ -12,14 +12,21 @@
...
-execute_process(COMMAND id COMMAND awk -F ")" "{ print $1 }" COMMAND awk -F "(" "{ print $2 }" OUTPUT_VARIABLE USER_NAME
-  OUTPUT_STRIP_TRAILING_WHITESPACE)
-execute_process(COMMAND awk -F= "/elasticsearchVersion/ {gsub(/-.*/,\"\"); print $2}"
-  $ENV{CPP_SRC_HOME}/gradle.properties OUTPUT_VARIABLE PRODUCT_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-if(${VERSION_QUALIFIER})
-  set(PRODUCT_VERSION "${PRODUCT_VERSION}-${VERSION_QUALIFIER}")
+string(TIMESTAMP BUILD_YEAR "%Y")
+
+if(MSVC)
```
droberts195, line 17: nit: Technically this is a feature of Windows, not Visual Studio, so maybe we should check for Windows here. (Obviously it would only matter if we decided to build with something else on Windows in the future.)

droberts195 (commented):
The direction of travel is great.

I just commented on a few low level details.

cmake/variables.cmake
edsavage: Good point! I'll make the change.

lib/ver/CMakeLists.txt
edsavage: I'll change this to `WIN32`

droberts195 (approved):
LGTM

PR #2314, opened 2022-06-20
[ML] Fix edge cases for early stopping for training classification and regression models

tveasey:
We were running into an issue where we decide during coarse tuning that there was no advantage in adjusting a hyperparameter further, but this wasn't reflected in vectors we captured for testing whether to skip fine tuning parameters. This caused a dimension mismatch when we came to perform ANOVA, which blows up an Eigen assertion in debug builds, and also caused the results to be wrong in this case.

Since this fixes a problem with #2298, which is not released yet, I've marked as a non-issue.

### DIFF
--- a/include/maths/analytics/CBoostedTreeHyperparameters.h
+++ b/include/maths/analytics/CBoostedTreeHyperparameters.h
@@ -641,6 +641,7 @@ private:
     using TDoubleVecVec = std::vector<TDoubleVec>;
     using TDoubleDoubleSizeTupleVec = std::vector<std::tuple<double, double, std::size_t>>;
     using TOptionalVector3x1 = boost::optional<TVector3x1>;
+    using TIndexVec = std::vector<TVector::TIndexType>;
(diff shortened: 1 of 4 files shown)


droberts195:
The first commit contains a compilation error:

```
CBoostedTreeHyperparameters.cc:367:51: error: too few arguments to function call, expected 2, have 1
    this->checkIfCanSkipFineTuneSearch(numberTrees);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~            ^
/private/var/lib/jenkins/workspace/elastic+machine-learning+pull-request+macos-aarch64/include/maths/analytics/CBoostedTreeHyperparameters.h:665:10: note: 'checkIfCanSkipFineTuneSearch' declared here
    void checkIfCanSkipFineTuneSearch(const TIndexVec& relevantParameters,
         ^
```

tveasey:
> The first commit contains a compilation error:

Yes I actually made the original change on feature/incremental-learning branch and introduced the bug resolving conflicts.

valeriy42 (approved):
LGTM. Good catch! :rocket:

PR #2319, opened 2022-06-22
[ML] Improve performance of early stopping steps

valeriy42:
We can get away with fewer kernel parameter optimization rounds to speed up training. Instead of recalibrating GP kernel parameters every time we get a new observation, we spend a couple of additional optimization rounds before calculating the early stopping criterion.


tveasey (approved):
Just a couple of v.minor things. Otherwise, LGTM.

PR #2322, opened 2022-06-22
[ML] Merge CMake branch to main

droberts195:
The CMake build system is now mature enough to become our primary build system.

If there are no problems reported after a few days the old GNU make system can be deleted in a followup PR against the main branch.


dev-tools/docker_build.sh
```diff
@@ -79,7 +79,7 @@ do
     DOCKERFILE="$TOOLS_DIR/docker/${PLATFORM}_builder/Dockerfile"
     TEMP_TAG=`git rev-parse --short=14 HEAD`-$PLATFORM-$$
 
-    prefetch_docker_base_image "$DOCKERFILE"
+    #prefetch_docker_base_image "$DOCKERFILE"
```
edsavage, line 82: This should be uncommented

edsavage (approved):
LGTM

Just the one item to attend to

3rd_party/3rd_party.cmake
```diff
@@ -0,0 +1,267 @@
...
+  set(BOOST_LIBRARIES "chrono" "date_time" "filesystem" "iostreams" "log" "log_setup" "program_options" "regex" "system" "thread" "unit_test_framework")
+  set(XML_LOCATION "${LOCAL_DRIVE}/usr/local/bin")
+  set(XML_EXTENSION ".dll")
+  set(GCC_RT_LOCATION)
+  # Read VCBASE from environment if defined, otherwise default to VS Professional 2019
+  set(VCBASE "Program Files (x86)/Microsoft Visual Studio/2019/Professional")
+  if (DEFINED ENV{VCBASE})
+    set(VCBASE $ENV{VCBASE})
+  endif()
+  set(VCVER "14.29.30133")
```
droberts195, line 153: I just noticed this.  Presumably it's correct for the current CI workers, but it changes on every Visual Studio patch so we cannot have this hardcoded long term.

The way we used to get it was `/bin/ls -1 /c/PROGRA~2/MICROS~2/2019/Professional/VC/Tools/MSVC | tail -1`.

@edsavage please can you think of a way to get this dynamically in the future.  Effectively it needs to list the sub-directories of `Program Files (x86)/Microsoft Visual Studio/2019/Professional/VC/Tools/MSVC` in sorted order and set `VCVER` to the name of the last one.

edsavage: We can use `CMake`'s `file(GLOB)` command and  then selecting the last item in the returned list. In fact we already do that in `cmake\compilers\vs2019.cmake`, e.g.
```
file(GLOB MSVC_VERS "${VCBASE}/VC/Tools/MSVC/*")
list(GET MSVC_VERS -1 MSVC_VER)
if(${MSVC_VER} MATCHES "/([^/]+)$")
  set(VCVER ${CMAKE_MATCH_1})
endif()
```

PR #2325, opened 2022-06-23
[ML] Fix edge case early stopping when parameters are fixed during coarse tuning

tveasey:
There was a mistake when we decided to lock down hyperparameter values during coarse search. We would skip the parameter when initialising BO but retain the loss points. This change simply discards the points for which the parameter value doesn't match the fixed value. (It would be better to update the GP code so it can handle zero width dimensions but that is a larger change.)

We now also recheck the stopping condition after one round of fine-tune. At this point we have an estimate of cross-fold variation. Generally, non-zero variance increases the chance of stopping since maximum likelihood will prefer a flat process if the variation in the loss between parameter values is within their margin of error.

Finally, I use the several rounds of maximum likelihood when we estimate parameters in the fine-tune loop. This increases the chance that the estimates are fully converged before computing quantities from them.

This all follows on from #2298 so marking as a non-issue.

### DIFF
--- a/lib/maths/analytics/CBoostedTreeHyperparameters.cc
+++ b/lib/maths/analytics/CBoostedTreeHyperparameters.cc
@@ -665,11 +665,11 @@ void CBoostedTreeHyperparameters::checkRestoredInvariants(bool expectOptimizerIn
 
 std::size_t CBoostedTreeHyperparameters::estimateMemoryUsage() const {
     std::size_t numberToTune{this->numberToTune()};
-    return sizeof(*this) + numberToTune * sizeof(int) +
-           (m_NumberRounds / 3 + 1) * numberToTune * sizeof(double) +
+    return sizeof(*this) + numberToTune * sizeof(int) + // m_TunableHyperparameters
+           (m_NumberRounds / 3 + 1) * numberToTune * sizeof(double) + // m_HyperparameterSamples
            common::CBayesianOptimisation::estimateMemoryUsage(numberToTune, m_NumberRounds) +
-           numberToTune * sizeof(std::size_t) +
-           numberToTune * maxLineSearchIterations() * sizeof(double);
+           numberToTune * sizeof(std::size_t) + // m_LineSearchRelevantParameters
+           numberToTune * maxLineSearchIterations() * sizeof(double); // m_LineSearchHyperparameterLosses
 }
 
 std::size_t CBoostedTreeHyperparameters::memoryUsage() const {
--- a/lib/maths/analytics/CBoostedTreeHyperparameters.cc
+++ b/lib/maths/analytics/CBoostedTreeHyperparameters.cc
@@ -27,6 +27,7 @@
 #include <boost/optional/optional_io.hpp>
 
 #include <cmath>
+#include <limits>
 #include <memory>
 
 namespace ml {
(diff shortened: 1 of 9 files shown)


lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -402,52 +376,53 @@ void CBoostedTreeHyperparameters::initializeFineTuneSearch(double lossGap, std::
...
-    // value computed in fine tune. However, an estimate of the test train loss gap
-    // is needed if we incrementally train.
-    m_BestForestLossGap = lossGap;
-
-    // The stored number of trees is used for final train when we train on
-    // all the data and so can't measure when to stop.
-    std::size_t numberTreesToRestore{m_MaximumNumberTrees.value()};
-    m_MaximumNumberTrees.set(numberTrees);
-    this->saveCurrent();
-    m_MaximumNumberTrees.set(numberTreesToRestore);
```
valeriy42, line 450: Why did this become irrelevant?

lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -687,13 +668,15 @@ std::size_t CBoostedTreeHyperparameters::estimateMemoryUsage() const {
     return sizeof(*this) + numberToTune * sizeof(int) +
            (m_NumberRounds / 3 + 1) * numberToTune * sizeof(double) +
            common::CBayesianOptimisation::estimateMemoryUsage(numberToTune, m_NumberRounds) +
+           numberToTune * sizeof(std::size_t) +
            numberToTune * maxLineSearchIterations() * sizeof(double);
```
valeriy42, line 672: I am starting to struggle remembering what summand is responsible for what. Could you maybe add an explanation comment?

lib/maths/analytics/CBoostedTreeHyperparameters.cc
```diff
@@ -1141,16 +1083,55 @@ void CBoostedTreeHyperparameters::setHyperparameterValues(TVector parameters) {
...
 }
 
-void CBoostedTreeHyperparameters::resetBayesianOptimization() {
-    if (m_BayesianOptimization != nullptr) {
-        auto boundingBox = m_BayesianOptimization->boundingBox();
-        m_BayesianOptimization = std::make_unique<common::CBayesianOptimisation>(
-            boundingBox, m_BayesianOptimisationRestarts.value_or(
-                             common::CBayesianOptimisation::RESTARTS));
+template<typename F>
+void CBoostedTreeHyperparameters::foreachTunableParameter(const F& f) const {
```
valeriy42, line 1093: Nice job refactoring this function!

lib/maths/common/CBayesianOptimisation.cc
```diff
@@ -623,69 +627,79 @@ CBayesianOptimisation::minusExpectedImprovementAndGradient() const {
...
-    m_KernelParameters = amax.cwiseAbs();
-    LOG_TRACE(<< "kernel parameters = " << m_KernelParameters.transpose());
-    LOG_TRACE(<< "likelihood = " << -lmax);
+        // Ensure that kernel lengths are always positive. It shouldn't change the results
+        // but improves traceability.
+        m_KernelParameters = amax.cwiseAbs();
+        LOG_TRACE(<< "kernel parameters = " << m_KernelParameters.transpose());
+        LOG_TRACE(<< "likelihood = " << -lmax);
+    } while ([&] {
+        bool result = (--numberRounds > 0 &&
```
valeriy42, line 697: nice!

valeriy42 (approved):
LGTM. Just a few minor comments.

lib/maths/analytics/CBoostedTreeHyperparameters.cc
tveasey: I just moved it to `initializeFineTuneSearch`, so I could also call this function from `selectNext`.

lib/maths/analytics/CBoostedTreeHyperparameters.cc
tveasey: I added extra documentation in [bcdd707](https://github.com/elastic/ml-cpp/pull/2325/commits/bcdd707c74d865ff71ca348ac765e963a5f3b9f6).

PR #2327, opened 2022-06-27
[ML] Fix cross compilation for Linux aarch64 using CMake

droberts195:
Followup to #2322.

The changes are simple typos. They don't show up in a PR build
because we build linux-aarch64 natively in PR builds so that we
can run the tests on that platform.


edsavage (approved):
LGTM

droberts195:
The PR CI failed because of #2330, which is clearly not caused by the changes of this PR.

Since the problem this PR is fixing is preventing our main build job from running I am going to merge it to see what effect it has on the main jobs.

PR #2328, opened 2022-06-27
[ML] Remove remnants of old make system

edsavage:
Remove hierarchy of Makefiles and related shell scripts, which have been superseded by the new `CMake` build system.


3rd_party/dependency_report.sh
```diff
@@ -1,60 +0,0 @@
-#!/bin/bash
```
droberts195, line 1: This file is still needed.

3rd_party/pull-rapidjson.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
```
droberts195, line 1: This file is still needed.

devbin/model_extractor/.objs/.gitignore
```diff
@@ -1 +0,0 @@
-model_extractor
```
droberts195, line 1: Before removing the `devbin` make system it's probably worth creating the replacement CMake system for the development programs, so that the old system is visible for reference.

droberts195 (commented):
There are some bits in the `build-setup` instructions that can be deleted too - we no longer need to build GNU make as part of the Windows build setup and there might be some aliases, symlinks or renames that can be removed for the other platforms.

devbin/model_extractor/.objs/.gitignore
edsavage: Yes, I'm currently working on a PR to do that

build-setup/windows_simple.md
```diff
@@ -35,6 +35,5 @@ VCVER=`/bin/ls -1 /c/$VCBASE/VC/Tools/MSVC | tail -1`
 VCBINDIR=/c/$VCBASE/VC/Tools/MSVC/$VCVER/bin/HostX64/x64:/c/$VCBASE/Common7/IDE:/c/$WINSDKBASE/8.0/bin/x64:/c/$WINSDKBASE/8.0/bin/x86:/c/$VCBASE/TEAMTO~1/PERFOR~1/x64:/c/$VCBASE/TEAMTO~1/PERFOR~1
 export JAVA_HOME=/c/PROGRA~1/Java/jdk1.8.0_121
 export PATH="$CPP_SRC_HOME/build/distribution/platform/windows-x86_64/bin:$VCBINDIR:/mingw64/bin:$JAVA_HOME/bin:/c/usr/local/bin:/c/usr/local/lib:/bin:/c/Windows/System32:/c/Windows"
```
droberts195, line 37: I think we need `/c/PROGRA~1/CMake/bin` adding to the path here.

Also, on line 35 of `windows.md` I think the path is wrong - it includes `/c/PROGRA~1/CMake/bin/cmake` but isn't the final `cmake` in that the program rather than the directory?

droberts195 (commented):
Several of the build setup files contain the phrase "as the Machine Learning Makefiles expect it".  Please change that to "as the Machine Learning build system requires it" in all the places.

build-setup/clion/using_clion.md
```diff
@@ -1,13 +1,9 @@
 # Using CLion with `ml-cpp`
```
droberts195, line 2: I think we should say which version of CLion you need.

I still had CLion 2020 on my machine and it didn't import our CMake project because its embedded CMake was version 3.17, which was too old.

After I upgraded to CLion 2022 everything is fine, as it comes with CMake 3.22 embedded.

I don't know about CLion 2021.  If you still have it and it works then we can say the minimum requirement is CLion 2021.  Or if you can quickly find out in a Google search which version of CMake CLion 2021 had then again, if it's 3.19.2 or above we can say 2021 is the minimum.

Otherwise, let's say CLion 2022 is the minimum required. It's not worth spending ages checking CLion 2021 if that would be time consuming.

build-setup/clion/using_clion.md
```diff
@@ -20,164 +16,50 @@ by simply calling `g++` or `clang++`.
...
+for the `ml-cpp` project.
 
-In the area **Toolchain** we have to specify custom tools for building and cleaning the project.
+![Default Configuration Options](./default_project_config_options.png)
 
-For **Build:** click on **...** to open **External Tools** window and then on **+** to create a new external tool.
+By default `CLion` will opt for generating a `Ninja` build system. While this is a perfectly valid option, for
+better compatibility with building from the command line it is preferred to configure `CLion` to generate `Unix Makefiles`.
+To do this either select `Unix Makefiles` from the `Generator` drop-down or type `-G "Unix Makefiles"` in the
+`CMake options` field. Add `cmake_build` to the `Build directory` field and e.g. `-j 10` to the `Build options` field
```
droberts195, line 41: This is an interesting conundrum.

CLion defaults to `cmake-build-release` and `cmake-build-debug` for its CMake build directories.  Everything else we're doing is using `cmake_build`.  I am wondering if we should actually use `cmake-build-release` everywhere where we've used `cmake_build` currently.

If we recommend CLion users hardcode their config to `cmake_build` then when they select a debug build we'll be potentially mixing stuff in the CMake build directory with release builds done from the command line.

I think CLion has actually got it right by selecting different CMake build directories for debug and release to make switching between the two safer.

Do you agree or is there a good argument the other way?

If we think we might ever want to change then now is the time, before people get used to one directory name.  I'm not sure how much work it is to change `cmake_build` to `cmake-build-release` everywhere.  Hopefully with a bit of `find`, `xargs` and `sed` it's not too hard.

The `.gitignore` and `.dockerignore` files should also list both `cmake-build-release` and `cmake-build-debug` if we do this.

build-setup/vscode/using_vscode.md
```diff
@@ -2,45 +2,45 @@
 
 Previously, we explained how to [set up CLion for `ml-cpp`](../clion/using_clion.md). You may find, that CLion has some
 performance problems when indexing `ml-cpp` project: it may take very long to open the project or CLion may freeze and
```
droberts195, line 4: I found CLion to be completely hopeless at reindexing the repo when we were using the old Makefiles solution.  I always had to kill it and regenerate `compile_commands.json` from scratch after a non-trivial change in the repo.

But with CMake it seems a lot better.  So maybe we could take out the bit about CLion freezing now.

build-setup/clion/using_clion.md
edsavage: That's an interesting point, yes I agree that it makes sense to go with the CLion naming scheme (plus its one less thing to configure when getting set up). I'll make the changes now.

build-setup/vscode/using_vscode.md
```diff
@@ -2,45 +2,45 @@
...
 
 In **Settings**, make sure *C_Cpp:Autocomplete* and *C_Cpp:Intelli Sense Engine* from the native C++ plugin of VS Code
 are disabled. Note that we don't want to disable the complete C/C++ Extension, since we are still going to use the
 debugging capabilities.
 
-Now for the `clangd` plugin, make sure that you have the correct path specified in **Settings** and *Clangd: Path*.
-*Clangd: Arguments* can look something like this:
+Now for the `CMake` plugin, change the `Build directory` specified in **Settings** > **Extensions** >
+**CMake Tools** to read `${workspaceFolder}/cmake_build`. You may also wish to set certain environment
+variables for the build. e.g. `ML_DEBUG=1`. To do so, under `Cmake: Build Environment` click the `Add item`
```
droberts195, line 26: Just seeing this here now makes me think we should change the `ML_DEBUG` functionality to instead be what you get when you do a debug build in CMake.  Then instead of setting environment variables a CLion user would just change the build type. Don't change it as part of this PR though.  We can come back to this at some time in the future.  It's not essential as hardly anyone does debug builds in reality, and it's also got dependencies on the Jenkins jobs, so needs some coordination there.

droberts195 (approved):
LGTM if CI passes

PR #2329, opened 2022-06-27
[ML] Alter build.gradle to allow for native windows builds

edsavage:
Set up the minimal environment variables to allow for building from a
native windows command prompt.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,7 +117,7 @@ clean {
     exec {
       environment makeEnvironment
       commandLine shell
-      args shell_flag, "${set_env} && set &&  + ${clean_cmd}"
+      args shell_flag, "${set_env} && ${clean_cmd}"
       workingDir "${projectDir}"
     }
   }
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ if ((isWindows && msystem == 'MINGW64') || isMacOsX || isLinux)
   project.ext.shell = "cmd.exe"
   project.ext.set_env = ".\\set_env.bat"
   project.ext.shell_flag = "/c"
-  project.ext.clean_cmd = "rmdir /s /q cmake_build 3rd_party\\eigen"
+  project.ext.clean_cmd = "if exist cmake_build\\ (rmdir /s /q cmake_build) && if exist 3rd_party\\eigen (rmdir /s /q 3rd_party\\eigen)"
   // Stripping the binaries is not necessary on windows, execute a no-op command instead
   project.ext.strip_cmd = "cd ."
 }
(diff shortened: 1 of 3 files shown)


set_env.bat
```diff
@@ -0,0 +1,29 @@
...
+cmd /c "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+
+REM Set %CPP_SRC_HOME% to be an absolute path to this script's location, as
+REM different builds will come from different repositories and go to different
+REM staging areas
+SET MY_DIR=%~dp0
+SET CPP_SRC_HOME=%~dp0
+
+REM Logical filesystem root
+if NOT "%LOCAL_DRIVE%"== "" (SET ROOT=%LOCAL_DRIVE%) else (SET ROOT=C:)
```
droberts195, line 25: Instead of hardcoding `C:` this should be able to use `%CD:~0,2%`.  Then it really is the local drive, whatever that happens to be.

This is what I was getting at in the comments about this on other PRs - using native Windows functionality we should not have to specify the current drive letter because Windows knows what it is.  It's a side effect of using Cygwin/MinGW that the paths incorporate the drive letters in a way that's not natural to Windows.

set_env.bat
```diff
@@ -0,0 +1,29 @@
...
+
+REM Set up a build environment, to ensure repeatable builds
+
+REM Initialize the Visual Studio command prompt environment variables
+cmd /c "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+
+REM Set %CPP_SRC_HOME% to be an absolute path to this script's location, as
+REM different builds will come from different repositories and go to different
+REM staging areas
+SET MY_DIR=%~dp0
```
droberts195, line 21: `MY_DIR` isn't needed here.  In `set_env.sh` it's used to simplify away symlinks when setting `CPP_SRC_HOME`, but used like it is here it's pointless.

build.gradle
```diff
@@ -99,39 +116,39 @@ clean {
   doLast {
     exec {
       environment makeEnvironment
-      commandLine bash
-      args '-c', 'source ./set_env.sh && rm -rf 3rd_party/eigen cmake_build'
+      commandLine shell
+      args shell_flag, "${set_env} && set &&  + ${clean_cmd}"
```
droberts195, line 120: The `set` command in here should be removed before merging.  Also, is the `+` a typo?

edsavage: Yep, the `+` was a hangnail from the previous construction of the commandline

PR #2331, opened 2022-06-27
[ML] Add CMake targets for "dev only" binaries

edsavage:


devlib/vflib/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+# files subject to the Elastic License 2.0 may only be used in production when
+# invoked by an Elasticsearch process with a license key installed that permits
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+
+project("ML vflib")
+
+include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/devinclude)
```
droberts195, line 14: I don't think our own headers should be treated as system headers.

droberts195 (approved):
LGTM

PR #2332, opened 2022-06-27
[ML] Address some bottlenecks training classification and regression models on large data sets

tveasey:
Investigating slowdowns on a couple of large QA data sets as a result of #2298 showed up some interesting things. We spend a lot of time updating quantiles and recomputing split indices for the data sets which slowed down. These data sets have relatively many metric features and the runtime share of these operations is larger in such cases. This change introduces some changes to address long run times in such cases. 

I'd planned to use a fast version of the quantile sketch, but a silly mistake meant it never actually got wired in. I've also made a further tweak to the fast version which gives it another 50% performance boost at the expense of slightly changing its behaviour. I tested accuracy fairly extensively for the new scheme, which still often produces the same results, and found no evidence for significantly different accuracy.

For many features we can gain a lot of performance with v.little impact on the QoR by updating candidate splits less frequently. This imposes a new limit on the frequency we fresh them based on feature count.

For large data sets we were allowing individual trees to become v.large. Such large trees both stresses our inference, we have to limit the model size based on its memory anyway, and means train time can become too high. I've capped the maximum tree size independent of the data set size. The limit is still pretty large and I measured no significant impact in QoR from this change.

### DIFF
--- a/lib/maths/analytics/unittest/CBoostedTreeTest.cc
+++ b/lib/maths/analytics/unittest/CBoostedTreeTest.cc
@@ -845,7 +845,7 @@ BOOST_AUTO_TEST_CASE(testNonUnitWeights) {
     LOG_DEBUG(<< "biasWithWeights    = " << biasWithWeights
               << ", rSquaredWithWeights    = " << rSquaredWithWeights);
 
-    BOOST_TEST_REQUIRE(std::fabs(biasWithWeights) < 0.2 * std::fabs(biasWithoutWeights));
+    BOOST_TEST_REQUIRE(std::fabs(biasWithWeights) < 0.25 * std::fabs(biasWithoutWeights));
     BOOST_TEST_REQUIRE(1.0 - rSquaredWithWeights < 0.8 * (1.0 - rSquaredWithoutWeights));
 }
 
@@ -1010,9 +1010,11 @@ BOOST_AUTO_TEST_CASE(testHoldoutRowMask) {
     auto frame = core::makeMainStorageDataFrame(cols, rows).first;
     fillDataFrame(rows, 0, cols, x, noise, target, *frame);
 
+    // We disable early stopping because we test hold losses from fine tuning.
     auto regression = maths::analytics::CBoostedTreeFactory::constructFromParameters(
                           1, std::make_unique<maths::analytics::boosted_tree::CMse>())
                           .numberHoldoutRows(numberHoldoutRows)
+                          .stopHyperparameterOptimizationEarly(false)
                           .buildForTrain(*frame, cols - 1);
 
     regression->train();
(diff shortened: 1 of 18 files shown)


lib/maths/common/unittest/CBasicStatisticsTest.cc
```diff
@@ -1121,63 +1121,14 @@ BOOST_AUTO_TEST_CASE(testCovariancesLedoitWolf) {
...
-        double sample[] = {3.0, 5.0, 10.0, 2.0, 6.0, 7.0, 1.0, 9.0, 4.0, 8.0};
-
-        maths::common::CBasicStatistics::TDoubleVec sampleVec(
-            sample, sample + sizeof(sample) / sizeof(sample[0]));
-
-        double median = maths::common::CBasicStatistics::median(sampleVec);
-
-        BOOST_REQUIRE_EQUAL(5.5, median);
-    }
+    BOOST_REQUIRE_EQUAL(0.0, maths::common::CBasicStatistics::median({}));
```
valeriy42, line 1124: nice simplification!

lib/maths/common/CQuantileSketch.cc
```diff
@@ -571,49 +588,68 @@ void CFastQuantileSketch::debugMemoryUsage(const core::CMemoryUsage::TMemoryUsag
...
 
-        this->reduce(m_Rng, m_MergeCosts, m_MergeCandidates, m_Stale);
-    }
-}
+        std::size_t numberToMerge{knots.size() - this->fastReduceTargetSize()};
 
-std::size_t CFastQuantileSketch::target() const {
-    return static_cast<std::size_t>(
-        m_ReductionFraction * static_cast<double>(this->maxSize()) + 1.0);
+        // This is pseudo greedy since unlike CQuantileSketch which is greedy.
```
valeriy42, line 630:
```suggestion
        // This is pseudo greedy unlike CQuantileSketch which is greedy.
```

lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -1618,12 +1620,8 @@ BOOST_AUTO_TEST_CASE(testMseIncrementalAddNewTrees) {
               << ", 5 new trees = " << testError5 << ", 10 new trees = " << testError10);
     // The initial model has too little capacity so we should get substantial
     // improvements for adding trees.
-    BOOST_TEST_REQUIRE(0.95 * testError0 >= testError5);
+    BOOST_TEST_REQUIRE(0.9 * testError0 >= testError5);
     BOOST_TEST_REQUIRE(0.9 * testError0 >= testError10);
```
valeriy42, line 1624: I guess, in this form we could remove `testError10` altogether to simplify.

tveasey: Actually good point. I think 5 vs 10 is just a fragile. This has failed for more under perturbations before. You have to arrange for the necessary capacity to be just right. Perhaps if we overrode more hyperparameter values we could make it more reliable, but I think 0 vs 5 is sufficient if you're happy?

lib/maths/common/CQuantileSketch.cc
tveasey: See [3671abb](https://github.com/elastic/ml-cpp/pull/2332/commits/3671abbe45844fe59591aeeb40ad435bf93c27ec).

valeriy42 (approved):
LGTM

PR #2333, opened 2022-06-27
[ML] Fixing CMake syntax for setting up debug builds

droberts195:
The CMake code to add debug and code coverage arguments was not
correctly checking the appropriate environment variables.


droberts195:
The first attempt doesn't work - it makes _all_ Linux builds into debug builds.  There's a difference between how Make and CMake treat environment variables that are defined but have empty values.

edsavage (approved):
LGTM

edsavage (approved):

PR #2334, opened 2022-06-27
[ML] Potential fix for RPATH issue for cross compiled macos builds

edsavage:
Unconditionally set `CMAKE_MACOSX_RPATH` to true for all shared libraries and executables


3rd_party/3rd_party.cmake
```diff
@@ -38,11 +38,11 @@ set(ARCH ${HOST_SYSTEM_PROCESSOR})
 set(CMAKE_SYSTEM_NAME ${HOST_SYSTEM_NAME})
```
droberts195, line 38: Why do we change the value of a built in CMake variable?  Can't we just use `${HOST_SYSTEM_NAME}` in the code below instead of using a built in variable with a changed value?

droberts195: Or could we just use `CMAKE_HOST_SYSTEM_NAME` throughout the code below so that we're using a built in variable that people won't have to check the meaning of?

3rd_party/3rd_party.cmake
```diff
@@ -38,11 +38,11 @@ set(ARCH ${HOST_SYSTEM_PROCESSOR})
 set(CMAKE_SYSTEM_NAME ${HOST_SYSTEM_NAME})
 
 message(STATUS "CMAKE_SYSTEM_NAME = ${CMAKE_SYSTEM_NAME}")
```
droberts195, line 40: This line is _mega_ confusing when you read it in the build log, because it makes it look like a built in variable has the wrong value.

3rd_party/3rd_party.cmake
```diff
@@ -38,11 +38,11 @@ set(ARCH ${HOST_SYSTEM_PROCESSOR})
 set(CMAKE_SYSTEM_NAME ${HOST_SYSTEM_NAME})
 
 message(STATUS "CMAKE_SYSTEM_NAME = ${CMAKE_SYSTEM_NAME}")
-if (CMAKE_SYSTEM_NAME STREQUAL "darwin")
+if ("${CMAKE_SYSTEM_NAME}" STREQUAL "darwin")
```
droberts195, line 41: In many other files the `if (variable STREQUAL "literal")` syntax is used.  Why change to `if ("${variable}" STREQUAL "literal")` here?  I think it would be best to find out which is considered best practice and use the same style everywhere.

PR #2337, opened 2022-06-28
[ML] Fix glob pattern in 3rd_party.cmake

edsavage:
A glob pattern in `3rd_party.cmake` was too wide ranging, it has been
narrowed to be more precise in the files it matches.


edsavage:
Relates https://github.com/elastic/elasticsearch/issues/88090

droberts195 (approved):
LGTM

Once this is merged into the feature branch together with all the other open PRs we can do some due diligence on a Linux dev machine to check that everything is now as expected.

PR #2338, opened 2022-06-29
[ML] Only set CMAKE_SYSTEM_NAME when cross compiling

edsavage:


cmake/darwin-x86_64.cmake
```diff
@@ -9,15 +9,13 @@
...
-set(CMAKE_SYSTEM_NAME Darwin)
-
-message(STATUS "CMAKE_SYSTEM_NAME ${CMAKE_SYSTEM_NAME}")
-
 set(CPP_PLATFORM_HOME $ENV{CPP_SRC_HOME}/build/distribution/platform/darwin-x86_64)
 
 if(DEFINED ENV{CPP_CROSS_COMPILE} AND "$ENV{CPP_CROSS_COMPILE}" STREQUAL "macosx")
 
+  # the name of the target operating system
+  set(CMAKE_SYSTEM_NAME Darwin)
```
droberts195, line 17:
```suggestion
  set(CMAKE_SYSTEM_NAME Darwin)
  # Darwin 18 is Mojave (macOS 10.14), which is what our macOS cross compilation environment currently uses.
  # Should be incremented when we upgrade the macOS cross compilation environment.
  set(CMAKE_SYSTEM_VERSION 18.7.0)
```

cmake/linux-aarch64.cmake
```diff
@@ -9,14 +9,12 @@
...
-# the name of the target operating system
-set(CMAKE_SYSTEM_NAME Linux)
-
-message(STATUS "CMAKE_SYSTEM_NAME ${CMAKE_SYSTEM_NAME}")
-
 set(CPP_PLATFORM_HOME $ENV{CPP_SRC_HOME}/build/distribution/platform/linux-aarch64)
 
 if(DEFINED ENV{CPP_CROSS_COMPILE} AND "$ENV{CPP_CROSS_COMPILE}" STREQUAL "aarch64")
+  # the name of the target operating system
+  set(CMAKE_SYSTEM_NAME Linux)
```
droberts195, line 16:
```suggestion
  set(CMAKE_SYSTEM_NAME Linux)
  # This is the system version of CentOS 7, which is what we currently cross compile for Linux on.
  # Should be incremented if the cross compile Docker image for linux-aarch64 is rebuilt.
  set(CMAKE_SYSTEM_VERSION 3.10.0-514.6.1.el7.x86_64)
```

droberts195 (commented):
The [docs](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html) say that if `CMAKE_SYSTEM_NAME` is set then `CMAKE_SYSTEM_VERSION` must also be set.

droberts195 (approved):
LGTM

PR #2341, opened 2022-07-01
[ML] Fix and simplify the way the version file is created

droberts195:
The version file was not being created correctly in all
scenarios following the CMake changes:

1. The "-SNAPSHOT" suffix was not always appropriately
   added to the version
2. Docker builds were considered to have uncommitted
   changes, so had a version of "DEVELOPMENT BUILD by root"

In addition to fixing those problems this PR also moves
away from making substitutions into a template to generate
the version library source code. Instead the necessary
information is passed using preprocessor macros.


edsavage (approved):
LGTM - Although it looks like the BUILD_YEAR is not being substituted correctly, causing the test case failure.

include/core/CStringUtils.h
```diff
@@ -272,7 +272,7 @@ class CORE_EXPORT CStringUtils : private CNonInstantiatable {
 //! Macro to convert a pre-processor symbol to a string constant - has to be
 //! done in a macro unfortunately as the # operator is only recognised by the
 //! pre-processor.
-#define STRINGIFY_MACRO(str) (#str)
+#define STRINGIFY_MACRO(str) #str
```
edsavage, line 275:
```suggestion
#define STRINGIFY_MACRO(str) XSTRINGIFY_MACRO(str)
#define XSTRINGIFY_MACRO(str) #str
```

droberts195: Thanks

PR #2342, opened 2022-07-04
[ML] Remove pre-tokenized categorization performance test

droberts195:
We had a test that asserted that categorization was faster
with pre-tokenized input than with tokenization done in the
C++ code.

With the recent CMake changes it's no longer possible to make
assertions about performance like this, because tests now run
in parallel with compilation, so the other activity on the
machine can distort the performance of different sections of
the test.

An example of a spurious failure is:
https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+main+windows+debug/337/testReport/junit/lib.model.CTokenListDataCategorizerTest/C__Users_jenkins_workspace_elastic+machine-learning+main+windows+debug_lib_model_unittest_CTokenListDataCategorizerTest/testPreTokenisedPerformance/

This test served its purpose at the time that pre-tokenized
input was introduced, but we'll never return to tokenizing for
categorization in the C++ code, so there's no value in
retaining this test now that it is doomed to be flakey.

### DIFF
--- a/lib/model/unittest/CTokenListDataCategorizerTest.cc
+++ b/lib/model/unittest/CTokenListDataCategorizerTest.cc
@@ -669,65 +669,6 @@ BOOST_FIXTURE_TEST_CASE(testPreTokenised, CTestFixture) {
     checkMemoryUsageInstrumentation(categorizer);
 }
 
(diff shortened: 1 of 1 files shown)


dimitris-athanasiou (approved):
LGTM

PR #2343, opened 2022-07-04
[ML] Minor tidyups to the CMake build

droberts195:
* The old 3rd_party.sh used to have an option to check if the
  required 3rd party libraries were installed on the machine, but
  not add them to the build. The replacement script,
  3rd_party.cmake, cannot do such a check, but still carried over
  some code related to it. Since that functionality is not used
  the script can be simplified by deleting the superfluous code.
* Some comments were incorrect in the version library header
  following replacement of the templates with simple macros.
* The ML_KEEP_GOING environment variable no longer exists.
* The make_info_plist.sh script is no longer used.
* We don't want to pick up the INCLUDE and LIBPATH environment
  variables on Windows.
* Now that unit tests run in parallel, different test suites need to
  use different test files. Named pipes are one example (though
  there may be more cases like this to fix in the future).


3rd_party/3rd_party.cmake
```diff
@@ -220,53 +224,32 @@ function(install_libs _target _source_dir _prefix _postfix)
   endif()
 endfunction()
 
-function(third_party _arg)
-  if("${_arg}" STREQUAL "check")
-    set(INSTALL_DIR "")
```
droberts195, line 225: This didn't work at all any more, so I've removed it and simplified the rest of the file as a result.

This file is best reviewed ignoring whitespace differences.

edsavage (approved):
LGTM

PR #2344, opened 2022-07-04
[ML] Windows build fixes

edsavage:
* Allow for Debug configuration builds
* Do not perform incremental builds
* Typo in lib/api/dump_state/CMakeLists.txt

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -157,6 +157,7 @@ function(ml_add_non_distributed_library _target _type)
         "${COVERAGE}")
     elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
       target_link_libraries(${_target} PRIVATE
+        "INCREMENTAL:NO"
         "-OPT:REF")
     endif()
   endif()
@@ -207,6 +208,7 @@ function(ml_add_library _target _type)
         "${COVERAGE}")
     elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
       target_link_libraries(${_target} PRIVATE
+        "INCREMENTAL:NO"
         "-OPT:REF")
     endif()
 
@@ -271,6 +273,7 @@ function(ml_add_executable _target)
       )
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     target_link_libraries(${_target} PRIVATE
+      "INCREMENTAL:NO"
       "-OPT:REF")
   endif()
 
@@ -309,6 +312,7 @@ function(ml_add_non_distributed_executable _target)
       )
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     target_link_libraries(${_target} PRIVATE
+      "INCREMENTAL:NO"
       "-OPT:REF")
   endif()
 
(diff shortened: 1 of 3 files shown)


cmake/functions.cmake
```diff
@@ -157,6 +157,7 @@ function(ml_add_non_distributed_library _target _type)
         "${COVERAGE}")
     elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
       target_link_libraries(${_target} PRIVATE
+        "INCREMENTAL:NO"
```
droberts195, line 160: I think the `Release` target adds this automatically and it's probably fine that debug builds don't have it.

edsavage: Without it the linker warns that:
```
LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/OPT:REF' specification [C:\Users\eds\src\ml-cpp\cmake-build-debug\lib\core\MlCore.vcxproj]
```
So its inclusion was just to quieten the linker warning. .. I'll remove it as its not necessary

droberts195: Oh, I see. Well then the solution is probably not to do `OPT:REF` either for debug builds. It makes sense that debug builds don't do optimisations. Does the syntax you used to only include `/O2` in release builds also work for linker arguments?

edsavage: > Does the syntax you used to only include `/O2` in release builds also work for linker arguments?

Yes, should do, I'll give it a whirl anyway

cmake/compiler/msvc.cmake
```diff
@@ -11,7 +11,7 @@
 
 
 if(NOT CRT_OPT)
-  set(CRT_OPT /MD)
+  set(CRT_OPT $<$<CONFIG:Release>:/MD>)
```
droberts195, line 14: We have to always use `/MD`, or else build different versions of our 3rd party dependencies like Boost and PyTorch.

The standard Windows way of doing things would be to use `/MDd` when doing a debug build, but that would require linking against Boost, PyTorch, etc, that were also built with `/MDd`.  It would be a lot of work to build two sets of all our 3rd party dependencies, especially given that we almost never debug anything on Windows.

So it would be better if instead of doing this we could tell CMake that we want it to use `/MD` instead of `/MDd` for debug builds on Windows.  Or, if that's too hard, just revert this line as the rest of the PR is making things better than they were.

edsavage: Ah ok, that makes sense - I'll back that change out.

cmake/variables.cmake
```diff
@@ -166,7 +166,7 @@ if(NOT "$ENV{ML_DEBUG}")
       list(APPEND ML_COMPILE_DEFINITIONS "_FORTIFY_SOURCE=2" "_REENTRANT")
```
droberts195, line 166: I just noticed something else that's wrong here: `_REENTRANT` should always be defined when building for Linux, whether it's a debug build or not.

In other words, `_REENTRANT` should be added to the compile definitions in a block that's only guarded by `CMAKE_SYSTEM_NAME STREQUAL "Linux"`.

Since this is a Windows PR I'm happy to leave this for a followup, but it needs fixing at some point.

PR #2345, opened 2022-07-05
[ML] Fix ML_DEBUG=1 behaviour

edsavage:
Perform a debug build based on the value of the ML_DEBUG environment variable in keeping with historical build behaviour


cmake/variables.cmake
```diff
@@ -153,13 +153,33 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
   list(APPEND ML_COMPILE_DEFINITIONS BOOST_ALL_NO_LIB)
 endif()
 
-if(UNIX AND "$ENV{ML_DEBUG}" AND "$ENV{ML_COVERAGE}")
+# Perform a "Release" build by default...
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
+message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
```
droberts195, line 161: This should go after line 169 so we find out the final value rather than one that potentially gets overridden.

cmake/variables.cmake
```diff
@@ -153,13 +153,33 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
+
+# However, to keep in step with our historical
+# build system if ML_DEBUG is set to a boolean
+# true value, e.g. ML_DEBUG=1, then override
+# the build type to be Debug
+if(DEFINED ENV{ML_DEBUG} AND $ENV{ML_DEBUG})
```
droberts195, line 167: This causes a syntax error if `$ENV{ML_DEBUG}` evaluates to nothing:

```
CMake Error at cmake/variables.cmake:167 (if):
  if given arguments:

    "DEFINED" "ENV{ML_DEBUG}" "AND"

  Unknown arguments specified
```

Since an empty string evaluates to false I think this is the way to do it:

```suggestion
if("$ENV{ML_DEBUG}")
```

cmake/variables.cmake
```diff
@@ -153,13 +153,33 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
 
-if(NOT "$ENV{ML_DEBUG}")
+if(CMAKE_BUILD_TYPE STREQUAL Debug)
+  message(STATUS "IN DEBUG MODE")
+else()
+  message(STATUS "NOT IN DEBUG MODE")
+  message(STATUS "Adding NDEBUG and EXCLUDE_TRACE_LOGGING to ML_COMPILE_DEFINITIONS")
   list(APPEND ML_COMPILE_DEFINITIONS "NDEBUG" "EXCLUDE_TRACE_LOGGING")
   if(UNIX)
+    message(STATUS "Adding -O3 to OPTCFLAGS")
```
droberts195, line 182: `-O3` is showing up twice on the compilation lines.  I think it's because we're appending `OPTCFLAGS` onto the final compilation flags in two places:

```
bash-3.2$ find . -name '*.cmake' -type f | xargs grep OPTCFLAGS
./cmake/compiler/clang.cmake:  ${OPTCFLAGS}
./cmake/compiler/msvc.cmake:  ${OPTCFLAGS}
./cmake/compiler/gcc.cmake:  ${OPTCFLAGS}
./cmake/variables.cmake:    list(APPEND OPTCFLAGS "-O3")
./cmake/variables.cmake:      list(APPEND OPTCFLAGS "-Wdisabled-optimization")
./cmake/variables.cmake:    set(OPTCFLAGS $<$<CONFIG:Release>:/O2> "/Qfast_transcendentals" "/Qvec-report:1")
./cmake/variables.cmake:list(APPEND ML_CXX_FLAGS ${OPTCFLAGS})
```

Since CMake can append to lists, maybe we don't even need an `OPTCFLAGS` variable now?  Or even if we do still want to separate out the optimization flags we should only append them once.

Maybe it would be cleaner to append all the compiler-dependent arguments in the `compiler/*.cmake` files using the `$<$<CONFIG:Release>:-O3>` syntax to select the ones appropriate to the current build type?

droberts195: Also, related to this, `msvc.cmake` references a variable `${OPTCPPFLAGS}` that I don't think is set anywhere.

We seem to have lost some Eigen-related Windows optimization flags from https://github.com/elastic/ml-cpp/blob/c86b7174f20c427c0ce3b19f6a280edbaefe7773/mk/windows.mk#L102

The ones that I cannot see any more are:

> `-DEIGEN_VECTORIZE_SSE3 -DEIGEN_VECTORIZE_SSE4_1 -DEIGEN_VECTORIZE_SSE4_2`

droberts195:
Another setting that's been lost from the make system is `-STACK:0x800000` from: https://github.com/elastic/ml-cpp/blob/c86b7174f20c427c0ce3b19f6a280edbaefe7773/mk/windows.mk#L112

Without that I think the unit tests will run out of stack space on Windows.

By default the stack is smaller on Windows than on macOS and Linux, but we set it to the same size for all platforms.

cmake/variables.cmake
```diff
@@ -153,13 +153,33 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
+# However, to keep in step with our historical
+# build system if ML_DEBUG is set to a boolean
+# true value, e.g. ML_DEBUG=1, then override
+# the build type to be Debug
+if("$ENV{ML_DEBUG}")
+  set(CMAKE_BUILD_TYPE Debug)
```
droberts195, line 166: I have a slight worry that this is going to cause CMake to want to use the `/MDd` option on Windows instead of `/MD`.  We always want to use `/MD`.

It seems like CMake has support for specifying which Microsoft C runtime library you want using the variable [`CMAKE_MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html).

So maybe we should replace the `CRT_OPT` stuff in `msvc.cmake` with this:

```
if (NOT CMAKE_MSVC_RUNTIME_LIBRARY)
  set(CMAKE_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
endif()
```

By telling CMake which variant we want in its own way hopefully there's no danger of some per-build type default kicking in.

cmake/variables.cmake
droberts195: If it's cleaner it wouldn't hurt to define `-DEIGEN_VECTORIZE_SSE3 -DEIGEN_VECTORIZE_SSE4_1 -DEIGEN_VECTORIZE_SSE4_2` on all platforms. We want to use these instructions on all platforms. With clang and gcc Eigen can define them itself based on the compiler flags. With Visual C++ it cannot. But it wouldn't hurt for us to define these ourselves always.

cmake/compiler/msvc.cmake
```diff
@@ -31,7 +30,6 @@ list(APPEND ML_COMPILE_DEFINITIONS
 list(APPEND ML_C_FLAGS
   "/X"
   "/nologo"
-  ${OPTCFLAGS}
   "/W4"
   ${CRT_OPT}
```
droberts195, line 34: This line should be deleted.

build.gradle
```diff
@@ -126,14 +140,14 @@ clean {
...
-  args shell_flag, "${set_env} && ${cmake} -B cmake-build-release ${cmakeFlags}"
+  args shell_flag, "${set_env} && ${cmake} -B ${cmake_build_dir} ${cmakeFlags}"
   workingDir "${projectDir}"
 }
 
 task compile(type: Exec) {
   environment makeEnvironment
   commandLine shell
-  args shell_flag, "${set_env} && ${cmake} --build cmake-build-release --config Release -t install -v -j ${numCpus}"
+  args shell_flag, "${set_env} && ${cmake} --build ${cmake_build_dir} --config ${cmake_build_type} -t MlSeccomp -v -j ${numCpus}"
```
droberts195, line 150: It looks like changing `install` to `MlSeccomp` was just for debugging and should be reverted now.

build.gradle
```diff
@@ -78,7 +78,21 @@ if (isWindows) {
...
+project.ext.cmake_build_dir = "cmake-build-release"
+project.ext.cmake_build_type = "Release"
+if (mlDebug == "1")
+{
+  cmakeFlags = cmakeFlags + ' -D CMAKE_BUILD_TYPE=Debug'
+  project.ext.cmake_build_dir = "cmake-build-debug"
+  project.ext.cmake_build_type = "Debug"
+} else {
+  cmakeFlags = cmakeFlags + ' -D CMAKE_BUILD_TYPE=Release'
+}
```
droberts195, line 94: I would prefer not to duplicate as much logic between the CMake files and build.gradle.  It's just more to remember to update if we ever change anything in the future.

The CMake files already choose the build type based on the value of `ML_DEBUG`, so why not leave that aspect up to the CMake files.

It seems all we really need to do here is select the correct CMake build directory and then should leave the rest up to the existing logic in the CMake files.

Also the `mlDebug == 1` condition is not ideal, as now that variable is being interpreted in 3 different ways in the CMake system, shell scripts and in this file.  CMake will treat `on`, `yes`, `true`, `y` or any non-zero number the same whereas this only accepts `1`.  Shell scripts treat anything with non-zero length as true.  I think this should match one or other of those rather than introducing a 3rd interpretation.  (Ideally all 3 would interpret it the same.)

Also, we need similar logic in `docker_entrypoint.sh` - that should now leave the build type up to the CMake system to decide.  For `docker_entrypoint.sh` I think it would also be fine to use `cmake-build-docker` as a single CMake build directory instead of trying to determine whether to use release or debug.  The container is entirely throwaway, so it's not like Gradle where somebody may want to share cache variables between CLion and Gradle for example.

droberts195 (commented):
I think you also need to change `cmake-build-release` to `cmake-build-*` in two places in `dev-tools/jenkins_ci.sh`.

build.gradle
edsavage: oops! 🤦

build.gradle
```diff
@@ -78,7 +78,17 @@ if (isWindows) {
...
 
-String cmakeFlags = '--no-warn-unused-cli -D CMAKE_BUILD_TYPE=Release -D CMAKE_TOOLCHAIN_FILE=cmake/' + artifactClassifier + '.cmake'
+// Ensure we have a consistent combination of build config type and build directory
+// based on whether we want a debug build or not.
+// By default a "Release" build is performed.
+String cmakeFlags = '--no-warn-unused-cli -D CMAKE_TOOLCHAIN_FILE=cmake/' + artifactClassifier + '.cmake'
+project.ext.cmake_build_dir = "cmake-build-release"
+project.ext.cmake_build_type = "Release"
+if (mlDebug.toBoolean())
+{
```
droberts195, line 88:
```suggestion
if (mlDebug.toBoolean()) {
```

droberts195 (approved):
LGTM

Let's try and get this merged today and then see what happens with tonight's debug builds.

PR #2347, opened 2022-07-06
[ML] More CMake tidyups

droberts195:
1. We no longer need references to make as we always use cmake
2. Use camelCase for variables in build.gradle
3. Don't require that BOOST_ROOT be an environment variable when
   building natively on Windows
4. Remove the last occurrence of $ML_KEEP_GOING


build.gradle
```diff
@@ -428,7 +425,8 @@ task buildDependencyReport(type: Exec) {
   outputs.file("${buildDir}/distributions/dependencies-${version}.csv")
   environment makeEnvironment
   commandLine shell
-  args '-c', "source ./set_env.sh && 3rd_party/dependency_report.sh --csv \"${outputs.files.singleFile}\""
+  // TODO: this won't work on Windows outside of Git bash
+  args shellFlag, "${setEnv} && 3rd_party/dependency_report.sh --csv \"${outputs.files.singleFile}\""
```
edsavage, line 429: I'm about to raise a PR which replaces this bash script with a portable cmake equivalent

PR #2348, opened 2022-07-07
[ML] Clean up CI test reports

droberts195:
Currently our CI test reports are not great.

- They mix the results from the three maths libraries together,
  which confuses the timings.
- Tests that don't make any assertions cause a very weird test
  result structure.

Screenshot 2022-07-07 at 10 21 23

This change:

- Splits out the results from the three maths libraries into
  different packages.
- Ensures every test makes at least one assertion.

### DIFF
--- a/lib/maths/time_series/unittest/CSignalTest.cc
+++ b/lib/maths/time_series/unittest/CSignalTest.cc
@@ -567,7 +567,7 @@ BOOST_AUTO_TEST_CASE(testReweightOutliers) {
 
     test::CRandomNumbers rng;
 
-    maths::time_series::CSignal::TFloatMeanAccumulatorVec values(144);
+    maths::time_series::CSignal::TFloatMeanAccumulatorVec values;
     TDoubleVec noise;
     TDoubleVec u01;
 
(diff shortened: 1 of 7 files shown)


lib/maths/time_series/unittest/CSignalTest.cc
```diff
@@ -567,7 +567,7 @@ BOOST_AUTO_TEST_CASE(testReweightOutliers) {
 
     test::CRandomNumbers rng;
 
-    maths::time_series::CSignal::TFloatMeanAccumulatorVec values;
+    maths::time_series::CSignal::TFloatMeanAccumulatorVec values(144);
```
droberts195, line 570: @tveasey of the 4 tests that were warning of no assertions this is the one that's a worry.  This test genuinely was not doing what it should have been doing, because everything was based on `values.size()` and `values.size()` was 0.  I copied the 144 here from a different test that was using the same pattern of basing everything on `values.size()`, but now the test fails. 🙀 

The failure is:

```
/Users/dave/ml-cpp/lib/maths/time_series/unittest/CSignalTest.cc:596: fatal error: in "CSignalTest/testReweightOutliers": critical check maths::common::CBasicStatistics::count(values[i]) < 1.0 has failed [1 >= 1]
```

Since this test seems to have been doing nothing up to now, please could you have a look at what it should be asserting now.

tveasey: Whoops! The issue is you can't use 144 for values size, otherwise 30 isn't an outlier for the second component. I would leave this line unchanged and update the line where we assign to values to be:
```
values.assign(100, maths::time_series::CSignal::TFloatMeanAccumulator{});
```

tveasey (approved):
LGTM (with suggested change to test).

PR #2349, opened 2022-07-07
[ML] Further improvements to early stopping for training regression and classification models

tveasey:
This completes the planned work to improve early stopping.

With this change we consider stopping fine tune after fewer rounds and also bail out of training uncompetitive models more aggressively.

This logically belongs with the other early stopping changes so I've marked as a non-issue.


lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -3776,11 +3776,11 @@ BOOST_AUTO_TEST_CASE(testStopAfterCoarseParameterTuning) {
...
                 .numberHoldoutRows(numberHoldoutRows)
                 .buildForTrain(*frame, cols - 1);
-        return regression->hyperparameters().optimisationMakingNoProgress();
+        return regression->hyperparameters().fineTuneSearchNotFinished();
     };
 
-    BOOST_REQUIRE_EQUAL(verify(0.001), false);
-    BOOST_REQUIRE_EQUAL(verify(1000.0), true);
+    BOOST_REQUIRE_EQUAL(verify(0.0), true);
+    BOOST_REQUIRE_EQUAL(verify(1000.0), false);
```
valeriy42, line 3783: Why are the check values swapped now?

valeriy42 (approved):
LGTM altogether. Just one question regarding the changes in the unit test.

lib/maths/analytics/unittest/CBoostedTreeTest.cc
tveasey: This is just because I swapped it to use `fineTuneSearchNotFinished()` which is the negation of what it was before. So this really means the same thing, i.e. true means we're going to carry on training and false means we'll stop.

PR #2351, opened 2022-07-07
[ML] Platform independent dependency report script

edsavage:
Replace dependency_report.sh bash script with cmake script equivalent


droberts195 (approved):
LGTM

PR #2352, opened 2022-07-08
[ML] Use ccache if it is found on the system

edsavage:


CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ endif()
 message (STATUS "CMAKE_MODULE_PATH: ${CMAKE_MODULE_PATH}")
 message (STATUS "CMAKE_VERSION: ${CMAKE_VERSION}")
 
+# Use ccache if it is available
+find_program(CCACHE_FOUND ccache)
+if(CCACHE_FOUND)
+  message(STATUS "ccache found: ${CCACHE_FOUND}")
```
droberts195, line 27: We don't want to use this in CI builds, so please can you check the build logs of the PR CI and make sure this message isn't present in any of them.  If it is then we'll have to take steps to stop it being used in CI builds.

edsavage: I've just confirmed that `ccache` was not found on any of the CI workers for this PR.

droberts195 (approved):
LGTM

PR #2354, opened 2022-07-08
[ML] Allow for more cmake generators and build types

edsavage:
A few tweaks to remove obstacles in the way of support for a wider range
of cmake build generators, such as "Xcode", "Ninja Multi-Config",
"Eclipse CDT4 - Ninja" etc. Also removed the assumption that the
build type was one of "Release" or "Debug".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,10 +79,3 @@ add_subdirectory(devlib)
 # Add a target to build Doxygen generated documentation
 # if the doxygen executable can be found
 ml_doxygen(${CMAKE_SOURCE_DIR}/build/doxygen)
-
-# Xcode code signs the binary artifacts by default,
-# which we don't want as it invalidates things when
-# the RPATH is updated upon install. Hence we disable
-# code signing here.
-set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
-set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED "NO")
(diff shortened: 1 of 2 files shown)


cmake/variables.cmake
```diff
@@ -156,7 +156,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
   list(APPEND ML_COMPILE_DEFINITIONS BOOST_ALL_NO_LIB)
 endif()
 
-# Dictate which flags to use for a "Release" build
+# Dictate which flags to use for "Release" and "Debug" builds
```
droberts195, line 159: At the moment we unconditionally add `-g` to these flags.

That was based on our earlier discovery that CMake's default `RelWithDebInfo` flags were pretty suboptimal.

But our primary builds really are "releases with debug info".

I don't think you should change anything more in this PR, but if this one goes through successfully then in a followup we should probably set an extra variable here for each platform to define our preferred `RelWithDebInfo` flags, take `-g` out of the unconditional flags, and switch our CI builds to do `RelWithDebInfo` builds. Then we'd be using CMake targets in the intended way.

But, like I said, don't try to do it all in this PR. Let's go one step at a time in case there's some unforeseen side effect of this one.

CMakeLists.txt
```diff
@@ -79,3 +79,10 @@ add_subdirectory(devlib)
 # Add a target to build Doxygen generated documentation
 # if the doxygen executable can be found
 ml_doxygen(${CMAKE_SOURCE_DIR}/build/doxygen)
+
+# Xcode code signs the binary artifacts by default,
+# which we don't want as it invalidates things when
+# the RPATH is updated upon install. Hence we disable
+# code signing here.
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED "NO")
```
droberts195, line 88: Should this go in one of the macOS-specific files rather than here?

droberts195 (approved):
LGTM

PR #2355, opened 2022-07-11
[ML] Default build configuation to  RelWithDebInfo

edsavage:
* Change the default build type to be 'RelWithDebInfo' as this makes
explicit the build flags we have been using all along i.e. flags for an
optimized release build but with additional debug symbols.

* Updated the CLion guide accordingly (and added in missing screen shots)


build-setup/clion/using_clion.md
```diff
@@ -39,7 +39,7 @@ By default `CLion` will opt for generating a `Ninja` build system. While this is
 better compatibility with building from the command line it is preferred to configure `CLion` to generate
 `Unix Makefiles`. To do this either select `Unix Makefiles` from the `Generator` drop-down or type 
 `-G "Unix Makefiles"` in the`CMake options` field. Add  e.g. `-j 10` to the `Build options` field, select your
-desired build type (either`Release` or `Debug`) and we're done. Make sure to leave the `Build directory` field
+desired build type (e.g.`Release` or `Debug`) and we're done. Make sure to leave the `Build directory` field
```
droberts195, line 42: I think we should recommend people work day-to-day with `RelWithDebInfo` and only switch to something different if absolutely necessary for some specific debugging exercise or investigation.  We've now got it set up so that `RelWithDebInfo` gives our desired blend of performance and debuggability, and it's what we ship to our users, so it's best that people develop with that day-to-day where possible.

Also, should we recommend _against_ using `Default` for the build type? Does selecting `Default` actually do a build that's identical to `RelWithDebInfo`? If it does then maybe `Default` is OK (although it does create a separate CMake build directory). But if it doesn't then it seems to be actively confusing.

droberts195 (commented):
I think it would be best if the file `build-setup/clion/default_project_config_options.png` showed `RelWithDebInfo` selected in the "Build type" dropdown.

`dev-tools/docker/docker_entrypoint.sh` also needs modifying.

build-setup/clion/using_clion.md
edsavage: > Also, should we recommend _against_ using `Default` for the build type?

A bit of experimentation indicates that the `Default` build type uses flags that are equivalent to those corresponding to `RelWithDebInfo` (at least on macOS), but I think it might be worth stating that it's best to explicitly choose one of the other build types, to avoid any confusion.

droberts195: > to avoid any confusion

It also has the advantage that builds from the command line can reuse object files already built by CLion builds, and vice-versa.  If CLion is using `Default` and the command line is using `RelWithDebInfo` then they have separate object files built with the same flags.

droberts195:
This line in `CONTRIBUTING.md` also needs editing:

> Note that we configure the build to be of type `Release`, and specify the compiler flag `-g` in the cmake configuration files in order to obtain a fully optimized build along with debugging symbols (for Windows builds using Visual Studio the equivalent is using the /Zi flag to generate PDB files). This is used in preference to `RelWithDebInfo` as `Release` generally gives a higher optimization level than `RelWithDebInfo` (`O3` vs `O2` respectively) and on Windows RelWithDebInfo omits inlining which is undesirable.

As does this line in `README.md`:

> * Run `cmake -B cmake-build-release --config Release` to generate the build system under the `cmake-build-release` directory (the `--config Release` option may be omitted on Linux and Mac).

droberts195 (commented):
In `functions.cmake` we have this a few times:

```
$<$<CONFIG:Release>:-OPT:REF>
```

`-OPT:REF` should also be included now for `RelWithDebInfo`, otherwise we'll lose it from our builds for customers.

Another thing to look at here is whether CMake automatically adds `-DEBUG` (or `/DEBUG`) to the link line when building a program/DLL with `RelWithDebInfo`.  If it does then I think we should also add `-OPT:ICF` for `RelWithDebInfo`, as `-DEBUG` switches that off, but allows it to be explicitly switched back on if desired.

droberts195 (approved):
LGTM

PR #2356, opened 2022-07-11
[ML] Migrate to std::optional

tveasey:
Since we're now using C++17 on main we can migrate `boost::optional` to `std::optional`. This change also makes use of in-place constructors in various places.


droberts195 (approved):
LGTM

PR #2357, opened 2022-07-11
[ML] Make C++ code match new meanings for threading inputs

droberts195:
The Java code now works on the basis that num_threads_per_allocation
should be multiplied by the number of allocations to give the total
number of threads required.

This PR changes the C++ code to reflect that.

There are a couple of subtle changes:

1. There's an assumption that either our outer thread pool _or_
   PyTorch's inner thread pool are busy, but not both at the same
   time. We are assuming that when a PyTorch inference fans out
   work across the threads of its inner thread pool the thread
   that requested the inference will block awaiting the inner
   results. This is consistent with how `parallel_for_each` works,
   so a reasonable assumption.
2. When multiple allocations are in the same process they are
   all sharing the same PyTorch inner thread pool. This means
   the allocations are sharing those "threads per allocation".
   For example, if number of allocations is 3 and threads per
   allocation is 4 then PyTorch's inner thread pool will have
   12 threads. But if 3 inferences are being done simultaneously
   then each will fan out large steps over 12 threads rather
   than 4. The 12 subtasks may then queue for a thread while the
   12 subtasks of the other inferences run. The total number of
   threads used is as specified, but they are not dedicated to
   particular allocations.

### DIFF
--- a/bin/pytorch_inference/CThreadSettings.cc
+++ b/bin/pytorch_inference/CThreadSettings.cc
@@ -31,6 +31,10 @@ std::int32_t CThreadSettings::numAllocations() const {
     return m_NumAllocations;
 }
 
+std::int32_t CThreadSettings::pyTorchThreadpoolThreads() const {
+    return m_NumAllocations * m_NumThreadsPerAllocation;
+}
+
 void CThreadSettings::numAllocations(std::int32_t numAllocations) {
     m_NumAllocations = numAllocations;
 }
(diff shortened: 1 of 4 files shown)


droberts195:
`>non-issue` as logically it's part of https://github.com/elastic/elasticsearch/pull/87366

droberts195:
> I wonder if this invalidates our assumptions when we dynamically increase the number allocations.

Yes, good point, it does. We haven't actually implemented that yet on the Java side as far as I am aware. The idea was that we could send a control message to change the number of allocations but we don't do that at the moment and it won't work correctly. So we should stick with restarting the process if we want to change either the number of allocations or the threads per allocation.

> I guess some work is done on the scheduling thread but this at least reduces the gain we expect from increasing allocations.

Yes, some work is done on the scheduling thread, and in fact it must be quite a high proportion because our tests in the inference threads/model threads days showed a considerable gain in throughput from increasing model threads while leaving inference threads alone.  (In the old terminology "inference threads" referred to PyTorch's inner thread pool.)

In the case where both settings are greater than 1 it all gets very complicated:

- If num allocations is 1 and threads per allocation is 16 then we have 17 threads, but the scheduling thread will be idle while the inner worker threads work, so all is good.
- If num allocations is 16 and threads per allocation is 1 then we have 17 threads again, but presumably OMP does the work of a `for_each_parallel` on the scheduling thread if the inner worker thread pool is configured with just one thread, so again all is good.
- If num allocations is 4 and threads per allocation is 4 then we have 20 threads, 4 for our outer thread pool and 16 for PyTorch's inner thread pool.  And it's possible that 3 of those outer threads could be working hard at the same time as the 16 inner threads, for a total of 19 threads active simultaneously.  Also, if single sporadic inferences are performed then they benefit from all 16 threads in the inner thread pool, thus presenting a misleading impression of the performance of 4 "threads per allocation".

It seems that we can only truly model "threads per allocation" if we put each allocation in a separate process, but that's mega inefficient in terms of memory usage and CPU cache performance.

We could potentially give PyTorch's inner thread pool 1 thread less than "threads per allocation" for each allocation if both settings are greater than 1.  But that has some weird edge cases too, for example num allocations 4, threads per allocation 2 would then end up doing exactly the same as num allocations 4, threads per allocation 1.

I'll wait until next week before merging this so that we can discuss it with @dimitris-athanasiou first.

PR #2360, opened 2022-07-12
[ML] Add an assertion to CCompressedLfuCacheTest

droberts195:
Similar to #2348, this PR is adding an assertion to a test
that didn't previously have one, so that it doesn't mess
up the CI test report format.


edsavage (approved):
LGTM

PR #2365, opened 2022-07-18
[ML] Size pytorch_inference thread pool properly

davidkyle:
Non issue as this fixes a bug in unreleased code


tveasey (approved):
LGTM. Although you just need to fix a format issue before merging.

davidkyle:
Even though the same change could be made in 8.3 I will not backport this PR as the function to dynamically change the number of allocations is not used or supported.

PR #2366, opened 2022-07-18
[ML] Update NumPy version for docker image

valeriy42:
Dependobot has bumped the version of NumPy to 1.22 in #2317. However, this change did not affect the docker images we created. And since NumPy 1.20 dropped the support for Python 3.6, and CentOS 7 doesn't provide packages for Python 3.7+, I needed to make a more significant change to the Docker files and compile Python explicitly.

I removed `diversipy` from the requirements since we do not require the sampling method, and the package has multiple outdated dependencies.

This change is not customer-facing, so I mark it as a non-issue.


droberts195 (approved):
LGTM, but please add some comments saying why `diversipy` and `sacred` are commented out requirements now.

tveasey (approved):
LGTM2

PR #2367, opened 2022-07-19
[ML] Improve quantile estimation accuracy and median accuracy for anomaly detection

### LINKED ISSUE #2364: [ML] Investigate approximate percentile accuracy for anomaly detection
Using the native code to compute quantiles uses a different approach with different error characteristics to the Elasticsearch percentiles aggregation. This can lead to somewhat confusing results. For example, if we don't enable model plot for a median job then the charts use the percentiles aggregation, but actual values reported with anomalies use the internal methodology based on histogram sketches. We can't eliminate discrepancy altogether, but we should revisit and benchmark the size of the disagreement for a range of data distributions. It may be, for example, we parameterise slightly differently to reduce errors at the expense of using more memory.

tveasey:
We use a histogram sketch for estimating data quantiles and also computing the time bucket median for anomaly detection.

Once data are merged into buckets we have to make some assumptions about how values are distributed on buckets. Previously, we assumed data are uniformly distributed on buckets whose endpoints are the midpoints between the bucket "centres" we track. In fact, the points we track are the weighted means of the values in each bucket, so it is more accurate to assume that roughly half the data in a bucket falls either side of this point.

This changes the interpolation scheme to use the same endpoints but to incorporate this new assumption. This gives a very nice additional property: a quantile which falls between two buckets each containing a single value is computed exactly. An immediate corollary is that all quantiles are exact if the data size is less than the sketch size. Previously, we used piecewise constant interpolation for estimating the median because it is exact in this case. We now cut over to linear interpolation. This is attractive because for large data, if the data distribution is smooth, linear interpolation is significantly more accurate for fixed memory usage than piecewise constant interpolation.

Closes #2364.

### DIFF
--- a/lib/model/unittest/CMetricModelTest.cc
+++ b/lib/model/unittest/CMetricModelTest.cc
@@ -600,7 +600,7 @@ BOOST_FIXTURE_TEST_CASE(testProbabilityCalculationForMedian, CTestFixture) {
         if (i == anomalousBucket) {
             values.push_back(0.0);
             values.push_back(mean * 3.0);
-            values.push_back(mean * 6.0);
+            values.push_back(mean * 3.0);
         } else {
             rng.generateNormalSamples(mean, variance, bucketCounts[i], values);
         }
--- a/lib/model/unittest/CMetricModelTest.cc
+++ b/lib/model/unittest/CMetricModelTest.cc
@@ -600,7 +600,7 @@ BOOST_FIXTURE_TEST_CASE(testProbabilityCalculationForMedian, CTestFixture) {
         if (i == anomalousBucket) {
             values.push_back(0.0);
             values.push_back(mean * 3.0);
-            values.push_back(mean * 3.0);
+            values.push_back(mean * 6.0);
         } else {
             rng.generateNormalSamples(mean, variance, bucketCounts[i], values);
         }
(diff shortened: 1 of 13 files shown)


lib/model/unittest/CMetricModelTest.cc
```diff
@@ -600,7 +600,7 @@ BOOST_FIXTURE_TEST_CASE(testProbabilityCalculationForMedian, CTestFixture) {
         if (i == anomalousBucket) {
             values.push_back(0.0);
             values.push_back(mean * 3.0);
-            values.push_back(mean * 3.0);
+            values.push_back(mean * 6.0);
```
droberts195, line 603: Why did this have to change?  Surely the median of 0, 3 and 3 is still 3?  If so it might be nice for the test to alternate the final value between 3 and 6 for consecutive buckets to prove that 0, 3, 3 and 0, 3, 6 both have median 3.

tveasey: This is just an artefact of how duplicate values are handled. Previously, we used piecewise constant which either uses the centre if "cumulative count" / "count" first exceeds the quantile at a centre or the mid point between two centres otherwise. Linear interpolation assumes values are uniformly distributed on the interval which extends from [1.5, 3] so it uses (1.5 + 0.25) * mean in the case that you have {0, 3, 3}. If I just present you with the buckets (0, 1) and (3, 2) they could just as well have generated from say {0, 2.5, 3.5} etc. So this is just the nature of approximation: sometimes we get better results one way sometimes another. If you look at CQuantileSketchTest you'll see for random data we get around 1/5 the error from linear interpolation. That said, I probably can be smarter here. The crux is whether we can work out that we have't actually merged any buckets based on class invariants. I'll have a think if I can do something better around this.

tveasey: Ok it turns out it is fairly easy to deduce whether we have merged values in nearly all cases and use exact interpolation in this case. See [a8ad541](https://github.com/elastic/ml-cpp/pull/2367/commits/a8ad54124454a056dc93d118eafc2c73fc1c2b61).

droberts195 (approved):
LGTM

PR #2369, opened 2022-07-20
[ML] Adjust PyTorch thread setting naming and validation

droberts195:
The Java code now works on the basis that `num_threads_per_allocation`
should be multiplied by the number of allocations to give the total
number of threads required.

This PR changes the C++ code to reflect that.

The logic for calling `at::set_num_threads()` and
`ml::core::defaultAsyncExecutor().numberThreadsInUse()` was already
correct given how libgomp on Linux does threading, but this PR
fixes the validation, adds comments and corrects the variable
names in the unit test.

### DIFF
--- a/bin/pytorch_inference/CThreadSettings.cc
+++ b/bin/pytorch_inference/CThreadSettings.cc
@@ -18,9 +18,12 @@
 namespace ml {
 namespace torch {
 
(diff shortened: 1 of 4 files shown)


dimitris-athanasiou (approved):
LGTM

PR #2373, opened 2022-07-25
[ML] Compile time improvements

tveasey:
Analysing the project dependency graph showed up some more low hanging fruit for compilation performance. This makes the following changes:
1. Remove `boost/range.hpp`. All our remaining uses are satisfied by `std::size()` and that way we don't need to include the all the range library headers.
2. Remove `boost/numeric/conversion/bounds.hpp`. We only needed `lowest` which arrived in C++11.
3. Migrate `boost::tuple` to `std::tuple` and `boost::any` to `std::any`.
4. Avoid including more rarely used `std` containers and `boost::multi_index` in `CMemory`. This is included in a lot of places. The only way to achieve this is to convert it to a namespace. This introduces a new problem: function lookup rules are different from class method definitions, so you need to be careful that all `dynamicSize` functions are declared first when using a namespace. I added in a static assert that we aren't resolving to the wrong implementation and introduce "dec" and "def" headers for this purpose.
5. Migrate `CMemoryUsage` to pimpl: it isn't performance critical and is included almost everywhere. I also migrate all cases where we only need `CMemoryUsage` to include that rather than `CMemoryFwd.h` which include `CMemoryUsage` and a little bit more.
6. Avoid unnecessary includes of `core::CVectorRange` and `CStoredStringPtr`.
7. Move `maths/common/CBasicStatistics.h` out of `maths/common/CTools.h`. It was only used to implement one template which is not widely used.
8. Use include what you use to find and remove unused headers. Windows requires explicit destructors (and sometimes constructors) when using forward declared types with `std::unique_ptr`. I had to make some changes for this.

Since I'm touching a whole load of files again I thought I'd tidy up the old cpp type conversion in assert stuff while I was about it.


droberts195 (approved):
LGTM

I think it would be nice to rename the `CMemory` and `CMemoryDebug` namespaces to `memory` and `memory_debug`, to make it clearer to future maintainers that they're not actually classes.  But this should definitely be left for a followup PR that _only_ does this renaming because this one is already huge and it will be harder to see the actual changes with all the renaming mixed in.

PR #2374, opened 2022-07-25
[ML] Adding a cache_hit indicator to inference results

droberts195:
This is a companion to https://github.com/elastic/elasticsearch/pull/88287.

In addition to adding a `cache_hit` indicator, the meaning of `time_ms` is
changed from "original inference time" to "cache lookup plus inference time
if necessary".

### DIFF
--- a/bin/pytorch_inference/CCommandParser.cc
+++ b/bin/pytorch_inference/CCommandParser.cc
@@ -11,9 +11,7 @@
 
 #include "CCommandParser.h"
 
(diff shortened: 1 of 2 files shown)


droberts195:
Like https://github.com/elastic/elasticsearch/pull/88287 this is `>non-issue`.  The PR that eventually exposes the number of cache hits and misses to the end user can be release noted.

dimitris-athanasiou (approved):
LGTM

PR #2375, opened 2022-07-27
[ML] Add a "clear cache" control message to pytorch_inference

droberts195:
This is likely to be useful for QA performance tests once
it's exposed on the Java side.

### DIFF
--- a/bin/pytorch_inference/CCommandParser.cc
+++ b/bin/pytorch_inference/CCommandParser.cc
@@ -101,7 +101,7 @@ bool CCommandParser::ioLoop(const TRequestHandlerFunc& requestHandler,
             }
             break;
         case EMessageType::E_ControlMessage:
-            controlHandler(jsonToControlMessage(doc));
+            controlHandler(*m_RequestCache, jsonToControlMessage(doc));
             break;
         case EMessageType::E_MalformedMessage:
             continue;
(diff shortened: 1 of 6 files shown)


bin/pytorch_inference/CCommandParser.cc
```diff
@@ -281,8 +291,18 @@ CCommandParser::jsonToInferenceRequest(const rapidjson::Document& doc) {
...
+    case E_NumberOfAllocations:
+        return {controlMessageType, doc[NUM_ALLOCATIONS].GetInt(),
+                doc[REQUEST_ID].GetString()};
+    case E_ClearCache:
+        return {controlMessageType, 0, doc[REQUEST_ID].GetString()};
+    case E_Unknown:
+        break;
+    }
+
+    LOG_ABORT(<< "Programmatic error - incorrect validation of control message");
```
droberts195, line 305: We should never get here as `validateControlMessageJson()` should have responded with an error if the intended methods are called in the correct order.

include/core/CCompressedLfuCache.h
```diff
@@ -161,7 +161,7 @@ class CCompressedLfuCache {
                         readValue(value);
                         return;
                     }
-                    m_LostCount += lastEvictedCount;
+                    m_RemovedCount += lastEvictedCount;
```
droberts195, line 164: I shifted this loss of count from `lost` to `removed` as it's similar to what happens in a resize.

include/core/CCompressedLfuCache.h
```diff
@@ -174,6 +174,18 @@ class CCompressedLfuCache {
         return false;
     }
 
+    void clear() {
+        this->guardWrite(DONT_TIME_OUT, [&] {
+            m_NumberLookups = 0;
+            m_NumberHits = 0;
+            m_LostCount = 0;
+            m_RemovedCount = 0;
```
droberts195, line 182: I decided to reset all the counters to zero here.  The alternative would be to increment the removed count instead to indicate everything being removed.  It's fairly academic anyway, as we don't expose these counters to the Java side - they are only for our unit tests at the moment.

bin/pytorch_inference/Main.cc
```diff
@@ -170,6 +171,18 @@ void writeThreadSettings(ml::core::CJsonOutputStreamWrapper& wrappedOutputStream
...
+void writeSimpleAck(ml::core::CJsonOutputStreamWrapper& wrappedOutputStream,
+                    const std::string& requestId) {
+    ml::core::CRapidJsonConcurrentLineWriter jsonWriter(wrappedOutputStream);
+    jsonWriter.StartObject();
+    jsonWriter.Key(ACK);
+    jsonWriter.StartObject();
+    jsonWriter.Key(ml::torch::CCommandParser::REQUEST_ID);
+    jsonWriter.String(requestId);
+    jsonWriter.EndObject();
+    jsonWriter.EndObject();
```
droberts195, line 183: The structure of this object highlights the benefit in hoisting `request_id` up to the top level.

I'll probably do that before the Java side of this change, so this object structure will never exist in the Java code, but didn't want to mix two changes in this PR.

dimitris-athanasiou (approved):
LGTM

PR #2376, opened 2022-07-28
[ML] Change inference cache to store only the inner part of results

droberts195:
Previously the inference cache stored complete results, including
a request ID and time taken. This was inefficient as it then meant
the original response had to be parsed and modified before sending
back to the Java side.

This PR changes the cache to store just the inner portion of the
inference result. Then the outer layer is added per request after
retrieving from the cache.

Additionally, the result writing functions are moved into a class
of their own, which means they can be unit tested.

Companion to elastic/elasticsearch#88901


dimitris-athanasiou (approved):
LGTM

droberts195:
retest

PR #2377, opened 2022-07-29
[ML] Fix inefficiency in bootstrap loss calculation

tveasey:
#2339 introduced an inefficiency where we calculate loss per row more times than we should. This showed up as surprisingly expensive in profiling. This hasn't been released so marking as a non-issue.


dimitris-athanasiou (approved):
LGTM

PR #2380, opened 2022-07-29
[ML] Boosted tree row split calculation performance

tveasey:
Profiling training on one of slowest internal benchmarks showed up two expensive operations which we have a good chance of improving:
1. `CBoostedTreeImpl::candidateSplits`
2. `CBoostedTreeImpl::refreshSplitsCache`

I plan to tackle these in two commits. This one focuses on `refreshSplitsCache` whose cost is dominated by `std::upper_bound`. clang in particular generates poor assembly for `std::upper_bound` for our usage: creating an (unpredictable) branch instead of a conditional move on the hot path. In any case, I wanted to to see if we can do better here using a search tree strategy.

The search tree can potentially benefit from better LOR when using branching factor greater than 2 because we can pack split values into consecutive memory. Another advantage is we always compile with SSE4.2 so we have single instructions for doing the comparison of 4 distinct values. The resulting data structure is implemented as `CSearchTree`. (Note that this is in `CBoostedTreeUtils` which was getting included more than it should because it contained functionality which wasn't really logically part of it anymore. I moved this into the correct places and cutdown the includes as a result.)

On synthetic benchmarks, with uniform random data, pretty much the worst case for the previous implementation, I get around a 4.5X speedup vs `std::upper_bound` for our typical usage set sizes. For the benchmark which motivated this change, I cut the cost of `refreshSplitsCache` by a little over 2X.


include/maths/analytics/CBoostedTreeUtils.h
```diff
@@ -50,37 +96,132 @@ enum EExtraColumnTag {
...
+            // Each branch point which is greater than x is out of order w.r.t. this
+            // point and must be subtracted from node to get the correct upper bound.
+            offset += 4 - branch;
+        }
+
+        std::size_t branch{selectBranch(&m_Values[node], vecx)};
+        LOG_TRACE(<< "x = " << x << ", node = " << node << "/" << this->printNode(node)
+                  << ", branch = " << branch << ", offset = " << offset);
+
+        return node + branch + 1 - offset;
```
valeriy42, line 180: The variables `node` and `branch` refer to the tree representation, while `offset` refers to the original array. It may create a confusion. It would be great if you could make the names more explicit.

include/maths/analytics/CBoostedTreeUtils.h
```diff
@@ -50,37 +96,132 @@ enum EExtraColumnTag {
...
+//! \brief A fast ordered search tree.
+//!
+//! DESCRIPTION:\n
+//! This provides a single query upperBound(x), i.e. find the smallest value in
+//! an ordered set greater than the specified query point x. When possible it
+//! uses SSE-like instructions to perform a 4 way comparison between the query
+//! point and candidate split points. This means it achieves a branch factor of
+//! 5 and complexity O(ceil(log(n) / log(5))) in the set size n.
+//!
+//! IMPLEMENTATION DECISIONS:\n
```
valeriy42, line 111: While you describe all important implementation decisions in words, I think it would be useful for posterity to document a specific example of input array and the tree representation.

include/core/CContainerPrinter.h
```diff
@@ -105,7 +105,12 @@ class CElementPrinter<false> {
...
 
-    //! std::ostringstream implementation.
+    //! std::to_string for float.
+    inline static std::string print_(float value, true_ /*is arithmetic*/) {
+        return std::to_string(value);
+    }
+
+    //! Fallback to std::ostringstream.
     template<typename T>
     inline static std::string print_(const T& value, false_ /*is arithmetic*/) {
```
valeriy42, line 115: should the comment be /*is not arithmetic*/?

include/core/CFloatStorage.h
```diff
@@ -93,40 +93,34 @@ class CORE_EXPORT CFloatStorage {
     //! \name Operators
     //@{
     CFloatStorage operator-() const { return CFloatStorage{-m_Value}; }
-    bool operator==(const CFloatStorage& rhs) const {
-        return m_Value == rhs.m_Value;
-    }
-    bool operator==(const double& rhs) const {
+    bool operator==(CFloatStorage rhs) const { return m_Value == rhs.m_Value; }
```
valeriy42, line 96: Why do you remove const reference here?

valeriy42 (commented):
Great change! I just have a few comments regarding documentation.

include/core/CFloatStorage.h
tveasey: For built in types it is actually better not to pass by reference. For example, this would require indirection when accessing rhs.m_Value and there is no overhead from the copy. Of course the compiler may elide this altogether by the as-if rule: copying would have no side-effects relative to pass by reference. However, I figured it is better not to rely on this behaviour.

include/core/CContainerPrinter.h
tveasey: In this case, it is `false_ /* is arithmetic */` with the false as well, i.e. the type isn't arithmetic.

lib/maths/analytics/unittest/CBoostedTreeUtilsTest.cc
```diff
@@ -21,6 +21,7 @@
 #include <test/CRandomNumbers.h>
 
 #include "BoostedTreeTestData.h"
+#include <core/CStopWatch.h>
```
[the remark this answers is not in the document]
tveasey: See [8f1c5d3](https://github.com/elastic/ml-cpp/pull/2380/commits/8f1c5d3618a84255bf44907364e0838d05e3ee5b).

include/maths/analytics/CBoostedTreeUtils.h
tveasey: See [8f1c5d3](https://github.com/elastic/ml-cpp/pull/2380/commits/8f1c5d3618a84255bf44907364e0838d05e3ee5b).

include/maths/analytics/CBoostedTreeUtils.h
tveasey: See [f690c42](https://github.com/elastic/ml-cpp/pull/2380/commits/f690c425b13a8eaa75f1c23b61f34191631249c4).

tveasey:
Thanks @valeriy42. I think I addressed all your comments.

include/maths/analytics/CBoostedTreeUtils.h
```diff
@@ -110,26 +102,47 @@ constexpr std::size_t UNIT_ROW_WEIGHT_COLUMN{std::numeric_limits<std::size_t>::m
...
+//! size 5^ceil(log(n)/log(5)), recusively into 5 equal sized ranges. That is
+//! using the 20th, 40th, 60th and 80th percentiles of each interval as the split
+//! points.
+//!
+//! To do this it uses a flat tree representation, which can then be searched
+//! very efficiently. (Note that the smallest value is not added to the tree;
+//! there is a check for out of range values upfront.) This means the search
+//! achieves a branch factor of 5. There is a simple one-to-one mapping between
+//! the positions of values in the flat tree representation and the sorted
+//! collection.
```
valeriy42, line 118: Great explanation! :rocket:

include/maths/analytics/CBoostedTreeUtils.h
```diff
@@ -158,10 +171,13 @@ class MATHS_ANALYTICS_EXPORT CSearchTree {
         }
 
         std::size_t node{0};
-        std::size_t offset{0};
+        // The tree structure doesn't arrange all values in order so we need to
+        // keep track of the number of out-of-order values preceding a position
+        // in the tree to map it back to its position in the sorted collection.
+        std::size_t numberOutOfOrderValues{0};
```
valeriy42, line 177: :+1:

valeriy42 (approved):
Good work! :rocket:

PR #2383, opened 2022-08-03
[ML] Fix main() functions for test suites

droberts195:
These have had a silly mistake ever since they were written.
The observer was unregistered after main() returned. Clearly
it didn't matter, as the program was about to exit anyway,
but it's nice to tidy up the code.


edsavage (approved):
LGTM

PR #2384, opened 2022-08-04
[ML] Fix debug build for raw JSON splicing

droberts195:
In #2376 the results writer for inference was changed
to splice a preformatted cached section of the final result
document into an outer wrapper that varies per request.

The approach used worked when assertions were disabled but
tripped assertions in RapidJSON due to its internal counters
thinking that the spliced portion was a single value instead
of a key and value together.

This PR fixes the code when assertions are enabled by
taking advantage of the fact that the internal value
count of the RapidJSON writer is accessible to derived
classes, so we can add a method that adds a key and
value together and increments the counter twice.


dimitris-athanasiou (approved):
LGTM

PR #2385, opened 2022-08-10
[ML] Fix potential out-of-bounds read

tveasey:
Debugging an intermittent SIGSEGV triggered by another change I'm working on showed up this potential out-of-bounds read. It would happen very infrequently. It was introduced by #1537.


droberts195 (approved):
LGTM

PR #2389, opened 2022-08-12
[ML] Detect debug-only compile errors in PR builds

droberts195:
People rarely build the code with debug enabled, so it's
possible for an error in the code that only affects debug
builds to get committed.

This change adds a step to build (but not unit test) the
code with debug enabled in every PR build. Doing this in
the cross compile sub-job should mean that it doesn't
increase the overall PR CI duration, as the cross compile
sub-job is faster than the sub-jobs that run unit tests.

### DIFF
--- a/dev-tools/jenkins_cross_compile.sh
+++ b/dev-tools/jenkins_cross_compile.sh
@@ -87,10 +87,15 @@ fi
 # Cross compile macOS
 ./docker_build.sh macosx
 
-# If this isn't a PR build cross compile aarch64 too
-if [ -z "$PR_AUTHOR" ] ; then
-    ./docker_build.sh linux_aarch64_cross
+# If this is a PR build then it's redundant to cross compile aarch64 (as
+# we build and test aarch64 natively for PR builds) but there's a benefit
+# to building one platform with debug enabled to detect code that only
+# compiles with optimisation
+if [ -n "$PR_AUTHOR" ] ; then
+    export ML_DEBUG=1
 fi
+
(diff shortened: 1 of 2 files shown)


tveasey (approved):
Great idea!

dev-tools/jenkins_cross_compile.sh
```diff
@@ -87,11 +87,16 @@ fi
...
-    ./docker_build.sh linux_aarch64_cross
+# If this is a PR build then it's redundant to cross compile aarch64 (as
+# we build and test aarch64 natively for PR builds) but there's a benefit
+# to building one platform with debug enabled to detect code that only
+# compiles with optimisation
+if [ -n "$PR_AUTHOR" ] ; then
+    export ML_DEBUG=1
 fi
 
+./docker_build.sh linux_aarch64_cross
```
dimitris-athanasiou, line 98: Doesn't this mean we cross compile even on PR builds?

droberts195: Yes, that's the whole point - we do an extra compile on an arbitrary platform with `ML_DEBUG=1`.  This means that as well as doing optimised builds for all platforms our PR builds will now fail if there are compilation errors that only show up when debug is enabled.

You can see the benefit in the CI for the first commit on this PR: https://elasticsearch-ci.elastic.co/job/elastic+machine-learning+pull-request+cross-compiles/1516/console. It has detected a compilation problem for debug builds that wasn't detected in the PR CI of the PR that introduced the problem.

PR #2390, opened 2022-08-12
[ML] Improve quantile estimation performance for train

tveasey:
This is the second part of #2380.

Quantile estimation during training is expensive for data sets with many numeric features. This change makes various speed ups in this area:
1. Batch computes merge costs using explicit vectorisation,
2. Use a larger reduce factor when merging buckets. I've held the reduced size fixed to maintain roughly similar accuracy, the upshot being more memory but less work per bucket,
3. Improve handling of duplicate values: we should deduplicate new values before merging because it reduces the cost of `inplace_merge`,
4. Be a bit more careful with how random numbers for tiebreaks are generated.

Together these drop the amortised add cost from around 100ns to 50ns per item on my i9. The remaining costs are 50% `std::nth_element` and 30% `std::sort` + `std::inplace_merge`. Short of dropping the requirement we extract the smallest k merge costs it is unlikely we'd be able to make big inroads on these costs. Doing this it was useful to extend `CFloatStorage` to be useable in constexprs. I also mark all methods noexcept which they are and always will be, it might in theory allow some extra optimisations, although the compiler should be able to deduce these are no throw.

Separately, I cap the maximum number of values we'll use to estimate quantiles. The accuracy of the quantiles we care about converges quickly with sample size. For example, the count of values less (greater) than the _sample_ 1st (99th) percentile would be ~ Binomial(n, 0.01), for sample size n, so relative error in count would be O(10.0 / n^(1/2)). I therefore cap the maximum sample size to 50000 for which this error is around 4%.

Finally, following on from #2364 we don't really need to expose interpolation outside the class (all uses were linear and that would be unlikely to change). I took the opportunity to remove the option from the constructor.

### DIFF
--- a/lib/maths/common/CQuantileSketch.cc
+++ b/lib/maths/common/CQuantileSketch.cc
@@ -68,7 +68,7 @@ ml_shuffle_128(float32x4_t a, float32x4_t b, MASK) {
     return result;
 }
 
-inline __attribute__((always_inline)) auto ml_rotate_128(x) {
+inline __attribute__((always_inline)) auto ml_rotate_128(float32x4_t x) {
     float32x2_t x21{vget_high_f32(vextq_f32(x, x), 3)};
     float32x2_t x03{vget_low_f32(vextq_f32(x, x), 3)};
     return vcombine_f32(x21, x03);
--- a/lib/maths/common/CQuantileSketch.cc
+++ b/lib/maths/common/CQuantileSketch.cc
@@ -49,7 +49,7 @@
 
 // clang-format off
 #define ml_unaligned_load_128(x) vld1q_f32(x)
-#define ml_unaligned_store_128(x) vst1q_f32(x)
+#define ml_unaligned_store_128(x, y) vst1q_f32(x, y)
 #define ml_minimum_128(x, y) vminq_f32(x, y)
 #define ml_subtract_128(x, y) vsubq_f32(x, y)
 #define ml_multiply_128(x, y) vmulq_f32(x, y)
@@ -69,8 +69,8 @@ ml_shuffle_128(float32x4_t a, float32x4_t b, MASK) {
 }
 
 inline __attribute__((always_inline)) auto ml_rotate_128(float32x4_t x) {
-    float32x2_t x21{vget_high_f32(vextq_f32(x, x), 3)};
-    float32x2_t x03{vget_low_f32(vextq_f32(x, x), 3)};
+    float32x2_t x21{vget_high_f32(vextq_f32(x, x, 3))};
+    float32x2_t x03{vget_low_f32(vextq_f32(x, x, 3))};
     return vcombine_f32(x21, x03);
 }
 
--- a/lib/maths/common/CQuantileSketch.cc
+++ b/lib/maths/common/CQuantileSketch.cc
@@ -47,22 +47,24 @@
 
 #include <arm_neon.h>
 
+// clang-format off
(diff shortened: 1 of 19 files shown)


lib/maths/common/CQuantileSketch.cc
```diff
@@ -99,19 +177,34 @@ std::ptrdiff_t nextDifferent(TFloatFloatPrVec& knots, std::size_t index) {
     return next.index();
 }
 
+std::size_t fastSketchSize(double reductionFactor, std::size_t size) {
+    size = static_cast<std::size_t>(
+        static_cast<double>(size) * CQuantileSketch::REDUCTION_FACTOR / reductionFactor + 0.5);
+    return size + (3 - (size + 1) % 3) % 3;
```
valeriy42, line 183: Where does `3` come from? Can we use a constant here?

valeriy42 (approved):
LGTM. Great results!

lib/maths/common/CQuantileSketch.cc
tveasey: This ensures that the loop which computes the weights never have any left over costs to compute, since we take steps of size 3.

lib/maths/common/CQuantileSketch.cc
```diff
@@ -24,8 +24,89 @@
 #include <algorithm>
 #include <cmath>
 #include <limits>
+#include <numeric>
+#include <optional>
 #include <random>
 
+#if defined(__SSE4_2__)
```
droberts195, line 31: `xmmintrin.h` is for the original SSE instructions, not SSE 4.2:

```
// instrinsics prototypes
#include <xmmintrin.h>     //SSE
#include <emmintrin.h>     //SSE2
#include <pmmintrin.h>     //SSE3
#include <tmmintrin.h>     //SSSE3
#include <smmintrin.h>     //SSE4.1
#include <nmmintrin.h>     //SSE4.2
```

So it should be this:

```suggestion
#if defined(__SSE__)
```

If that then shows a problem on Windows then I'd rather explicitly exclude Windows rather than use a check for a macro that Windows doesn't have to sneakily exclude it.

tveasey: Windows was a mistake in the test. In this case, I do only use SSE instructions since it compiles after all. I think I'll probably just keep including the minimum set I need, but agree swapping to __SSE__ probably makes sense. (I went with __SSE4_2__ because it is also understood by all our compilers and we compile with SSE4.2 enabled.)

droberts195: I don't think Windows was compiling the code in this branch before.  Windows doesn't have an option to use SSE4.2 instructions unless you are prepared to use AVX.  There's no option in between SSE2 and AVX - see https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-160.  So I think on Windows checking `__SSE4_2__` without using the `/arch:AVX` compiler option would mean this optimisation wasn't used when it could be.  And we don't want to use `/arch:AVX` because that stops us working on a whole load of AMD processors that some users may still be using.

tveasey: Right that's worth knowing. I updated the ifdef to your suggestion [here](https://github.com/elastic/ml-cpp/pull/2390/commits/f5cec9737cdf6326d6ece2bf39dde4e99ce25468). Note that the real fix for Windows was [this](https://github.com/elastic/ml-cpp/pull/2390/commits/c98cf0c1e5c21ee43210549a03b9f2ad179f234b) one.

PR #2391, opened 2022-08-17
[ML] Improve layout of aggregate derivatives for training

tveasey:
This migrates the count of values we maintain for splits into the buffer we use for storing accumulated derivatives. This has the advantage that we can reset and copy it in the same operations we use for derivates rather than separately iterating over the collection to update counts. Furthermore, we often get these operations for free since we pad buffers to maintain alignment.

This change showed up some deficiencies in `CChecksum` particularly for pointers. I fixed these and pulled non template implementations out of the header since it is widely included. This logically belongs with #2388. I don't think we need to document it separately, so I'm marking this as a non-issue.


valeriy42 (approved):
LGTM

PR #2392, opened 2022-08-18
[ML] Use CMake for configuration Python prototyping code

valeriy42:
This PR replaces the Makefile by CMakeLists.txt to be consistent with the rest of the project. I noticed that some Python unit tests were failing. I fixed and updated them.


jupyter/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/..
+    COMMENT "Create docker image for the specified in IMAGE_NAME"
+)
+
+add_custom_target(run_docker
+    COMMAND docker run -p 9999:9999 $ENV{IMAGE_NAME}:latest
+    COMMENT "Run docker container for specified IMAGE_NAME"
+)
+
+add_custom_target(clean_build
```
edsavage, line 93: Unfortunately this isn't portable as it assumes the presence of a unix-like environment and commands. Maybe consider replacing with a combination of `CMake`'s `file(GLOB_RECURSE` and `file(REMOVE` ?

edsavage (commented):
Looking pretty good Valeriy.. there's just a few places left that aren't platform independent. It would be great if we could replace that functionality using `CMake` native commands.

jupyter/CMakeLists.txt
valeriy42: Thank you. I updated the cleaning procedure

jupyter/CMakeLists.txt
```diff
@@ -90,24 +90,8 @@ add_custom_target(run_docker
...
-    COMMAND @find ${SRC_DIR} -name '*.pyc' -delete \;
-	COMMAND @find ${SRC_DIR} -name '__pycache__' -delete \;
-	COMMAND @find ${SRC_DIR} -name '*.egg-info' -exec rm -r {} +
-	COMMAND @find ${TEST_DIR} -name '*.pyc' -delete 2>/dev/null || true \;
-	COMMAND @find ${TEST_DIR} -name '__pycache__' -delete 2>/dev/null || true \;
-	COMMAND -@rm -rf __pycache__ \;
-    COMMAND -@rm ${REQUIREMENTS_LOG}
-    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-)
+file(GLOB CLEAN_BUILD_FILES LIST_DIRECTORIES true ${SRC_DIR}/*.egg-info ${SRC_DIR}/**/__pycache__ ${TEST_DIR}/**/__pycache__ ${TEST_DIR}/**/*.pyc .pytest_cache .ipynb_checkpoints)
```
edsavage, line 93: Because of the fact that CMake is a build system generator, this `GLOB` command gets run only when `cmake .` is run, so it may or may not find the files and directories desired to be cleaned up. It may seem clunkier but the recommended way of doing this sort of thing is to have a script e.g. `clean_build.cmake` with contents something like...
```
set(SRC_DIR src)
set(TEST_DIR tests)

function(clean)
  foreach(file ${CLEAN_BUILD_FILES})
    message(STATUS ${file})
  endforeach() 
  if(CLEAN_BUILD_FILES)
    file(REMOVE_RECURSE ${CLEAN_BUILD_FILES})
  endif()
endfunction()

file(GLOB CLEAN_BUILD_FILES LIST_DIRECTORIES true ${SRC_DIR}/*.egg-info ${SRC_DIR}/**/__pycache__ ${TEST_DIR}/**/__pycache__ ${TEST_DIR}/**/*.pyc .pytest_cache .ipynb_checkpoints)

clean()
```

and then the `clean_build` custom target becomes

```
add_custom_target(clean_build
    COMMAND ${CMAKE_COMMAND} -P clean_build.cmake
)
```
that way, when the `clean_build` target is called it performs the search at run time - not configure time

valeriy42: Thank you for this example! I updated the code as you suggested.

edsavage (approved):
LGTM

PR #2394, opened 2022-08-18
[ML] Remove the source of CI debug build failures

edsavage:
Remove a call to a protected method called in TRACE logging


dimitris-athanasiou (approved):
LGTM

PR #2395, opened 2022-08-22
[ML] Raise the limit on maximum number of classes for multi-class classification

### LINKED ISSUE #2246: [ML] Increase the maximum number of classes we will support
Because we compute the Hessian of the cross-entropy, for classification we don't scale well to many classes and have limited the maximum number to 30.

We can likely already increase that limit as a result of performance enhancements. However, to increase it significantly we need to think about how to tackle this bottleneck. A reasonable option to explore is identifying classes which are well separated in feature space. For any given class pair (i,j) for which there aren't regions where you are choosing between i and j then one could safely zero {H(i,j), H(j,i)}. This would in effect create block diagonal Hessians and we can bound the maximum block size as we do now.

tveasey:
This addresses the bottleneck preventing us from increasing the maximum number of distinct classes we'll support and raises the limit from 30 to 100.

The issue is computing the hessian which scales like `n(n+1)/2` for `n` distinct classes. Here I've taken the approach of targeting just the classes which satisfy `arg top-k_{c \in C}{ \sum_i 1{c_i = c} loss(p_i, c_i) }` where `p_i` are the prediction probabilities for the i'th example, `c_i` is the true class, `C` is the set of all classes and `1{.}` denotes the indicator function, when computing best splits. This limits the hessian to `k(k+1)/2` and I've capped `k` at 20. We simply lump all classes not selected into a hold all other class when we chose the best splits and compute the expected derivatives of the loss for them. However, since we choose classes afresh for each round of boosting, we should still find splits which discriminate between all classes in the forest as a whole.

I tested this on synthetic data with 30 and 40 classes generated by the distribution `p = softmax(W x + n)` for `x` the features, `n` noise and `p` the class probabilities (see testMultinomialLogisticRegressionForManyClasses). Since I reduced `k` the maximum number of classes we consider per tree to 20 both are in the approximation regime. For the "old" case with 40 classes I increased the threshold. The results for the test data are as follows:

 approach | # classes|runtime /s | KL vs generating distribution
---- | ---- | ---- | ----
old | 30 | 450 | 0.31
old | 40 | 1954 | 0.67
new | 30 | 285 | 0.27
new | 40 | 550 | 0.66

Note that lower KL divergence is better so these represent a win for both runtime and accuracy.

This closes #2246.

### DIFF
--- a/include/maths/analytics/CBoostedTreeLoss.h
+++ b/include/maths/analytics/CBoostedTreeLoss.h
@@ -987,7 +987,7 @@ private:
     const TNodeVec* m_Tree{nullptr};
 };
 
-//!  \brief The loss for multinomial logistic regression.
+//! \brief The loss for multinomial logistic regression.
 //!
 //! DESCRIPTION:\n
 //! This targets the cross-entropy loss using the forest to predict the class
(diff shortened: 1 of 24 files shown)


include/maths/analytics/CBoostedTreeLoss.h
```diff
@@ -991,7 +1059,36 @@ class MATHS_ANALYTICS_EXPORT CPseudoHuber final : public CLoss {
     bool acceptRestoreTraverser(core::CStateRestoreTraverser& traverser) override;
 
 private:
-    double m_Delta;
+    std::size_t m_NumberClasses;
+};
+
+class MATHS_ANALYTICS_EXPORT CMultinomialLogisticSubsetLoss final
```
valeriy42, line 1065: Can you please add a comment about the purpose of this class?

valeriy42 (approved):
LGTM. Great job on improving the performance and accuracy! 🚀

include/maths/analytics/CBoostedTreeLoss.h
tveasey: Done in [48903cd](https://github.com/elastic/ml-cpp/pull/2395/commits/48903cd58633aaacd60ad6d7ad64b7cf9acaa034).

PR #2396, opened 2022-08-24
[ML] Explain anomalies

### LINKED ISSUE #2393: [ML] Explaining anomalies
When looking at the anomalies in the anomaly viewer, users often ask themselves why a certain anomaly has received a seemingly unexpected score. This results in multiple channels of communication. With the new Anomaly Explanations, we give these users meaningful and just-in-time information to explain the scoring algorithm. 

In the anomaly viewer, next to the listed anomalies, you can see the description of different factors that influenced the anomaly score.

## What is included in the explanation?

The explanation includes
- Detector function: e.g. rare_count, rare, mean, sum, etc.
- What happened:
  - The detector identifies the function value in the bucket as unusually high/low
  - The detector identifies the value for the first time for this field
- Which buckets are considered
  - The current bucket function value
  - The current multi-bucket function value
  - Both of the above
- What affects the anomaly score
  - Variation at similar times is larger than is typical => the score is reduced
  - The bucket contains fewer values than is typical => the score is reduced
  - The anomaly is shorter than other anomalies seen before (short spike) => the score is reduced.
  - The function value is unusual for the multi-bucket, but the current bucket is normal => the score is reduced.

## Examples

- This is an infrequently seen value for <X>. The count in the time bucket is unusually high. The score is reduced because variation at similar times, such as the hour of the day, is larger than is typical. The score is increased because the anomaly is longer than others we’ve seen for <Y>.
- This is the first time we’ve seen value <Y> for field <X>.
- The mean of the values in the time bucket is unusually low. The score is reduced because the bucket contains fewer values than is typical, so our prediction is less confident. The score is further reduced because the anomaly is shorter than others we have seen for <Y>.
- Both the sum of values in the current time bucket and the sum of values over recent time buckets are unusually high.
The max value has been unusually high for recent time buckets. The score is reduced because the current bucket max is normal.

valeriy42:
This PR surfaces new information about the impact of the factors on the initial anomaly score:
- single bucket impact is determined by the deviation between actual and typical in the current bucket
- multi-bucket impact is determined by the deviation between actual and typical in the past 12 buckets 
- anomaly characteristics are statistical properties of the current anomaly compared to the historical observations
- high variance penalty is the reduction of anomaly score in the buckets with large confidence intervals.
- incomplete bucket penalty is the reduction of anomaly score in the buckets with fewer samples than historically expected.

Additionally, we compute lower- and upper-confidence bounds and the typical value for the anomaly records. This improves the explainability of the cases where the model plot is not activated with only a slight overhead in performance (1-2%).

This PR accompanies elastic/elasticsearch#90675, which documents the enhancement. Therefore, I label this PR as a non-issue.

Closes #2393


lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -273,6 +274,20 @@ class CTimeSeriesAnomalyModel {
     //! Persist by passing information to \p inserter.
     void acceptPersistInserter(core::CStatePersistInserter& inserter) const;
 
+    double sumPredictionError() const {
+        if (!m_Anomaly) {
```
tveasey, line 278: nit (conventions)
```suggestion
        if (m_Anomaly == std::nullopt) {
```
And same below.

include/maths/common/CModel.h
```diff
@@ -245,6 +246,37 @@ class MATHS_COMMON_EXPORT CModelProbabilityParams {
...
+    enum EAnomalyType { E_UNKNOWN = 0, E_DIP, E_SPIKE };
+
+    EAnomalyType s_AnomalyType{E_UNKNOWN};
+    std::size_t s_AnomalyLength{0};
+    int s_SingleBucketImpact{0};
+    int s_MultiBucketImpact{0};
+    int s_AnomalyCharacteristicsImpact{0};
+    double s_LowerConfidenceBound{std::numeric_limits<double>::infinity()};
+    double s_TypicalValue{std::numeric_limits<double>::infinity()};
+    double s_UpperConfidenceBound{std::numeric_limits<double>::infinity()};
```
tveasey, line 260: It feels more natural that these are initialised to quiet_NaN for missing, but don't mind too much if this is a pain to change.

lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -1007,11 +1046,31 @@ bool CUnivariateTimeSeriesModel::uncorrelatedProbability(
...
+                static_cast<int>(std::round(-std::log10(pAnomaly)));
+        }
+
         probabilities.push_back(pAnomaly);
         featureProbabilities.emplace_back(
             common::SModelProbabilityResult::E_AnomalyModelProbability, pAnomaly);
     }
 
+    if (pOverall < common::LARGEST_SIGNIFICANT_PROBABILITY) {
+        LOG_DEBUG(<< "Computing confidence bounds");
```
tveasey, line 1066:
```suggestion
        LOG_TRACE(<< "Computing confidence bounds");
```

lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -1007,11 +1046,31 @@ bool CUnivariateTimeSeriesModel::uncorrelatedProbability(
 
         double pAnomaly;
         std::tie(pOverall, pAnomaly) = m_AnomalyModel->probability(pSingleBucket, pOverall);
+        if (pAnomaly < common::LARGEST_SIGNIFICANT_PROBABILITY) {
+            result.s_AnomalyScoreExplanation.s_AnomalyType =
+                (m_AnomalyModel->sumPredictionError() > (0.0))
+                    ? common::SAnomalyScoreExplanation::E_SPIKE
+                    : common::SAnomalyScoreExplanation::E_DIP;
+            result.s_AnomalyScoreExplanation.s_AnomalyLength =
+                std::min(m_AnomalyModel->length(time), 12UL);
```
tveasey, line 1055: I'm not sure this should be min. I would make this something like:
```
result.s_AnomalyScoreExplanation.s_MultiBucketImpact != 0 ? std::max(m_AnomalyModel->length(time), 12UL) : m_AnomalyModel->length(time);
```
i.e. if we know multi-bucket is important we know last 12 buckets are relevant, but it could be longer. Also, 12 feels like a magic number. I would add an accessor to `m_MultibucketFeature` which returns `m_SlidingWindow.capacity()`.

lib/model/CProbabilityAndInfluenceCalculator.cc
```diff
@@ -625,6 +625,37 @@ void CProbabilityAndInfluenceCalculator::add(const CProbabilityAndInfluenceCalcu
...
+                 other.m_AnomalyScoreExplanation.s_TypicalValue);
+    m_AnomalyScoreExplanation.s_UpperConfidenceBound =
+        std::min(m_AnomalyScoreExplanation.s_UpperConfidenceBound,
+                 other.m_AnomalyScoreExplanation.s_UpperConfidenceBound);
+    m_AnomalyScoreExplanation.s_HighVariancePenalty =
+        m_AnomalyScoreExplanation.s_HighVariancePenalty ||
+        other.m_AnomalyScoreExplanation.s_HighVariancePenalty;
+    m_AnomalyScoreExplanation.s_IncompleteBucketPenalty =
+        m_AnomalyScoreExplanation.s_IncompleteBucketPenalty ||
+        other.m_AnomalyScoreExplanation.s_IncompleteBucketPenalty;
```
tveasey, line 658: I think it is more logical to reset the explanation if p is less than this probability. So before updating compute:
```
double pThis{1.0};
m_Probability.calculate(pThis);
```
then here:
```
if (p < pThis) {
  m_AnomalyScoreExplanation = other.m_AnomalyScoreExplanation;
}
```

tveasey (commented):
Overall looks good just a couple of small comments.

valeriy42:
Thank you for the review, @tveasey. I addressed your comments, and it would be great if you could take another look.

tveasey (approved):
Thanks @valeriy42. LGTM

PR #2397, opened 2022-09-01
[ML] Tidy up utility functions using parameter packs

tveasey:
Some of our utility functions are ideally suited to parameter packs and were implemented before we adopted C++11. This migrates:
1. `COrderings::lexicographical_compare`,
2. `COrderings::simultaneousSort`, and
3. `CSetTools::simultaneousRemoveIf`.

Also, `CTools::inplace_set_difference` can just be replaced with `std::set_differerence`, `COrderings::lexicographical_compare` should follow our repository naming conventions, and we should prefer std::less<> and std::greater<> in template comparison operators since they are required to provide a total order of pointers.


droberts195 (approved):
LGTM if you can get it to compile on all the platforms

PR #2398, opened 2022-09-02
[ML] Don't keep categorization tokens when existing category matches

droberts195:
In categorization, when a new message matches an existing category
we don't need to keep a record of any new distinct tokens found in
the new message, as the category definition doesn't need them.

We used to store all unique tokens ever seen, but this can lead to
unbounded memory build-up if a particular type of logs has some
sort of unique ID that doesn't get filtered by the token filter of
the categorization analyzer.

This PR removes new distinct tokens from the token store for all
code paths other than a message causing a new category to be
created.

### DIFF
--- a/lib/model/CTokenListDataCategorizerBase.cc
+++ b/lib/model/CTokenListDataCategorizerBase.cc
@@ -69,11 +69,13 @@ CTokenListDataCategorizerBase::computeCategory(bool isDryRun,
     std::size_t workWeight{0};
     std::size_t minReweightedWorkWeight{0};
     std::size_t maxReweightedWorkWeight{0};
+    std::size_t preExistingTokenCount{m_TokenIdLookup.size()};
     auto preTokenisedIter = fields.find(PRETOKENISED_TOKEN_FIELD);
     if (preTokenisedIter != fields.end()) {
         if (this->addPretokenisedTokens(preTokenisedIter->second, m_WorkTokenIds,
                                         m_WorkTokenUniqueIds, workWeight, minReweightedWorkWeight,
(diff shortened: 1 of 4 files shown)


lib/model/CTokenListDataCategorizerBase.cc
```diff
@@ -467,6 +472,12 @@ void CTokenListDataCategorizerBase::addCategoryMatch(bool isDryRun,
     }
 }
 
+void CTokenListDataCategorizerBase::discardLatestTokens(std::size_t previousTokenCount) {
+    while (m_TokenIdLookup.size() > previousTokenCount) {
```
tveasey, line 476: I thought a random access `multi_index` has [resize](https://www.boost.org/doc/libs/1_53_0/libs/multi_index/doc/reference/rnd_indices.html)?

tveasey (approved):
LGTM (although I do have a question about why you use pop_back rather than resize).

lib/model/CTokenListDataCategorizerBase.cc
droberts195: It does, but that would require that `CTokenInfoItem` has a default constructor, or that we construct a `CTokenInfoItem` with some special values that would be copied if the container was resized to be bigger.  (Obviously it never would get copied, since we only resize to smaller sizes.)

droberts195: I had a look and `erase` would be a way to do this in one function call from this code.  But internally `erase` is implemented by moving the elements to be deleted to the end of the vector and then repeatedly calling `pop_back`, so it wouldn't be any faster than what's here now.

PR #2402, opened 2022-09-06
[ML] Add missing Windows export

droberts195:
This only shows up as a problem when inlining is disabled.

Followup to #2397


edsavage (approved):
LGTM

PR #2404, opened 2022-09-21
[ML] Fix seed parameter name

valeriy42:
Java backend was passing the seed configuration parameter using the name `randomize_seed`, while C++ code was expecting `seed`. This PR fixes the misalignment.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -40,10 +40,6 @@
 * Increase the limit on the maximum number of classes to 100 for training classification
   models. (See {ml-pull}2395[#2395] issue: {ml-issue}2246[#2246].)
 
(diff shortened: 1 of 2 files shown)


droberts195 (approved):
LGTM

I think this should probably be backported to 8.5.0 and 7.17.7 as well, because users who think they're fixing the analysis currently aren't.

droberts195:
The Linux test failures are because ML is on 8.6.0 on main but Elasticsearch is still on 8.5.0.  If you push a release note tomorrow morning then that will trigger a retest and hopefully the repos will be consistent by then.

droberts195:
retest

droberts195:
We discussed in another channel that this only affects incremental training. Since that isn't realised this PR doesn't need to be release noted or backported.

PR #2406, opened 2022-09-27
[ML] Fix failure computing leaf values with multithreading for multi-class classification

tveasey:
In the case that all the samples are duplicates, merging our stream dissimilar vector sampler discards all samples. This behaviour was introduced by #2401: subsequently categorical sample without replacement will not select any sample with zero weight. The upshot is that leaf weight estimation for multi-class classification, which makes use of this class, could yield undefined behaviour. I now explicitly trap the error case and fail gracefully. I also explicitly handle duplicates in both sampling and merging dissimilar vector samplers.

This fixes fallout from #2401. Since this isn't released I've marked as a non-issue.


valeriy42 (approved):
LGTM

PR #2410, opened 2022-09-30
[ML] Fix for 'No statistics' error message

edsavage:
An anomaly detector job ignores records that fall in the initial, incomplete bucket. However, a model is created with a start time of -1 in the expectation that more records will be received that fall into subsequent buckets. In the case that no more records are received the model continues to exist with start time of -1.

In the scenario that a job is created initially as lookback and has such models with start time of -1 in existence at the point the job closes, and then is re-opened as a realtime job at some point in the future, any attempt to sample the model will fail with the 'No statistics at ...' error message due to the start time (-1) falling outside the current bucket.

This change always updates the model's current bucket stats start time to that of the current record, even if the record data is not to be added to the associated data gatherer.

### DIFF
--- a/devbin/model_extractor/Main.cc
+++ b/devbin/model_extractor/Main.cc
@@ -147,7 +147,7 @@ int main(int argc, char** argv) {
     while (restoredJob.restoreState(restoreSearcher, completeToTime)) {
         assert(completeToTime > prevCompleteToTime);
         prevCompleteToTime = completeToTime;
-        LOG_DEBUG(<< "Restore complete to time " << completeToTime << std::endl);
+        LOG_DEBUG(<< "Restore complete to time " << completeToTime);
 
         core::CNamedPipeFactory::TOStreamP persistStrm{&ioMgr.outputStream(),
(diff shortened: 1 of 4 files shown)


droberts195:
> This change always updates the model's start time to that of the current record, even if the record data is not to be added to the associated data gatherer.

Shouldn't it be the model's _current bucket stats_ start time?

edsavage:
Yes, I'll update the description.

droberts195 (approved):
LGTM

Since Tom's off I'll approve this so we can get it merged and monitor for any unforeseen side effects over the longest possible period before release.  Please ask Tom to review once he's back and if he has any suggestions they can go into a followup PR.

davidkyle:
I am seeing this error in serverless logs

```
[gallery_sum_bytes] [autodetect/179] [CIndividualModel.cc@136] No statistics at 1719197550 for sum bytes partition=clientip, current bucket = [-1,899), partitionFieldValue = 175.144.21.153, personName = | repeated [5]
```

Looks very siimilar

PR #2412, opened 2022-10-18
[ML] Add option to run pytorch_inference in low priority

dimitris-athanasiou:
This commit adds an option `--lowPriority` to the pytorch_inference program so it can be run in low priority.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -131,12 +131,13 @@ int main(int argc, char** argv) {
     std::int32_t numAllocations{1};
     std::size_t cacheMemorylimitBytes{0};
     bool validElasticLicenseKeyConfirmed{false};
+    bool lowPriority{false};
 
     if (ml::torch::CCmdLineParser::parse(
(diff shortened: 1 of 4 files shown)


bin/pytorch_inference/Main.cc
```diff
@@ -198,6 +199,10 @@ int main(int argc, char** argv) {
 
     // Reduce memory priority before installing system call filters.
     ml::core::CProcessPriority::reduceMemoryPriority();
+    if (lowPriority == true) {
+        LOG_DEBUG(<< "Running with low priority");
+        ml::core::CProcessPriority::reduceCpuPriority();
```
droberts195, line 204: In the other programs we do this after connecting the named pipes, because named pipe timeouts stop the process working altogether. For example see https://github.com/elastic/ml-cpp/blob/59242db5f3e7666ddcd08397f9367e5e9bf1f4b4/bin/autodetect/Main.cc#L173-L176

droberts195 (commented):
There’s a format error in `Main.cc` - run `dev-tools/clang-format.sh` to fix it.

PR #2414, opened 2022-10-28
[ML] Fix for "No counts available" error message

### LINKED ISSUE #2411: [ML] Error message 'No counts available' seen when re-opening lookback job as realtime
On occasion, when a lookback job has completed and has then been re-opened, and error message similar to

```
[2022-09-28T11:51:25,925][ERROR][o.e.x.m.p.l.CppLogMessageHandler] [node-0] [gallery-partial-2020-sum-bytes-12] [autodetect/97529] [CBucketGatherer.cc@464] No counts available at 1585698300, current bucket = [1585699200,1585700100)
```

is seen. The error message appears to be a one off and due to the fact that the time `1585698300` is earlier than the bucket start time `1585699200`.

edsavage:
When restarting a look-back job as a real-time one on occasion error messages similar to
```
[CBucketGatherer.cc@464] No counts available at 1585698300, current bucket = [1585699200,1585700100)
```
 are seen.

Investigation reveals that these error messages stem from the situation where an incomplete initial bucket for a partition was persisted at the close of the look-back job. When the job is re-opened as real-time the start time of the anomaly detector in question is still before that of the associated bucket gatherer, triggering the error.

This PR adds a check to determine if an incomplete initial bucket has been restored from persisted state, and if so to skip any attempt to output results for that bucket (which is essentially what would have happened prior to this change except without logging the error).

Fixes #2411


tveasey (approved):
LGTM

PR #2420, opened 2022-11-07
[ML] Fix 3rd party license sort order and make MKL more searchable

droberts195:
To get the 3rd party licenses to sort correctly in alphabetical order we need the filenames to sort in the same order as the license names.

This change also includes the abbreviation MKL for Math Kernel Library so that people searching for "MKL" in the 3rd party license list will find it.


edsavage (approved):
LGTM

PR #2423, opened 2022-11-23
[ML] First foray into Buildkite CI

edsavage:
This first iteration of `ml-cpp` BuildKite pipelines supports linux x86_64 builds (along with unit testing) only.

The boost unit test framework has been modified to output results in `junit` format alongside the current `xml` output. Test results are uploaded to BuildKite's "Test Analytics" dashboard - https://buildkite.com/organizations/elastic/analytics/suites/ml-cpp-testing-testing?branch=all+branches

The subset of ES integration tests normally run by PR builds in Jenkins have been disabled under BuildKite. The tests themselves run and pass, however the test cluster fails to be torn down cleanly at the conclusion of the test run. The reason for this is as yet unknown.

Subsequent PRs will add support for other supported architectures and operating systems.


.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -0,0 +1,66 @@
...
+# 2.0 and the following additional limitation. Functionality enabled by the
+# files subject to the Elastic License 2.0 may only be used in production when
+# invoked by an Elasticsearch process with a license key installed that permits
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+
+# This script gets run within the Docker container when a build is done in a
+# Docker container.
```
droberts195, line 14: Most of what's in this script looks very similar/identical to what's in `docker_entrypoint.sh`.

Could this file be changed to be a very thin wrapper around `docker_entrypoint.sh`?

Eventually BuildKite will have to be our CI system for _all_ supported branches, including 7.17, and 7.17 is not using CMake. So the BuildKite files will be easier to backport if they call other scripts that vary between branches rather than duplicating them.

If this is absolutely impossible then we can duplicate.  But another thing is that we'll need to maintain the ability to build locally in a Docker container for debugging and iterating on the Docker builds themselves.  If we have two scripts in two different directories that contain very similar code but one hardly ever gets used and the other gets used all the time then the one that hardly ever gets used will fall into disrepair.

edsavage:
buildkite test this

dev-tools/docker/build_linux_jdk.sh
```diff
@@ -0,0 +1,48 @@
...
+#
+
+# Builds the Docker image that can be used to run a subset of the ES
+# integration tests.
+#
+# This script is not intended to be run regularly.
+
+if [ `uname -m` != x86_64 ] ; then
+    echo "Native build images must be built on the correct hardware architecture"
+    echo "Required: x86_64, Current:" `uname -m`
```
droberts195, line 20: We'll need a similar image for `aarch64` soon too.

It might be worth thinking about the container naming here. With our existing images `aarch64` came along a long time after we first created the images, hence we have `ml_linux_build` and `ml_linux_native_aarch64_build`. But if we'd thought from the start that we'd be supporting multiple architectures we wouldn't have named them like this.

I think there are 2 options:

1. Fit with our current image naming, so `ml_cpp_jdk` -> `ml_cpp_linux_jdk`, and later introduce `ml_cpp_linux_aarch64_jdk`.
2. Use architecture in all names from the start, so `ml_cpp_jdk` -> `ml_cpp_linux_x86_64_jdk`, and later introduce `ml_cpp_linux_aarch64_jdk`.

edsavage: > 2\. Use architecture in all names from the start, so `ml_cpp_jdk` -> `ml_cpp_linux_x86_64_jdk`, and later introduce `ml_cpp_linux_aarch64_jdk`.

I like this latter naming scheme, even though it is at odds with the other existing images

edsavage:
jenkins retest this

.buildkite/job-build-test-all-debug.yml
```diff
@@ -0,0 +1,36 @@
...
+  CMAKE_FLAGS: "-DCMAKE_TOOLCHAIN_FILE=cmake/linux-x86_64.cmake"
+  ML_DEBUG: 1
+steps:
+  - label: "Testing of the Machine Learning main branch with full debug and assertions."
+    timeout_in_minutes: 180
+    agents:
+      cpu: "6"
+      ephemeralStorage: "20G"
+      memory: "64G"
+        image: "docker.elastic.co/ml-dev/ml-linux-build:24.bk"
```
droberts195, line 24:
```suggestion
        image: "docker.elastic.co/ml-dev/ml-linux-build:25"
```

Or if that doesn't work then whatever stops it working needs to be fixed.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -0,0 +1,52 @@
...
+# 2.0 and the following additional limitation. Functionality enabled by the
+# files subject to the Elastic License 2.0 may only be used in production when
+# invoked by an Elasticsearch process with a license key installed that permits
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+
+set -euo pipefail
+
+export CMAKE_QUIET="yes"
```
droberts195, line 13: Does this mean we can never see the compile commands?

That's a pretty big drawback compared to Jenkins - it after making some build update it won't be possible to use CI to see the effect on all platforms without having to find appropriate machines and manually build on all platforms.

The optimum here is probably to log the build commands to a file, then store that file as an artifact, so that the primary build output is not cluttered with compile commands but they're available when needed.

dev-tools/aws_creds_from_vault.sh
```diff
@@ -42,12 +42,17 @@ case $- in
         ;;
 esac
 
-export VAULT_TOKEN=$(vault write -field=token auth/approle/login role_id="$VAULT_ROLE_ID" secret_id="$VAULT_SECRET_ID")
+PREFIX=""
+if [ "$BUILDKITE" = true ]; then
+    PREFIX="secret/ci/elastic-ml-cpp/"
+else
+    export VAULT_TOKEN=$(vault write -field=token auth/approle/login role_id="$VAULT_ROLE_ID" secret_id="$VAULT_SECRET_ID")
```
droberts195, line 49: Does this imply BuildKite automatically sets `VAULT_TOKEN`?

dev-tools/run_es_tests.sh
```diff
@@ -89,8 +89,10 @@ if [ `uname -m` = aarch64 ] ; then
     export ES_BUILD_JAVA=$(echo $ES_BUILD_JAVA | sed 's/.*jdk/jdk/')
 fi
 
-echo "Setting JAVA_HOME=$HOME/.java/$ES_BUILD_JAVA"
-export JAVA_HOME="$HOME/.java/$ES_BUILD_JAVA"
+if [ -z "${BUILDKITE}" ]; then
```
droberts195, line 92: I think lines 73-91 can also go inside this `if`.

dev-tools/vault-cp.sh
```diff
@@ -0,0 +1,65 @@
...
+# Where the first path belongs to https://secrets.elastic.co:8200
+# and the second to https://vault-ci-prod.elastic.dev.
+# It is likely that this script will only need to be run once.
+#
+
+set -eo pipefail
+
+# ensure we were given two command line arguments
+if [[ $# -ne 2 ]]; then
+	echo 'usage: vault-cp SOURCE DEST' >&2
```
droberts195, line 26: This whole file is indented using tabs

.buildkite/scripts/steps/build_and_test.sh
edsavage: Yes, it turns off cmake's verbose flag, resulting in a summary of the build commands e.g.
```
  | [  0%] Building CXX object lib/seccomp/CMakeFiles/MlSeccomp.dir/CSystemCallFilter_MacOSX.cc.o
  | [  0%] Building CXX object bin/controller/CMakeFiles/Mlcontroller.dir/CBlockingCallCancellingStreamMonitor.cc.o
  | [  0%] Building CXX object bin/categorize/CMakeFiles/Mlcategorize.dir/CCmdLineParser.cc.o
  | [  0%] Building CXX object lib/ver/CMakeFiles/MlVer.dir/CBuildInfo.cc.o
  | [  0%] Building CXX object bin/data_frame_analyzer/CMakeFiles/Mldata_frame_analyzer.dir/CCmdLineParser.cc.o
  | [  0%] Building CXX object bin/controller/CMakeFiles/Mlcontroller.dir/CCmdLineParser.cc.o
  | [  0%] Building CXX object bin/normalize/CMakeFiles/Mlnormalize.dir/CCmdLineParser.cc.o
  | [  1%] Building CXX object bin/controller/CMakeFiles/Mlcontroller.dir/CCommandProcessor.cc.o
  | [  1%] Building CXX object bin/controller/CMakeFiles/Mlcontroller.dir/CResponseJsonWriter.cc.o
  | [  1%] Building CXX object bin/autodetect/CMakeFiles/Mlautodetect.dir/CCmdLineParser.cc.o
  | [  1%] Building CXX object bin/pytorch_inference/CMakeFiles/Mlpytorch_inference.dir/CBufferedIStreamAdapter.cc.o
  | [  1%] Building CXX object bin/pytorch_inference/CMakeFiles/Mlpytorch_inference.dir/CCmdLineParser.cc.o
  | [  3%] Building CXX object bin/pytorch_inference/CMakeFiles/Mlpytorch_inference.dir/CCommandParser.cc.o
  | [  3%] Building CXX object bin/pytorch_inference/CMakeFiles/Mlpytorch_inference.dir/CThreadSettings.cc.o
  | [  3%] Building CXX object bin/pytorch_inference/CMakeFiles/Mlpytorch_inference.dir/CResultWriter.cc.o
```
With full build output BuildKite's console buffer fills up and you can no longer "follow" the output. However that's really just a niggle. I'll turn verbose output back on.

dev-tools/aws_creds_from_vault.sh
edsavage: Yes, behind the scenes access to CI vault is automagically set up for us.

.buildkite/scripts/steps/build_and_test.sh
droberts195: > With full build output BuildKite's console buffer fills up and you can no longer "follow" the output.

That sounds like another bad problem.

Is there no way to log the full commands somewhere else but not log them to the console?

If there isn't then another alternative might be to have a build parameter so that usually the commands don't get logged but it's easy to kick off a job manually and set the parameter that switches on verbose output.

edsavage: Buildkite still retains the full log file and it is available for download, so there's no need to separately archive it.

As for optionally setting a build parameter controlling the flag, yes that is possible either via labels or comments. As it will require a little work on the buildkite PR hook and some additional code to parse the variables I propose to do this in a follow up PR...

droberts195 (approved):
LGTM

PR #2424, opened 2022-11-29
[ML] Improve performance of closing files before spawning

droberts195:
Before spawning new processes from the `controller` we close all open file descriptors except for stdin, stdout and stderr.

Previously this was done by checking every possible file descriptor to see if it was open, but this is very expensive if the file descriptor limit is high.  During a lookback a large number of `normalize` processes get started, and this could lead to significant CPU usage by the `controller` process.

This change makes the file closure code learn the highest open file descriptor each time it is used, and work on the basis that no more than 10 files will be opened in between calls to it. This significantly reduces `controller` CPU usage on machines that have high file descriptor limits and run a lot of `normalize` processes.


edsavage (approved):
LGTM

PR #2428, opened 2022-12-21
[ML] Use rapidjson's key function where possible

### LINKED ISSUE #383: [ML] Use rapidjson's key function where possible
Rapidjson introduced in a later version a `Key(string)` function. It is no different that then `String(string)` function. However, it does communicate better the intention of writing a field name instead of a string value. It would make our JSON writing code more readable if we replaced all calls to `String` with `Key` where field names are written.

dimitris-athanasiou:
Rapidjson introduced a `Key()` function on the writers which is effectively the same as `String()`. However, using it where keys are written improves readability of the code.

Closes #383


droberts195 (approved):
LGTM

PR #2433, opened 2023-01-13
[ML] Fix instructions for building PyTorch on Windows

### LINKED ISSUE #2432: [ML] Initial build of PyTorch 1.13.1 crashes on Windows unless Visual Studio is installed
The initial build of PyTorch 1.13.1 for Windows crashes with the following error if used on a machine that doesn't have Visual Studio installed:

image (10)

This happens because the build linked to an experimental Microsoft library that should be linked when the `/openml:llvm` option is used during compilation.

The PyTorch build does not use that option. However, the CMake FindOpenMP module is deciding to link `libomp.lib`, which is the import library for `libomp140.x86_64.dll`. The correct import library to go with the GA `libvcomp140.dll` is `vcomp.dll`, but that should get linked implicitly when required. The `libtorch_cpu.dll` we built on the first attempt actually shows _both_ of the OMP DLLs are linked to it. Hopefully we can find a way to only use functionality from the GA library `libvcomp140.dll`.

This problem needs to be resolved before 8.7 feature freeze, or else we will have to revert the PyTorch upgrade.

droberts195:
On Windows, recent updates of Visual Studio 2019 include an experimental improved OMP library. Because this library is experimental redistribution is not permitted.

When we built PyTorch 1.13.1 CMake's FindOpenMP module decided to link to this experimental OMP library, which then means the resulting library only works on machines with Visual Studio installed.

It's not very clear where in the PyTorch build configuration the fault lies. It could be in CMake itself or maybe there's some option PyTorch could be using to control CMake more tightly. CMake was _also_ linking to Microsoft's GA OMP library, so this PR works around the problem by temporarily renaming the experimental libraries during the build so that CMake does not find them.

Fixes #2432

### DIFF
--- a/build-setup/windows.md
+++ b/build-setup/windows.md
@@ -247,6 +247,11 @@ doing and never want to do for security reasons. Replacing the calls to
 potentially dangerous function calls in our shipped product will not encounter
 these functions that run external processes.
 
+In `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30133\lib\x64`
+rename `libomp.lib` to `libomp.lib.bak` and rename `libompd.lib` to `libompd.lib.bak`.
+This is to prevent CMake's FindOpenMP module from linking to Microsoft's experimental
+library that we are not allowed to redistribute.
+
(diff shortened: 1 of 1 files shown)


edsavage (approved):
LGTM

PR #2434, opened 2023-01-13
Test buildkite with linux aarch64 docker container

edsavage:
Experimenting with native linux aarch64 builds


.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -32,14 +32,31 @@ if [[ -n "$BUILDKITE_PULL_REQUEST" && -n "$VERSION_QUALIFIER" ]] ; then
...
+  GLIBC_VERSION=`ldconfig --version | head -1 | sed 's/ldconfig//'`
+
+  if [ "$RUN_TESTS" = false ] ; then
+    ${REPO_ROOT}/dev-tools/docker_build.sh linux_aarch64_native
+  else
+    ${REPO_ROOT}/dev-tools/docker_test.sh --extract-unit-tests linux_aarch64_native
+    echo "Re-running seccomp unit tests outside of Docker container - kernel: $KERNEL_VERSION glibc: $GLIBC_VERSION"
+    (cd ${REPO_ROOT}/cmake-build-docker/test/lib/seccomp/unittest && LD_LIBRARY_PATH=`cd ../../../../../build/distribution/platform/linux-aarch64/lib && pwd` ./ml_test_seccomp)
+  fi
+  exit $?
```
droberts195, line 57: I don't think it should exit here, as we want the artifact upload at the end.

But the `docker_entrypoint.sh` part below should only be called if running in Kubernetes.

dev-tools/docker_build.sh
```diff
@@ -67,6 +67,9 @@ cd "$TOOLS_DIR/.."
 # necessary network access
 3rd_party/pull-eigen.sh
 
+du -sh .
+df -h .
```
droberts195, line 71: Was this just temporary debug that can be removed now?

Or if you think this would be useful to keep forever then please add some `echo`s with titles for the output.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -32,14 +32,31 @@ if [[ -n "$BUILDKITE_PULL_REQUEST" && -n "$VERSION_QUALIFIER" ]] ; then
...
+  
+  KERNEL_VERSION=`uname -r`
+  GLIBC_VERSION=`ldconfig --version | head -1 | sed 's/ldconfig//'`
+
+  if [ "$RUN_TESTS" = false ] ; then
+    ${REPO_ROOT}/dev-tools/docker_build.sh linux_aarch64_native
+  else
+    ${REPO_ROOT}/dev-tools/docker_test.sh --extract-unit-tests linux_aarch64_native
+    echo "Re-running seccomp unit tests outside of Docker container - kernel: $KERNEL_VERSION glibc: $GLIBC_VERSION"
+    (cd ${REPO_ROOT}/cmake-build-docker/test/lib/seccomp/unittest && LD_LIBRARY_PATH=`cd ../../../../../build/distribution/platform/linux-aarch64/lib && pwd` ./ml_test_seccomp)
```
droberts195, line 55: Since this is running on a VM, we can run the ES integration tests on the same VM.

Just like we did in Jenkins in https://github.com/elastic/ml-cpp/blob/976a2a293ec815ea1b869fd61d243ff1dab30eb5/dev-tools/jenkins_ci.sh#L151-L153

That will mean we need a JVM installed on the VM though.  Eventually that should be achieved by using the same VM image as Elasticsearch, but for now maybe the appropriate `yum` command (or whatever package manager Amazon Linux uses) could be added to the `commands` section of `pipeline.yml`.

.buildkite/pipeline.yml
```diff
@@ -25,24 +25,55 @@ steps:
...
+    timeout_in_minutes: 180
+    agents:
+      provider: aws
+      instanceType: m6g.2xlarge
+      imagePrefix: ci-amazonlinux-2-aarch64
+      diskSizeGb: 100
+      diskName: '/dev/xvda'
+    commands:
+      - 'du -sh'
+      - 'df -h'
```
droberts195, line 57: Please remove these if they were temporary debug, and also on lines 28-29.

.buildkite/pipeline.yml
```diff
@@ -25,24 +25,55 @@ steps:
...
+      CMAKE_FLAGS: "-DCMAKE_TOOLCHAIN_FILE=/ml-cpp/cmake/linux-aarch64.cmake"
+      CPP_CROSS_COMPILE: ""
+      RUN_TESTS: "true"
+      BOOST_TEST_OUTPUT_FORMAT_FLAGS: "--logger=JUNIT,error,boost_test_results.junit"
+    plugins:
+      - test-collector#v1.2.0:
+          files: "lib/*/unittest/boost_test_results.junit"
+          format: "junit"
+
+  - wait
```
droberts195, line 75: As with the other PR, I don't think it's worth blocking until immediately before the final upload to S3.

dev-tools/docker_build.sh
edsavage: > Was this just temporary debug that can be removed now?

Yes, just temporary while I was trying to correctly size the VM's disk. I'll remove it.

.buildkite/pipeline.yml
```diff
@@ -79,21 +102,19 @@ steps:
...
     command: 
-      - "echo 'Java Integration Tests step is disabled, awaiting fix for cluster tear down.'"
-        #- 'buildkite-agent artifact download "build/*" .'
-        #- "mkdir -p ~/.ssh"
-        #- "ssh-keyscan github.com >> ~/.ssh/known_hosts"
-        #- '.buildkite/scripts/steps/run_es_tests.sh || (cd ../elasticsearch && find x-pack -name logs | xargs tar cvzf logs.tgz && buildkite-agent artifact upload logs.tgz && false)'
-    depends_on: "build_test_linux-x86_64-RelWithDebInfo"
+      - "sudo yum -y install java-17-amazon-corretto-devel"
+      - "java -version"
+      - 'buildkite-agent artifact download "build/*" . --step build_test_linux-aarch64-RelWithDebInfo'
```
droberts195, line 107: Would this work?

```suggestion
      - 'buildkite-agent artifact download "build/*linux-aarch64*" . --step build_test_linux-aarch64-RelWithDebInfo'
```

If it would then we'd avoid downloading pointless stuff, like artifacts for Mac and Windows.

.buildkite/pipeline.yml
```diff
@@ -50,15 +75,14 @@ steps:
...
-      - 'echo "checkout succeeded"'
       - ".buildkite/scripts/steps/build_and_test.sh"
     depends_on: "check_style"
     key: "build_linux_aarch64_cross-RelWithDebInfo"
     env:
       CPP_CROSS_COMPILE: "aarch64"
       CMAKE_FLAGS: "-DCMAKE_TOOLCHAIN_FILE=cmake/linux-aarch64.cmake"
       RUN_TESTS: "false"
-
+
```
droberts195, line 85: Nit: trailing whitespace added

droberts195 (approved):
LGTM, although please think about whether the integration test step can avoid downloading everything for the irrelevant platforms. If it's not easy to do this them we can come back to it in a followup PR.

.buildkite/pipeline.yml
edsavage: I don't think this is necessary as that is what the `--step` parameter is doing (see https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts-example).

Checking the BuildKite logs I can confirm that it is only downloading artifacts uploaded from the `build_test_linux-aarch64-RelWithDebInfo` step:

```
  | 2023-01-19 15:02:22 INFO   Searching for artifacts: "build/*" within step: "build_test_linux-aarch64-RelWithDebInfo" 
  | 2023-01-19 15:02:22 INFO   Found 2 artifacts. Starting to download to: /opt/buildkite-agent/builds/bk-agent-elastic-aws-1674140483898757664/elastic/ml-cpp
  | 2023-01-19 15:02:24 INFO   Successfully downloaded "build/distributions/ml-cpp-8.7.0-SNAPSHOT-linux-aarch64.zip" 44593845 bytes
  | 2023-01-19 15:02:40 INFO   Successfully downloaded "build/distributions/ml-cpp-8.7.0-SNAPSHOT-debug-linux-aarch64.zip" 838695567 bytes
```

droberts195: Great, thanks for confirming

PR #2435, opened 2023-01-13
[ML] Add cross compilation steps to BuildKite pipeline

edsavage:


.buildkite/pipeline.yml
```diff
@@ -43,6 +41,42 @@ steps:
...
+      - 'echo "checkout succeeded"'
+      - ".buildkite/scripts/steps/build_and_test.sh"
+    depends_on: "check_style"
+    key: "build_macos_aarch64_cross-RelWithDebInfo"
+    env:
+      CPP_CROSS_COMPILE: "macosx"
+      CMAKE_FLAGS: "-DCMAKE_TOOLCHAIN_FILE=cmake/darwin-x86_64.cmake"
+      RUN_TESTS: "false"
+
+  - wait
```
droberts195, line 78: I think it's best if the `wait` goes just before uploading artifacts at the very end. Running the Linux integration tests doesn't depend on macOS being built for example.

.buildkite/pipeline.yml
```diff
@@ -51,7 +85,6 @@ steps:
         #- "mkdir -p ~/.ssh"
         #- "ssh-keyscan github.com >> ~/.ssh/known_hosts"
         #- '.buildkite/scripts/steps/run_es_tests.sh || (cd ../elasticsearch && find x-pack -name logs | xargs tar cvzf logs.tgz && buildkite-agent artifact upload logs.tgz && false)'
-    depends_on: "build_test_linux-x86_64-RelWithDebInfo"
```
droberts195, line 54: I think you should keep this and only `wait` before uploading artifacts to S3.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -32,12 +34,6 @@ if [[ -n "$BUILDKITE_PULL_REQUEST" && -n "$VERSION_QUALIFIER" ]] ; then
...
 fi
 
-# Tests must be run in PR builds
-if [[ -n "$BUILDKITE_PULL_REQUEST" && "$RUN_TESTS" = false ]] ; then
-    echo "RUN_TESTS should not be false PR builds"
-    exit 3
-fi
-
 echo "environment variables:"
 env
```
droberts195, line 38: This should only be temporary. Printing every environment variable could give away sensitive information. Please remove this once debugging is complete.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -32,12 +34,6 @@ if [[ -n "$BUILDKITE_PULL_REQUEST" && -n "$VERSION_QUALIFIER" ]] ; then
     exit 2
 fi
 
-# Tests must be run in PR builds
-if [[ -n "$BUILDKITE_PULL_REQUEST" && "$RUN_TESTS" = false ]] ; then
-    echo "RUN_TESTS should not be false PR builds"
-    exit 3
-fi
-
```
droberts195, line 40: Please can you add in this functionality: https://github.com/elastic/ml-cpp/blob/5b3696ade2f54e4c3c7164d9e252df55111a9c72/dev-tools/jenkins_cross_compile.sh#L90-L96

In other words, in PR builds always do a debug build when cross compiling for linux-aarch64. The reasoning is in the code comment.

PR #2437, opened 2023-01-16
Fix s3 upload script

edsavage:
Take into account that BuildKite always sets a value for BUILDKITE_PULL_REQUEST and that it is set to "false" for non PR builds.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -68,4 +68,12 @@ steps:
     depends_on: "java_integration_tests"
     command:
       - 'buildkite-agent artifact download "build/*" .'
+      - "env"
       - "./.buildkite/scripts/steps/upload_to_s3.sh"
+    agents:
+      cpu: "16"
+      ephemeralStorage: "20G"
+      memory: "32G"
+      image: "docker.elastic.co/ml-dev/ml_cpp_linux_x86_64_jdk17:1"
+      # Run as a non-root user
+      imageUID: "1000"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -71,9 +71,9 @@ steps:
       - "env"
       - "./.buildkite/scripts/steps/upload_to_s3.sh"
     agents:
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/steps/upload_to_s3.sh
```diff
@@ -9,7 +9,7 @@
 # limitation.
 #
 # If this isn't a PR build and isn't a debug build then upload the artifacts
-if [[ -z "$BUILDKITE_PULL_REQUEST" && -z "$ML_DEBUG" ]] ; then
+if [[ -z "$BUILDKITE_PULL_REQUEST" || x"$BUILDKITE_PULL_REQUEST" = "xfalse" && -z "$ML_DEBUG" ]] ; then
```
droberts195, line 12: This is a very complex condition, mixing `||` and `&&` when the operator precedence of those varies between `[` `]` and `[[` `]]`.

I think it would be better to refactor it to take out that complexity.

```suggestion
if [[ x"$BUILDKITE_PULL_REQUEST" != xtrue && -z "$ML_DEBUG" ]] ; then
```

Plus also add a comment to say that we think BuildKite always sets `BUILDKITE_PULL_REQUEST` but want to treat not set as equivalent to `false`.

droberts195 (approved):
LGTM

.buildkite/pipeline.yml
```diff
@@ -68,4 +68,12 @@ steps:
     depends_on: "java_integration_tests"
     command:
       - 'buildkite-agent artifact download "build/*" .'
+      - "env"
       - "./.buildkite/scripts/steps/upload_to_s3.sh"
+    agents:
+      cpu: "16"
+      ephemeralStorage: "20G"
+      memory: "32G"
```
droberts195, line 76: This step shouldn't need anything like 16 CPUs or 32GB of RAM to work.

I would think 2 CPUs and 4GB of RAM would be fine.  Or if not, why is uploading a file so memory hungry?

PR #2439, opened 2023-01-17
[ML] Correct path to AWS credentials in CI vault

edsavage:
Fix a typo affecting the path to the AWS credentials used for uploading artifacts to S3


droberts195 (approved):
LGTM

PR #2440, opened 2023-01-17
[ML] Identify multimodal prior distributions for anomaly explanation

### LINKED ISSUE #2416: [ML] Identify multimodal prior distributions for anomaly explanation
As an enhancement for explaining anomalies, it would be helpful to inform the users whether the prior distribution of the observed time series is multi-modal or has a single mode. For instance, if the distribution has two modes, then the _typical_ value may lie between the modes and hence not constitute the most probable value.

In particular, we can add a method
```c++
COneOfN::isMultimodal(){
  double sumOfWeights = ...
  for (const auto& model : m_Models){
    if(model.second.type() == E_Multimodal && model.first/sumOfWeights > 0.1) {
      return true;
    }
  }
  return false;
}
```

And then in the corresponding code in  `CtimSeriesModel`:
```c++
if (m_ResidualModel.type() == E_OneOfN) {
  return m_ResidualModel.isMultimodal();
}
return false;
```

valeriy42:
As an enhancement for explaining anomalies, it would be helpful to inform the users whether the prior distribution of the observed time series is multi-modal or has a single mode. For instance, if the distribution has two modes, the typical value may lie between the modes and hence not constitute the most probable value.

Closes #2416.

### DIFF
--- a/lib/maths/common/COneOfNPrior.cc
+++ b/lib/maths/common/COneOfNPrior.cc
@@ -769,13 +769,10 @@ void COneOfNPrior::sampleMarginalLikelihood(std::size_t numberSamples,
     LOG_TRACE(<< "samples = " << samples);
 }
 
-bool COneOfNPrior::isMultimodal() const {
-    double sumOfWeights{0.0};
-    for (const auto& weight : this->weights()) {
-        sumOfWeights += weight;
-    }
-    for (const auto& model : m_Models) {
-        if (model.second->type() == EPrior::E_Multimodal && model.first / sumOfWeights > 0.1) {
+bool COneOfNPrior::isSelectedModelMultimodal() const {
+    auto weights = this->weights();
+    for (std::size_t i = 0; i < weights.size(); ++i) {
+        if (weights[i] > 0.1 && m_Models[i].second->type() == EPrior::E_Multimodal) {
             return true;
         }
     }
--- a/lib/maths/common/COneOfNPrior.cc
+++ b/lib/maths/common/COneOfNPrior.cc
@@ -769,6 +769,19 @@ void COneOfNPrior::sampleMarginalLikelihood(std::size_t numberSamples,
     LOG_TRACE(<< "samples = " << samples);
 }
 
+bool COneOfNPrior::isMultimodal() const {
+    double sumOfWeights{0.0};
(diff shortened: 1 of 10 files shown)


lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -1075,13 +1076,25 @@ bool CUnivariateTimeSeriesModel::uncorrelatedProbability(
...
     result.s_Probability = pOverall;
     result.s_FeatureProbabilities = std::move(featureProbabilities);
     result.s_Tail = {tail};
 
     return true;
 }
 
+bool CUnivariateTimeSeriesModel::multimodalPrior() const {
+    LOG_DEBUG(<< "Residual model type " << m_ResidualModel->type());
+    if (m_ResidualModel->type() == common::CPrior::E_OneOfN) {
```
tveasey, line 1092: So I can see why you did it like this. My inclination would be we have an `bool isSelectedModelMultimodal()` defined on `CPrior` which returns false and override it in `COneOfOneN` and ditch this function altogether.

tveasey (approved):
Just one real observation which relates to the implementation of the check to see if the selected residual model is multimodal. I have a couple of other very minor things I'll leave up to you. Happy for you to commit without another review.

lib/maths/common/COneOfNPrior.cc
```diff
@@ -769,6 +769,19 @@ void COneOfNPrior::sampleMarginalLikelihood(std::size_t numberSamples,
...
+    double sumOfWeights{0.0};
+    for (const auto& weight : this->weights()) {
+        sumOfWeights += weight;
+    }
+    for (const auto& model : m_Models) {
+        if (model.second->type() == EPrior::E_Multimodal && model.first / sumOfWeights > 0.1) {
+            return true;
+        }
+    }
+    return false;
```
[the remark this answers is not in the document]
valeriy42: Thank you for this suggestion. 👍

lib/maths/time_series/CTimeSeriesModel.cc
valeriy42: I see how this can simplify the code. I updated the code according to your suggestion.

PR #2441, opened 2023-01-18
[ML] Add Windows step to BuildKite pipeline

edsavage:
Add a step that will be used to build and test on windows x86_64.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,94 +12,94 @@ env:
   VERSION_QUALIFIER: ""
 steps:
 
-  - label: "Validate formatting with clang-format :white_check_mark:"
-    key: "check_style"
-    command: ".buildkite/scripts/steps/check-style.sh --all"
-    agents:
-      image: "docker.elastic.co/ml-dev/ml-check-style:2"
-
-  - label: "Build & test :cpp: for linux-x86_64-RelWithDebInfo :linux:"
-    timeout_in_minutes: 180
-    agents:
-      cpu: "6"
-      ephemeralStorage: "20G"
-      memory: "64G"
-      image: "docker.elastic.co/ml-dev/ml-linux-build:25"
-    commands:
-      - ".buildkite/scripts/steps/build_and_test.sh"
-    depends_on: "check_style"
-    key: "build_test_linux-x86_64-RelWithDebInfo"
-    artifact_paths:
-      - "lib/*/unittest/boost_test_results.junit"
-      - "bin/*/unittest/boost_test_results.junit"
-    env:
-      CPP_CROSS_COMPILE: ""
-      CMAKE_FLAGS: "-DCMAKE_TOOLCHAIN_FILE=cmake/linux-x86_64.cmake"
-      RUN_TESTS: "true"
-      BOOST_TEST_OUTPUT_FORMAT_FLAGS: "--logger=JUNIT,error,boost_test_results.junit"
-    plugins:
-      - test-collector#v1.2.0:
-          files: "lib/*/unittest/boost_test_results.junit"
-          format: "junit"
-
-  - label: "Build & test :cpp: for linux-aarch64-RelWithDebInfo :linux: :aws:"
-    timeout_in_minutes: 180
-    agents:
-      provider: aws
-      instanceType: m6g.2xlarge
-      imagePrefix: ci-amazonlinux-2-aarch64
-      diskSizeGb: 100
-      diskName: '/dev/xvda'
-    commands:
-      - ".buildkite/scripts/steps/build_and_test.sh"
-    depends_on: "check_style"
-    key: "build_test_linux-aarch64-RelWithDebInfo"
-    artifact_paths:
-      - "lib/*/unittest/boost_test_results.junit"
-      - "bin/*/unittest/boost_test_results.junit"
-    env:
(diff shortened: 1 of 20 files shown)


.buildkite/pipeline.yml
```diff
@@ -76,6 +76,24 @@ steps:
...
+  - label: "Build :cpp: for Windows_x86_64-RelWithDebInfo"
+    timeout_in_minutes: 180
+    agents:
+      provider: gcp
+      image: family/ci-windows-2016
+    commands:
+      - 'echo "checkout succeeded"'
+      - Get-ChildItem
+      - "dev-tools\\download_windows_deps.ps1"
+      - 'Install-Binary -Url "https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&channel=Release&version=VS2022&source=VSLandingPage&cid=2030&passive=false" -Name "VisualStudioSetup.exe" -ArgumentList ("/quiet")'
```
droberts195, line 89: We shouldn't be installing our dependencies on every build. Doesn't `family/ci-windows-2016` already have this installed? If not, the solution is to use the right base image. If we could use the base image we use in Jenkins that would be perfect.

droberts195: Try `family/elastic-elasticsearch-ci-windows-2016-ml-pinned`

edsavage: Sadly trying `family/elastic-elasticsearch-ci-windows-2016-ml-pinned` results in:

```
googleapi: Error 404: The resource 'projects/elastic-images-qa/global/images/family/elastic-elasticsearch-ci-windows-2016-ml-pinned' was not found, notFound
```

The VM images must now be defined here in the new `ci-agent-images` repo by way of extending one of the many existing base images there.

.buildkite/pipeline.yml
```diff
@@ -11,125 +11,149 @@ env:
...
-      diskName: '/dev/xvda'
+      provider: gcp
+      image: family/ml-cpp-1-windows-2016
     commands:
-      - ".buildkite/scripts/steps/build_and_test.sh"
-    depends_on: "check_style"
-    key: "build_test_linux-aarch64-RelWithDebInfo"
-    artifact_paths:
-      - "lib/*/unittest/boost_test_results.junit"
-      - "bin/*/unittest/boost_test_results.junit"
```
droberts195, line 59: On Linux we're preserving the JUnit output as artifacts as well as passing to the test collector. I think we should do this on Windows too. Anything we can keep to make understanding test failures easier is good. If it's not possible to have the same files be artifacts and also be passed to the test collector then maybe we could make the Boost XML output format the artifacts and pass the JUnit format to the test collector.

Also, it seems like it should be relatively simple to make all the unit test log files artifacts. Currently we `cat` the log for any test suite that fails, but it's not possible to view it without the risk of output from parallel processes being mixed in, and it's not possible to view the output of other test suites that ran simultaneously to look for possible interactions. So it would be nice if all the unit test logs got preserved (for all platforms, not just Windows).

lib/core/unittest/CDetachedProcessSpawnerTest.cc
```diff
@@ -64,7 +64,7 @@ BOOST_AUTO_TEST_CASE(testSpawn) {
 
     BOOST_TEST_REQUIRE(spawner.spawn(PROCESS_PATH1, args));
 
-    // Expect the copy to complete in less than 1 second
+    // Expect the copy to complete in less than 1 seconds
```
droberts195, line 67: It's probably best to delete the final "s" so that this PR doesn't pointlessly appear in the change history of this file.

.buildkite/pipeline.yml
```diff
@@ -99,7 +99,33 @@ steps:
...
+  - label: "Build & test :cpp: for Windows_x86_64-RelWithDebInfo :windows:"
+    timeout_in_minutes: 180
+    agents:
+      provider: gcp
+      machineType: "c2-standard-16"
+      minCpuPlatform: "Intel Cascade Lake"
+      image: family/ml-cpp-1-windows-2016
+    commands:
+      - 'echo "checkout succeeded"'
+      - "& Get-ComputerInfo"
```
droberts195, line 111: If `Get-ComputerInfo` is a built-in PowerShell command then should it be called without the leading `&`? I don't know if that's needlessly running it in a different invocation of PowerShell for example.

.buildkite/pipeline.yml
```diff
@@ -99,7 +99,33 @@ steps:
...
+      - "& .buildkite\\scripts\\steps\\build_and_test.ps1"
+    depends_on: "check_style"
+    key: "build_test_Windows-x86_64-RelWithDebInfo"
+    artifact_paths:
+      - "lib/*/unittest/boost_test_results.junit"
+      - "bin/*/unittest/boost_test_results.junit"
```
droberts195, line 117: Is there a good reason not to use `"*/*/unittest/boost_test_results.junit"` here? Then it would be totally clear that the idea is to preserve all the same files that are being interpreted as unit test results.

Same for the other platforms (unless it's impossible to use a double wildcard for some reason).

droberts195 (approved):
LGTM apart from a few nits

.buildkite/pipeline.yml
edsavage: Removed it as it was transient debug anyway

.buildkite/pipeline.yml
edsavage: Worth a try..

edsavage:
buildkite retest this

PR #2444, opened 2023-01-27
[ML] Experiment with dynamic pipelines

edsavage:
Investigate the potential of dynamic pipelines to:
 * Maximize code re-use
 * Configure builds at run-time, with github comments and/or labels


edsavage:
buildkite debug this on linux

PR #2445, opened 2023-01-30
Buildkite dynamic pipelines

edsavage:
Use dynamically generated `BuildKite` pipelines to:

- Promote code re-use between pipelines
- Customize build based on PR labels and/or comments


edsavage:
buildkite debug on linux

droberts195 (commented):
One thing I am not completely clear on is when we do native builds with testing on `linux-aarch64`, and when we do cross compiles (which can't run unit tests). Which variations are we running in the different types of builds?

edsavage:
> One thing I am not completely clear on is when we do native builds with testing on `linux-aarch64`, and when we do cross compiles (which can't run unit tests). Which variations are we running in the different types of builds?

Currently, we do native builds with testing for both PR builds and branch builds. Cross compilation of `linux-aarch64` is also triggered for both PR and branch builds, but only for non-debug builds. 

As discussed, I'll change the logic so that we always perform a cross compilation with debug enabled on every PR build - even if the rest of the build is optimised. Cross compilation of `linux-aarch64` will not be performed under any other circumstances.

.buildkite/ml_pipeline/config.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env Python
```
droberts195, line 1: Should it be lower case?

```suggestion
#!/usr/bin/env python
```

droberts195 (approved):
LGTM

The logic is so complex that I am sure this will need another iteration to get the branch builds working perfectly rather than just the PR builds. But we have to merge it to do that, so I'm happy to merge and see what happens.

PR #2449, opened 2023-02-07
[ML] Anomaly score explanation for rare detector

valeriy42:
This PR provides additional fields to the `anomaly_score_explanation` of the anomaly records:
- `by_field_first_occurrence` is `true` if the rare category is seen for the first time
- `by_field_relative_rarity`  indicates how much less frequently the rare category occurs compared to the median frequent category in the bucket.


include/model/CAnnotatedProbability.h
```diff
@@ -75,9 +75,6 @@ struct MODEL_EXPORT SAttributeProbability {
     //! Restore the probability reading state from \p traverser.
     bool acceptRestoreTraverser(core::CStateRestoreTraverser& traverser);
 
-    //! Add the descriptive data \p value for \p key.
-    void addDescriptiveData(annotated_probability::EDescriptiveData key, double value);
-
```
tveasey, line 80: Can't we now remove the enum  `EDescriptiveData` at the start of this header?

include/model/CAnnotatedProbability.h
```diff
@@ -124,9 +119,6 @@ struct MODEL_EXPORT SAnnotatedProbability {
     SAnnotatedProbability();
     SAnnotatedProbability(double p);
 
-    //! Add the descriptive data \p value for \p key.
-    void addDescriptiveData(annotated_probability::EDescriptiveData key, double value);
```
tveasey, line 128: There are some typedefs which should be removed as well. I guess these should surface when you remove `EDescriptiveData`.

include/model/CAnnotatedProbabilityBuilder.h
```diff
@@ -47,46 +47,34 @@ class MODEL_EXPORT CAnnotatedProbabilityBuilder : private core::CNonCopyable {
...
-                                 function_t::EFunction function,
-                                 std::size_t numberOfPeople);
+                                 function_t::EFunction function);
 
     void attributeProbabilityPrior(const maths::common::CMultinomialConjugate* prior);
     void personAttributeProbabilityPrior(const maths::common::CMultinomialConjugate* prior);
-    void personFrequency(double frequency, bool everSeenBefore);
     void probability(double p);
     void anomalyScoreExplanation(SAnnotatedProbability::TAnomalyScoreExplanation explanations);
+    SAnnotatedProbability::TAnomalyScoreExplanation& anomalyScoreExplanation();
```
tveasey, line 56: I think now you have a return by writeable reference I would remove `void anomalyScoreExplanation(SAnnotatedProbability::TAnomalyScoreExplanation explanations);`: you should be able to replace it with `builder.anomalyScoreExplanation() = ...`

lib/maths/time_series/CTimeSeriesModel.cc
```diff
@@ -1077,7 +1077,7 @@ bool CUnivariateTimeSeriesModel::uncorrelatedProbability(
 
     result.s_AnomalyScoreExplanation.s_MultimodalDistribution =
         m_ResidualModel->isSelectedModelMultimodal();
-    LOG_DEBUG(<< "Multimodel distribution: "
```
tveasey, line 1080: Good catch! Wonder when this sneaked in.

lib/model/CModelTools.cc
```diff
@@ -301,6 +301,29 @@ bool CModelTools::CCategoryProbabilityCache::lookup(std::size_t attribute, doubl
...
+    auto concentrations = m_Prior->concentrations();
+    // compute element concentrations[size/2]
+    auto middle_element1 = concentrations.begin() + concentrations.size() / 2;
+    std::nth_element(concentrations.begin(), middle_element1, concentrations.end());
+    m_MedianConcentration = *middle_element1;
+    if (concentrations.size() % 2 == 0) {
+        auto middle_element2 = concentrations.begin() + concentrations.size() / 2 - 1;
+        std::nth_element(concentrations.begin(), middle_element2, concentrations.end());
+        m_MedianConcentration = (m_MedianConcentration.value() + *middle_element2) / 2;
+    }
```
tveasey, line 322: See `maths::CBasicStatistics::median`.

lib/model/unittest/CAnnotatedProbabilityBuilderTest.cc
```diff
@@ -157,64 +150,43 @@ BOOST_AUTO_TEST_CASE(testAddAttributeProbabilityGivenPopulationRare) {
...
-        6.0, result.s_AttributeProbabilities[0].s_DescriptiveData[1].second);
+    // BOOST_REQUIRE_EQUAL(2, result.s_AttributeProbabilities[0].s_DescriptiveData.size());
+    // BOOST_REQUIRE_EQUAL(annotated_probability::E_ATTRIBUTE_CONCENTRATION,
+    //                     result.s_AttributeProbabilities[0].s_DescriptiveData[0].first);
+    // BOOST_REQUIRE_EQUAL(
+    //     3.0, result.s_AttributeProbabilities[0].s_DescriptiveData[0].second);
+    // BOOST_REQUIRE_EQUAL(annotated_probability::E_ACTIVITY_CONCENTRATION,
+    //                     result.s_AttributeProbabilities[0].s_DescriptiveData[1].first);
+    // BOOST_REQUIRE_EQUAL(
+    //     6.0, result.s_AttributeProbabilities[0].s_DescriptiveData[1].second);
```
tveasey, line 184: Let's just delete.

lib/model/CEventRateModel.cc
```diff
@@ -477,15 +476,23 @@ bool CEventRateModel::computeProbability(std::size_t pid,
...
         resultBuilder.multiBucketImpact(multiBucketImpact);
     }
 
+    resultBuilder.anomalyScoreExplanation(pJoint.anomalyScoreExplanation());
     bool everSeenBefore = this->firstBucketTimes()[pid] != startTime;
-    resultBuilder.personFrequency(this->personFrequency(pid), everSeenBefore);
+    auto typicalConcentration = m_Probabilities.medianConcentration();
+    double actualConcentration;
+    if (m_ProbabilityPrior.concentration(pid, actualConcentration) &&
+        typicalConcentration.has_value()) {
```
tveasey, line 490: It seems more principled to test the simple condition first
```suggestion
    if (typicalConcentration.has_value() &&
        m_ProbabilityPrior.concentration(pid, actualConcentration)) {
```

tveasey: Also should we just grab `auto& explanation = resultBuilder.anomalyScoreExplanation();` upfront. Might make the code a little cleaner.

lib/model/CEventRatePopulationModel.cc
```diff
@@ -634,7 +634,7 @@ bool CEventRatePopulationModel::computeProbability(std::size_t pid,
 
     CAnnotatedProbabilityBuilder resultBuilder(
         result, std::max(numberAttributeProbabilities, std::size_t(1)),
-        function_t::function(gatherer.features()), gatherer.numberActivePeople());
```
tveasey, line 637: I would expect us to at least set if this is the first time we've seen a category. But if this isn't trivially accessible at the moment this could be a follow on PR.

tveasey (commented):
Overall looks good. I think you can maybe do a little more tidying up. Also, I wonder if we should at least be setting ever seen for population model categories. Let's discuss offline what we can use to get this info.

lib/model/unittest/CAnnotatedProbabilityBuilderTest.cc
valeriy42: Indeed, missed this one.

lib/model/CEventRateModel.cc
valeriy42: changed

include/model/CAnnotatedProbabilityBuilder.h
valeriy42: changed

tveasey (approved):
LGTM

PR #2452, opened 2023-02-15
Add separate BuildKite pipeline for branch builds

edsavage:
Separate the concerns of branch builds away from those of PR builds. This simplifies and fixes a problem of the latter whereby builds were being triggered twice, once by BuildKite directly and again from an API call from within a PR webhook.

### DIFF
--- a/.buildkite/branch.json.py
+++ b/.buildkite/branch.json.py
@@ -38,8 +38,6 @@ def main():
     if config.build_linux:
         build_linux = pipeline_steps.generate_step_template("Linux", "build", None, None)
         pipeline_steps.append(build_linux)
-    pipeline_steps.append(pipeline_steps.generate_step("Upload ES tests runner pipeline",
-                                                       ".buildkite/pipelines/run_es_tests.yml.sh"))
     pipeline_steps.append({"wait": None})
     pipeline_steps.append(pipeline_steps.generate_step("Upload artifact uploader pipeline",
                                                        ".buildkite/pipelines/upload_to_s3.yml.sh"))
--- a/.buildkite/branch.json.py
+++ b/.buildkite/branch.json.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0 and the following additional limitation. Functionality enabled by the
+# files subject to the Elastic License 2.0 may only be used in production when
+# invoked by an Elasticsearch process with a license key installed that permits
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+#
+# This script generates a pipeline JSON for the ml-cpp-branch pipeline.
+# Builds for this pipeline may be triggered by commit or comment.
+#
+import json
+
+from ml_pipeline import (
+    step,
+    config as buildConfig,
+)
+
+def main():
+    pipeline = {}
+    pipeline_steps = step.PipelineStep([])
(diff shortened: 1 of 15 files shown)


.buildkite/pipeline.json.py
```diff
@@ -48,8 +47,10 @@ def main():
     if config.build_linux:
         build_linux = pipeline_steps.generate_step_template("Linux", config.action, config.snapshot, config.version_qualifier)
         pipeline_steps.append(build_linux)
-    pipeline_steps.append(pipeline_steps.generate_step("Upload ES tests runner pipeline",
-                                                       ".buildkite/pipelines/run_es_tests.yml.sh"))
+    # Never run the ES integration tests if VERSION_QUALIFIER has been set.
```
droberts195, line 50: I can see why this won't work (it would actually if we passed the same qualifier through to the ES build) but it was never a problem before because:

1. Nobody ever used version qualifiers on PR builds
2. Release branch builds never ran the ES integration tests

I think it's best to preserve the old behaviour of release branch builds not running the ES tests. In PR builds we want to catch incompatibilities between the C++ and Java that have been accidentally introduced. In the release branch builds we're often under pressure to produce artifacts for a new version. We don't want these branch builds to fail because Elasticsearch is at a different stage of its version bumping process to ML.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -11,7 +11,7 @@
 set -euo pipefail
 
 # Default to a snapshot build
-if [ -z "$BUILD_SNAPSHOT" ] ; then
+if [ -z "${BUILD_SNAPSHOT:=true}" ] ; then
```
droberts195, line 14:
```suggestion
if [ "${BUILD_SNAPSHOT:=true}" != false ] ; then
```

edsavage:
jenkins retest this please

.buildkite/branch.json.py
```diff
@@ -0,0 +1,61 @@
...
+    if config.build_macos:
+        build_macos = pipeline_steps.generate_step_template("MacOS", "build")
+        pipeline_steps.append(build_macos)
+    if config.build_linux:
+        build_linux = pipeline_steps.generate_step_template("Linux", "build")
+        pipeline_steps.append(build_linux)
+
+    pipeline_steps.append({"wait": None})
+
+	# Trigger the DRA pipeline to create and upload artifacts to S3 and GCS
```
droberts195, line 44: nit: this is indented with a tab

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -29,7 +29,7 @@ else
 fi
 
 # Version qualifier shouldn't be used in PR builds
-if [[ x"$BUILDKITE_PULL_REQUEST" != xfalse && -n "$VERSION_QUALIFIER" ]] ; then
+if [[ x"$BUILDKITE_PULL_REQUEST" != xfalse && -n "${VERSION_QUALIFIER:=""}" ]] ; then
```
droberts195, line 32: The `:=""` in `${VERSION_QUALIFIER:=""}` seems redundant, as `bash` will expand `${VERSION_QUALIFIER}` as an empty string if the environment variable is not set.

droberts195 (approved):
LGTM apart from a couple of minor nits.

But then please watch the branch build that is triggered after merging this closely to make sure it works as expected, as the PR build is only testing the PR pipeline.

.buildkite/scripts/steps/build_and_test.sh
edsavage: The script calls `set -euo pipefail` on line 11. The `-u` option treats unset variables as errors, so in this case explicitly setting `VERSION_QUALIFIER` to the empty string if it's not set is necessary.

PR #2455, opened 2023-02-24
7.17 buildkite backport

edsavage:
Backporting `BuildKite` related changes to 7.17.

Looking for a review of this backport as it was ever so slightly non-trivial (mainly due to the `CMake` build system not being present in `7.17`)


.buildkite/pipelines/build_linux.json.py
```diff
@@ -0,0 +1,159 @@
...
+              'tar xvzf eigen.tgz',
+              ".buildkite/scripts/steps/build_and_test.sh"
+            ],
+            "depends_on": [
+              "check_style",
+              "clone_eigen"
+              ],
+            "key": f"build_test_linux-{arch}-{build_type}",
+            "env": {
+              "CMAKE_FLAGS": f"-DCMAKE_TOOLCHAIN_FILE=cmake/linux-{arch}.cmake",
```
droberts195, line 82: 7.17 doesn't use CMake - but is something else needed instead?

.buildkite/pipelines/build_linux.json.py
```diff
@@ -0,0 +1,159 @@
...
+              "image": "docker.elastic.co/ml-dev/ml-linux-aarch64-cross-build:1"
+            },
+            "commands": [
+              ".buildkite/scripts/steps/build_and_test.sh"
+            ],
+            "depends_on": "check_style",
+            "key": "build_linux_aarch64_cross-RelWithDebInfo",
+            "env": {
+              "CPP_CROSS_COMPILE": "aarch64",
+              "CMAKE_FLAGS": "-DCMAKE_TOOLCHAIN_FILE=cmake/linux-aarch64.cmake",
```
droberts195, line 125: Another CMake reference

droberts195 (approved):
LGTM

I saw a couple of CMake environment variables left over. They'll be harmless but just made me wonder if 7.17 needed something else instead in those two places.

.buildkite/pipelines/build_linux.json.py
edsavage: Thanks Dave, I think these `CMake` references can simply be removed as there is no equivalent configuration in the GNU make system.

PR #2456, opened 2023-02-28
Parameterise CConcurrentQueue work queue size 

davidkyle:
`CstaticThreadPool` uses a `CConcurrentQueue` to buffer work. The size of the work queue is a template parameter which makes it difficult to change in `CstaticThreadPool`. This change makes the queue size an ctor argument which is passed down to `CConcurrentQueue`

The `pytorch_inference` app is changed to use a queue of 1.

The motivation for this change is to avoid queuing work in the `pytorch_inference` process as requests are already queued on the Java side and certain features require the ability to jump to the front of the queue. Also the number of items in the Java queue is reported via the stats API but that number is not accurate if work is also queued in `pytorch_inference`.

### DIFF
--- a/include/core/CConcurrentQueue.h
+++ b/include/core/CConcurrentQueue.h
@@ -35,19 +35,18 @@ namespace core {
 //! for the price of blocking.
 //!
 //! @tparam T the objects of the queue
-//! @tparam QUEUE_CAPACITY fixed queue capacity
-//! @tparam NOTIFY_CAPACITY special parameter, for signaling the producer in blocking case
-template<typename T, size_t QUEUE_CAPACITY, size_t NOTIFY_CAPACITY = QUEUE_CAPACITY>
+template<typename T>
 class CConcurrentQueue final : private CNonCopyable {
 public:
     using TOptional = std::optional<T>;
 
 public:
-    CConcurrentQueue() : m_Queue(QUEUE_CAPACITY) {
-        static_assert(NOTIFY_CAPACITY > 0, "NOTIFY_CAPACITY must be positive");
-        static_assert(QUEUE_CAPACITY >= NOTIFY_CAPACITY,
-                      "QUEUE_CAPACITY cannot be less than NOTIFY_CAPACITY");
-    }
+    explicit CConcurrentQueue(std::size_t queueCapacity, std::size_t notifyCapacity)
+        : m_QueueCapacity(queueCapacity), m_NotifyCapacity(notifyCapacity),
+          m_Queue(queueCapacity) {}
+
+    explicit CConcurrentQueue(std::size_t queueCapacity)
+        : CConcurrentQueue(queueCapacity, queueCapacity) {}
 
     //! Pop an item out of the queue, this blocks until an item is available
     T pop() {
@@ -109,7 +108,7 @@ public:
     //! Forward \p item to the queue, if the queue is full this fails and returns false
     bool tryPush(T&& item) {
         std::unique_lock<std::mutex> lock(m_Mutex);
-        if (m_Queue.size() >= QUEUE_CAPACITY) {
+        if (m_Queue.size() >= m_QueueCapacity) {
             return false;
         }
 
@@ -141,12 +140,12 @@ private:
     }
     void waitWhileFull(std::unique_lock<std::mutex>& lock) {
         m_ProducerCondition.wait(
-            lock, [this] { return m_Queue.size() < QUEUE_CAPACITY; });
+            lock, [this] { return m_Queue.size() < m_QueueCapacity; });
     }
 
     void notifyIfNoLongerFull(std::unique_lock<std::mutex>& lock, std::size_t oldSize) {
         lock.unlock();
-        if (oldSize >= NOTIFY_CAPACITY) {
+        if (oldSize >= m_NotifyCapacity) {
             m_ProducerCondition.notify_all();
         }
     }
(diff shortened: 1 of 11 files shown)


davidkyle:
I added debug logging to Elasticsearch and could see high priority search requests go to the front of the queue Java side but there were still 5/6 requests in front of the search request before it was processed. This is most likely due to the buffer in the named pipe holding the requests before they are read. A follow up change will be to reduce the size of the buffer used in the named pipe.

lib/core/CStaticThreadPool.cc
```diff
@@ -21,11 +22,18 @@ std::size_t computeSize(std::size_t hint) {
...
 
-CStaticThreadPool::CStaticThreadPool(std::size_t size)
-    : m_Busy{false}, m_Cursor{0}, m_TaskQueues{computeSize(size)} {
-    m_NumberThreadsInUse.store(m_TaskQueues.size());
-    m_Pool.reserve(m_TaskQueues.size());
-    for (std::size_t id = 0; id < m_TaskQueues.size(); ++id) {
+CStaticThreadPool::CStaticThreadPool(std::size_t size, std::size_t queueCapacity)
+    : m_Busy{false}, m_Cursor{0} {
+
+    std::size_t poolSize = computeSize(size);
```
tveasey, line 28: Nit coding conventions:
```suggestion
    std::size_t poolSize{computeSize(size)};
```

include/core/CConcurrentWrapper.h
```diff
@@ -30,24 +30,25 @@ namespace core {
...
+template<typename T>
 class CConcurrentWrapper final : private CNonCopyable {
 public:
     //! Wrap and return the wrapped object
     //!
     //! The object has to wrapped once and only once, pass the reference around in your code.
     //! This starts a background thread.
-    explicit CConcurrentWrapper(T& resource)
-        : m_Resource(resource), m_Done(false) {
+    explicit CConcurrentWrapper(T& resource, std::size_t queueCapacity, std::size_t notifyCapacity)
```
tveasey, line 40: How about just defaulting parameters here and ditching extra constructor:
```suggestion
    explicit CConcurrentWrapper(T& resource, std::size_t queueCapacity = 100, std::size_t notifyCapacity = 50)
```

include/core/CConcurrentQueue.h
```diff
@@ -160,6 +159,9 @@ class CConcurrentQueue final : private CNonCopyable {
     static bool always(const T&) { return true; }
 
 private:
+    std::size_t m_QueueCapacity;
```
tveasey, line 162: How about moving documentation from @tparam here, e.g. `fixed queue capacity` and same for `m_NotifyCapacity`.

include/core/CStaticThreadPool.h
```diff
@@ -82,8 +83,9 @@ class CORE_EXPORT CStaticThreadPool {
         TOptionalSize m_ThreadId;
     };
     using TOptionalTask = std::optional<CWrappedTask>;
-    using TWrappedTaskQueue = CConcurrentQueue<CWrappedTask, 50>;
-    using TWrappedTaskQueueVec = std::vector<TWrappedTaskQueue>;
+    using TWrappedTaskQueue = CConcurrentQueue<CWrappedTask>;
+    using TWrappedTaskQueueUPtr = std::unique_ptr<TWrappedTaskQueue>;
+    using TWrappedTaskQueueUPtrVec = std::vector<TWrappedTaskQueueUPtr>;
```
tveasey, line 88: As per comment above I think this change is unneeded.

lib/core/Concurrency.cc
```diff
@@ -55,16 +56,16 @@ class CThreadPoolExecutor final : public CExecutor {
 class CExecutorHolder {
 public:
     CExecutorHolder()
-        : m_ThreadPoolSize{0}, m_Executor(std::make_unique<CImmediateExecutor>()) {}
+        : m_ThreadPoolSize(0), m_Executor(std::make_unique<CImmediateExecutor>()) {}
```
tveasey, line 59: Let's keep the type checking in place here:
```suggestion
        : m_ThreadPoolSize{0}, m_Executor(std::make_unique<CImmediateExecutor>()) {}
```

tveasey (commented):
Overall changes all look correct. I have a few minor C++ style comments. Seems like we can and should avoid wrapping the queues in a pointer in the thread pool as well. Other than that looks good to go.

lib/core/CStaticThreadPool.cc
```diff
@@ -21,11 +22,18 @@ std::size_t computeSize(std::size_t hint) {
...
+    m_Pool.reserve(poolSize);
+
+    for (std::size_t id = 0; id < poolSize; ++id) {
+        m_TaskQueues.push_back(std::make_unique<TWrappedTaskQueue>(queueCapacity));
+    }
```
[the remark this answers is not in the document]
davidkyle: I should have commented on this in the description sorry.

`CConcurrentQueue` not move insertable and it doesn't have a copy ctor either. The unique pointer was the only way I could think of adding the class to the container. 

It worked before because `CConcurrentQueue` had a default no parameter ctor, the queues were initialised by  `m_TaskQueues{computeSize(size)}`. No that no longer works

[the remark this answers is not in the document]
tveasey: Oh it is mutex which stops it being moveable; that's a pain. We should probably make `CConcurrentQueue` explicitly not moveable, by deleting its move operations. At the moment it looks like it is moveable reading the code.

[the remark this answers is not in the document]
davidkyle: I deleted the move and copy options (rather than inheriting CNonCopyable) and added a comment

PR #2461, opened 2023-03-06
[ML] Don't expect commit status in PR config

edsavage:
Since "publish_commit_status" is set to "false" on the ml-cpp pipeline it should also be disabled in the pull request config with "set_commit_status: false"

This is because the PR service only sends a commit status when it triggers the build, so if the pipeline isn't configured to send a commit status, a "complete" status will never get sent

This fixes a problem with a PR check for "ml-cpp-ci" continually stuck in the "pending" state.


droberts195 (approved):
LGTM

edsavage:
Merging without waiting for Jenkins CI to finish as the change has no impact on it

PR #2464, opened 2023-03-07
[ML] Include debug artifacts in DRA uploads

edsavage:
Account for naming convention differences between v8 and v7


droberts195 (approved):
LGTM

PR #2465, opened 2023-03-09
Fix gradle uploadAll task to include debug artifacts

edsavage:
Tweak the gradle `uploadAll` task to pick up "debug" artifacts on the `7.17` branch.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -352,7 +352,7 @@ task uploadAll(type: UploadS3Task) {
   bucket 'prelert-artifacts'
   // Upload ALL artifacts (including the dependency report) in this task
   def fileDir = fileTree("${buildDir}/distributions").matching {
-    include "ml-cpp-*${project.version}*.zip", "dependencies-${version}.csv"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -352,7 +352,7 @@ task uploadAll(type: UploadS3Task) {
   bucket 'prelert-artifacts'
   // Upload ALL artifacts (including the dependency report) in this task
   def fileDir = fileTree("${buildDir}/distributions").matching {
-    include "ml-cpp-${project.version}*.zip", "dependencies-${version}.csv"
+    include "ml-cpp-*${project.version}*.zip", "dependencies-${version}.csv"
```
droberts195, line 355: In #2464 the pattern was `ml-cpp*${project.version}*.zip` instead of `ml-cpp-*${project.version}*.zip`. It would be nice to use exactly the same pattern in both places, to avoid confusion in the future.

droberts195 (approved):
LGTM, but before merging please could you make the patterns consistent across the different files.

PR #2472, opened 2023-04-11
Update Gradle wrapper to 8.0.2

breskeby:
This updates the gradle wrapper to use Gradle 8.0.2 as we do for the elasticsearch build. 

Some updates in the root build script have been required. 

This will unblock users to build ml-cpp users to build along the elasticsearch build again as included build via 

```
./gradlew --write-verification-metadata sha256 localDist --include-build ../ml-cpp
```

which was broken due to Gradle version 7 build scripts in ml-cpp have been incompatible with Gradle 8


droberts195 (approved):
LGTM

Thanks for upgrading this repo @breskeby!

PR #2475, opened 2023-05-03
[ML] Remove unneeded yml files to simplify BuildKite pipelines

edsavage:
Experiment with removing top level yaml file from the BuildKite pipeline as they _should_ be unneeded.


edsavage:
Closing this as I believe the benefit is not worth the effort of pursuing it further at this late stage of development

PR #2478, opened 2023-05-04
[7.17][ML] Add native MacOS aarch64 build support in BuildKite (#2470)

edsavage:
Add a step to build natively on the MacOS aarch64 VM

Backports #2470

### DIFF
--- a/.buildkite/scripts/steps/build_and_test.sh
+++ b/.buildkite/scripts/steps/build_and_test.sh
@@ -90,7 +90,7 @@ else
      # need to hardcode the version that was used to build Boost for that
      # version of Elasticsearch.
      if [ "$HARDWARE_ARCH" = aarch64 ] ; then
-         export BOOSTCLANGVER=13
+         export BOOSTCLANGVER=120
      fi
 
      (cd ${REPO_ROOT} && ./gradlew --info -Dbuild.version_qualifier=${VERSION_QUALIFIER:-} -Dbuild.snapshot=$BUILD_SNAPSHOT -Dbuild.ml_debug=$ML_DEBUG $TASKS)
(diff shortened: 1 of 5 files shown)


edsavage:
buildkite build this

edsavage:
buildkite build this

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -93,7 +93,7 @@ else
          export BOOSTCLANGVER=120
      fi
 
-     (cd ${REPO_ROOT} && ./gradlew --info -Dbuild.version_qualifier=${VERSION_QUALIFIER:-} -Dbuild.snapshot=$BUILD_SNAPSHOT -Dbuild.ml_debug=$ML_DEBUG $TASKS)
+     (cd ${REPO_ROOT} && ./gradlew --info -Dbuild.version_qualifier=${VERSION_QUALIFIER:-} -Dbuild.snapshot=${BUILD_SNAPSHOT:-} -Dbuild.ml_debug=${ML_DEBUG:-} $TASKS)
```
droberts195, line 96: Is this intended to say "default is empty string"? That's what the previous code was saying too, because `bash` expands unset environment variables to empty strings.

If you find it much clearer the way it is now then I don't mind leaving it, but personally I found the old way easier to read.

edsavage: > Is this intended to say "default is empty string"?
Yes, that's correct. The script calls `set -u` near the top, which will exit on undefined variables, rather than expanding to the empty string. I do find it much clearer (and much easier to debug) this way, but it is really a matter of taste. I'd prefer to leave it as it is at the moment, mainly because it's now working!

droberts195: Oh, I forgot about the `set -u`. That explains why the explicit default is needed.

PR #2480, opened 2023-05-05
[ML] Fix Linux PATH in environment file

droberts195:
Our Linux build setup instructions say that cmake should be installed in /usr/local/cmake/bin, but if you do this then you could not build with Gradle because our environment file did not put this directory on the PATH.

This PR changes the PATH to include /usr/local/cmake/bin.


droberts195:
retest

PR #2485, opened 2023-05-16
[7.17][ML] Handle precession and time shifts of seasonality in anomaly detection

tveasey:
Backport #2202.


lib/maths/time_series/CSeasonalComponent.cc
```diff
@@ -107,15 +118,17 @@ bool CSeasonalComponent::acceptRestoreTraverser(double decayRate,
...
-    inserter.insertLevel(BUCKETING_TAG,
-                         std::bind(&CSeasonalComponentAdaptiveBucketing::acceptPersistInserter,
-                                   &m_Bucketing, std::placeholders::_1));
+    inserter.insertLevel(BUCKETING_TAG, [this](core::CStatePersistInserter& inserter_) {
+        m_Bucketing.acceptPersistInserter(inserter_);
+    });
     inserter.insertValue(LAST_INTERPOLATION_TAG, m_LastInterpolationTime);
+    inserter.insertValue(MAX_TIME_SHIFT_PER_PERIOD_TAG, m_MaxTimeShiftPerPeriod);
+    inserter.insertValue(TOTAL_SHIFT_TAG, m_TotalShift);
+    inserter.insertValue(CURRENT_MEAN_SHIFT_TAG, m_CurrentMeanShift.toDelimited());
```
droberts195, line 131: By making these persistence changes in 7.17 it's theoretically possible that somebody could upgrade to an 8.x that _doesn't_ understand the new fields. #2202 went in 8.2, so the chance of this is very slim. But there's no protection against it, so if somebody _did_ upgrade from 7.17.11 to 8.0 or 8.1 would the models still work, just discarding the extra fields? Or would `autodetect` then crash?

docs/CHANGELOG.asciidoc
```diff
@@ -28,7 +28,12 @@
 
 //=== Regressions
 
-== {es} version 7.17.9
+== {es} version 7.17.11
```
droberts195, line 31:
```suggestion
== {es} version 7.17.12
```

docs/CHANGELOG.asciidoc
```diff
@@ -28,7 +28,12 @@
...
+
+=== Enhancements
+
+* Better handle small shifts of the seasonal patterns in time series data.
+  (See {ml-pull}2202[#2202].)
```
droberts195, line 37:
```suggestion

== {es} version 7.17.9

```

tveasey:
retest

docs/CHANGELOG.asciidoc
droberts195: @tveasey please can you include this change too. As it stands you've moved a 7.17.9 bug fix to 7.17.13.

lib/maths/time_series/CSeasonalComponent.cc
tveasey: I checked the code. Because there is no name collision it should be fine. We will just discard the fields, but other than these extra fields the state document is unchanged, so the 8.0 and 8.1 will be able to deserialise. So we will lose the fix which was present in 7.17 (unfortunate), but I think we can live with this behaviour.

droberts195: Agreed. Thanks for checking.

PR #2487, opened 2023-05-16
[ML] retrieve real memory usage of pytorch inference

hendrikmuhs:
add functionality to get the real memory usage and integrate it into pytorch inference


lib/core/CProcessStats_Linux.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+std::size_t CProcessStats::residentSetSize() {
+    std::string statm;
+    if (readFromSystemFile("/proc/self/statm", statm) == true) {
+        std::vector<std::string> tokens;
+        std::string remainder;
+        core::CStringUtils::tokenise(" ", statm, tokens, remainder);
+
+        return static_cast<std::size_t>(std::stol(tokens[1]));
```
droberts195, line 63: `stol` can throw an exception, and our code is written to avoid exceptions as much as possible.

I think it's safer to use `CStringUtils::stringToType` and log an error and return 0 if that fails. Otherwise this code is relying on the format of `/proc/self/statm` never changing and if it does change then we might find our process crashing just because it couldn't get memory info.

Also, please check that at least 2 tokens were found before accessing `tokens[1]`.

docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,12 @@
 
 //=== Regressions
 
+== {es} version 8.9.0
+
+=== Enhancements
+
+* Retrieve real memory usage of PyTorch inference (See {ml-pull}2487[#2487].)
```
droberts195, line 35: On it's own this isn't a user-facing change. There will need to be a Java PR to make this do anything as far as end users are concerned. So I would add the release note for the overall functionality in the Java PR. The C++ release notes have to be manually copied into the overall release notes, whereas the Java release notes are auto-generated, so less work.

hendrikmuhs: I made this entry, because of the github workflow check. I happily revert and ignore the gh check. 

Maybe it can be documented in the process that this check isn't mandatory to pass.

droberts195: If you change `>enhancement` to `>non-issue` then a release note won't be expected. Then for the Java PR set the labels the other way around.

lib/core/CProcessStats_Linux.cc
```diff
@@ -0,0 +1,92 @@
...
+
+        if (tokens.size() < 2) {
+            LOG_DEBUG(<< "unexpected output from /proc/self/statm, missing rss: " << statm);
+            return 0;
+        }
+
+        if (core::CStringUtils::stringToTypeSilent(tokens[1], rss) == false) {
+            LOG_DEBUG(<< "failed to convert rss value: " << tokens[1]
+                      << " from /proc/self/statm: " << statm);
+        }
```
droberts195, line 74:
```suggestion
            return 0;
        }
```

(Or alternatively intialize `rss` to 0 on line 59.)

lib/core/CProcessStats_Linux.cc
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    return rss;
+}
+
+std::size_t CProcessStats::maxResidentSetSize() {
+    struct rusage rusage;
+
+    if (getrusage(RUSAGE_SELF, &rusage) != 0) {
+        LOG_DEBUG(<< "faild to get resource usage(getrusage): " << ::strerror(errno));
```
droberts195, line 84:
```suggestion
        LOG_DEBUG(<< "failed to get resource usage(getrusage): " << ::strerror(errno));
```

lib/core/CProcessStats_MacOSX.cc
```diff
@@ -0,0 +1,39 @@
...
+std::size_t CProcessStats::residentSetSize() {
+    // not supported on osx
+    return 0;
+}
+
+std::size_t CProcessStats::maxResidentSetSize() {
+    struct rusage rusage;
+
+    if (getrusage(RUSAGE_SELF, &rusage) != 0) {
+        LOG_DEBUG(<< "faild to get resource usage(getrusage): " << ::strerror(errno));
```
droberts195, line 31:
```suggestion
        LOG_DEBUG(<< "failed to get resource usage(getrusage): " << ::strerror(errno));
```

droberts195 (approved):
LGTM if you could just fix 3 minor nits before merging

PR #2489, opened 2023-05-18
[ML] Upgrade to latest AWS SDK version

droberts195:
Our Gradle build uses the AWS SDK. We should upgrade to the latest version.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,8 +10,7 @@ repositories {
 dependencies {
   compileOnly gradleApi()
   compileOnly localGroovy()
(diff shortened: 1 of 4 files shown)


edsavage (approved):
LGTM

droberts195:
The `sleep` commands added to the Docker build scripts are coincidental to the AWS SDK upgrade. The CI team has recently upgraded from Docker version 23 to 24, and it looks like the problem of https://github.com/moby/moby/issues/3631 might have been reintroduced. The `sleep` commands prevent the Docker container terminating its `stdout` before the `stdin` of the next command has had a chance to consume it. It's not an ideal fix so we should revisit how we extract files from our build container when we have a bit more time.

PR #2493, opened 2023-05-22
[ML] Improve calendar cyclic component modelling for long bucket lengths

tveasey:
Our test to find important calendar components to model was not reliable for long bucket lengths. This change addresses some of the short comings:
1. We take too long to warm up the error quantiles,
2. We can't calculate a p-value per repeat because we don't get to observe enough errors per repeat to estimate this properly,
3. We can't estimate high error percentiles accurately with long buckets. (Instead we switch to counting events with fixed expected count as a function of long bucket length rather than fixed probability.)

We also now estimate a lower bound on the right-tail probability for the feature directly from its sum error. This can fall between the error distribution points we count and improves the test sensitivity. Finally, it is useful to be able to detect multiple components per test, otherwise it can take a long time to initialise every component which is needed. This adds the ability to initialise multiple statistically significant components per test run.

### DIFF
--- a/lib/maths/time_series/CCalendarCyclicTest.cc
+++ b/lib/maths/time_series/CCalendarCyclicTest.cc
@@ -408,7 +408,7 @@ double CCalendarCyclicTest::errorsPValue(double n, double nl, double nv) const {
 
 double CCalendarCyclicTest::sufficientCountToMeasureLargeErrors() const {
     // Cap the how long we'll wait identify large errors.
-    return std::min(static_cast<double>(10 * core::constants::DAY) / m_BucketLength, 100.0);
+    return std::min(static_cast<double>(20 * core::constants::DAY) / m_BucketLength, 100.0);
 }
 
 double CCalendarCyclicTest::largeErrorPercentile() const {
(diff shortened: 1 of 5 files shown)


lib/maths/time_series/CCalendarCyclicTest.cc
```diff
@@ -313,33 +400,49 @@ double CCalendarCyclicTest::survivalFunction(double error) const {
...
+
+double CCalendarCyclicTest::largeErrorPercentile() const {
+    // To test for calendar components we check if they align with excess
+    // counts in the right tail of the error distribution. For long bucket
+    // lengths we expect to see few high percentile errors in the window
+    // as a whole and the expected variance of high percentile counts is
+    // too large to be useful. Instead we adjust the threshold above which
+    // we count errors so the expected counts are fixed for long bucket
+    // lengths.
+    return 100.0 - (100.0 - LARGE_ERROR_PERCENTILE) *
```
valeriy42, line 433: Does it make sense to factor out the code that repeats in `largeErrorPercentile` and `veryLargeErrorPercentile`?

valeriy42 (approved):
LGTM. Just a minor comment.

PR #2496, opened 2023-05-26
Better handle test suite failures

edsavage:
Simulate a test suite failure in order to test better failure handling

### DIFF
--- a/.buildkite/scripts/steps/build_and_test.sh
+++ b/.buildkite/scripts/steps/build_and_test.sh
@@ -63,12 +63,14 @@ if [[ x"$BUILDKITE_PULL_REQUEST" != xfalse && "$CPP_CROSS_COMPILE" = "aarch64" ]
     export ML_DEBUG=1
 fi
 
+TEST_FAILURE=0
 # For now, re-use our existing CI scripts based on Docker
 # Don't perform these steps for native linux aarch64 builds as
 # they are built using docker, see above.
 if ! [[ "$HARDWARE_ARCH" = aarch64 && -z "$CPP_CROSS_COMPILE" ]] ; then 
   if [ "$RUN_TESTS" = "true" ]; then
     ${REPO_ROOT}/dev-tools/docker/docker_entrypoint.sh --test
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -109,3 +112,5 @@ if [[ -z "$CPP_CROSS_COMPILE" ]] ; then
   find . -path  "*/**/ml_test_*.out" -o -path "*/**/*.junit" | xargs tar cvzf ${TEST_RESULTS_ARCHIVE}
   buildkite-agent artifact upload "${TEST_RESULTS_ARCHIVE}"
```
droberts195, line 113: If a macOS test fails on line 101 then the script will exit there and not upload the test artifacts here.

Also, on line 106 we probably _don't_ want to upload a Linux build if the tests were run and failed.

.buildkite/scripts/steps/build_and_test.ps1
```diff
@@ -58,10 +58,12 @@ $ErrorActionPreference="Continue"
...
 & ".\gradlew.bat" --info "-Dbuild.version_qualifier=$Env:VERSION_QUALIFIER" "-Dbuild.snapshot=$Env:BUILD_SNAPSHOT" $DebugOption $Tasks 2>&1 | % { "$_" }
-if ($LastExitCode -ne 0) {
-    Exit $LastExitCode
-}
+$ExitCode=$?
 
 Get-ChildItem -Path . -Include *.out -File -Recurse -ErrorAction SilentlyContinue | Compress-Archive -DestinationPath windows-x86_64-unit_test_results.zip
+buildkite-agent artifact upload "windows-x86_64-unit_test_results.zip"
+if ($ExitCode -ne "True") {
+    Exit 1
```
droberts195, line 66: This is losing information. Why not set `ExitCode` to `LastExitCode` and preserve the exit code rather than converting any error to 1? The shell script version preserves the exit code.

droberts195 (approved):
LGTM

PR #2502, opened 2023-06-02
[ML] Fix BUILD_SNAPSHOT env vars in catalog-info.yaml

edsavage:
The value of the BUILD_SNAPSHOT env vars must be a quoted string, either "true" or "false"


droberts195 (approved):
LGTM

PR #2504, opened 2023-06-05
[ML] Better name the ml-cpp PR build pipeline

edsavage:


edsavage:
Ignoring the CI checks as this change is not directly related to them.

PR #2508, opened 2023-06-06
[ML] BuildKite fix linux x86_64 dependencies

edsavage:
The dependency libraries for the linux x86_64 build are not being collected correctly. An error in the build log indicates that this is due to incorrect interpretation of the `CPP_CROSS_COMPILE` environment variable.


droberts195 (approved):
LGTM

PR #2509, opened 2023-06-06
ML] Do not upload unnecessary artifacts

edsavage:
* Do not upload artifacts stemming from cross compilation builds.
 * Correct the default name of the PR build pipeline.


.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -100,15 +100,15 @@ else
...
-fi
-
 if [[ -z "$CPP_CROSS_COMPILE" ]] ; then 
-  OS=$(uname -s | tr "A-Z" "a-z")
-  TEST_RESULTS_ARCHIVE=${OS}-${HARDWARE_ARCH}-unit_test_results.tgz
-  find . -path  "*/**/ml_test_*.out" -o -path "*/**/*.junit" | xargs tar cvzf ${TEST_RESULTS_ARCHIVE}
-  buildkite-agent artifact upload "${TEST_RESULTS_ARCHIVE}"
+  if [[ $TEST_OUTCOME -eq 0 ]] ; then 
+    buildkite-agent artifact upload "build/distributions/*.zip"
+  else
```
droberts195, line 106: I think it's best to upload the test output even if the build succeeds. If we only upload test results when there are failures then we'll never be able to compare successful output against failure output.

.buildkite/scripts/steps/build_and_test.sh
```diff
@@ -100,15 +100,15 @@ else
...
-  buildkite-agent artifact upload "build/distributions/*.zip"
-fi
-
 if [[ -z "$CPP_CROSS_COMPILE" ]] ; then 
-  OS=$(uname -s | tr "A-Z" "a-z")
-  TEST_RESULTS_ARCHIVE=${OS}-${HARDWARE_ARCH}-unit_test_results.tgz
-  find . -path  "*/**/ml_test_*.out" -o -path "*/**/*.junit" | xargs tar cvzf ${TEST_RESULTS_ARCHIVE}
-  buildkite-agent artifact upload "${TEST_RESULTS_ARCHIVE}"
+  if [[ $TEST_OUTCOME -eq 0 ]] ; then 
+    buildkite-agent artifact upload "build/distributions/*.zip"
```
droberts195, line 105: By moving this here will we still upload cross compiled Intel macOS artifacts?

droberts195 (commented):
I'm thinking `build_and_test.sh` was better as it was before. Maybe what it really needed was a comment on line 102 to say why we don't upload cross compiled artifacts _for `aarch64` only_.

.buildkite/scripts/steps/build_and_test.sh
edsavage: > 

Hmm no. In hindsight I think you're right, this file was better as it was. I'll revert these changes and stop playing with it.

droberts195 (approved):
LGTM

PR #2513, opened 2023-06-07
[ML] Add missing contexts for github commit statuses

edsavage:
The format validation step and the ES integration tests step were missing configuration to notify github about their status. This PR remedies that.


droberts195 (approved):
LGTM

PR #2519, opened 2023-06-12
[7.17][ML] Upload build artifacts from BuildKite (#2511)

edsavage:
For the snapshot and staging builds:

 * Disable uploading from Jenkins
 * Enable uploading from BuildKite

Backports #2511


edsavage:
Merging even though there are failing tests, as these changes relate only to the build infrastructure, not business logic.

PR #2526, opened 2023-06-15
[ML] How to profile autodetect on MacOS

valeriy42:
This PR adds bash scripts and a description of how to profile the anomaly detection process started from the Elasticsearch Java backend.

### DIFF
--- a/dev-tools/autodetect_profiling_macos/autodetect_profiling_on_macos.md
+++ b/dev-tools/autodetect_profiling_macos/autodetect_profiling_on_macos.md
@@ -1,0 +1,155 @@
+# Anomaly detection profiling on MacOS <!-- omit in toc -->
+
+---
+- Author: Valeriy Khakhutskyy
+- Date: 2023-06-15
+---
+
+- [Objective](#objective)
+- [Recording Information](#recording-information)
+  - [Using `leaks`](#using-leaks)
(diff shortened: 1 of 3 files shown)


dev-tools/autodetect_profiling_macos/autodetect_profiling_on_macos.md
```diff
@@ -0,0 +1,155 @@
...
+## Common issues
+
+### Changing the Executable Signature
+
+You may receive the following error message when you select the executable:
+```
+Required kernel recording resources are in use by another document
+```
+
+To fix the permission problems, I need to apply the **in zsh** following command to the executable (instead of `pytorch_inference`):
```
tveasey, line 110:
```suggestion
To fix the permission problems, you need to apply **in zsh** the following command to the executable (instead of `pytorch_inference`):
```

tveasey (commented):
Overall looks great. One top-level comment: do we need `record_memgraph.sh` and `record_xctrace.sh`? I wonder if we can just run both commands at once.

valeriy42:
@tveasey IMO, we usually don't want to record more information than needed as it is resource-expensive. The scripts are useful examples. I wouldn't expect them to be comprehensive tools. For instance, using `xctrace`, you may want to use a different template or even to create your own template that captures counters you are interested in for a particular problem.

PR #2527, opened 2023-06-15
[ML] Replace CResourceMonitor::m_AllocationFailures to reduce the memory footprint

### LINKED ISSUE #2524: [ML] Replace CResourceMonitor::m_AllocationFailures to reduce the memory footprint
The map `CResourceMonitor::m_AllocationFailures` collects the times of all allocation failures due to the hard_limit state. Although we only require the total number of failures and the time of the last failure in code, we keep the time of all failures in the map.

Hence, we can easily reduce the memory footprint of the anomaly jobs by refactoring this member variable.

valeriy42:
Fixes #2524

Mark it as a non-issue since it is not customer-facing.

### DIFF
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -179,7 +179,6 @@ public:
 private:
     using TMonitoredResourcePtrSizeUMap =
         boost::unordered_map<CMonitoredResource*, std::size_t>;
-    using TSizeTimePr = std::pair<std::size_t, core_t::TTime>;
     using TMeanVarAccumulator =
         maths::common::CBasicStatistics::SSampleMeanVar<double>::TAccumulator;
 
--- a/include/model/CResourceMonitor.h
+++ b/include/model/CResourceMonitor.h
@@ -258,8 +258,11 @@ private:
     //! Callback function to fire when memory usage increases by 1%
     TMemoryUsageReporterFunc m_MemoryUsageReporter;
 
-    //! Keep track of times of anomaly detector allocation failures
-    TSizeTimePr m_AllocationFailures;
+    //! Keep track of the number of allocation failures
+    std::size_t m_AllocationFailuresCount{0};
+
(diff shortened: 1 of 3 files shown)


include/model/CResourceMonitor.h
```diff
@@ -259,7 +259,7 @@ class MODEL_EXPORT CResourceMonitor {
     TMemoryUsageReporterFunc m_MemoryUsageReporter;
 
     //! Keep track of times of anomaly detector allocation failures
-    TTimeSizeMap m_AllocationFailures;
+    TSizeTimePr m_AllocationFailures;
```
droberts195, line 262: The comment is now wrong.

But I think the code would be much clearer overall if there were two separate member variables, namely `m_AllocationFailureCount` and `m_LastAllocationFailureTime`. Readers should be able to guess what those mean without even referring to the comment.

dimkots (approved):
LGTM 🚀 (But let's wait for a more exp. c++ person)

include/model/CResourceMonitor.h
valeriy42: This makes sense. Thank you.

droberts195 (approved):
LGTM

Thanks for renaming the variables. It's much clearer now than when it was using the `map` originally as well as being clearer than the `pair`.

include/model/CResourceMonitor.h
```diff
@@ -258,8 +257,11 @@ class MODEL_EXPORT CResourceMonitor {
     //! Callback function to fire when memory usage increases by 1%
     TMemoryUsageReporterFunc m_MemoryUsageReporter;
 
-    //! Keep track of times of anomaly detector allocation failures
-    TTimeSizeMap m_AllocationFailures;
+    //! Keep track of the number of allocation failures
```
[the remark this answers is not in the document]
valeriy42: comment updated

PR #2531, opened 2023-06-15
Add comment about preserving important log message

davidkyle:
Elasticsearch elastic/elasticsearch#96877 uses the specific text in the error messages about the unexpected termination of a process to trigger auditing the message in the UI.  This change adds a comment to say please don't modify the text. Additionally a method is renamed to be less triggering.


tveasey (approved):
LGTM

davidkyle:
Closing without merge as the related change in Elasticsearch was not merged

PR #2532, opened 2023-06-15
[ML] Add internal flag to flush api determining whether to refresh indices or not.

edsavage:
When datafeeds send flush requests they don't require the indices to be refreshed.
    
Round trip the "should_refresh" flag from ES to ml-cpp code and back again.

### DIFF
--- a/lib/api/CAnomalyJob.cc
+++ b/lib/api/CAnomalyJob.cc
@@ -366,7 +366,6 @@ bool CAnomalyJob::handleControlMessage(const std::string& controlMessage) {
         if (core::CStringUtils::stringToType(controlMessage.substr(1), refreshRequired) == false) {
             LOG_ERROR(<< "Received request to flush with invalid control message '"
                       << controlMessage << "'");
-            break;
         } else {
             m_RefreshRequired = refreshRequired;
         }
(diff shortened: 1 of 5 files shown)


lib/api/CAnomalyJob.cc
```diff
@@ -323,6 +323,9 @@ bool CAnomalyJob::handleControlMessage(const std::string& controlMessage) {
         return false;
     }
 
+    bool refreshRequired{false};
+    bool ret{true};
```
droberts195, line 327: Please remove.

lib/api/CAnomalyJob.cc
```diff
@@ -358,6 +359,17 @@ bool CAnomalyJob::handleControlMessage(const std::string& controlMessage) {
         break;
     case 'w':
         this->processPersistControlMessage(controlMessage.substr(1));
+        break;
+    case 'z':
+        LOG_INFO(<< "Received control message '" << controlMessage << "'");
```
droberts195, line 364: I should have noticed this before, but this is going to be unacceptably spammy in production. If you want to leave it as INFO for a day or two to observe the effect that’s OK. But if you do that we need another PR before feature freeze that sets it to TRACE. Or set it to TRACE in this PR if you prefer.

droberts195 (approved):
LGTM, although at some point something will need to be done about the logging.

PR #2533, opened 2023-06-19
[NLP] Update evaluate.py for results format changes

davidkyle:
Adapt the test script for the result format changes in #2376


droberts195 (approved):
LGTM

PR #2537, opened 2023-06-19
[ML] Remove the now unneeded DRA pipeline definition

edsavage:
The logic in the standalone DRA pipeline definition has been amalgamated into the snapshot and staging branch build pipelines and can now safely be removed.


droberts195 (approved):
LGTM

PR #2543, opened 2023-06-20
[ML] Preparing for 8.9 feature freeze

droberts195:
- Bump version on `main` branch
- Set up branch builds for the new `8.9` branch

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -117,7 +117,7 @@ spec:
         build_branches: true
         build_pull_request_forks: true
         cancel_deleted_branch_builds: true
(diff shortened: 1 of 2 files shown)


edsavage (approved):
LGTM

Although the failing ES integration test setup may require some investigation

droberts195:
> It may prove simpler to somehow ignore the ES integration tests during version bumping.. perhaps with the aid of a label?

This isn't high priority. We can just merge with the failures that we know the reason for. It's much more important to get serverless working than to avoid a PR build failure once every 2 months.

PR #2544, opened 2023-06-20
[ML] Remove the ML_KEEP_GOING functionality

edsavage:
The 7.17 branch still has mentions of the ML_KEEP_GOING functionality which affected behaviour of GNU make builds run under Jenkins. Remove that now we are using BuildKite for CI.


droberts195 (approved):
LGTM

Those make files look a bit silly now, setting `FAILED=1` and then unconditionally exiting.

But it's OK given that this is in the 7.17 branch which has very little activity now.

edsavage:
Merging even with a Jenkins build failure as the BuildKite jobs have all passed

PR #2545, opened 2023-06-21
[NLP] Add memory benchmark mode to evaluate.py

davidkyle:
The script uses the functionality added in #2487 to get the max RSS memory usage from the `pytorch_inference` process. The benchmark sends batches of inferences to be evaluated followed by a get memory usage request and prints a summary of memory usage vs batch size.

The only complication is that `pytorch_inference` handles control messages on the main thread while model evaluation is off-loaded to a thread pool. To ensure that the inference requests and get memory request are processed sequentially and in order I added an  `--useImmediateExecutor` flag to `pytorch_inference`. When set the immediate executor is used to process inference requests. This option should only be used for benchmarking.

### DIFF
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -9,9 +9,6 @@
  * limitation.
  */
 
-#include <ATen/core/ATen_fwd.h>
-#include <ATen/core/TensorBody.h>
-#include <ATen/ops/cat.h>
 #include <core/CBlockingCallCancellingTimer.h>
 #include <core/CLogger.h>
 #include <core/CProcessPriority.h>
@@ -21,7 +18,6 @@
 #include <core/CStringUtils.h>
 #include <core/Concurrency.h>
 
-#include <cstddef>
 #include <seccomp/CSystemCallFilter.h>
 
 #include <ver/CBuildInfo.h>
@@ -47,8 +43,6 @@
 torch::Tensor infer(torch::jit::script::Module& module_,
                     ml::torch::CCommandParser::SRequest& request) {
 
-    LOG_INFO(<< "request " << request.s_RequestId << ", " << request.s_NumberInferences);
-
     std::vector<torch::jit::IValue> inputs;
     inputs.reserve(1 + request.s_SecondaryArguments.size());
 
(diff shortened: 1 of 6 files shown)


bin/pytorch_inference/Main.cc
```diff
@@ -274,29 +277,41 @@ int main(int argc, char** argv) {
...
+        ml::core::startDefaultAsyncExecutor(0, 1);
+        // Set the number of threads to use
+        ml::core::defaultAsyncExecutor().numberThreadsInUse(threadSettings.numAllocations());
+    }
+    ml::core::CExecutor& executor{ml::core::defaultAsyncExecutor()};
+    ml::core::CImmediateExecutor immediateExecutor{ml::core::CImmediateExecutor()};
+
+    if (useImmediateExecutor) {
+        executor = immediateExecutor;
+    }
```
tveasey, line 293: If you want immediate execution this is the default if you don't start the async executor or to be explicit you could explicitly stop.
```suggestion
    if (useImmediateExecutor == false) {
        // Size the threadpool to the number of hardware threads
        // so we can grow and shrink the threadpool dynamically.
        // The task queue size is set to 1.
        ml::core::startDefaultAsyncExecutor(0, 1);
        // Set the number of threads to use
        ml::core::defaultAsyncExecutor().numberThreadsInUse(threadSettings.numAllocations());
    } else {
        // Make sure we're using immediate execution.
        ml::core::stopDefaultAsyncExecutor();
    }
```

bin/pytorch_inference/Main.cc
```diff
@@ -83,9 +83,10 @@ torch::Tensor infer(torch::jit::script::Module& module_,
 bool handleRequest(ml::torch::CCommandParser::CRequestCacheInterface& cache,
                    ml::torch::CCommandParser::SRequest request,
                    torch::jit::script::Module& module_,
+                   ml::core::CExecutor& executor,
                    ml::torch::CResultWriter& resultWriter) {
 
-    ml::core::async(ml::core::defaultAsyncExecutor(), [
+    ml::core::async(executor, [
```
tveasey, line 89: My inclination is to leave this using the default executor and not pass explicitly then update the start logic as before.

bin/pytorch_inference/evaluate.py
```diff
@@ -51,6 +51,25 @@
...
+this script. The requests are hard-coded to 512 tokens (padded) and the number of requests
+in each batch is also hardcoded. See the `memory_usage(args)` function to change these settings.
+
+NOTE The generated tokens, and hence the benchmark are only compatible with BERT based models.
+
+Memory Benchmark model spawns the pytorch_inference process with the `--useImmediateExecutor` 
+flag to ensure that processing the get memory usage control message and model evaluation
+occurs in sequence.
+
+Switch to memeory benchmark mode by passing the `--memory_benchmark` argument.
```
tveasey, line 71: typo
```suggestion
Switch to memory benchmark mode by passing the `--memory_benchmark` argument.
```

bin/pytorch_inference/evaluate.py
```diff
@@ -51,6 +51,25 @@
 
 Switch to threading benchmark mode by passing the `--threading_benchmark` argument.
 
+MEMORY_BENCHMARK MODE
+-----------
+
+Evalutes the model over a number of batches of increasing size. After each batch is
+processed a get memory usage request is sent and the Max RSS returned. A summary of
+memory usage againts batch size is printed at the end of the  benchmark.
```
tveasey, line 59:
```suggestion
memory usage against batch size is printed at the end of the  benchmark.
```

bin/pytorch_inference/evaluate.py
```diff
@@ -120,7 +151,7 @@ def launch_pytorch_app(args):
         '--restore=' + args.restore_file,
         '--input=' + args.input_file,
         '--output=' + args.output_file,
-        '--validElasticLicenseKeyConfirmed=true'
+        '--validElasticLicenseKeyConfirmed=true',
```
tveasey, line 154: Formatting
```suggestion
        '--validElasticLicenseKeyConfirmed=true',
```

include/core/Concurrency.h
```diff
@@ -82,6 +82,15 @@ class CExecutor {
...
 };
 
+//! \brief Executes a function immediately (on the calling thread).
+class CImmediateExecutor final : public CExecutor {
+public:
+    void schedule(std::function<void()>&& f) override { f(); }
+    bool busy() const override { return false; }
+    void busy(bool) override {}
+    void numberThreadsInUse(std::size_t) override {}
+};
```
tveasey, line 92: Using the default executor means this can stay private.

tveasey (commented):
Just a couple of small things. Also, I think it is perhaps cleaner to just keep on using the default executor.

tveasey (approved):
LGTM

PR #2547, opened 2023-06-29
Ipex investigation

valeriy42:


3rd_party/licenses/ipex-INFO.csv
```diff
@@ -0,0 +1,2 @@
+name,version,revision,url,license,copyright,sourceURL
+Intel® Extension for PyTorch,1.13.100+cpu,,https://github.com/intel/intel-extension-for-pytorch,Apache 2.0,(c) Intel Corporation,http://intel-optimized-pytorch.s3.cn-north-1.amazonaws.com.cn/libipex/cpu/libintel-ext-pt-cxx11-abi-1.13.100%2Bcpu.run
```
droberts195, line 2: The license needs to be `Apache-2.0` with a hyphen, so that it will match the identifier on SPDX: https://spdx.org/licenses/Apache-2.0.html

For Apache 2 dependencies we don't need the `sourceURL` link. That causes us to take a copy of the source code and make it available for download on our servers, and that's not necessary for Apache 2. So please leave `sourceURL` blank for this dependency.

I am also not sure about including the "®" symbol. It leaves us vulnerable to a character encoding bug far in the depths of the unified release process. I think it would be better to use "(R)", which won't cause a problem that's detected only on release day.

Finally, this file will get processed by shell commands, which like newline characters at the end of lines, so please make sure the second line ends with a newline character.

build-setup/macos.md
```diff
@@ -102,8 +102,8 @@ to:
 To complete the build, type:
 
 ```
-./b2 -j8 --layout=versioned --disable-icu cxxflags="-std=c++17 -stdlib=libc++ $SSEFLAGS" linkflags="-std=c++17 -stdlib=libc++ -Wl,-headerpad_max_install_names" optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC
-sudo ./b2 install --layout=versioned --disable-icu cxxflags="-std=c++17 -stdlib=libc++ $SSEFLAGS" linkflags="-std=c++17 -stdlib=libc++ -Wl,-headerpad_max_install_names" optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC
```
droberts195, line 106: The macOS files shouldn't be changed by this PR.

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -161,11 +164,42 @@ RUN \
...
+  /usr/local/bin/python3.10 setup.py build_clib /usr/src/pytorch/torch
+
+
+RUN \
+  cd ${build_dir}/pytorch && \
   mkdir /usr/local/gcc103/include/pytorch && \
   cp -r torch/include/* /usr/local/gcc103/include/pytorch/ && \
   cp torch/lib/libtorch_cpu.so /usr/local/gcc103/lib && \
   cp torch/lib/libc10.so /usr/local/gcc103/lib && \
+  cp /usr/src/intel-extension-for-pytorch/build/Release/packages/intel_extension_for_pytorch/lib/libintel-ext-pt-cpu.so /usr/local/gcc103/lib && \
```
droberts195, line 202:
```suggestion
  cp ${build_dir}/intel-extension-for-pytorch/build/Release/packages/intel_extension_for_pytorch/lib/libintel-ext-pt-cpu.so /usr/local/gcc103/lib && \
```

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -161,11 +164,42 @@ RUN \
...
   cd .. && \
   rm -rf pytorch
```
droberts195, line 204: You've split the building of PyTorch over multiple `RUN` steps. The reason it was all in one before was that trying to delete the `pytorch` directory after having built it up via multiple file system layers sometimes caused errors when we first tried that. Maybe it's OK with certain versions of Docker but not others.

The other consideration with using multiple `RUN` steps is the disk space needed by Docker. The build script warns if less than 50GB is available, but does it now need to say that more is required?

dev-tools/docker/build_linux_build_image.sh
```diff
@@ -34,7 +34,7 @@ sleep 5
 HOST=docker.elastic.co
 ACCOUNT=ml-dev
 REPOSITORY=ml-linux-build
-VERSION=25
+VERSION=26
```
edsavage, line 37: FYI: I've already created a ml-linux-build:26 in order to upgrade to Boost 1.83 in the PR https://github.com/elastic/ml-cpp/pull/2560

droberts195: I think we _could_ override that with the version from this PR if it was ready to merge first.

When it's clearer which PR will be fit to merge first we can decide which one will create the final image 26, and then update the second PR to incorporate the changes of the first and bump the version to 27.

3rd_party/licenses/ipex-INFO.csv
valeriy42: Thank you for all this pointers!

build-setup/macos.md
valeriy42: Oops-a-daisy...

dev-tools/docker/linux_image/Dockerfile
valeriy42: Indeed, thank you for the hint. As this was a WIP I needed to be able to skip some more expensive steps repeating the same process multiple times. I squashed all the steps for torch and ipex in one RUN.

valeriy42:
@droberts195, thank you for the review. It would be great if you could give it another pass.

droberts195 (commented):
The Linux CI is failing with this:

```
gmake[2]: *** No rule to make target `/usr/local/gcc103/lib/libintel-ext-pt-cpu.so', needed by `bin/pytorch_inference/pytorch_inference'.  Stop.
```

I think you also need to change 25 to 26 in `.buildkite/pipelines/build_linux.json.py`.

bin/pytorch_inference/Main.cc
```diff
@@ -105,6 +105,9 @@ bool handleRequest(ml::torch::CCommandParser::CRequestCacheInterface& cache,
                 } catch (std::runtime_error& e) {
                     resultWriter.writeError(request_.s_RequestId, e.what());
                     return std::nullopt;
+                } catch (...) {
```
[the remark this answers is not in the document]
valeriy42: The idea was to have a catch-all for other errors related to inference and write them out using `resultWriter`.  From the client point (e.g. `evaluate.py`), it's very difficult to understand otherwise, why processor doesn't return any results.

bin/pytorch_inference/Main.cc
```diff
@@ -283,6 +286,9 @@ int main(int argc, char** argv) {
     } catch (const c10::Error& e) {
         LOG_FATAL(<< "Error loading the model: " << e.msg());
         return EXIT_FAILURE;
+    } catch (...) {
```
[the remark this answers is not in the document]
valeriy42: Ok, I'll update the catch block 👍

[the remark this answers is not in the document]
valeriy42: I removed the `catch (...) `  block.

bin/pytorch_inference/Main.cc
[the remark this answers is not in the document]
droberts195: If an exception we had no idea existed happens then it's reasonable to terminate the program. At least that will mean that we find out what the exception was from the stack trace. If we ever saw a whole series of `Unknown Error` messages that wouldn't be very helpful in diagnosing the problem. Maybe what we should do in the long term is to propagate unknown exceptions to the main thread (using `exception_ptr` and `current_exception()`) and rethrow them there so they actually terminate the program. But let's not add that to this PR because it would be good to get it merged today to allow the follow-on work to progress. So for this PR just add a `TODO` to improve this in the future.

droberts195 (approved):
LGTM

PR #2549, opened 2023-06-29
[ML] Add getrusage to list of allowed syscall

### LINKED ISSUE #2546: [NLP] Missing Seccomp filters for getrsusage?
#2487 added control message to `pytorch_inference` that reports memory usage based on a call to `getrusage`. On some OS's the call fails most likely due to the Seccomp system call filter. 

`getrusage` is never called in production code only in the memory benchmark mode of `evaluate.py`. Before the functionality is ever used in production code investigate that it works on Linux distributions that are supported.

davidkyle:
In support of https://github.com/elastic/ml-cpp/pull/2487

Closes #2546

Non-issue as no production code calls getrusge


droberts195 (approved):
LGTM

PR #2550, opened 2023-06-29
[ML] Display info for all branches for PR builds

edsavage:
Currently the default value of "main" is used for the default_branch configuration option for the "ml-cpp-pr-build" pipeline. This results in no useful information about PR build history being displayed for that pipeline in the BuildKite UI.

By explicitly setting the default_branch config option to nul BuildKite will display build information for all PR branches, which is much more useful.


edsavage:
Merging without waiting for CI as this change is unrelated to the `ml-cpp` code itself

PR #2552, opened 2023-08-11
[NLP] Respond to the server when an unknown command is received 

davidkyle:
pytorch_inference must respond to every request as Elasticsearch is waiting for a response. The switch statement in the command handler code logged an error when an unknown command was received but did not write a response. This change adds an error response. 

In practice I cannot see how an unrecognised command could be sent as the Elasticsearch side is closely tied to pytorch_inference, I've labelled this a non-issue rather than a bug for that reason.


droberts195 (approved):
LGTM

PR #2554, opened 2023-08-14
[ML] Download Windows build dependencies from GCS

edsavage:
Change Windows build scripts to obtain dependency archives from a new location on GCS, rather than AWS S3, which is proving unreliable.

Labelling as a non issue as this is not a notable change for end users.


droberts195:
This looks like it's worked, but please could you add a line to the script that prints out the date and time as well, so that we have information to test the other theory. The PR CI will run again when that change is made which will be an extra stability test too.

Also, this change will need backporting to 8.9 and 7.17, as those branches are also suffering download failures.

droberts195 (approved):
LGTM

PR #2560, opened 2023-08-23
Upgrade to boost 1 83

edsavage:
Upgrade to Boost 1.83 libraries

We have been using Boost 1.77 for approximately 2 years now, and it's good practice to upgrade occasionally.
    
As Boost 1.83 has just been released it makes sense to attempt to upgrade to it.

Labelling as a `>non-issue` as this change should have no outwardly facing effects.


build-setup/windows.md
```diff
@@ -119,33 +119,21 @@ nmake
...
-Edit `boost/unordered/detail/implementation.hpp` and change line 287 from:
-
-```
-    (17ul)(29ul)(37ul)(53ul)(67ul)(79ul) \
-```
-
-to:
-
-```
-    (3ul)(17ul)(29ul)(37ul)(53ul)(67ul)(79ul) \
```
droberts195, line 142: Why is this substitution no longer used? Did the Boost maintainers decide to start off with a smaller hash table in the official code? This optimisation was originally included because the anomaly detection code creates large numbers of hash maps with one item in them, and reserving 17 slots in such cases noticeably increases the memory usage of the `autodetect` process for high cardinality jobs.

edsavage: The `BOOST_UNORDERED_PRIMES` definition is no longer present after `Boost 1.79`, so there's no longer anything to substitute.

droberts195: Please can you grep the unordered code and check if another file has a similar set of primes in it. If you grep for 17 and 29 on the same line that might find it. I’m a bit worried we’ll get reports of greatly increased model memory sizes if the memory usage of a size 1 hash map dramatically increases.

droberts195: I suggest adding `(3ul)` to the beginning of that line and see if everything still works and memory usage is reduced.

dev-tools/download_windows_deps.ps1
```diff
@@ -9,7 +9,7 @@
 # limitation.
 #
 $ErrorActionPreference="Stop"
-$Archive="usr-x86_64-windows-2016-9.zip"
+$Archive="usr-x86_64-windows-2016-10.zip"
 $Destination="C:\"
 # If PyTorch is not version 1.13.1 then we need the latest download
```
droberts195, line 14: The condition here needs to be changed to look for a file that only exists in Boost 1.83.

https://github.com/elastic/ml-cpp/pull/638/files#diff-0a356335cbaadf7c9897e08805d11e4d124df92f6b166867982739296cd008b1 is an example from a previous upgrade.

(It's much easier to detect if latest Boost is present than to detect latest PyTorch.)

build-setup/macos.md
```diff
@@ -69,36 +69,37 @@ xcode-select --install
...
+In the resulting `boost_1_83_0` directory, run:
 
 ```
 ./bootstrap.sh --with-toolset=clang --without-libraries=context --without-libraries=coroutine --without-libraries=graph_parallel --without-libraries=mpi --without-libraries=python --without-icu
 ```
 
 This should build the `b2` program, which in turn is used to build Boost.
 
-Edit `boost/unordered/detail/implementation.hpp` and change line 287 from:
+Edit boost/unordered/detail/prime_fmod.hpp and change line 134 from
```
droberts195, line 90:
```suggestion
Edit `boost/unordered/detail/prime_fmod.hpp` and change line 134 from
```

cmake/variables.cmake
```diff
@@ -226,7 +226,7 @@ set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_DEBUG_RUNTIME OFF)
 set(Boost_COMPILER "${ML_BOOST_COMPILER_VER}")
 
-find_package(Boost 1.77.0 REQUIRED COMPONENTS iostreams filesystem program_options regex date_time log log_setup thread unit_test_framework)
+find_package(Boost 1.83.0 REQUIRED COMPONENTS iostreams filesystem program_options regex date_time log log_setup thread unit_test_framework json)
```
droberts195, line 229:
```suggestion
find_package(Boost 1.83.0 EXACT REQUIRED COMPONENTS iostreams filesystem program_options regex date_time log log_setup thread unit_test_framework json)
```

build-setup/linux.md
```diff
@@ -171,41 +171,41 @@ sudo make install
...
+In the resulting `boost_1_83_0` directory, run:
 
 ```
 ./bootstrap.sh --without-libraries=context --without-libraries=coroutine --without-libraries=graph_parallel --without-libraries=mpi --without-libraries=python --without-icu
 ```
 
 This should build the `b2` program, which in turn is used to build Boost.
 
-Edit `boost/unordered/detail/implementation.hpp` and change line 287 from:
+Edit boost/unordered/detail/prime_fmod.hpp and change line 134 from
```
droberts195, line 192:
```suggestion
Edit `boost/unordered/detail/prime_fmod.hpp` and change line 134 from:
```

build-setup/windows.md
```diff
@@ -119,33 +119,33 @@ nmake
...
 Assuming you chose the `.bz2` version, extract it in a Git bash shell using the GNU tar that comes with Git for Windows, e.g.:
 
 ```
 cd /c/tools
-tar jxvf /z/cpp_src/boost_1_77_0.tar.bz2
+tar jxvf /z/cpp_src/boost_1_83_0.tar.bz2
 ```
 
-Edit `boost/unordered/detail/implementation.hpp` and change line 287 from:
+Edit boost/unordered/detail/prime_fmod.hpp and change line 134 from
```
droberts195, line 133:
```suggestion
Edit `boost/unordered/detail/prime_fmod.hpp` and change line 134 from:
```

build-setup/linux.md
```diff
@@ -171,41 +171,41 @@ sudo make install
...
+./b2 -j6 --layout=versioned --disable-icu pch=off optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC define=_FORTIFY_SOURCE=2 cxxflags='-std=gnu++17 -fstack-protector -msse4.2 -mfpmath=sse' cflags='-D__STDC_FORMAT_MACROS' linkflags='-std=gnu++17 -Wl,-z,relro -Wl,-z,now'
+sudo env PATH="$PATH" LD_LIBRARY_PATH="$LD_LIBRARY_PATH" ./b2 install --prefix=/usr/local/gcc103 --layout=versioned --disable-icu pch=off optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC define=_FORTIFY_SOURCE=2 cxxflags='-std=gnu++17 -fstack-protector -msse4.2 -mfpmath=sse' cflags='-D__STDC_FORMAT_MACROS' linkflags='-std=gnu++17 -Wl,-z,relro -Wl,-z,now'
```
droberts195, line 208: The new flag `__STDC_FORMAT_MACROS` isn't being used with the `linux-aarch64` Docker image. I don't see why we'd need the flag on `x86_64` but not `aarch64` when we're using the same compiler and standard library version on both.

dev-tools/docker/build_macosx_build_image.sh
```diff
@@ -11,7 +11,7 @@
 #
 
 # Builds the Docker image that can be used to compile the machine learning
-# C++ code for Linux
+# C++ code for macOS
```
droberts195, line 14:
```suggestion
# C++ code for Intel macOS
```

dev-tools/docker/linux_aarch64_native_image/Dockerfile
```diff
@@ -75,14 +75,14 @@ RUN \
...
+  sed -i -e 's|(13ul)(29ul)(53ul)(97ul)(193ul)(389ul)(769ul)(1543ul)(3079ul)(6151ul)( \\|(3ul)(13ul)(29ul)(53ul)(97ul)(193ul)(389ul)(769ul)(1543ul)(3079ul)(6151ul)(       \\|' boost/unordered/detail/prime_fmod.hpp && \
   ./b2 -j`nproc` --layout=versioned --disable-icu pch=off optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC define=_FORTIFY_SOURCE=2 cxxflags='-std=gnu++17 -fstack-protector -march=armv8-a+crc+crypto' linkflags='-std=gnu++17 -Wl,-z,relro -Wl,-z,now' && \
   ./b2 install --prefix=/usr/local/gcc103 --layout=versioned --disable-icu pch=off optimization=speed inlining=full define=BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS define=BOOST_LOG_WITHOUT_DEBUG_OUTPUT define=BOOST_LOG_WITHOUT_EVENT_LOG define=BOOST_LOG_WITHOUT_SYSLOG define=BOOST_LOG_WITHOUT_IPC define=_FORTIFY_SOURCE=2 cxxflags='-std=gnu++17 -fstack-protector -march=armv8-a+crc+crypto' linkflags='-std=gnu++17 -Wl,-z,relro -Wl,-z,now' && \
```
droberts195, line 83: These are the lines where I'd expect to see `cflags='-D__STDC_FORMAT_MACROS'` if it's needed at all.

lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -3085,7 +3085,8 @@ BOOST_AUTO_TEST_CASE(testEstimateMemory) {
         LOG_DEBUG(<< "estimated memory usage = " << estimatedMemory);
         LOG_DEBUG(<< "high water mark = " << predictInstrumentation.maxMemoryUsage());
 
-        BOOST_TEST_REQUIRE(predictInstrumentation.maxMemoryUsage() <= estimatedMemory);
+        BOOST_TEST_REQUIRE(predictInstrumentation.maxMemoryUsage() <=
+                           (estimatedMemory * 101) / 100);
```
droberts195, line 3089: Instead of adding the uplift here it should be added in https://github.com/elastic/ml-cpp/blob/c28c82c0c29e13990d169fbea5a6e1836e3a1b91/include/core/CMemoryDef.h#L208-L210 https://github.com/elastic/ml-cpp/blob/c28c82c0c29e13990d169fbea5a6e1836e3a1b91/include/core/CMemoryDef.h#L220-L221  https://github.com/elastic/ml-cpp/blob/c28c82c0c29e13990d169fbea5a6e1836e3a1b91/include/core/CMemoryDef.h#L492-L493 and https://github.com/elastic/ml-cpp/blob/c28c82c0c29e13990d169fbea5a6e1836e3a1b91/include/core/CMemoryDef.h#L529-L530

The new memory layout is detailed here: https://www.boost.org/doc/libs/1_83_0/libs/unordered/doc/html/unordered.html#structures_closed_addressing_containers

The formulas should be updated to match this. As far as I can see the key part is:

> there’s only 4 bits of space overhead per bucket introduced by the bucket groups.

So we should add on an extra half a byte per bucket (rounded up to the next higher multiple of `sizeof(std::size_t))`) to account for these bucket groups.

Please give this a try and see if it allows the 1% hack here to be removed. If it does then please also add a link to that section of the Boost docs as part of a comment to explain why we're adding this strange amount.

dev-tools/docker/linux_aarch64_native_image/Dockerfile
edsavage: It seems to be only needed for ancient versions of RH/CentOS e.g. `CentOS Linux release 7.1.1503 (Core)`, where certain macros defined  in `/usr/include/inttypes.h` are guarded by `_STDC_FORMAT_MACROS`. Hence its not strictly required for our docker images which are on a later version of the OS. I'll update the Docker files anyway, for consistency.

droberts195: So are you saying CentOS 7.8 has it but CentOS 7.1 doesn't? Both our Docker images are based on CentOS 7 latest patch.

It's fine to just include it in the commands then. Don't actually spend time rebuilding the images.

edsavage: Cool, yes I've found its only needed on our `marple` server. None of the docker images require it.

droberts195: In that case please add a TODO to the instructions to remove it once we no longer build on CentOS 7.

include/core/CMemoryDef.h
```diff
@@ -207,7 +207,9 @@ template<typename K, typename V, typename H, typename P, typename A>
 std::size_t dynamicSize(const boost::unordered_map<K, V, H, P, A>& t) {
     return memory::elementDynamicSize(t) +
            (t.bucket_count() * memory::storageNodeOverhead(t)) +
-           (t.size() * (sizeof(K) + sizeof(V) + memory::storageNodeOverhead(t)));
+           (t.size() * (sizeof(K) + sizeof(V) + memory::storageNodeOverhead(t))) +
+           (static_cast<std::size_t>(std::ceil(t.bucket_count() / 2.0 / sizeof(std::size_t))) *
+            sizeof(std::size_t));
```
droberts195, line 212: You can do this without resorting to floating point functions like this:

```
((t.bucket_count() + 2 * sizeof(std::size_t) - 1) / 2 * sizeof(std::size_t)) * sizeof(std::size_t)
```

The compiler can simplify that better at compile time and avoid a function call at runtime.

Also, this formula could go in a new function `std::size_t bucketGroupOverhead(std::size_t bucketCount)` that could then have a Doxygen comment explaining what it's doing and linking to the Boost unordered documentation. That function should get inlined by the compiler if it's defined higher up the same file, so won't really add any overhead.

include/core/CMemoryDef.h
```diff
@@ -489,8 +493,11 @@ void dynamicSize(const char* name,
     std::string componentName(name);
     componentName += "_umap";
 
-    std::size_t mapSize = (t.bucket_count() * sizeof(std::size_t) * 2) +
-                          (t.size() * (sizeof(K) + sizeof(V) + 2 * sizeof(std::size_t)));
+    std::size_t mapSize =
+        (t.bucket_count() * sizeof(std::size_t) * 2) +
+        (t.size() * (sizeof(K) + sizeof(V) + 2 * sizeof(std::size_t))) +
```
droberts195, line 498: Even though it's not a problem you introduced, I just noticed that this method is not using a helper function that the others are.

```suggestion
        (t.bucket_count() * memory::storageNodeOverhead(t)) +
        (t.size() * (sizeof(K) + sizeof(V) + memory::storageNodeOverhead(t))) +
```

lib/maths/analytics/unittest/CBoostedTreeTest.cc
```diff
@@ -3085,7 +3085,16 @@ BOOST_AUTO_TEST_CASE(testEstimateMemory) {
         LOG_DEBUG(<< "estimated memory usage = " << estimatedMemory);
         LOG_DEBUG(<< "high water mark = " << predictInstrumentation.maxMemoryUsage());
 
-        BOOST_TEST_REQUIRE(predictInstrumentation.maxMemoryUsage() <= estimatedMemory);
+        // Beginning with Boost 1.80, unordered containers require more memory per
+        // bucket than those of previous versions. Even though this has largely been
+        // accounted for in our memory estimations there still is a small discrepancy
+        // between actual and estimated memory. We account for that maximum percentage
+        // difference here in boostUnorderedContainerMemorySlackPercent.
+        static constexpr double boostUnorderedContainerMemorySlackPercent{0.9};
```
droberts195, line 3093: This fudge factor shouldn't be in the test - it should be in the core code in `core::CDataFrame::estimateMemoryUsage`. Otherwise this test will be asserting that customers will have problems in production.

lib/core/CDataFrame.cc
```diff
@@ -495,13 +495,29 @@ std::size_t CDataFrame::estimateMemoryUsage(bool inMainMemory,
...
+    // between actual and estimated memory. We account for that maximum percentage
+    // difference here in boostUnorderedContainerMemorySlackPercent.
+    static constexpr double boostUnorderedContainerMemorySlackPercent{2.5};
+
+    // calculate the additional memory rounded up to the nearest multiple of sizeof(std::size_t) bytes
+    std::size_t additionalMemory =
+        static_cast<int>((2 * estimatedMemoryUsage * boostUnorderedContainerMemorySlackPercent / 100 +
+                          2 * sizeof(std::size_t) - 1) /
+                         (2 * sizeof(std::size_t))) *
+        sizeof(std::size_t);
```
droberts195, line 518: Since this is only a crude estimate, is it really worth rounding to a multiple of `sizeof(std::size_t)`?

```suggestion
    std::size_t additionalMemory{
        static_cast<std::size_t>(estimatedMemoryUsage * boostUnorderedContainerMemorySlackPercent / 100)};
```

lib/core/CDataFrame.cc
```diff
@@ -495,13 +495,29 @@ std::size_t CDataFrame::estimateMemoryUsage(bool inMainMemory,
...
+        core::memory::dynamicSize(TBoolVec(numberColumns)) +
+        (inMainMemory ? numberRows * CAlignment::roundupSizeof<CFloatStorage>(alignment, numberColumns)
+                      : 0);
+
+    // Beginning with Boost 1.80, unordered containers require more memory per
+    // bucket than those of previous versions. Even though this has largely been
+    // accounted for in our memory estimations there still is a small discrepancy
+    // between actual and estimated memory. We account for that maximum percentage
+    // difference here in boostUnorderedContainerMemorySlackPercent.
+    static constexpr double boostUnorderedContainerMemorySlackPercent{2.5};
```
droberts195, line 511: I think there was also underestimation already in the `TStrVecVec` value, since the existing estimate assumes all the nested vectors are empty. It's probably worth making this fudge factor less specific to the Boost changes, and call the variable something more generic.

lib/api/CAnomalyJobConfig.cc
```diff
@@ -674,6 +676,59 @@ void CAnomalyJobConfig::CDataDescription::parse(const rapidjson::Value& analysis
...
+                            "field varies between detectors");
+                    }
+                }
+            }
+            ++fallbackDetectorIndex;
+        }
+    }
+}
+
+void CAnomalyJobConfig::CAnalysisConfig::parse() {
```
droberts195, line 718: Please give this method a more precise name. It doesn't appear to parse the full analysis config. Maybe something like `reparseDetectorsFromStoredConfig` (or something else that says what it does).

include/api/CAnomalyJobConfig.h
```diff
@@ -248,9 +248,11 @@ class API_EXPORT CAnomalyJobConfig {
 
         void parse(const rapidjson::Value& json);
 
-        bool updateFilters(const boost::property_tree::ptree& propTree);
+        void parse();
```
droberts195, line 251: Please add Doxygen comments to this method and the one immediately above saying why they do and why they're different.

droberts195 (approved):
LGTM if this passes CI

🤞

PR #2566, opened 2023-09-05
[ML] Treat Boost version as exact required, not minimum

droberts195:
A Boost upgrade is not trivial and we don't want CMake deciding to do one for us.


maxhniebergall (approved):
LGTM

PR #2568, opened 2023-09-05
[8.9] [ML] Treat Boost version as exact required, not minimum

droberts195:
A Boost upgrade is not trivial and we don't want CMake deciding to do one for us.

Backport of #2566


droberts195:
Java integration test failure is because the Elasticsearch version bump is out-of-sync with ML. Merging anyway.

Issue #1899, opened 2021-05-13
[ML] Don't try to change the OOM score adjustment when running under Kubernetes


droberts195:
Kubernetes stops processes from changing their OOM score adjustment.  There is more detail about this in https://github.com/kubernetes/kubernetes/issues/90973.

This causes the ML processes to log warning messages when running under Kubernetes, which can worry users.  To avoid this we should change our code to somehow detect that it is running under Kubernetes and then not attempt to change the OOM score adjustment.  We should also keep an eye on https://github.com/kubernetes/kubernetes/issues/90973 to see if this policy of Kubernetes ever changes.

The reason the ML processes increase their OOM score adjustment is similar to the reasoning for PostgreSQL: if the OOM killer has to choose a process it's less catastrophic for it to choose an ML job process than the Elasticsearch JVM.

Assuming our memory accounting is accurate the OOM killer is less likely to be a problem in a container that's dedicated to running Elasticsearch than on a general machine, because there's no chance that the user starts up some unrelated process that consumes a lot of memory.  Therefore this adjustment is less important in containers than on general purpose machines, such as development laptops.  It only really matters in a container if our memory accounting is wrong under some circumstance, and this makes a job use more memory than its model memory limit plus allowed process overhead.

droberts195:
As we move towards running on Kubernetes more and more, we should revisit whether to remove the OOM priority functionality altogether if there's still no solution for setting it on Kubernetes.

PR #2569, opened 2023-09-06
[ML] Reduce log level of OOM killer adjustment warning

droberts195:
Kubernetes doesn't allow OOM killer adjustments. (More on this in https://github.com/kubernetes/kubernetes/issues/90973.)

We try to make these adjustments so that if the kernel has to kill a process it will prefer to kill an ML process rather than the Elasticsearch JVM.

However, with the Kubernetes restriction, the current warning message about not being able to make the adjustment shows up more and more as log spam as ECK and serverless become more prevalent. Therefore, it's best that the message isn't logged by default.

Relates #1899


przemekwitek (approved):
LGTM

PR #2570, opened 2023-09-21
[ML] Increase testMurmurHash slack factor to 8%

edsavage:
The upgrade to Boost 1.83 has resulted in testMurmurHash failing more often. Occasional failures are expected and accounted for in the test with a "slack factor". Increase that factor to 8% to make the test more resilient.


droberts195 (approved):
LGTM

droberts195:
Maybe Boost changed their default string hash function between 1.77 and 1.83? That would explain why murmur hash no longer has such a benefit over the Boost default.

edsavage:
> Maybe Boost changed their default string hash function between 1.77 and 1.83? That would explain why murmur hash no longer has such a benefit over the Boost default.

Yes, I think that is the case, see https://www.boost.org/doc/libs/1_83_0/libs/container_hash/doc/html/hash.html#recent_boost_1_82_0

PR #2572, opened 2023-09-25
[ML] Add extra debug handling for the seccomp test.

edsavage:
It appears that the seccomp test may hang on occasion on macOS when instrumented for debug builds.

This PR attempts to gather extra information (temporarily) regarding the tests in general and the seccomp tests in particular.


edsavage:
> I guess only the 4th file change is temporary, not all of them?

Yes, that's right. The changes in the other files are to upload the junit output files from the maths library unit tests - not directly related to this issue I don't think, but something I noticed needed addressing.

Issue #2574, opened 2023-09-29
[ML] After Boost 1.83 upgrade murmur hash is no longer faster than Boost hash


droberts195:
This is not a bug, more of an observation that's interesting to make searchable in an issue.

Since we upgraded to Boost 1.83 the test that asserts that murmur hash lookup performance is better than the default Boost hash lookup performance regularly fails, even with an 8% tolerance on the timings.

https://github.com/elastic/ml-cpp/blob/0ee9833fbc05658d2ea6c4f1dc82b8c9fe1aa08f/lib/core/unittest/CHashingTest.cc#L316

Some examples of timings from before the Boost upgrade:

```
2023-08-16 09:08:30,805755 UTC [4304] DEBUG /buildkite/builds/bk-agent-prod-k8s-1692174718318253470/elastic/ml-cpp-staging-builds/lib/core/unittest/CHashingTest.cc@301 default insert runtime = 1855ms, murmur insert runtime = 1467ms
2023-08-16 09:08:30,805765 UTC [4304] DEBUG /buildkite/builds/bk-agent-prod-k8s-1692174718318253470/elastic/ml-cpp-staging-builds/lib/core/unittest/CHashingTest.cc@303 default lookup runtime = 5110ms, murmur lookup runtime = 2882ms

2023-06-12 14:12:54,350673 UTC [4280] DEBUG /buildkite/builds/bk-agent-elastic-k8s-1686576616312423140/elastic/ml-cpp-staging-builds/lib/core/unittest/CHashingTest.cc@301 default insert runtime = 2067ms, murmur insert runtime = 1683ms
2023-06-12 14:12:54,350683 UTC [4280] DEBUG /buildkite/builds/bk-agent-elastic-k8s-1686576616312423140/elastic/ml-cpp-staging-builds/lib/core/unittest/CHashingTest.cc@303 default lookup runtime = 6868ms, murmur lookup runtime = 4048ms

2023-09-21 08:36:59,722971 UTC [4304] DEBUG /buildkite/builds/bk-agent-prod-k8s-1695283442532840168/elastic/ml-cpp-staging-builds/lib/core/unittest/CHashingTest.cc@301 default insert runtime = 1214ms, murmur insert runtime = 864ms
[...]

PR #2575, opened 2023-09-29
[ML] Removing assertion on murmur hash lookup performance

droberts195:
Since the upgrade from Boost 1.77 to 1.83 murmur hash is no longer significantly faster than the default Boost hash.

More details in #2574


edsavage (approved):
LGTM

PR #2580, opened 2023-10-03
[ML] Revert Boost 1.83 upgrade for 8.11

droberts195:
The upgrade to Boost 1.83 has caused an unexplained increase in model memory size for a number of nightly test jobs.

While we investigate this PR reverts to Boost 1.77 for the upcoming 8.11 release.

The main branch will continue to use Boost 1.83.

Reverts #2570 and #2560


edsavage (approved):
LGTM

PR #2581, opened 2023-10-03
[ML] Add a release note for Boost 1.83 upgrade

droberts195:
Previously we've release noted major library upgrades like this one.

Relates #2560


edsavage (approved):
LGTM

PR #2582, opened 2023-10-04
[ML] Remove inaccurate log message

davidkyle:
#2576 added an error message when a null value was is read in `CCompressedLfuCache::lookup` as the cache should not contain null values. This was based on a misreading of the code, a null value means the item was not found in the cache.

The curious thing is that the message was only logged once by my application when I would expect to see it every time `CCompressedLfuCache::lookup` was called, I'm assuming this is due to the log throttler removing exact duplicate messages.


droberts195 (approved):
LGTM

davidkyle:
buildkite build this

PR #2585, opened 2023-10-10
[ML] Adjust memory calculations for unordered containers

edsavage:
Boost 1.83 has an entirely different implementation of its unordered containers compared to Boost 1.77 (the previous version used by `ml-cpp`).

This PR adjusts the calculations for memory usage of the unordered containers (map/set) to be commensurate with the new implementation.

### DIFF
--- a/include/core/CMemoryDec.h
+++ b/include/core/CMemoryDec.h
@@ -80,7 +80,10 @@ std::size_t bucketGroupOverhead(const T&) {
 }
 
 //! The overhead associated with the unordered container's \em bucket_groups.
-//! This is calculated as 4 bits per bucket_group pointer, of which there are 4.
+//! On average this is 4 bits per bucket. This is because the bucket_group
+//! structure consists of a 64 bit bitmask representing up to 64 buckets,
+//! plus 3 pointers, so a 32 byte structure stores the information for up to
+//! 64 buckets.
 //! See https://www.boost.org/doc/libs/1_83_0/libs/unordered/doc/html/unordered.html#structures_closed_addressing_containers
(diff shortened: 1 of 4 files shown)


include/core/CMemoryDec.h
```diff
@@ -115,12 +117,12 @@ constexpr std::size_t storageNodeOverhead(const T&) {
 
 template<typename K, typename V, typename H, typename P, typename A>
```
droberts195, line 118: I think it's worth mentioning the Boost version here, so that anyone doing a future Boost upgrade will find this and consider if it's still true.

```suggestion
//! In Boost 1.83 the bucket lists and node lists are singly linked, hence
//! only 1 pointer is needed here
template<typename K, typename V, typename H, typename P, typename A>
```

The same comment can be added to the next method too.

include/core/CMemoryDec.h
```diff
@@ -80,25 +80,27 @@ std::size_t bucketGroupOverhead(const T&) {
 }
 
 //! The overhead associated with the unordered container's \em bucket_groups.
-//! This is calculated as 4 bits per bucket_group.
+//! This is calculated as 4 bits per bucket_group pointer, of which there are 4.
```
droberts195, line 83:
```suggestion
//! On average this is 4 bits per bucket. This is because the bucket_group
//! structure consists of a 64 bit bitmask representing up to 64 buckets,
//! plus 3 pointers, so a 32 byte structure stores the information for up to
//! 64 buckets.
```

And the same comment change can be made below too.

droberts195 (approved):
LGTM if CI passes

PR #2586, opened 2023-10-12
[ML] Ensure estimated latitude is within the allowed range

### LINKED ISSUE #2578: [ML] Incorrect constraint applied to latitude for lat_long anomaly detection
We must constrain all latitudes to be in the range [-90, 90]. Generating values outside of this range will trigger an error when we try to index the result data. We should also reject input outside of this range and generate a log error.

valeriy42:
Fixes #2578


lib/model/unittest/CMetricModelTest.cc
```diff
@@ -2349,4 +2358,78 @@ BOOST_FIXTURE_TEST_CASE(testIgnoreSamplingGivenDetectionRules, CTestFixture) {
                         modelNoSkipTime);
 }
 
+class MyFakeModel : public ml::maths::common::CModelStub {
+public:
+    MyFakeModel(model_t::TDouble2Vec latLong)
+        : ml::maths::common::CModelStub(), m_LatLong(latLong) {}
```
tveasey, line 2364: nit:
```suggestion
        : m_LatLong(latLong) {}
```

include/model/CIndividualModel.h
```diff
@@ -25,6 +25,8 @@
 #include <utility>
 #include <vector>
 
+class CTestHelper;
```
tveasey, line 28: I don't think we should have something this specific in the global namespace called `CTestHelper`. I would move this into `ml::model` and probably call it `CIndividualModelTestHelper`.

tveasey (approved):
LGTM. (It's a shame we need to introduce `CTestHelper`, but I can't think of a better approach.)

include/model/CIndividualModel.h
valeriy42: Renamed

PR #2253, opened 2022-04-08
[ML] Upgrade to zlib 1.2.12 on Windows

droberts195:
Version 1.2.12 has these key improvements over 1.2.11:
- Fix a deflate bug when using the Z_FIXED strategy that can result
  in out-of-bound accesses.
- Fix a deflate bug when the window is full in deflate_stored().
- Speed up CRC-32 computations by a factor of 1.5 to 3.
- Use the hardware CRC-32 instruction on ARMv8 processors.
- Speed up crc32_combine() with powers of x tables.
- Add crc32_combine_gen() and crc32_combine_op() for fast combines.

For Linux and macOS we use the zlib provided by the OS install.
Users on Linux and macOS are advised to install the latest OS
updates if they wish to take advantage of these improvements.

PR #2588, opened 2023-10-25
[ML] Upgrade to Pytorch 2.1.2 and zlib 1.2.13

### LINKED ISSUE #2587: [ML] Upgrade to PyTorch 2.1
PyTorch 2.1 shipped a couple of weeks ago, and we should upgrade to it providing it does not introduce significant backwards compatibility issues.

Our last upgrade was to PyTorch 1.13.1, and #2430 should provide inspiration for how to upgrade to PyTorch 2.1.

PyTorch 2.1 supports Python 3.11 as a beta feature. However, it also still supports Python 3.10. We should be able to continue to use Python 3.10 during the PyTorch build process to minimise the amount of build system changes and hence expedite the upgrade.

edsavage:
Pytorch 2.1.2 has recently been released. Opening this PR to upgrade to it and potentially flush out any BWC issues via the unit and ES integration tests.

Closes #2587


3rd_party/licenses/pytorch-INFO.csv
```diff
@@ -1,2 +1,2 @@
 name,version,revision,url,license,copyright,sourceURL
-PyTorch,1.11.0,bc2c6edaf163b1a1330e37a6e34caf8c553e4755,https://pytorch.org,BSD-3-Clause,,
+PyTorch,2.1.0,bc2c6edaf163b1a1330e37a6e34caf8c553e4755,https://pytorch.org,BSD-3-Clause,,
```
droberts195, line 2: The Git commit (3rd value) should also be changed here.

build-setup/linux.md
```diff
@@ -389,32 +389,43 @@ Begin by cloning the IPEX repository:
...
+IPEX expects that PyTorch build directory contains a file `build-hash`, which contains the PyTorch git revision. This file is not created by the PyTorch build process, so we need to create it manually:
 ```bash
- echo "1.13.1+cpu\n" > ${PYTORCH_SRC_DIR}/torch/build-version
+ (cd ${PYTORCH_SRC_DIR}/torch/ && git rev-parse HEAD > build-hash)
 ```
 
 This assumes that you have cloned PyTorch in the directory `${PYTORCH_SRC_DIR}` in the step above. Make sure that this path is correct.
 
-Building IPEX requires a lot of memory. To reduce this requirement, we can patch the IPEX build system to lower the number of parallel processes. We use `sed` to replace the call to `multiprocessing.cpu_count()` with a constant `1` in the file `setup.py`:
+Building IPEX requires a lot of memory. To reduce this requirement, we can set the MAX_JOBS environment variable to lower the number of parallel processes:
```
droberts195, line 405:
```suggestion
Building IPEX requires a lot of memory. To reduce this requirement, we can set the `MAX_JOBS` environment variable to lower the number of parallel processes:
```

build-setup/linux.md
```diff
@@ -389,32 +389,43 @@ Begin by cloning the IPEX repository:
...
 ```bash
-sed -i 's/multiprocessing.cpu_count()/1/g' setup.py
+export MAX_JOBS=1
+```
+
+IPEX expects that the `blas-devel` library package be installed:
+```bash
+yum install blas-devel.x86_64
```
droberts195, line 412: This needs some due diligence. The library that gets installed here is `libblas.so.3`. Does IPEX end up requiring that at runtime? If so then that's a problem as we cannot guarantee that it will be present on end user systems.

Or if IPEX doesn't need to link `libblas` at runtime, why is it needed at build time? Are there environment variables that can be set to say not to use this?

build-setup/windows.md
```diff
@@ -65,25 +65,25 @@ As well as providing a Git implementation, Git for Windows comes with Windows po
...
+CMake version 3.19.2 is the minimum required to build ml-cpp. Download the MSI installer for version 3.23.3 from <https://github.com/Kitware/CMake/releases/download/v3.23.3/cmake-3.23.3-windows-x86_64.msi> (or get a more recent version).
 
 Install it mainly using the default options _except_ on the "Install Options" dialog check "Add CMake to the system PATH for all users".
 
 ### zlib
 
 Whilst it is possible to download a pre-built version of `zlib1.dll`, for consistency we want one that links against the Visual Studio 2019 C runtime library. Therefore it is necessary to build zlib from source.
 
-Download the source code from <http://zlib.net/> - the file is called `zlib1212.zip`. Unzip this file under `C:\tools`, so that you end up with a directory called `C:\tools\zlib-1.2.12`.
+Download the source code from <http://zlib.net/> - the file is called `zlib1213.zip`. Unzip this file under `C:\tools`, so that you end up with a directory called `C:\tools\zlib-1.2.13`.
```
droberts195, line 76: This means `3rd_party/licenses/zlib-INFO.csv` needs updating too.

dev-tools/docker/linux_image/Dockerfile
```diff
@@ -19,7 +19,7 @@ MAINTAINER David Roberts <dave.roberts@elastic.co>
 # libffi is required for building Python
 RUN \
   rm /var/lib/rpm/__db.* && \
-  yum install -y bzip2 gcc gcc-c++ git libffi-devel make texinfo unzip wget which xz zip zlib-devel
+  yum install -y bzip2 gcc gcc-c++ git libffi-devel make texinfo unzip wget which xz zip zlib-devel blas-devel.x86_64
```
droberts195, line 22: This list is no longer in alphabetical order.

But also, please investigate my other comment about whether this is really needed, if so why, and is it causing a runtime complication.

droberts195:
We usually release-note major library upgrades like this, so please add a release note like https://github.com/elastic/ml-cpp/pull/2430/files#diff-6eabb97e98640429cb68b61823f99438d287d604d72ecf0e93ed2e9ecbec7479R35 and https://github.com/elastic/ml-cpp/pull/2253/files#diff-6eabb97e98640429cb68b61823f99438d287d604d72ecf0e93ed2e9ecbec7479R37.

build-setup/linux.md
```diff
@@ -407,25 +407,23 @@ Building IPEX requires a lot of memory. To reduce this requirement, we can set t
 export MAX_JOBS=1
 ```
 
-IPEX expects that the `blas-devel` library package be installed:
-```bash
-yum install blas-devel.x86_64
-```
+The IPEX third party library dependency `LIBXSMM` link stage has a dependency on `libblas`. This dependency can be removed by setting `BLAS=0` in the environment.
```
droberts195, line 410: Please can you confirm with `ldd` that the `BLAS=0` workaround works. Also check if there are any other niche OS library dependencies of either IPEX or PyTorch 2.1.

I am a bit worried about the side effects of setting `BLAS=0` for `LIBXSMM`. It might mean that then certain operations of `LIBXSMM` simply don't work. We'd only detect this at runtime when a model that uses IPEX needs to do one of these operations. Our PR build tests are nowhere near up to the job of detecting subtle problems like this.

For 8.12 this doesn't actually matter, as we're not using IPEX. However, this is an extra piece of work to do before we ever do use in in the future.

This line in the `LIBXSMM` `Makefile` makes me wonder if it's possible to force the BLAS to always be MKL, and then we could set the build up like that: https://github.com/libxsmm/libxsmm/blob/54b8bb32042d204ff6fe550fa70d961f22c0b99e/Makefile#L246

Please can you add a TODO in this file to revisit `BLAS=0` before actually using IPEX in production, and also open an issue to investigate the best way to build IPEX that's both portable to all supported Linux distributions _and_ means that IPEX won't have any missing areas of functionality that cause runtime errors.

But don't try and solve this as part of this PR.

droberts195: Those dynamic dependencies you've listed look good.

The BLAS documentation for `LIBXSMM` isn't 100% clear to me, and then there's the extra complexity on top of when/how IPEX uses `LIBXSMM`.

Thanks for opening #2589. Nothing more is needed for this PR though.

edsavage:
buildkite build this please and thank you

edsavage:
buildkite build this

droberts195:
Looks good, but we need to run the QA suite against these changes before merging to `main`, as `main` could get promoted into serverless production immediately and we don't want to find out there are problems after this.

I think some manual compatibility testing with the current version of Eland should also be done before merging this. So build an Elasticsearch locally that incorporates a C++ build from this PR branch, then use the current version of Eland (which uses PyTorch 1.13.1) to download a model from HuggingFace, install it, and start it. Then perform at least one inference against that model. (The latest main branch snapshot of Kibana should be able to connect to the locally built Elasticsearch for this testing.)

Finally, with that same local install of Kibana and Elasticsearch, check that you can download and install ELSER v2 and do an inference using that.

edsavage:
> manual compatibility testing

Sanity checks were performed using `elastic/distilbert-base-uncased-finetuned-conll03-english` for named entity recognition and `typeform/distilbert-base-uncased-mnli` for zero shot classification. Both of these models were imported  from `huggingface` via the `eland_import_hub_model` script. No issues were encountered in either of these tasks.

Likewise, no issues were encountered downloading ELSER v2 and testing against it.

droberts195:
TODO: Set `QAF_TESTS_TO_RUN` to the appropriate value and then trigger the QA framework CI job.

droberts195:
The crashes we've observed with the Intel-optimised variant of ELSER are caused by IPEX. I have opened https://github.com/intel/intel-extension-for-pytorch/issues/484 for this. Depending on the outcome of that we will either have to upgrade to a patched version of IPEX in this PR, or remove it entirely.

docs/CHANGELOG.asciidoc
```diff
@@ -41,6 +43,8 @@
 === Enhancements
 
 * Upgrade Boost libraries to version 1.83. (See {ml-pull}2560[#2560].)
+* Improve forecasting for time series with step changes. (See {ml-pull}#2591[2591],
+  issue: {ml-issue}2466[#2466]).
```
droberts195, line 47: Please delete these two lines. They're duplicated about 10 lines below in the 8.11.2 section, and 8.11.2 seems to be correct. (I think this might be cause by one of my attempts to resolve a merge clash - sorry about that.)

droberts195 (approved):
LGTM if CI passes and if you could just adjust the release notes before merging - it's probably worth waiting for the current CI to finish before adjusting the release notes to expedite smoking out any other problems, then merge first thing tomorrow if all is good.

PR #2591, opened 2023-10-30
[ML] Improve forecasting for time series with step changes

### LINKED ISSUE #2466: Improve forecast with step changes
When creating forecasts for time series which have step changes we create a model of conditions under which we expect the time series to step, specifically the values at and interval between steps, based on historical data. This is a probabilistic model so we run a number of roll outs to estimate an expected value and distribution. We have seen this misbehaving when the forecast time series value is too far from the values for which we have a reasonable characterisation of this distribution. It would be more appropriate to be cautious in such cases: at the moment the behaviour depends on the choice of characterisation of distribution tail values. This issue covers the work to detect and avoid stepping the time series in such cases.

tveasey:
We model the level of a time series which we've observed having step discontinuities via a Markov process for forecasting. Specifically, we estimate the historical step size distribution and the distribution of the steps in time and as a function of the time series value. For this second part we use an online naive Bayes model to estimate the probability that at any given point in a roll out for forecasting we will get a step.

This approach generally works well unless we're in the tails of the distribution values we've observed for the time series historically when we roll out. In this case, our prediction probability are very sensitive to the tail behaviour of the distributions we fit to the time series values where we saw a step and sometimes we predict far too many steps as a result. We can detect this case: when we're in the tails of time series value distribution.

This change does this and stops predicting changes in such cases, which avoids pathologies. This fixes #2466.

### DIFF
--- a/include/maths/common/CNaiveBayes.h
+++ b/include/maths/common/CNaiveBayes.h
@@ -157,8 +157,10 @@ private:
 //! \brief Implements a Naive Bayes classifier.
 class MATHS_COMMON_EXPORT CNaiveBayes {
 public:
+    using TDoubleDoublePr = std::pair<double, double>;
     using TDoubleSizePr = std::pair<double, std::size_t>;
     using TDoubleSizePrVec = std::vector<TDoubleSizePr>;
+    using TDoubleSizePrVecDoublePr = std::pair<TDoubleSizePrVec, double>;
     using TDouble1Vec = core::CSmallVector<double, 1>;
     using TDouble1VecVec = std::vector<TDouble1Vec>;
     using TOptionalDouble = std::optional<double>;
@@ -183,6 +185,9 @@ public:
 
     //! Check if any training data has been added initialized.
     bool initialized() const;
+
+    //! Get the number of classes.
+    std::size_t numberClasses() const;
 
     //! This can be used to optionally seed the class counts
     //! with \p counts. These are added on to data class counts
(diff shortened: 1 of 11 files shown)


lib/maths/common/unittest/CNaiveBayesTest.cc
```diff
@@ -270,6 +281,40 @@ BOOST_AUTO_TEST_CASE(testPropagationByTime) {
     }
 }
 
+BOOST_AUTO_TEST_CASE(testExtrapolation) {
+    // Test that we detect when we're extrapolating the class conditional
+    // distributions to predict a class labels.
```
valeriy42, line 286: When reading through the test, it's difficult to see where exactly the extrapolation detection is asserted. Can you maybe make it more clear?

lib/maths/time_series/CTrendComponent.cc
```diff
@@ -52,6 +52,24 @@ const core_t::TTime UNSET_TIME{0};
...
+    void add(std::size_t class_, double logLikelihood) override {
+        if (class_ == NO_CHANGE_LABEL) {
+            m_LogLikelihood = logLikelihood;
+        }
+    }
+
+    double calculate() const override {
+        // Downweight features for which we don't have sufficient examples
+        // of the time series not changing.
+        return common::CTools::logisticFunction((4.5 + m_LogLikelihood) / 4.5, 0.1);
```
valeriy42, line 66: Maybe, you can add some more information abou the implementation decision here: where does 4.5 come from?

lib/maths/time_series/CTrendComponent.cc
```diff
@@ -786,14 +804,22 @@ CTrendComponent::TDouble3Vec
...
+            return weight;
+        };
         for (std::size_t i = 0; i < m_Levels.size(); ++i) {
             double dt{static_cast<double>(time - m_TimesOfLastChange[i])};
             double x{m_Levels[i] + prediction};
-            double p{m_Probability.classProbability(LEVEL_CHANGE_LABEL, {{dt}, {x}})};
-            m_ProbabilitiesOfChange[i] = std::max(m_ProbabilitiesOfChange[i], p);
+            auto[p, pConfidence] = m_Probability.classProbability(
+                LEVEL_CHANGE_LABEL, {{dt}, {x}}, weightProvider);
+            if (pConfidence > 0.5) {
```
valeriy42, line 820: Maybe you can add a commment here why do we need this check (for posterity).

valeriy42 (approved):
Good work on fixing the problem with fat tail distributions. I have just a couple of comments regarding readability.

include/maths/common/CNaiveBayes.h
```diff
@@ -210,27 +237,39 @@ class MATHS_COMMON_EXPORT CNaiveBayes {
     //!
     //! \param[in] n The number of class probabilities to estimate.
     //! \param[in] x The feature values.
+    //! \return The class probabilities and the minimum feature weight.
```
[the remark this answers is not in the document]
tveasey: Done.

lib/maths/common/CNaiveBayes.cc
```diff
@@ -288,62 +279,64 @@ void CNaiveBayes::propagateForwardsByTime(double time) {
...
         p.emplace_back(CTools::fastLog(class_.second.count()), class_.first);
     }
+    double confidence{1.0};
 
     TDoubleVec logLikelihoods;
     for (std::size_t i = 0; i < x.size(); ++i) {
-        if (x[i].size() > 0) {
-            TMaxAccumulator maxLogLikelihood;
+        if (x[i].empty() == false) {
+            auto& featureWeight = weightProvider();
```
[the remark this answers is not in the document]
tveasey: I added an explanation.

lib/maths/common/CNaiveBayes.cc
```diff
@@ -288,62 +279,64 @@ void CNaiveBayes::propagateForwardsByTime(double time) {
...
-                    (maxLogLikelihood[0] - *m_MinMaxLogLikelihoodToUseFeature) /
-                        std::fabs(*m_MinMaxLogLikelihoodToUseFeature),
-                    0.1);
-            }
+            double featureWeight_{featureWeight.calculate()};
             for (std::size_t j = 0; j < logLikelihoods.size(); ++j) {
-                p[j].first += weight * logLikelihoods[j];
+                p[j].first += featureWeight_ * logLikelihoods[j];
             }
+            confidence = std::min(confidence, featureWeight_);
```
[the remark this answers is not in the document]
tveasey: I didn't think of anything better than `featureWeight`: it is the weight we assign a specific feature in class selection (but I did add an explanation). I renamed to `minFeatureWeight`.

lib/maths/time_series/CTrendComponent.cc
tveasey: I added an explanation of the fact that 4.5 = 3 sigma point of the distribution.

lib/maths/time_series/CTrendComponent.cc
tveasey: Done.

lib/maths/common/unittest/CNaiveBayesTest.cc
tveasey: Oh actually this is a mistake. I was testing this in a different way in an earlier incarnation and this comment got left behind. I fixed.

yomduf:
Do you think it would be possible that Fix could be applied to v8.11.?

tveasey:
Done in #2597.

PR #2593, opened 2023-11-03
[7.17][ML] Improve forecasting for time series with step changes (#2591)

### LINKED ISSUE #2466: Improve forecast with step changes

tveasey:
We model the level of a time series which we've observed having step discontinuities via a Markov process for forecasting. Specifically, we estimate the historical step size distribution and the distribution of the steps in time and as a function of the time series value. For this second part we use an online naive Bayes model to estimate the probability that at any given point in a roll out for forecasting we will get a step.

This approach generally works well unless we're in the tails of the distribution values we've observed for the time series historically when we roll out. In this case, our prediction probability are very sensitive to the tail behaviour of the distributions we fit to the time series values where we saw a step and sometimes we predict far too many steps as a result. We can detect this case: when we're in the tails of time series value distribution.

This change does this and stops predicting changes in such cases, which avoids pathologies. This fixes #2466.

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,13 @@
 
 //=== Regressions
 
+== {es} version 7.17.15
+
+=== Enhancements
+
+* Improve forecasting for time series with step changes. (See {ml-pull}#2591[2591],
+  issue: {ml-issue}2466[#2466]).
+
 == {es} version 7.17.13
 
 === Enhancements
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,7 +28,7 @@
 
 //=== Regressions
 
-== {es} version 7.17.15
+== {es} version 7.17.16
 
 === Enhancements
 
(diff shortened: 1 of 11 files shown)


docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,13 @@
 
 //=== Regressions
 
+== {es} version 7.17.15
```
droberts195, line 31:
```suggestion
== {es} version 7.17.16
```

tveasey: Done

PR #2596, opened 2023-11-20
[ML] Set QAF_TESTS_TO_RUN to ml_cpp_pr

edsavage:
Set the BuildKite environment variable `QAF_TESTS_TO_RUN` to be `ml_cpp_pr` in the step that triggers QA tests to run when the PR in question is labelled with `ci:run-qa-tests`.

Which specific tests will be run may change over time.


droberts195 (approved):
LGTM

PR #2601, opened 2023-12-12
[ML] Use a custom checkout script in 7.17

droberts195:
Latest Buildkite has an optimised checkout, but it requires Git 2.5, and CentOS 6 uses 1.7.


droberts195:
This doesn't work because `post-checkout` is the earliest hook that can be run in the repository "context", and hence be defined in the repository itself.

PR #2602, opened 2023-12-13
[ML] Build Linux x86_64 image in VM

droberts195:
Up to now we've built the Linux x86_64 artifacts in a Docker container hosted on Kubernetes. This means the BuildKite agent needs to run in the build Docker image and use its Git to get the correct code to build. This no longer works with the ancient version of Git on CentOS 6. By building in a VM, BuildKite can run on the (modern) VM distribution and just the compiler needs to run in the ancient Docker image (which is needed for 7.x platform support).

### DIFF
--- a/.buildkite/pipelines/build_linux.json.py
+++ b/.buildkite/pipelines/build_linux.json.py
@@ -33,11 +33,12 @@ actions = [
 ]
 agents = {
    "x86_64": {
-      "cpu": "6",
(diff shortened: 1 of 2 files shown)


edsavage (approved):
LGTM

droberts195:
Java integration tests failed because ES is still on 7.17.16 and ML is on 7.17.17.

PR #2603, opened 2023-12-14
[ML] Another build fix for old OS used with 7.17

droberts195:
Ubuntu 16.04 can no longer clone the Eigen repo without this extra option.


edsavage (approved):
LGTM

PR #2604, opened 2023-12-22
[ML] Workaround issues with second inference when IPEX is linked

droberts195:
We have observed a couple of issues where IPEX is linked where the first inference call works but the second does not:

- https://github.com/intel/intel-extension-for-pytorch/issues/484 happens with ELSER and PyTorch 2.1
- https://github.com/elastic/elasticsearch/issues/102541 happens with the multilingual E5 base and large models and PyTorch 1.13.1

Disabling JIT profiling avoids the problems.

### DIFF
--- a/bin/pytorch_inference/CMakeLists.txt
+++ b/bin/pytorch_inference/CMakeLists.txt
@@ -27,6 +27,7 @@ if (LINK_IPEX)
     ${C10_LIB}
     ${IPEX_LIB} 
     )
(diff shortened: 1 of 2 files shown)


droberts195:
Important - before we merge this we need to check whether the fix of disabling JIT profiling with IPEX makes inference slower than enabling JIT profiling without IPEX. If it does then we should instead stop linking IPEX.

droberts195:
Superseded by #2605 - we decided it's better to remove IPEX altogether.

PR #2607, opened 2024-01-02
[ML] Better fallback when hardware_concurrency fails for pytorch inference

tveasey:
The advice is to treat this as a hint only. It will return 0 if it can't be computed for the platform and in this case we should size the thread pool for the requested number of allocations.

Also, we were never actually outputting how many allocations a `pytorch_inference` process has available so if `std::thread::hardware_concurrency()` fails then we don't get to find out.


droberts195 (approved):
LGTM

davidkyle (approved):
LGTM

The Java side is well informed about the number of threads that can be used so it does make sense to fall back to the num allocations if `std::thread::hardware_concurrency()` returns 0

PR #2611, opened 2024-01-10
[ML] Create identical dependency archives for identical contents

edsavage:
Ensure that the zip archive files created for the daily releasable artifacts are identical when the contents are identical. This means:

 * Set the timestamps of the contents to be a fixed value
 * Sort the files before archiving
 * Exclude any extra file attributes from the archive

Creating as a non-issue as this affects the build/CI processes only


droberts195 (approved):
LGTM

PR #2613, opened 2024-01-10
[ML] Run Java integration tests on both Intel and ARM

droberts195:
Currently the Java integration tests we run against PRs are only run on ARM. Since Intel is our primary platform for Cloud we should also run these tests on Intel.


edsavage (approved):
LGTM

PR #2614, opened 2024-01-17
[ML] Switch from RapidJSON to Boost.JSON

edsavage:
Migration from `RapidJSON` to `boost::json` .

For various reasons it is desirable to replace the use of the `RapidJSON` library in the `ml-cpp` codebase. The `boost::json` library has been selected as the replacement.

While the work is still not complete, opening this PR as a draft for visibility of the direction and progress.


lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -349,96 +395,133 @@ bool CJsonStateRestoreTraverser::SRapidJsonHandler::Bool(bool b) {
...
-                                                        rapidjson::SizeType length,
-                                                        bool) {
+bool CJsonStateRestoreTraverser::SBoostJsonHandler::on_key( std::string_view s, std::size_t n, json::error_code& ec ) {
     s_Type = E_TokenKey;
+    LOG_DEBUG(<< "on_key: '" << s_Name[s_NextIndex] << "', n: " << n << ", s_NextIndex: " << s_NextIndex);
+    if (ec) {
+        return false;
+    }
     if (s_RememberValue) {
-        s_NextIndex = 1 - s_NextIndex;
```
droberts195, line 427: It seems wrong that this has been removed.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -349,96 +395,133 @@ bool CJsonStateRestoreTraverser::SRapidJsonHandler::Bool(bool b) {
...
+    LOG_DEBUG(<< "on_object_end");
     s_Type = E_TokenObjectEnd;
+    if (ec) {
+        return false;
+    }
 
     if (s_RememberValue) {
         s_NextIndex = 1 - s_NextIndex;
-        s_Level[s_NextIndex] = s_Level[1 - s_NextIndex] - 1;
+        if (s_Level[1 - s_NextIndex] > 0) {
```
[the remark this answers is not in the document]
edsavage: When parsing character by character, as is the case here, results in the `on_string_part`, `on_key_part` handlers being called. This changes the semantics somewhat from the previous RapidJSON handlers.

If parsing is done with larger buffers the `*_part` handlers _should_ not be called and these changes (hopefully) won't be necessary.

include/core/CJsonStateRestoreTraverser.h
```diff
@@ -140,7 +294,12 @@ class CORE_EXPORT CJsonStateRestoreTraverser : public CStateRestoreTraverser {
             E_TokenObjectStart = 9,
             E_TokenObjectEnd = 10,
             E_TokenArrayStart = 11,
-            E_TokenArrayEnd = 12
+            E_TokenArrayEnd = 12,
+            E_TokenKeyPart = 13,
+            E_TokenStringPart = 14,
+            E_TokenComma,
+            E_TokenColon,
+            E_TokenWhiteSpace
```
droberts195, line 302: It should be possible to write the rest of the parser in such a way that these extra constants are not needed.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -223,15 +228,58 @@ const std::string& CJsonStateRestoreTraverser::nextValue() const {
...
+            break;
+        }
+
+        m_Handler.s_RememberValue = remember;
+
+        char c = '\0';
+        m_ReadStream.get(c);
+        if (c == '\0') {
+            break;
+        }
```
droberts195, line 246: For efficiency it would be best to read, say, 256 characters into a stack buffer and use `memchr` to check for zero bytes within it. Reading one character at a time is very inefficient.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -223,15 +228,58 @@ const std::string& CJsonStateRestoreTraverser::nextValue() const {
...
+            m_Handler.s_Type = SBoostJsonHandler::E_TokenKey;
+        }
+
+        if (c == ',') {
+            m_Handler.s_Type = SBoostJsonHandler::E_TokenComma;
+        }
+
+        if (c == ':') {
+            m_Handler.s_Type = SBoostJsonHandler::E_TokenColon;
+        }
```
droberts195, line 264: It shouldn't be necessary to peek the characters here _except_ checking for `\0` for end-of-document. Doing that would be more conducive to reading blocks of characters rather than one at a time.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -349,96 +395,133 @@ bool CJsonStateRestoreTraverser::SRapidJsonHandler::Bool(bool b) {
...
     return true;
 }
 
-bool CJsonStateRestoreTraverser::SRapidJsonHandler::Int64(std::int64_t i) {
-    s_Type = E_TokenInt64;
+bool CJsonStateRestoreTraverser::SBoostJsonHandler::on_double(double d, std::string_view/* s*/, json::error_code&/* ec*/) {
+    s_Type = E_TokenDouble;
     if (s_RememberValue) {
-        s_Value[s_NextIndex].assign(CStringUtils::typeToString(i));
+        s_Value[s_NextIndex].assign(CStringUtils::typeToString(d));
```
droberts195, line 419: Presumably in this case `s` already contains the string representation of the `double`, so round-tripping through an actual `double` variable is pointless. Could we just assign `s` here?

Same for other types of numbers.

According to https://www.boost.org/doc/libs/1_83_0/libs/json/doc/html/json/ref/boost__json__parse_options/numbers.html, if you use `number_precision::none` in the options then it just validates and doesn't actually parse the numbers, which should make parsing our number-heavy state _much_ faster.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -349,96 +395,133 @@ bool CJsonStateRestoreTraverser::SRapidJsonHandler::Bool(bool b) {
...
+
     if (s_RememberValue) {
-        ++s_Level[s_NextIndex];
-        s_Value[s_NextIndex].clear();
+        if (m_NewToken) {
+            m_NewToken = false;
+            s_NextIndex = 1 - s_NextIndex;
+            s_Name[s_NextIndex].clear();
+        }
+        s_Name[s_NextIndex].push_back(s.front());
```
droberts195, line 490: This would avoid propagating knowledge of how many characters were parsed in one go:

```suggestion
        s_Name[s_NextIndex].append(s);
```

Same in the other methods that are currently pushing `s.front()`.

lib/core/CJsonStateRestoreTraverser.cc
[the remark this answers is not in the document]
droberts195: > If parsing is done with larger buffers the `*_part` handlers should not be called and these changes (hopefully) won't be necessary.

Even if parsing is done with larger buffers the `*_part` handlers will _sometimes_ be called, if the end of a buffer happens to fall in the middle of a key name or value.

lib/api/CNdJsonInputParser.cc
```diff
@@ -111,24 +111,44 @@ bool CNdJsonInputParser::readStreamIntoVecs(const TVecReaderFunc& readerFunc,
     return true;
 }
 
-bool CNdJsonInputParser::parseDocument(char* begin, rapidjson::Document& document) {
-    // Parse JSON string using Rapidjson
-    if (document.ParseInsitu<rapidjson::kParseStopWhenDoneFlag>(begin).HasParseError()) {
-        LOG_ERROR(<< "JSON parse error: " << document.GetParseError());
+bool CNdJsonInputParser::parseDocument(char* begin, json::value& document) {
```
droberts195, line 114: You can easily add a `length` argument to this method, as `CNdInputParser::parseLine` returns both a start and a length. Only the start is currently passed, because that's all RapidJSON used. But by passing the length as well you can avoid the extra buffer and loop in this method. All of that can just be replaced with:

```
    p.write_some(begin, sizeof(length), ec)
```

where `begin` is the existing argument and `length` is the new one.

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -1308,19 +1364,29 @@ CBoostedTreeFactory CBoostedTreeFactory::constructFromDefinition(
 
     CBoostedTreeFactory factory{constructFromParameters(numberThreads, std::move(loss))};
 
+    auto is = validInputStream(dataSearcher);
```
droberts195, line 1367: Calling this only once instead of twice is skipping over this code (as well as breaching encapsulation by making assumptions about which implementation an interface points to):

https://github.com/elastic/ml-cpp/blob/e8a44b7c4fe79d1c5ea07c6114ce7c0a65a9164f/lib/api/CSingleStreamSearcher.cc#L41-L44

I think you should put this method back to how it was before.

The cause of the crash is probably different interpretation of the zero byte between the two JSON documents. We probably need to wrap up Boost.JSON in some way so that our wrapper treats zero bytes as document separators. Otherwise there will probably be several places that need changing in Java as well as in C++.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -240,19 +240,21 @@ bool CJsonStateRestoreTraverser::parseNext(bool remember) {
...
         m_Handler.s_RememberValue = remember;
 
-        char c = '\0';
-        m_ReadStream.get(c);
-        if (c == '\0') {
-            break;
+        if (m_BytesRemaining == 0) {
+            m_BytesRemaining = BUFFER_SIZE - 1;
+            ::memset(m_Buffer, '\0', BUFFER_SIZE);
+            m_ReadStream.get(m_Buffer, BUFFER_SIZE, '\0');
```
droberts195, line 246:
```suggestion
            ::memset(m_Buffer, '\0', BUFFER_SIZE);
            m_ReadStream.get(m_Buffer, BUFFER_SIZE, '\0');
            if (m_ReadStream.bad()) {
                // handle error
            }
            m_BytesRemaining = m_ReadStream.gcount();
```

What we're doing here is very similar to this: https://github.com/elastic/ml-cpp/blob/e7f6029489fc162b7461d30a26cc395dc82b895f/lib/api/CCsvInputParser.cc#L150-L168

That is tried and trusted code (used in the old Prelert Splunk app for years).

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -240,19 +240,21 @@ bool CJsonStateRestoreTraverser::parseNext(bool remember) {
...
+            m_BufferPtr = m_Buffer;
         }
 
         json::error_code ec;
-        m_Reader.write_some(true, &c, 1, ec);
+        std::size_t written = m_Reader.write_some(true, m_BufferPtr++, 1, ec);
```
droberts195, line 251: Does this work?

```suggestion
        std::size_t written = m_Reader.write_some(m_ReadStream.eof() == false, m_BufferPtr, m_BytesRemaining, ec);
        m_BufferPtr += written;
```

If it does it will be much more efficient.

lib/core/CJsonStateRestoreTraverser.cc
```diff
@@ -240,19 +240,21 @@ bool CJsonStateRestoreTraverser::parseNext(bool remember) {
...
         if (ec) {
             this->logError();
             ret = false;
             break;
         }
+        m_BytesRemaining -= written;
```
droberts195, line 257: By modifying `m_BytesRemaining` after the error handling but `m_BufferPtr` before, they could be out-of-sync next time the method is called. Is this a problem? Even if it's not now it might be better to modify them together to avoid pitfalls for future maintainers.

lib/core/CJsonStateRestoreTraverser.cc
edsavage: Sadly no. This results in `parseNext` not returning for each complete JSON token that has been processed. In this respect the `boost::json` SAX parser is different from the `RapidJSON` parser that supplies the `IterativeParseNext` function that parses one single JSON token from its input.

droberts195: Oh yes, I see. Multiple callbacks can get called but they'll just trample each other's changes _before_ `parseNext` returns.

But then it seems doubly important to keep `m_BufferPtr` and `m_RemainingBytes` perfectly in sync. So we should have `m_BufferPtr += written` and `m_RemainingBytes -= written` as adjacent lines immediately after the `write_some` call, and not use `m_BufferPtr++` before we know the result of `write_some`.

lib/api/CNdJsonInputParser.cc
edsavage: Got it. Thanks!

include/api/CSerializableToJson.h
```diff
@@ -135,73 +132,87 @@ class API_EXPORT CSerializableFromCompressedChunkedJson {
...
     }
 
-    static std::uint64_t getAsUint64From(const rapidjson::Value& value) {
-        if (value.IsUint64()) {
-            return value.GetUint64();
+    static std::int64_t getAsInt64From(const json::value& value) {
+        json::error_code ec;
+        std::int64_t ret = value.to_number<std::int64_t>(ec);
+        if (ec) {
+            throw std::runtime_error{"is not a uint64"};
```
droberts195, line 185:
```suggestion
            throw std::runtime_error{"is not an int64"};
```

include/core/CStreamWriter.h
```diff
@@ -0,0 +1,44 @@
...
+* or more contributor license agreements. Licensed under the Elastic License
+* 2.0 and the following additional limitation. Functionality enabled by the
+* files subject to the Elastic License 2.0 may only be used in production when
+* invoked by an Elasticsearch process with a license key installed that permits
+* use of machine learning features. You may not use this file except in
+* compliance with the Elastic License 2.0 and the foregoing additional
+* limitation.
+*/
+
+#ifndef ML_CSTREAMWRITER_H
```
droberts195, line 12: I think it's best to stick to a pattern across all files:

```suggestion
#ifndef INCLUDED_ml_core_CStreamWriter_h
```

include/core/CStringBufWriter.h
```diff
@@ -0,0 +1,38 @@
...
+* or more contributor license agreements. Licensed under the Elastic License
+* 2.0 and the following additional limitation. Functionality enabled by the
+* files subject to the Elastic License 2.0 may only be used in production when
+* invoked by an Elasticsearch process with a license key installed that permits
+* use of machine learning features. You may not use this file except in
+* compliance with the Elastic License 2.0 and the foregoing additional
+* limitation.
+*/
+
+#ifndef ML_CSTRINGBUFWRITER_H
```
droberts195, line 12:
```suggestion
#ifndef INCLUDED_ml_core_CStringBufWriter_h
```

cmake/variables.cmake
```diff
@@ -139,6 +139,7 @@ list(APPEND ML_SYSTEM_INCLUDE_DIRECTORIES
   ${TORCH_INC}
   ${CMAKE_SOURCE_DIR}/3rd_party/include
   ${CMAKE_SOURCE_DIR}/3rd_party/eigen
+  ${CMAKE_SOURCE_DIR}/3rd_party/valijson/include
```
droberts195, line 142: Do we really need this? Is it just used for one test? Could we just make that test do something a bit different? Our integration tests with Java should be pretty good at confirming we really do output valid JSON. If we don't _have_ to include another 3rd party dependency then it's best not to.

lib/maths/analytics/CDataFrameCategoryEncoder.cc
```diff
@@ -23,16 +23,24 @@
...
 
+namespace json = boost::json;
+
 namespace ml {
 namespace maths {
 namespace analytics {
 namespace {
+template<typename T>
+bool getEncodingAttribute(const json::object& obj, const std::string& tag, T& value) {
+    std::string tmp{obj.at(tag).as_string().c_str()};
```
droberts195, line 41: Why is this going via `c_str()`? Is it because we expect the string representation of `obj` to contain an embedded zero byte and we want to truncate at that point? Whatever the reason is, a comment to explain what's going on would be good.

edsavage: Replaced with the recommended method of converting `json::string` to `std::string`

3rd_party/pull-valijson.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Script to get the appropriate version of Valijson, if not already present.
+#
+# If updating this script ensure the license information is correct in the
+# licenses sub-directory.
+
+cd `dirname "$BASH_SOURCE"`
+
+if [ ! -d "valijson" ] ; then
+    git -c advice.detachedHead=false clone --depth=1 --branch=v1.0.2 --recurse-submodules git@github.com:tristanpenman/valijson.git
```
droberts195, line 21: `--recurse-submodules` here will pull in RapidJSON 🙀

edsavage: Arrrgh! 

I'll see what I can do to excise it from the dependencies.

include/core/CStateDecompressor.h
```diff
@@ -85,36 +86,241 @@ class CORE_EXPORT CStateDecompressor : public CDataSearcher {
...
-            bool StartObject();
-            bool Key(const char* str, rapidjson::SizeType length, bool);
-            bool EndObject(rapidjson::SizeType);
-            bool StartArray();
-            bool EndArray(rapidjson::SizeType);
+        //! <a https://www.boost.org/doc/libs/1_83_0/libs/json/doc/html/json/ref/boost__json__basic_parser.html#json.ref.boost__json__basic_parser.handler0">Handler</a>
+        //! for events fired by boost::json during parsing.
+        struct SBaseBoostJsonHandler {
+
+            /// The maximum number of elements allowed in an array
```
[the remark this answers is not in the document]
edsavage: Of course, it does look messy otherwise.

edsavage:

lib/api/CDataFrameAnalysisInstrumentation.cc
```diff
@@ -512,12 +512,12 @@ void CDataFrameTrainBoostedTreeInstrumentation::writeValidationLoss(json::object
     auto* writer = this->writer();
     if (writer != nullptr) {
         writer->addMember(VALIDATION_LOSS_TYPE_TAG, json::value(m_LossType), parentObject);
-        json::array lossValuesArray{writer->makeArray()};
+        json::array lossValuesArray = writer->makeArray();
```
droberts195, line 515: If the braces notation here compiled but produced incorrect results then it's worth adding a comment to say why. Otherwise someone might change it back in the future without realising the subtlety.

Same in the other 3 places.

lib/api/CInferenceModelDefinition.cc
```diff
@@ -24,17 +24,22 @@ namespace api {
 
 namespace {
 
-auto toRapidjsonValue(std::size_t value) {
-    return rapidjson::Value{static_cast<std::uint64_t>(value)};
+auto toBoostjsonValue(std::size_t value) {
+    //    uint64_t val = static_cast<std::uint64_t>(value);
+    //    json::value ret = val;
+    //    return ret;
```
droberts195, line 30: Please delete this redundant code.

lib/api/CNdJsonOutputWriter.cc
```diff
@@ -72,15 +70,14 @@ bool CNdJsonOutputWriter::writeRow(const TStrStrUMap& dataRowFields,
 
         this->writeField(name, value, doc);
     }
-
     m_Writer.write(doc);
-    m_Writer.Reset(m_WriteStream);
+    //    m_Writer.reset(m_OutStream);
```
droberts195, line 74: Delete if not needed.

lib/api/CRetrainableModelJsonReader.cc
```diff
@@ -22,15 +21,219 @@
...
+    struct handler {
+        static inline std::string IDENTITY_ENCODING_TAG = "identity_encoding";
+        static inline std::string ONE_HOT_ENCODING_TAG = "one_hot_encoding";
+        static inline std::string FREQUENCY_ENCODING_TAG = "frequency_encoding";
+        static inline std::string TARGET_MEAN_ENCODING_TAG = "target_mean_encoding";
+
+        constexpr static std::size_t max_object_size = std::size_t(-1);
+        constexpr static std::size_t max_array_size = std::size_t(-1);
+        constexpr static std::size_t max_key_size = std::size_t(-1);
+        constexpr static std::size_t max_string_size = std::size_t(-1);
```
droberts195, line 48: `std::numeric_limits<std::size_t>::max()` would be less hacky.

Alternatively, maybe we _should_ be setting some limits that are lower than the absolute max, yet still very large. For example, I don't think we'd ever expect a key longer than 1KB, or a string longer than 1GB. And I cannot imagine an object or array with more than a million elements. Any larger values would likely indicate corrupt input.

lib/api/CSerializableToJson.cc
```diff
@@ -12,40 +12,42 @@
...
-#include <rapidjson/rapidjson.h>
+#include <boost/json.hpp>
 
 #include <boost/iostreams/copy.hpp>
 #include <boost/iostreams/device/array.hpp>
 #include <boost/iostreams/filter/gzip.hpp>
 #include <boost/iostreams/filtering_stream.hpp>
 #include <boost/iostreams/stream.hpp>
 
+#include "core/CStreamWriter.h"
```
droberts195, line 25: Use angle brackets and move with the other `core` includes.

lib/api/CSerializableToJson.cc
```diff
@@ -12,40 +12,42 @@
...
+namespace json = boost::json;
 namespace ml {
 namespace api {
 namespace {
 namespace io = boost::iostreams;
 using TFilteredInput = io::filtering_stream<io::input>;
 using TFilteredOutput = io::filtering_stream<io::output>;
-using TGenericLineWriter = core::CRapidJsonLineWriter<rapidjson::OStreamWrapper>;
+using TStreamWriter = core::CStreamWriter;
+//using TGenericLineWriter = core::CBoostJsonLineWriter<std::ostream>;
```
droberts195, line 41: Delete unused code.

lib/api/CSerializableToJson.cc
```diff
@@ -71,8 +73,8 @@ const std::string JSON_EOS_TAG{"eos"};
 std::string CSerializableToJsonStream::jsonString() const {
     std::ostringstream jsonStream;
     {
-        rapidjson::OStreamWrapper osw{jsonStream};
-        TGenericLineWriter writer{osw};
+        std::ostringstream& osw = jsonStream;
```
droberts195, line 76: Why introduce another reference variable? Just use `jsonStream` in place of `osw`.

lib/api/CSerializableToJson.cc
```diff
@@ -129,18 +130,42 @@ CSerializableFromCompressedChunkedJson::rawJsonStream(const std::string& compres
                                                       TIStreamPtr inputStream,
                                                       std::iostream& buffer) {
     if (inputStream != nullptr) {
-        core::CRapidJsonUnbufferedIStreamWrapper isw{*inputStream};
+        //        core::CBoostJsonUnbufferedIStreamWrapper isw{*inputStream};
+        //        LOG_DEBUG(<< "inputStream: " << inputStream->rdbuf());
```
droberts195, line 134: Delete commented out code.

lib/api/CSerializableToJson.cc
```diff
@@ -129,18 +130,42 @@ CSerializableFromCompressedChunkedJson::rawJsonStream(const std::string& compres
...
+                assertIsJsonObject(doc);
+                LOG_DEBUG(<< "doc: " << doc);
+                try {
+                    auto chunk = ifExists(compressedDocTag, getAsObjectFrom,
+                                          doc.as_object());
+                    buffer.write(ifExists(payloadTag, getAsStringFrom, chunk),
+                                 ifExists(payloadTag, getStringLengthFrom, chunk));
+                    done = chunk.contains(JSON_EOS_TAG);
+                } catch (const std::runtime_error& e) {
+                    LOG_WARN(<< "Caught exception: " << e.what());
```
droberts195, line 165: This isn't great. It's catching and silently ignoring errors like `std::bad_alloc`. But I know it's done like this elsewhere in the code, so no need to fix in this PR. As a future project it would be good to review exception handling in the code. The style guide says not to use exceptions but everyone ignores that now. So it ought to be done with a bit more thought than just assuming all exceptions mean the same thing when we catch them.

lib/api/unittest/CAnomalyJobTest.cc
```diff
@@ -403,36 +404,36 @@ BOOST_AUTO_TEST_CASE(testOutputBucketResultsUntilGivenIncompleteInitialBucket) {
...
+    //        BOOST_REQUIRE_EQUAL(uint64_t(0), job.numRecordsHandled());
+    //
+    //        dataRows["."] = "f";
+    //        BOOST_TEST_REQUIRE(job.handleRecord(dataRows));
+    //        BOOST_REQUIRE_EQUAL(uint64_t(0), job.numRecordsHandled());
+    //
+    //        dataRows["."] = "f1";
+    //        BOOST_TEST_REQUIRE(job.handleRecord(dataRows));
+    //        BOOST_REQUIRE_EQUAL(uint64_t(0), job.numRecordsHandled());
+    //    }
```
droberts195, line 436: Why don't control messages work now?

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
```diff
@@ -129,244 +127,253 @@ BOOST_FIXTURE_TEST_CASE(testMemoryState, ml::test::CProgramCounterClearingFixtur
...
-                LOG_ERROR(<< "Invalid document: " << sb.GetString());
-                BOOST_FAIL("Schema validation failed");
-            }
-        }
-    }
-    BOOST_TEST_REQUIRE(hasOutlierDetectionStats);
-    BOOST_TEST_REQUIRE(initialMemoryReport);
-}
+//
+//BOOST_FIXTURE_TEST_CASE(testTrainingRegression, ml::test::CProgramCounterClearingFixture) {
```
droberts195, line 160: There are several commented out tests here.

lib/api/unittest/CDataFrameTrainBoostedTreeRegressionRunnerTest.cc
```diff
@@ -52,18 +52,24 @@ BOOST_AUTO_TEST_CASE(testPredictionFieldNameClash) {
...
+    BOOST_TEST_REQUIRE(jsonParameters.is_object());
+
     api::CDataFrameTrainBoostedTreeRegressionRunnerFactory factory;
     auto placeholder = factory.make(*spec, jsonParameters);
 
     BOOST_TEST_REQUIRE(errors.size() == 1);
     BOOST_TEST_REQUIRE(errors[0] == "Input error: prediction_field_name must not be equal to any of [is_training].");
 }
 
+//TODO
```
droberts195, line 72: What is this TODO for?

lib/core/CBoostJsonUnbufferedIStreamWrapper.cc
```diff
@@ -9,17 +9,17 @@
...
 
 namespace ml {
 namespace core {
 
-CRapidJsonUnbufferedIStreamWrapper::CRapidJsonUnbufferedIStreamWrapper(std::istream& strm)
+CBoostJsonUnbufferedIStreamWrapper::CBoostJsonUnbufferedIStreamWrapper(std::istream& strm)
     : m_Stream{strm} {};
 
-char CRapidJsonUnbufferedIStreamWrapper::Take() {
+char CBoostJsonUnbufferedIStreamWrapper::Take() {
```
droberts195, line 20: Another place where method names could be updated to comply with the coding standards in a followup.

lib/core/CJsonOutputStreamWrapper.cc
```diff
@@ -69,30 +70,31 @@ void CJsonOutputStreamWrapper::releaseBuffer(TGenericLineWriter& writer,
     }
 }
 
-void CJsonOutputStreamWrapper::flushBuffer(TGenericLineWriter& writer,
-                                           rapidjson::StringBuffer*& buffer) {
-    writer.Flush();
+void CJsonOutputStreamWrapper::flushBuffer(TGenericLineWriter& writer, std::string*& buffer) {
+    //    writer.flush();
```
droberts195, line 74: Is this really no longer needed?

lib/core/CJsonOutputStreamWrapper.cc
```diff
@@ -69,30 +70,31 @@ void CJsonOutputStreamWrapper::releaseBuffer(TGenericLineWriter& writer,
...
 
     m_ConcurrentOutputStream([this, buffer](std::ostream& o) {
+        std::string& str = *buffer;
         if (m_FirstObject) {
             m_FirstObject = false;
         } else {
             o.put(JSON_ARRAY_DELIMITER);
         }
-        o.write(buffer->GetString(), buffer->GetLength());
+        o.write(str.c_str(), str.size());
```
droberts195, line 83: Does `o << *buffer` work?

lib/core/CMakeLists.txt
```diff
@@ -61,8 +61,8 @@ ml_add_library(MlCore SHARED
   CProcessStats.cc
   CProgName.cc
   CProgramCounters.cc
-  CRapidJsonConcurrentLineWriter.cc
-  CRapidJsonUnbufferedIStreamWrapper.cc
+        CBoostJsonConcurrentLineWriter.cc
+        CBoostJsonUnbufferedIStreamWrapper.cc
```
droberts195, line 65: Please indent like the others and put in alphabetical order.

lib/core/CStateDecompressor.cc
```diff
@@ -13,14 +13,21 @@
...
-#include <rapidjson/error/en.h>
-
 #include <boost/iostreams/filter/gzip.hpp>
+// This file must be manually included when
+// using basic_parser to implement a parser.
+#include <boost/json/basic_parser_impl.hpp>
+
+// This file must be manually included when
+// using basic_parser to implement a parser.
+//#include <boost/json/basic_parser_impl.hpp>
```
droberts195, line 23: Delete redundant comment.

lib/core/CStateDecompressor.cc
```diff
@@ -98,21 +107,85 @@ std::streamsize CStateDecompressor::CDechunkFilter::read(char* s, std::streamsiz
...
-        return false;
-    }
+    bool ret{true};
+    SBoostJsonHandler::ETokenType currentTokenType = m_Reader->handler().s_Type;
+    do {
+        //        if (m_Reader->last_error()) {
+        //            LOG_ERROR(<< "Error parsing JSON");
+        //            ret = false;
+        //            break;
+        //        }
```
droberts195, line 117: Delete redundant code.

lib/core/unittest/CJsonStateRestoreTraverserTest.cc
```diff
@@ -11,10 +11,174 @@
...
+#include <boost/json.hpp>
 #include <boost/test/unit_test.hpp>
 
 #include <sstream>
 
+// This file must be manually included when
+// using basic_parser to implement a parser.
+#include <boost/json/basic_parser_impl.hpp>
+
+using namespace std;
```
droberts195, line 23: Historically we never use `using namespace std`.

lib/core/unittest/CJsonStateRestoreTraverserTest.cc
```diff
@@ -11,10 +11,174 @@
...
+            ec = error::extra_data;
+        }
+        if (ec) {
+            LOG_ERROR(<< "Json parsing failed: " << ec.message() << ": " << data);
+            return n;
+        }
+
+        //        auto const n = p_.write_some( false, data, size, ec );
+        //        if(! ec && n < size)
+        //            ec = error::extra_data;
```
droberts195, line 148: Delete redundant code.

lib/core/unittest/CMakeLists.txt
```diff
@@ -53,9 +53,9 @@ set (SRCS
   CProcessTest.cc
   CProgNameTest.cc
   CProgramCountersTest.cc
-  CRapidJsonLineWriterTest.cc
+        CBoostJsonLineWriterTest.cc
   CRapidJsonUnbufferedIStreamWrapperTest.cc
-  CRapidJsonWriterBaseTest.cc
+        CBoostJsonWriterBaseTest.cc
```
droberts195, line 58: Indent like the others and put in alphabetical order.

lib/core/unittest/CMemoryUsageJsonWriterTest.cc
```diff
@@ -41,7 +41,12 @@ BOOST_AUTO_TEST_CASE(test) {
         writer.addItem(description);
         writer.endObject();
         writer.finalise();
-        BOOST_REQUIRE_EQUAL(std::string("{\"Hello\":{\"memory\":223}}\n"), ss.str());
+        std::string expected = "{\"Hello\":{\"memory\":223}}\n";
+        std::string actual = ss.str();
+        LOG_INFO(<< "expected: " << expected);
+        LOG_INFO(<< "actual  : " << actual);
+        BOOST_REQUIRE_EQUAL(expected, actual);
+        //        BOOST_REQUIRE_EQUAL(std::string("{\"Hello\":{\"memory\":223}}\n"), ss.str());
```
droberts195, line 49: Delete commented out code.

lib/core/unittest/CPersistUtilsTest.cc
```diff
@@ -193,134 +195,134 @@ BOOST_AUTO_TEST_CASE(testPersistContainers) {
...
-
-        testPersistRestore(collection5);
-    }
-    LOG_DEBUG(<< "*** unordered_map ***");
-    {
-        TStrIntUMap map{
-            {"hello", 66}, {"thing", 999}, {"wanton destruction", 2}, {"0", 1}, {"hola", 43}};
-        testPersistRestore(map);
-    }
+    //    {
```
droberts195, line 198: Many commented out tests here - do they work now?

lib/maths/analytics/CBoostedTreeFactory.cc
```diff
@@ -130,6 +130,54 @@ double minBoundary(const CBoostedTreeHyperparameters::TDoubleParameter& paramete
     T minBoundary{maxBoundary - interval};
     return parameter.fromSearchValue(minBoundary);
 }
+
+template<typename streambuf>
+class peekbuf : public streambuf {
```
droberts195, line 135: It's mega confusing to have the template argument and the base class have the same name.

lib/model/CHierarchicalResultsNormalizer.cc
```diff
@@ -261,6 +261,7 @@ CHierarchicalResultsNormalizer::fromJsonStream(std::istream& inputStream) {
                 return E_Ok;
             }
         }
+        LOG_DEBUG(<< "name: " << traverser.name());
```
droberts195, line 264: Is this really going to be useful in production?

lib/test/CDataFrameAnalysisSpecificationFactory.cc
```diff
@@ -266,9 +266,10 @@ CDataFrameAnalysisSpecificationFactory::regressionLossFunctionParameter(double l
 
 std::string CDataFrameAnalysisSpecificationFactory::outlierParams() const {
 
-    rapidjson::StringBuffer parameters;
-    TRapidJsonLineWriter writer;
-    writer.Reset(parameters);
+    std::string parameters;
+    std::stringbuf sb;
+    std::ostream os(&sb);
```
droberts195, line 271: Is `std::ostringstream` simpler?

lib/test/CDataFrameAnalysisSpecificationFactory.cc
```diff
@@ -287,9 +288,11 @@ std::string CDataFrameAnalysisSpecificationFactory::outlierParams() const {
         writer.Double(0.0);
     }
     writer.EndObject();
-    writer.Flush();
+    writer.flush();
 
-    return parameters.GetString();
+    parameters = sb.str();
+
+    return parameters;
```
droberts195, line 295: `return os.str()` and get rid of the redundant `parameters` variable?

lib/test/CDataFrameAnalysisSpecificationFactory.cc
```diff
@@ -317,9 +320,10 @@ CDataFrameAnalysisSpecificationFactory::predictionParams(const std::string& anal
     using TClassificationRunner = api::CDataFrameTrainBoostedTreeClassifierRunner;
     using TRegressionRunner = api::CDataFrameTrainBoostedTreeRegressionRunner;
 
-    rapidjson::StringBuffer parameters;
-    TRapidJsonLineWriter writer;
-    writer.Reset(parameters);
+    std::string parameters;
+    std::stringbuf sb;
+    std::ostream os(&sb);
```
droberts195, line 325: As above with using `std::ostringstream` to replace all these.

lib/api/CRetrainableModelJsonReader.cc
edsavage: I'll switch to using `std::numeric_limits<std::size_t>::max()` for now and experiment with lower values in a follow up PR.

lib/api/unittest/CDataFrameAnalysisInstrumentationTest.cc
edsavage: These tests rely on a schema validator which, as previously discussed, was decided to be non-essential for the initial PR. I'll add an explanatory comment and come back to these time permitting.

lib/api/unittest/CDataFrameTrainBoostedTreeRegressionRunnerTest.cc
edsavage: It's redundant now. Removed.

lib/maths/analytics/CBoostedTreeFactory.cc
edsavage: Removed - not needed

lib/model/CHierarchicalResultsNormalizer.cc
edsavage: No, removed

lib/core/unittest/CPersistUtilsTest.cc
edsavage: Thankfully they do!

lib/core/CJsonOutputStreamWrapper.cc
edsavage: It does and is much cleaner, thanks.

lib/core/CJsonOutputStreamWrapper.cc
edsavage: re-instated.

include/core/CStreamWriter.h
```diff
@@ -0,0 +1,44 @@
...
+using TBoostJsonLineWriterStream = core::CBoostJsonLineWriter<std::ostream>;
+
+class CStreamWriter : public TBoostJsonLineWriterStream {
+public:
+    CStreamWriter(std::ostream& os) : TBoostJsonLineWriterStream(os) {}
+    CStreamWriter() : TBoostJsonLineWriterStream() {}
+
+    void append(const std::string_view& str) override {
+        (*m_Os) << str;
+        m_DebugBuffer.append(str);
```
droberts195, line 30: Are we using this class in production? If so, it seems like it will cause problems to build up the entire output in a string. It will be fine in tests but then will blow up at a customer with a 20GB model state.

Alternatively, if this class is _only_ for use in tests, please add a Doxygen comment to say this. Also, could it be moved to the `test` library if that's not too complicated?

include/core/CStateDecompressor.h
```diff
@@ -85,36 +86,225 @@ class CORE_EXPORT CStateDecompressor : public CDataSearcher {
...
+            static constexpr std::size_t max_array_size = -1;
+
+            //! The maximum number of elements allowed in an object
+            static constexpr std::size_t max_object_size = -1;
+
+            //! The maximum number of characters allowed in a string
+            static constexpr std::size_t max_string_size = -1;
+
+            //! The maximum number of characters allowed in a key
+            static constexpr std::size_t max_key_size = -1;
```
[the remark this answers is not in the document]
edsavage: I'll replace all with `std::numeric_limits<std::size_t>::max()` for now but have made a note to return to this (and elsewhere) to use better tailored values.

include/core/CStreamWriter.h
edsavage: Removed - it was only ever useful during development.

include/core/CStateDecompressor.h
```diff
@@ -85,36 +86,225 @@ class CORE_EXPORT CStateDecompressor : public CDataSearcher {
...
+                E_TokenColon = 11,
+                E_TokenSpace = 12
             };
 
             //! the last token type extracted
             ETokenType s_Type;
 
             //! the last string (c string) as pointer (only valid till next call)
-            const char* s_CompressedChunk;
+            char s_CompressedChunk[4096 * 400];
```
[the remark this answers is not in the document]
edsavage: The choice of the size of the buffer to hold the compressed chunk was arbitrary - `4096 * 400` just seemed "big enough" without being overly so. The values don't appear as magic numbers elsewhere.

include/core/CStateDecompressor.h
```diff
@@ -11,17 +11,21 @@
...
+#include <boost/json.hpp>
 
-#include <rapidjson/reader.h>
+namespace json = boost::json;
 
 namespace ml {
 namespace core {
 
+//! Maximum size of the buffer (in bytes) used to hold a compressed chunk
+static const int COMPRESSED_CHUNK_BUFFER_LENGTH = 1638400;
```
droberts195, line 27:
```suggestion
static const std::size_t COMPRESSED_CHUNK_BUFFER_LENGTH = 1638400;
```

Also, should it be a class member rather than loose in the namespace?

include/core/CJsonStateRestoreTraverser.h
```diff
@@ -107,40 +112,183 @@ class CORE_EXPORT CJsonStateRestoreTraverser : public CStateRestoreTraverser {
...
+        static constexpr std::size_t max_array_size = -1;
+
+        //! The maximum number of elements allowed in an object
+        static constexpr std::size_t max_object_size = -1;
+
+        //! The maximum number of characters allowed in a string
+        static constexpr std::size_t max_string_size = -1;
+
+        //! The maximum number of characters allowed in a key
+        static constexpr std::size_t max_key_size = -1;
```
droberts195, line 130: More places where `std::numeric_limits<std::size_t>::max()` would be better.

lib/core/CJsonOutputStreamWrapper.cc
```diff
@@ -42,25 +42,26 @@ CJsonOutputStreamWrapper::~CJsonOutputStreamWrapper() {
...
-                                             rapidjson::StringBuffer* buffer) {
+void CJsonOutputStreamWrapper::releaseBuffer(TGenericLineWriter& writer, std::string* buffer) {
     writer.Flush();
 
     // check for data that has to be written
-    if (buffer->GetLength() > 0) {
+    if (buffer->size() > 0) {
         m_ConcurrentOutputStream([this, buffer](std::ostream& o) {
+            std::string& str = *buffer;
+
```
droberts195, line 57: I think these lines are unused now.

include/api/CDataFrameAnalysisSpecificationJsonWriter.h
```diff
@@ -12,8 +12,9 @@
 #ifndef INCLUDED_ml_api_CDataFrameAnalysisSpecificationJsonWriter_h
 #define INCLUDED_ml_api_CDataFrameAnalysisSpecificationJsonWriter_h
 
+#include "core/CStreamWriter.h"
```
droberts195, line 15:
```suggestion
#include <core/CStreamWriter.h>
```

include/core/CBoostJsonPoolAllocator.h
```diff
@@ -33,52 +34,52 @@ namespace core {
...
+    void clear() { m_JsonPoolAllocator.release(); }
 
     //! \return document pointer suitable for storing in a container
     //! Note: The API is designed to emphasise that the client does not own the document memory
     //! i.e. The document will be invalidated on destruction of this allocator
     TDocumentWeakPtr makeStorableDoc() {
-        TDocumentPtr newDoc = std::make_shared<rapidjson::Document>(&m_JsonPoolAllocator);
-        newDoc->SetObject();
+        // TODO sort out the allocator pool?
+        //        TDocumentPtr newDoc = std::make_shared<boost::json::object>(&m_JsonPoolAllocator);
```
droberts195, line 55: Please can you open an issue listing the nice-to-have followups to this PR. This is one. Changing the method naming to match our style guide is another. There may be some others from previous reviews.

droberts195 (approved):
LGTM if you could just fix a couple of final nits.

Please merge this on Monday after first merging latest main into the PR branch and ensuring CI is still green.

include/core/CBoostJsonPoolAllocator.h
edsavage: Will do. Thanks Dave.

PR #2615, opened 2024-01-17
Update QA tests pipeline slug

dolaru:
The target QA tests pipeline has been renamed to solve an issue in CI, and the pipeline slug has changed in the process.

This updates the `run_qa_tests.yml.sh` pipeline generator file with the new slug.


droberts195 (approved):
LGTM

edsavage (approved):
LGTM2

PR #2618, opened 2024-01-29
[ML] Fix incorrect auto type

droberts195:
This is something that was flagged as a compiler warning.

There was a test where an `auto` variable resolved to an `int`, but it really should be 64 bit. It's being used to store a time in epoch seconds, so we get away with this type of error in general until 2038. In the exact test where the problem occurred it will never be a problem as the times being used are very small. However, it's not good practice to use a technically incorrect pattern that may get copied to other places where it matters.


edsavage (approved):
LGTM

PR #2628, opened 2024-02-15
[backport] Remove debug info log

maxhniebergall:
backport of https://github.com/elastic/ml-cpp/pull/2627

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 org.gradle.daemon=false
 
-elasticsearchVersion=8.13.0
+elasticsearchVersion=8.14.0
 
 artifactName=ml-cpp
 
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -1,6 +1,6 @@
 org.gradle.daemon=false
 
-elasticsearchVersion=8.13.0
+elasticsearchVersion=8.14.0
```
droberts195, line 3: This file must not be in this PR. It would set the 8.13 branch to the wrong version.

droberts195 (changes requested):
You've added 2 extra files to the backport, one of which doesn't matter but the other one would ruin the 8.13 branch.

gradle.properties
maxhniebergall: yikes, I should have looked into why Jan's commit showed up here. My bad.

PR #2630, opened 2024-02-15
[ML] Actually delete RapidJSON from the repo

droberts195:
Our dependency on RapidJSON was removed in #2614, but it wasn't actually deleted from the repo.

This PR does that, and also adds valijson to the .gitignore file so that it doesn't clutter the "git status" output.


przemekwitek (approved):
LGTM

PR #2634, opened 2024-02-19
[ML] RapidJSON removal followup - remove license for nested msinttypes

droberts195:
The msinttypes headers were a sub-component of RapidJSON. Now that we have removed RapidJSON there is no need to mention the msinttypes license, as the code was not used anywhere else in the ML C++ codebase.


davidkyle (approved):
LGTM

PR #2641, opened 2024-03-17
[ML] Set buildkite_branch_name_separator to "+"

edsavage:
To work around a problem with BuildKite PR builds on bare metal machines (currently only macOS aarch64 builds), set the value of the character separating username and branchname to be "+" rather than the default ":"

Marking as `>non-issue` as this only affects the internal build system and is not worthy of a release note.

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -13,7 +13,8 @@
       "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))",
       "skip_ci_labels": ["skip-ci", "jenkins-ci", ">test-mute", ">docs"],
       "skip_target_branches": ["6.8", "7.11", "7.12"],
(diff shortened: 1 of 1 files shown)


edsavage:
buildkite build this

droberts195 (approved):
LGTM if it fixes the problem 🤞

droberts195:
I see that the PR CI failed with the same error:

> 2024-03-17 15:15:17 ERROR  Could not obtain OIDC token for default audience
>🚨 Error: Error setting up job executor: The global environment hook exited with status 1

Is the problem that the PR bot is always getting its config from `main`, so the change of this PR won't be picked up until after it's merged? If so, let's just merge this PR with red CI and hope the change works once it's on `main`.

jonahbull:
> Is the problem that the PR bot is always getting its config from main, so the change of this PR won't be picked up until after it's merged?

@droberts195 Yes, that's correct

PR #2643, opened 2024-03-18
[8.12][ML] Set buildkite_branch_name_separator to "+" (#2641)

edsavage:
To work around a problem with BuildKite PR builds on bare metal machines (currently only macOS aarch64 builds), set the value of the character separating username and branchname to be "+" rather than the default ":"

Backports #2641


edsavage:
buildkite build this

PR #2649, opened 2024-04-03
[ML] Do not remove eigen directory as part of "clean" step

edsavage:
The scripts that clone the eigen repo already have logic to remove an existing directory structure prior to cloning, and removing it as part of the clean step forces a cmake reconfigure before subsequent builds will succeed (as the eigen headers will be missing)

Labelling as `>non-issue` as this only affects the build scripts.


jan-elastic:
LGTM

PR #2652, opened 2024-04-10
[ML] Remove CStringStore

jan-elastic:
It's probably most convenient to review this commit by commit. All commits are self-contained and the tests pass after each commit.


lib/model/CSearchKey.cc
```diff
@@ -60,20 +59,17 @@ CSearchKey::CSearchKey(int detectorIndex,
...
                        bool useNull,
                        model_t::EExcludeFrequent excludeFrequent,
-                       std::string fieldName,
-                       std::string byFieldName,
-                       std::string overFieldName,
-                       std::string partitionFieldName,
+                       const std::string& fieldName,
+                       const std::string& byFieldName,
+                       const std::string& overFieldName,
+                       const std::string& partitionFieldName,
```
tveasey, line 65: My inclination would be to pass by value and move into place, this is more idiomatic, and allows us to potentially avoid a copy, although probably not given how we use this object.

tveasey (commented):
Overall looks good so far (and thanks for assorted clean ups!).

I think my main comment would be let's try and move away from `optional<std::string>` where we can. I think the only real use for this is in the hierarchical results.

include/core/CStoredStringPtr.h
```diff
@@ -16,102 +16,14 @@
...
-//! Hash function named such that it will work automatically with Boost
-//! unordered containers.
-CORE_EXPORT
-std::size_t hash_value(const CStoredStringPtr& ptr);
-
-//! Swap for use by generic code.
-CORE_EXPORT
-void swap(CStoredStringPtr& lhs, CStoredStringPtr& rhs);
+// TODO(jan): rename
+using CStoredStringPtr = std::optional<std::string>;
```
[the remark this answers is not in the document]
jan-elastic: I did so. I think I've captured everything that can be null in an `optional<string>` and most things that cannot be null in `string` now. I tried to err on the safe side, so there probably are optionals in the code right now that are always set.

lib/model/CSearchKey.cc
jan-elastic: I agree

include/core/CStoredStringPtr.h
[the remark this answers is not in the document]
jan-elastic: @tveasey If you see `TOptionalStr`s that you think can be just `string`s, please let me know

dev-tools/clang-format.sh
```diff
@@ -23,7 +23,7 @@ if [ $? != 0 ] ; then
 fi
 
 REQUIRED_CLANG_FORMAT_VERSION=5.0.1
-FOUND_CLANG_FORMAT_VERSION=$(expr "`clang-format --version`" : ".* \([0-9].[0-9].[0-9]\)")
+FOUND_CLANG_FORMAT_VERSION=$(expr "`clang-format --version`" : ".* \([0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}\)")
```
jan-elastic, line 26: To match version `10.0.0` and newer.

docs/CHANGELOG.asciidoc
```diff
@@ -28,6 +28,12 @@
 
 //=== Regressions
 
+== {es} version 8.14.0
+
+=== Bug Fixes
+
+* Remove CStringStore. (See {ml-pull}2652[#2652].)
```
jan-elastic, line 35: @tveasey What's a sensible changelog comment for this change?

lib/api/unittest/CAnomalyJobLimitTest.cc
```diff
@@ -481,7 +481,7 @@ BOOST_AUTO_TEST_CASE(testModelledEntityCountForFixedMemoryLimit) {
             LOG_DEBUG(<< "Memory usage = " << used.s_Usage);
             LOG_DEBUG(<< "Memory limit bytes = " << memoryLimit * 1024 * 1024);
             BOOST_TEST_REQUIRE(used.s_OverFields > testParam.s_ExpectedOverFields);
-            BOOST_TEST_REQUIRE(used.s_OverFields < 7100);
+            BOOST_TEST_REQUIRE(used.s_OverFields < 8700);
```
jan-elastic, line 484: @tveasey Does this change sound reasonable to you?

jan-elastic:
@tveasey What end-to-end testing do you think is good to do to verify these changes?

include/core/CStoredStringPtr.h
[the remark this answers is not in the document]
tveasey: My inclination is you've done a reasonable effort first pass. I do suspect we may have some unnecessary nullable strings in a few places, but equally nothing you've left is trivial to unpick. I would be keen to get this through the overnights since it already simplifies a lot of code.

tveasey:
> @tveasey What end-to-end testing do you think is good to do to verify these changes?

Let's first see if any results change in the overnights (this should be a no op in terms of behaviour). This should also give us visibility of any performance degradations.

Once we are happy with that, we can try a local test with one of the overnights with many distinct partitions + over field values. I would test:
1. Memory is controlled effectively (also check the running process memory monitored independently) for different settings of limit
2. We use broadly similar memory (of course it doesn't matter if we actually use less)

docs/CHANGELOG.asciidoc
jan-elastic: Thanks, SGTM

wwang500:
buildkite run_qa_tests

wwang500:
buildkite run_qa_tests

wwang500:
buildkite run_qa_tests

Issue #2653, opened 2024-04-11
[ML] Windows build warnings


edsavage:
Examination of the Windows build logs shows a number of instances of the following warnings.

- [x] C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\bin\pytorch_inference\CCommandParser.cc(26,68): warning C4273: std::ostream& operator<<(std::ostream& os, const json::value& doc) { [C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\cmake-build-relwithdebinfo\bin\pytorch_inference\Mlpytorch_inference.vcxproj] - #2654
- [x]     12>C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\lib\model\CTokenListDataCategorizerBase.cc(429,1): warning C4573: the usage of 'ml::model::CTokenListDataCategorizerBase::acceptPersistInserter' requires the compiler to capture 'this' but the current default capture mode does not allow it [C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\cmake-build-relwithdebinfo\lib\model\MlModel.vcxproj] - #2655
- [x]     13>zlib.lib(deflate.obj) : warning LNK4099: PDB 'zlib.pdb' was not found with 'zlib.lib(deflate.obj)' or at 'C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\cmake-build-relwithdebinfo\lib\core\RelWithDebInfo\zlib.pdb'; linking object as if no debug info [C:\buildkite-agent\builds\bk-agent-prod-gcp-1712709148528061232\elastic\ml-cpp-snapshot-builds\cmake-build-relwithdebinfo\lib\core\MlCore.vcxproj] - #2659 
[...]

PR #2655, opened 2024-04-12
[ML] Fix lambda capture triggering warning on Windows

edsavage:
Windows build logs contain the warning

```
lib\model\CTokenListDataCategorizerBase.cc(429,1): warning C4573: the usage of 'ml::model::CTokenListDataCategorizerBase::acceptPersistInserter' requires the compiler to capture 'this' but the current default capture mode does not allow it
```

This PR fixes that by explicitly capturing the "this" pointer.

Relates #2653


edsavage:
Annoyingly this now produces warnings on other platforms that the "this" pointer capture is not needed.
Another way must be found.

edsavage:
buildkite build this

PR #2656, opened 2024-04-16
[ML] Suppress LNK4217 linker warning for libMlCore

edsavage:
The link step for libMlCore generates many [LNK4217](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4217?view=msvc-170) warnings referencing `Boost::Json` symbols. The warning is indicating that the definitions of the symbols have already been imported in other translation units in the library.

These warnings are harmless and are safe to suppress (and are not avoidable by any other means as far as I can tell)

Relates #2653

### DIFF
--- a/lib/core/CMakeLists.txt
+++ b/lib/core/CMakeLists.txt
@@ -112,7 +112,7 @@ ml_add_library(MlCore SHARED
   )
 
 if (WIN32)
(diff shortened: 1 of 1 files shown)


jan-elastic (approved):
I'm fine with merging this, but to be honest, suppressing warnings doesn't seem the optimal solution. I don't know enough about the root cause and/or Windows builds to suggest a better solution though.

PR #2662, opened 2024-05-09
[ML] Use Orka for macOS ARM builds.

edsavage:
At the moment macOS ARM builds are performed on bare metal Mac Minis provided by MacStadium. For consistency with the rest of engineering, ease of maintenance, reproducibility of builds, automation of provisioning builders etc. it is preferred that we migrate to using VMs provided by MacStadium Orka.

This PR makes use of an image built using "packer". Where the packer config will live is still TBD.

Labelling as a non-issue as this is purely build related.


jan-elastic (approved):
LGTM

PR #2670, opened 2024-05-22
script to continuously evaluate elser

davidkyle:
First download the elser model locally. Either

- https://ml-models.elastic.co/elser_model_2.pt
- https://ml-models.elastic.co/elser_model_2_linux-x86_64.pt

The script runs pytorch_inference, loads the model then continuously runs inference on it. Logging is to std out, the model output is written to a json file. Every 100 request the script asks the pytorch_inference how much memory it is using and this is written to the same json file. `grep mem out.json` will show that data.

Run with 
```
python3 signal9.py '/PATH/TO/elser_2/elser_model_2.pt' --num_allocations=4
``` 

`--num_threads_per_allocation` and `--num_allocations` are the parameters to tweak. Increasing either of those will make inference faster and changes in memory should be seen sooner.

### DIFF
--- a/bin/pytorch_inference/signal9.py
+++ b/bin/pytorch_inference/signal9.py
@@ -1,0 +1,181 @@
+import argparse
+import json
+import os
+import platform
+import random
+import stat
+import subprocess
+import time
+
+
+#
+# python3 signal9.py '/Users/davidkyle/Development/NLP Models/elser_2/elser_model_2.pt' --num_allocations=4
+#
+
+def parse_arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('model', help='A TorchScript model with .pt extension')
+    # The pipes are created by the C++ app
+    # The names must match those passed to app
+    parser.add_argument('--output_file', default='out.json')
+    parser.add_argument('--log_file', default='log.txt')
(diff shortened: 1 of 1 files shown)


bin/pytorch_inference/signal9.py
```diff
@@ -0,0 +1,181 @@
...
+
+def parse_arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('model', help='A TorchScript model with .pt extension')
+    # The pipes are created by the C++ app
+    # The names must match those passed to app
+    parser.add_argument('--output_file', default='out.json')
+    parser.add_argument('--log_file', default='log.txt')
+    parser.add_argument('--num_threads_per_allocation', type=int, help='The number of inference threads used by LibTorch. Defaults to 1.')
+    parser.add_argument('--num_allocations', type=int, help='The number of allocations for parallel forwarding. Defaults to 1')
```
jonathan-buttner, line 23: We should probably include setting the cache size here. Does it default to 0 if `--cacheMemorylimitBytes` isn't passed?

edsavage: > Does it default to 0 if --cacheMemorylimitBytes isn't passed?

Yes, the default cache value is `0`:

```
  --cacheMemorylimitBytes arg           Optional memory in bytes that the 
                                        inference cache can use - default is 0 
                                        which disables caching
```

PR #2672, opened 2024-05-26
[ML] Add CMake config for Address Sanitizer builds

edsavage:
Add configuration for compiling with Address Sanitizer flags set for Linux and Windows (they're already present for macOS).

To use:

 ## Linux
 * cmake -B cmake-build-sanitizer -DCMAKE_BUILD_TYPE="Sanitizer"
 * cmake --build cmake-build-sanitizer -j\`nproc\` -t install

 ## Windows
 * cmake -B cmake-build
 * cmake --build cmake-build --config Sanitizer -j\`nproc\` -t install

In addition, certain flags may want to be set via the environment variable `ASAN_OPTIONS`, see
https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags for more details.

### DIFF
--- a/cmake/variables.cmake
+++ b/cmake/variables.cmake
@@ -155,17 +155,27 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
   list(APPEND ML_COMPILE_DEFINITIONS BOOST_ALL_NO_LIB)
 endif()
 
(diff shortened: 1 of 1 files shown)


valeriy42 (approved):
LGTM

PR #2674, opened 2024-05-28
[ML] Use custom Boost::JSON allocator

edsavage:
The current code uses the monotonic resource allocator, for allocating memory to `boost::json` objects, which allocates memory in ever increasing chunks, which can lead to over allocation. The image below shows a typical series of memory allocations when using the monotonic resource allocator

Screenshot 2024-05-29 at 15 07 42

The other disadvantage of the monotonic resource allocator is that no deallocations are performed until the resource allocator is destroyed - hence the name `monotonic` as resource allocations can only increase during its lifetime.

These factors make the choice of the monotonic resource allocator unsuitable for its current use.

This PR introduces a very simplistic custom allocator that allocates and deallocates individual objects upon request using standard `operator` `::new` and `::delete`. This gives a much better experience as only as much memory is allocated at any point in time as absolutely needs to be, and gives a much more predictable memory profile

Screenshot 2024-05-29 at 15 06 02

On small data sets this change appears performant, but I do think it would be wise to run the QA tests against this PR, before merging.


edsavage:
buildkite run_qa_tests

wwang500:
buildkite run_qa_tests

include/api/CJsonOutputWriter.h
```diff
@@ -238,6 +238,8 @@ class API_EXPORT CJsonOutputWriter {
     //! from the CResourceMonitor via a callback
     void reportMemoryUsage(const model::CResourceMonitor::SModelSizeStats& modelSizeStats);
 
+    std::size_t getAllocatorMemUsage() const;
```
valeriy42, line 241: Can you please add a short documentation?

valeriy42 (commented):
Good work on figuring out and changing the allocator behavior.  I think we need to report the allocator memory usage as part of the memory_stats since it may be considerable.

When considering whether or not to update the model, we use the following code:
```
std::size_t CResourceMonitor::allocationLimit() const {
    return this->highLimit() - std::min(this->highLimit(), this->totalMemory());
}
```

Hence, we need to be transparent why the job gets into the hard_limit state while model_bytes  is lower.

valeriy42:
> > report the allocator memory usage as part of the memory_stats
> 
> Just to clarify @valeriy42 , by `memory_stats` do you mean the `Model size stats` (as reported in the `counts` tab in the AD job results in Kibana)? i.e. include the JSON allocator mem usage in `model::CResourceMonitor::SModelSizeStats`?

Exactly. Sorry for mixing up `memory_stats` and `model_size_stats`.

edsavage:
buildkite build this

lib/api/CModelSizeStatsJsonWriter.cc
```diff
@@ -45,6 +45,7 @@ const std::string LOG_TIME{"log_time"};
 const std::string CATEGORIZER_STATS{"categorizer_stats"};
 const std::string PARTITION_FIELD_NAME{"partition_field_name"};
 const std::string PARTITION_FIELD_VALUE{"partition_field_value"};
+const std::string JSON_MEMORY_ALLOCATOR_BYTES("json_memory_allocator_bytes");
```
valeriy42, line 48: nit: I am wondering if this name is not too specific. Maybe `output_memory_allocator_bytes`?

valeriy42 (approved):
LGTM.

DaveCTurner:
Sorry to say that https://github.com/elastic/elasticsearch/pull/109833 thoroughly breaks the ES wire protocol, I'm going to have to revert it to fix the ES build. I guess that means something needs to be reverted here too, but I'm not qualified to address that.

Issue #2664, opened 2024-05-15
[ML] Investigate/fix reported memory usage


jan-elastic:
Anomaly detection uses more memory than is reported by the system (at least on MacOS).

E.g. for a job from the QA test suite, the anomaly detection system reports:

```
        "model_bytes": 257363572,
        "peak_model_bytes": 257363572,
```

while MacOS reports (via activity monitor / top / ps) that the autodetect process uses up to approximately 1GB. 

Additionally, ECK shows a gradual increase in memory consumed by pods containing the autodetect process.

![Screenshot from 2024-06-04 12-51-40](https://github.com/elastic/ml-cpp/assets/1223797/500d0c95-76f5-4d2a-ac19-21fa7ebd3dde)

PR #2676, opened 2024-06-03
[ML] Release allocator cache

valeriy42:
This PR regularly removes allocators from the pool. This ensures that memory is released and reduces the job's memory consumption. 

Related to #2664


valeriy42:
test

valeriy42:
Merge cleaned up version in #2679

Issue #1386, opened 2020-07-06
[ML] Improve handling of bucket count variation for mean value anomaly detection


tveasey:
Currently, we use a worst case estimate of the impact of changes in the count of values in a bucket on their mean variance. This is safe in the sense of not generating false positives, but can lead to large changes in the model plot bounds and potentially false negatives when the count of values in the bucket is low.

Specifically, we assume all measurements are independent so that the variance of the mean statistic will be proportional to 1 / "number samples" in the bucket. If the rate of values is highly variable this can lead to large increases in the width of the model bounds when the count is low. Unfortunately, this is not calibrated to the actual data behaviour. For example, in the other extreme, if all measurements in each bucket were perfectly correlated then we would get no change in variation of the mean statistic as a function of bucket count.

It would be possible to estimate the relationship between the bucket count and the sufficient statistics related to data variation for all the residual distributions we fit since we know the sample count for each bucket. This would also be a more accurate way of calibrating heavy tailed distributions like the log-normal to observed changes in the seasonal variation.

A computationally feasible formulation would be to use linear regression. For example, for the normal model we could fit the linear model `(x_i - m)^2 = [c_i s_i] [p_1 p_2]^t` for parameters `p_1` and `p_2`, observed bucket values `x_i`, mean of `x_i` `m` and bucket count and seasonal variance scale `c_i` and `s_i`, respectively. If we solve this in the least squares sense we only need to maintain a small set of statistics rather than all bucket values, which is the key to this being tractable for us in the streaming setting. The same formulation carries over for the log mean and log variance we estimate for the log-normal distribution and so on.

sophiec20:
Above applies to functions mean, median and variance.

LucaWintergerst:
Tom was just helping me troubleshoot this problem with one of our out of the box APM jobs 

This is what this behaviour looks like:

<img width="456" alt="Screenshot 2022-06-30 at 13 14 08" src="https://user-images.githubusercontent.com/11661400/176721409-20c9f0c2-b5a1-4bc5-8cd8-b2ae8cd2f47c.png">

and this is a fixed version where we used a `summary_count_field_name` field that always had the value 1. In this particular case I was doing a cardinality agg on a field with a cardinality of 1, but there should be lots of other options too. Some kind of scripted agg might even be quicker. 

[...]

PR #2677, opened 2024-06-03
Don't try and correct for sample count when estimating statistic variances for anomaly detection

tveasey:
Currently, we:
1. Try to ensure each metric sample contains the same number of values
2. Correct for the variance assuming the raw values are independent samples from some distribution when we compute time bucket anomaly scores

This adds significant complexity to sampling metrics and creates a disconnect between the data we show in visualisations and the data we use for anomaly detection. Furthermore, the independence assumption frequently does not hold. In this case our current behaviour can lead to false negatives. For data where outages are associated with a significant fall in data rate this is particularly problematic. The choice to try to correct the variance predated modelling periodic variance, which now better accounts for the most common case, that the data rate is periodic.

In this PR I have reverted to using the raw time bucket statistics for model update and anomaly detection. I rely on periodic variance estimation to deal with (common instances of) time varying data rate. This is a step towards #1386.


valeriy42:
Behaviour before the change:
image
Behaviour after the change:
image

include/core/CSmallVector.h
```diff
@@ -142,6 +142,17 @@ class CSmallVector : public boost::container::small_vector<T, N> {
         return *this;
     }
 
+    std::string toDelimited(const std::string& delimiter = ", ") const {
+        std::string result;
+        for (size_type i = 0; i < this->size(); ++i) {
+            result += std::to_string((*this)[i]);
+            if (i < this->size() - 1) {
+                result += delimiter;
+            }
```
tveasey, line 151: As I recall appending has quite some overhead for `std::string` due to reallocations. How about something like
```
if (i < this->size() - 1) {
    result += delimiter;
}
if (i == 0) {
    result.reserve(result.size() * this->size());
}
```

tveasey (commented):
Thanks for finishing this Valeriy! One small suggestion, but LGTM.

PR #2679, opened 2024-06-06
[ML] Remove allocators from cache after usage

valeriy42:
In the `CBoostJsonWriteBase` class, we maintain a cache of allocators that should be reused when we write new results. Although we took measures to free memory in these allocators, it was still not completely possible. As a result, we noticed a gradual increase in memory usage by native processes due to the increasing size of the allocators in the cache.

This PR ensures that the allocator is removed from the cache and the memory is freed.

Related to #2664

### DIFF
--- a/include/core/CBoostJsonWriterBase.h
+++ b/include/core/CBoostJsonWriterBase.h
@@ -132,10 +132,10 @@ public:
     }
 
     void removeAllocator(const std::string& allocatorName) {
-        if (m_AllocatorCache.find(allocatorName) != m_AllocatorCache.end()) {
-            TPoolAllocatorPtr allocator = m_AllocatorCache[allocatorName];
-            allocator.reset();
-            m_AllocatorCache.erase(allocatorName);
+        auto allocator = m_AllocatorCache.find(allocatorName);
+        if (allocator != m_AllocatorCache.end()) {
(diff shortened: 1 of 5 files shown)


valeriy42:

include/core/CBoostJsonWriterBase.h
```diff
@@ -134,6 +131,14 @@ class CBoostJsonWriterBase {
         return this->getAllocator()->get();
     }
 
+    void removeAllocator(const std::string& allocatorName) {
+        if (m_AllocatorCache.find(allocatorName) != m_AllocatorCache.end()) {
+            TPoolAllocatorPtr allocator = m_AllocatorCache[allocatorName];
```
jan-elastic, line 136: It looks slightly inefficient to do `cache.find(key)` followed by `cache[key]` and `erase(key)`. It does the map lookup 3x.

Why not do:
```
allocator = cache.find(name);
if (allocator != cache.end()) {
  allocator->reset();
  cache.erase(allocator);
}
```

jan-elastic (approved):
LGTM

include/core/CBoostJsonWriterBase.h
valeriy42: Good point. I adjusted the code.

PR #2680, opened 2024-06-07
[ML] Catch any exception thrown by inference 

davidkyle:
The `pytorch_inference` process must respond to every request as the Java side is waiting for a response. Partly this is ensured by a try catch block, we can't be certain what types of exception might be thrown so better to catch the base `std::exception` rather than derived classes.


jan-elastic (approved):
LGTM

PR #2681, opened 2024-06-10
[Backport][ML] Remove allocators from cache after usage (#2679)

valeriy42:
In the CBoostJsonWriteBase class, we maintain a cache of allocators that should be reused when we write new results. Although we took measures to free memory in these allocators, it was still not entirely possible. As a result, we noticed a gradual increase in memory usage by native processes due to the increasing size of the allocators in the cache.

This PR ensures that the allocator is removed from the cache and the memory is freed.

Backport of #2679 to 8.14

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -28,6 +28,30 @@
 
 //=== Regressions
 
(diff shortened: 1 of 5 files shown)


valeriy42:

PR #2683, opened 2024-06-21
[ML] Add `codesign` CMake target

edsavage:
This PR adds the `codesign_autodetect` and `codesign_pytorch_inference` targets to CMake. These targets add the `get-task-allow` entitlement to the respective executables, which is necessary to allow debugging with e.g. `Instruments` on `macOS`

The `codesign` targets are only applicable on `macOS`. Attempting to run them on any other platform will fail with an explanatory message.

Labelling as a `>non-issue` as this is purely a development aid.

### DIFF
--- a/.entitlements.xml
+++ b/.entitlements.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd"\>
+<plist version="1.0">
+    <dict>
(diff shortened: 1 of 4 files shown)


davidkyle:
Thanks Ed this change will be helpful. One question, are the new cmake targets invoked automatically or do they have to be called explicitly when building debug on macOS?

edsavage:
> One question, are the new cmake targets invoked automatically or do they have to be called explicitly when building debug on macOS?

The new targets need to be called manually @davidkyle - something like this (for `pytorch_inference`)

```
cmake --build cmake-build-relwithdebinfo -j 8 -t codesign_pytorch_inference
```

davidkyle (approved):
LGTM

PR #2684, opened 2024-06-25
[ML] Log 'No statistics at.. ' as a warning

davidkyle:
The condition is considered harmless and should not be logged as an error

### DIFF
--- a/docs/CHANGELOG.asciidoc
+++ b/docs/CHANGELOG.asciidoc
@@ -30,6 +30,10 @@
 
 == {es} version 8.15.0
 
+=== Enhancements
+
+* Log 'No statistics at.. ' message as a warning. (See {ml-pull}2684[#2684].)
+
 === Bug Fixes
 
 * Fix "stack use after scope" memory error. (See {ml-pull}2673[#2673].)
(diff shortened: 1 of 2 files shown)


davidkyle:
@edsavage this change was provoked by the error message below. Looking at the current bucket time `[-1,899)` I think it should have been fixed in #2410. Do you think there is still an issue here or we can go ahead and change the log level?

```
[gallery_sum_bytes] [autodetect/179] [CIndividualModel.cc@136] No statistics at 1719197550 for sum bytes partition=clientip, current bucket = [-1,899), partitionFieldValue = 175.144.21.153, personName = | repeated [5]
```

edsavage:
> @edsavage this change was provoked by the error message below. Looking at the current bucket time `[-1,899)` I think it should have been fixed in #2410. Do you think there is still an issue here or we can go ahead and change the log level?
> 
> ```
> [gallery_sum_bytes] [autodetect/179] [CIndividualModel.cc@136] No statistics at 1719197550 for sum bytes partition=clientip, current bucket = [-1,899), partitionFieldValue = 175.144.21.153, personName = | repeated [5]
> ```

I think that it's safe and appropriate to downgrade the log level for that message @davidkyle, thanks!

edsavage (approved):
LGTM

PR #2694, opened 2024-07-16
[ML] Pass the required build architecture args to the pipeline steps

edsavage:
Add the now required architecture flags to the dynamic pipeline step generator


edsavage:
Merging as a matter of urgency as this bug affects all our branch builds.

PR #2695, opened 2024-07-16
[ML] Force time shift with detection rule

valeriy42:
This PR implements a new time of customer rule `force_time_shift`. Execution of the detection rule action `force_time_shift` will shift the time inside the anomaly detector by a specified amount. This is useful, e.g. to quickly adapt to the daylight saving time events that are known beforehand.


include/model/CDetectionRule.h
```diff
@@ -81,6 +101,12 @@ class MODEL_EXPORT CDetectionRule {
 
     //! The conditions that trigger the rule.
     TRuleConditionVec m_Conditions;
+
+    //! Callback function
```
tveasey, line 105:
```suggestion
    //! Callback function to apply a change to a model based on the rule action.
```

include/maths/common/CModel.h
```diff
@@ -539,6 +539,8 @@ class MATHS_COMMON_EXPORT CModel {
     //! Returns true
     virtual bool shouldPersist() const;
 
+    virtual void shiftTime(core_t::TTime time, core_t::TTime shift) = 0;
```
tveasey, line 542:
```suggestion
    //! Apply a time shift at the time \p time by \p shift amount of seconds.
    virtual void shiftTime(core_t::TTime time, core_t::TTime shift) = 0;
```

include/maths/time_series/CTimeSeriesModel.h
```diff
@@ -195,6 +195,9 @@ class MATHS_TIME_SERIES_EXPORT CUnivariateTimeSeriesModel : public common::CMode
     //! Unpack the weights in \p weights.
     static TDoubleWeightsAry unpack(const TDouble2VecWeightsAry& weights);
 
+    //! Apply time shift at the time port \p time by \p shift amount of seconds.
```
tveasey, line 198: This and others below.
```suggestion
    //! Apply time shift at the time \p time by \p shift amount of seconds.
```

include/model/CDetectionRule.h
```diff
@@ -54,6 +64,9 @@ class MODEL_EXPORT CDetectionRule {
     //! Add a condition.
     void addCondition(const CRuleCondition& condition);
 
+    //! Set callback function.
```
tveasey, line 67:
```suggestion
    //! Set callback function to apply some action to a supplied time series model.
```

lib/api/CDetectionRulesJsonParser.cc
```diff
@@ -163,6 +182,45 @@ bool CDetectionRulesJsonParser::parseRuleActions(const json::object& ruleObject,
     return true;
 }
 
+bool CDetectionRulesJsonParser::parseParameters(const json::object& ruleObject,
+                                                model::CDetectionRule& rule,
+                                                int action) {
+    if (ruleObject.contains(PARAMETERS.c_str()) == false) {
```
tveasey, line 188: Shouldn't this use `hasObjectMember`?

lib/api/CDetectionRulesJsonParser.cc
```diff
@@ -153,6 +170,8 @@ bool CDetectionRulesJsonParser::parseRuleActions(const json::object& ruleObject,
             action |= model::CDetectionRule::E_SkipResult;
         } else if (parsedAction == SKIP_MODEL_UPDATE) {
             action |= model::CDetectionRule::E_SkipModelUpdate;
+        } else if (parsedAction == FORCE_TIME_SHIFT) {
```
tveasey, line 173: Handling for callback action?

lib/model/CDetectionRule.cc
```diff
@@ -54,6 +65,31 @@ bool CDetectionRule::apply(ERuleAction action,
     return true;
 }
 
+void CDetectionRule::executeCallback(CAnomalyDetectorModel& model, core_t::TTime time) const {
+
+    if (m_Callback) {
```
tveasey, line 70: We tend to avoid implicit conversions
```suggestion
    if (m_Callback != std::nullptr) {
```

lib/model/CDetectionRule.cc
```diff
@@ -54,6 +65,31 @@ bool CDetectionRule::apply(ERuleAction action,
...
+        bool executeCallback{true};
+        for (const auto& condition : m_Conditions) {
+            executeCallback &= condition.test(time);
+        }
+        if (executeCallback) {
+            m_Callback(model);
+        }
```
tveasey, line 77: Maybe use the efficient early return strategy (although clearly doesn't matter in this case)
```suggestion
        for (const auto& condition : m_Conditions) {
            if (condition.test(time) == false) { // Disjunction
                return;
            }
        }
        m_Callback(model);
```

lib/model/CDetectionRule.cc
```diff
@@ -54,6 +65,31 @@ bool CDetectionRule::apply(ERuleAction action,
...
+    }
+}
+
+void CDetectionRule::addTimeShift(core_t::TTime timeShift) {
+    this->setCallback([timeShift, this](CAnomalyDetectorModel& model) {
+        if (this->m_TimeShiftApplied == false) {
```
tveasey, line 83: I think this can be a mutable lambda and make `m_TimeShiftApplied` a member of it rather than the enclosing class. Seems cleaner. Also does the current approach work with multiple time shift rules?

tveasey (commented):
Mainly minor comments. Although I wonder about the correctness of `m_TimeShiftApplied` as per comment. This is basically LGTM.

lib/api/CDetectionRulesJsonParser.cc
```diff
@@ -163,6 +182,45 @@ bool CDetectionRulesJsonParser::parseRuleActions(const json::object& ruleObject,
...
+            return true;
+        } else {
+            LOG_ERROR(<< "Missing rule field: " << PARAMETERS
+                      << " for the force_time_shift action");
+            return false;
+        }
+    }
+    const json::object& parametersObject = ruleObject.at(PARAMETERS).as_object();
+    if (parametersObject.empty()) {
+        LOG_ERROR(<< "Parameters must not be empty");
```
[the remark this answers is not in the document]
valeriy42: 👍🏻

include/model/CEventRatePopulationModel.h
```diff
@@ -322,6 +322,9 @@ class MODEL_EXPORT CEventRatePopulationModel : public CPopulationModel {
     //! Get the annotations produced by this model.
     const TAnnotationVec& annotations() const override;
 
+    //! Apply time shift at the time port \p time by \p shift amount of seconds.
```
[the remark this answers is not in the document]
valeriy42: 👍🏻

include/model/CDetectionRule.h
valeriy42: 👍🏻

lib/api/CDetectionRulesJsonParser.cc
valeriy42: I removed `model::CDEtectionRule::E_Callback` instead. I will use specific callback action enums instead of a generic one.

lib/api/CDetectionRulesJsonParser.cc
valeriy42: I missed this one. Thank you! 👍🏻

lib/model/CDetectionRule.cc
valeriy42: 👍🏻

lib/model/CDetectionRule.cc
valeriy42: 👍🏻

lib/model/CDetectionRule.cc
valeriy42: Multiple time shifts should live in the different detection rules, since every detection rule will get only on force_time_shift action. I'll add a unit test to verify the functionality 👍🏻

valeriy42: A modified the lambda function to contain its own state. You are right, it's cleaner this way.

include/maths/common/CModel.h
valeriy42: 👍🏻

include/maths/time_series/CTimeSeriesModel.h
```diff
@@ -678,6 +681,9 @@ class MATHS_TIME_SERIES_EXPORT CMultivariateTimeSeriesModel : public common::CMo
     //! Unpack the weights in \p weights.
     static TDouble10VecWeightsAry unpack(const TDouble2VecWeightsAry& weights);
 
+    //! Apply time shift at the time port \p time by \p shift amount of seconds.
```
tveasey, line 684: There are still a few of these typos (if you search for port \p)
```suggestion
    //! Apply time shift at the time \p time by \p shift amount of seconds.
```

tveasey (approved):
There were just a few more of the same typo in the function comments but otherwise LGTM.

PR #2697, opened 2024-07-17
[ML] Fix the regex used to trigger PR builds with flexible configuration

edsavage:
Fix the `trigger_comment_regex` regular expression that is used to provide a high degree of flexibility when triggering PR builds. 

See https://github.com/elastic/buildkite-pr-bot/blob/c229bf261668611aa11d47455fd980cc0eb86002/README.md for more info.


edsavage:
Merging to main because that is the only way of fully testing it

PR #2699, opened 2024-07-17
[ML] Productionize the PyTorch Docker build

edsavage:
Tidy the PyTorch Docker build scripts and configuration, in order to:

* Enable full logging from the docker build process - to simplify investigation should something go wrong
* Read Docker credentials from Vault.
* Limit the scope during which the Docker credentials are in play.
* Push PyTorch Docker image to registry upon successful build.
* Simplify Buildkite pipeline steps so that the QA tests are only run upon success of `Linux x86_64 builds`
* Use a `c2-standard-16` GCP machine for the Docker build step - this greatly improves build times.
* Trigger the `appex-qa-stateful-custom-ml-cpp-build-testing ` pipeline after successfully building the `ml-cpp` code in the newly built PyTorch Docker image - This may be subject to change, but does serve the purpose of some QA testing on the `ml-cpp` code.


valeriy42 (approved):
LGTM. Good work, Ed! 🚀

PR #2700, opened 2024-07-29
[ML] Refactor add annotation

valeriy42:
This PR adds a simple protected interface to add annotations to the model. It is factored out of #2695 for a simpler review.

This functionality will be unit-tested in #2695, where the new `addAnnotation` function will be used.

### DIFF
--- a/include/model/CEventRateModel.h
+++ b/include/model/CEventRateModel.h
@@ -279,6 +279,11 @@ public:
     //! Get the annotations produced by this model.
     const TAnnotationVec& annotations() const override;
 
+protected:
+    void addAnnotation(core_t::TTime time,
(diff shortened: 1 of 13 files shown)


include/model/CEventRateModel.h
```diff
@@ -279,6 +279,11 @@ class MODEL_EXPORT CEventRateModel : public CIndividualModel {
     //! Get the annotations produced by this model.
     const TAnnotationVec& annotations() const override;
 
+protected:
```
tveasey, line 282: I think this can/should be private here and same for other derived versions.

tveasey (approved):
LGTM. Just a minor nit on scoping.

PR #2703, opened 2024-08-12
[ML] Migrate to S3 bucket in elastic-dev AWS account for 3rd party dependencies

edsavage:
Use the `elastic-dev` account for the S3 bucket containing the dependencies for the macOS x86 build.

### DIFF
--- a/dev-tools/docker/linux_aarch64_cross_image/Dockerfile
+++ b/dev-tools/docker/linux_aarch64_cross_image/Dockerfile
@@ -27,7 +27,7 @@ RUN \
 RUN \
   mkdir -p /usr/local/sysroot-aarch64-linux-gnu/usr && \
   cd /usr/local/sysroot-aarch64-linux-gnu/usr && \
-  wget --quiet -O - https://s3-eu-west-1.amazonaws.com/prelert-artifacts/dependencies/usr-aarch64-linux-gnu-13.tar.bz2 | tar jxf - && \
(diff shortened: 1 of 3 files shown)


davidkyle (commented):
There are a few more references in `build.gradle`

https://github.com/elastic/ml-cpp/blob/f8197a0e1c1d21ca387fbdbcba2e04d55a2fae48/build.gradle#L450

edsavage:
> There are a few more references in `build.gradle`

Thanks Dave. These remaining references are to the built `ml-cpp` artifacts themselves, that will later be referenced by the Elasticsearch builds.

I thought that to simplify matters I'd handle the migration to a new S3 bucket in 2 stages:
1. Migrate the 3rd party dependencies that the `ml-cpp` builds require to the new S3 bucket (this PR)
2. Update where the `ml-cpp` artifacts get written to. Because this step will also require updating references in e.g. the `elasticsearch` repo some coordination is required. I'm thinking that a good approach may be to:
  *  update the `ml-cpp` scripts to briefly store the artifacts in both the old `prelert-artifacts` bucket and also in the new `ml-cpp-artifacts` bucket
  * update the `elasticsearch` repo appropriately to refer to the new bucket.
  * finally remove all references to the `prelert-artifacts` bucket from the `ml-cpp` repo.

davidkyle (approved):
LGTM

The plan to incrementally switch to the new bucket makes sense to me

PR #2713, opened 2024-08-29
[ML] Add CMake config to integrate gperftools

edsavage:
Add options to link in libprofiler and/or libtcmalloc to autodetact and pytorch_inference.
 *  For Linux builds only.
 *  Not for production releases.


valeriy42 (approved):
LGTM

PR #2717, opened 2024-09-08
[ML] Trigger a "staging" build on merge to branch

edsavage:
This PR addresses the timeliness of the `ml-cpp` "Staging" (release) builds.

With this change a Buildkite Staging build will be triggered each time a code change is merged to a branch matching the filter condition, i.e. branches names consisting of two or three groups of numbers separated by dots. 

Up until now we have relied upon human processes to trigger these types of builds. However this approach does allow scope to fail.

It it is not necessary to perform Staging builds any more frequently than upon code change. This is due to the fact that the same code is incorporated into Snapshot builds, that are built daily, regardless of code change (in order to run a frequent barrage of tests).

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -219,7 +219,7 @@ spec:
 
           '
         filter_enabled: true
-        trigger_mode: none
+        trigger_mode: code
       repository: elastic/ml-cpp
       skip_intermediate_builds: true
       teams:


valeriy42 (commented):
LGTM in general. Can you please add a bit more explanation/motivation in the PR description so we have it captured for posterity? 

I don't understand the `filter_condition`. Does it mean that the branches like `8.15.2` are included or excluded? What about `main` and `8.x`?

edsavage:
> LGTM in general. Can you please add a bit more explanation/motivation in the PR description so we have it captured for posterity?
> 
> I don't understand the `filter_condition`. Does it mean that the branches like `8.15.2` are included or excluded? What about `main` and `8.x`?

Good spot on the filter condition.

It is an "inclusive" filter and is intended to ensure that only branches consisting of two or three groups of digits separated by dots, so e.g. `8.16`; `9.0.0`, will be built for staging (release) builds. And to explicitly rule out branches such as `main`, `8.x` for that purpose.

valeriy42 (commented):
> And to explicitly rule out branches such as main, 8.x for that purpose.

I think, we actually want to include `main` and `^[0-9]+\.x` here to build the snapshots automatically. Otherwise, when we have a feature freeze of 9.0 and we suppose to have a snapshot with all related changes for BC, we still would need to remember to trigger it automatically. Same for 8.16.0 that will be branched of 8.x.

edsavage:
> I think, we actually want to include `main` and `^[0-9]+\.x` here to build the snapshots automatically.

Hmmm.. I actually think the filter is ok, as it stands. We would never perform a release build off e.g. `main` or `8.x` directly. We would first create a say `9.0.0` or `8.16.0` branch first, and build off that. And since those branches match the filter, and (with this PR) trigger a build on code change, they should be built upon creation.

valeriy42 (approved):
LGTM

edsavage:
Removing the `v8.16.0` label as changes to `catalog-info.yaml` are only required on `main` to take effect.

PR #2718, opened 2024-09-09
[ML] Increase timeout for macOS debug builds

edsavage:
The macOS x86_64 debug build has been timing out recently (since migrating to building on Orka instances). Bump the timeout for all macOS builds by another hour.


edsavage:
Removing the `v8.16.0` label as this change only applies to CI builds of the `main` branch, which is now `v9.0.0`.

PR #2719, opened 2024-09-09
Fix typo in build-setup/macos.md

maxhniebergall:
There was an extra "install" in the Pytorch line which caused it to fail


edsavage (approved):
LGTM

PR #2720, opened 2024-09-09
Change result writing to allow for a single value

maxhniebergall:
Allow for single dimensional vectors / zero dimensional tensors in pytorch inference

### DIFF
--- a/bin/pytorch_inference/CResultWriter.h
+++ b/bin/pytorch_inference/CResultWriter.h
@@ -191,7 +191,7 @@ private:
         jsonWriter.onObjectEnd();
     }
 
-        //! Write a 2D inference result
+    //! Write a 1D inference result
     template<typename T>
     void writeInferenceResults(const ::torch::TensorAccessor<T, 1UL>& accessor,
                                TStringBufWriter& jsonWriter) {
@@ -199,10 +199,12 @@ private:
         jsonWriter.onKey(RESULT);
         jsonWriter.onObjectBegin();
         jsonWriter.onKey(INFERENCE);
-        // The Java side requires a 3D array, so wrap the 2D result in an
-        // extra outer array.
+        // The Java side requires a 3D array, so wrap the 1D result in an
+        // extra outer array twice.
+        jsonWriter.onArrayBegin();
         jsonWriter.onArrayBegin();
         this->writeTensor(accessor, jsonWriter);
+        jsonWriter.onArrayEnd();
         jsonWriter.onArrayEnd();
         jsonWriter.onObjectEnd();
     }
--- a/bin/pytorch_inference/CResultWriter.h
+++ b/bin/pytorch_inference/CResultWriter.h
@@ -191,6 +191,22 @@ private:
         jsonWriter.onObjectEnd();
     }
 
+        //! Write a 2D inference result
+    template<typename T>
+    void writeInferenceResults(const ::torch::TensorAccessor<T, 1UL>& accessor,
+                               TStringBufWriter& jsonWriter) {
+
+        jsonWriter.onKey(RESULT);
+        jsonWriter.onObjectBegin();
+        jsonWriter.onKey(INFERENCE);
+        // The Java side requires a 3D array, so wrap the 2D result in an
+        // extra outer array.
+        jsonWriter.onArrayBegin();
+        this->writeTensor(accessor, jsonWriter);
+        jsonWriter.onArrayEnd();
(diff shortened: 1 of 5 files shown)


bin/pytorch_inference/CResultWriter.h
```diff
@@ -191,6 +191,22 @@ class CResultWriter : public TStringBufWriter {
...
+    void writeInferenceResults(const ::torch::TensorAccessor<T, 1UL>& accessor,
+                               TStringBufWriter& jsonWriter) {
+
+        jsonWriter.onKey(RESULT);
+        jsonWriter.onObjectBegin();
+        jsonWriter.onKey(INFERENCE);
+        // The Java side requires a 3D array, so wrap the 2D result in an
+        // extra outer array.
+        jsonWriter.onArrayBegin();
+        this->writeTensor(accessor, jsonWriter);
```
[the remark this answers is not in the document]
maxhniebergall: Seems like yes, we need to have two arrayBegin()

bin/pytorch_inference/Main.cc
```diff
@@ -69,11 +69,18 @@ torch::Tensor infer(torch::jit::script::Module& module_,
...
         auto output = module_.forward(inputs);
+        LOG_DEBUG(<< "output_" << i << ": " << output);
         if (output.isTuple()) {
             // For transformers the result tensor is the first element in a tuple.
             all.push_back(output.toTuple()->elements()[0].toTensor());
         } else {
-            all.push_back(output.toTensor());
+            auto outputTensor = output.toTensor();
+            auto sizes = outputTensor.sizes();
+            if (sizeof(sizes) > 1) {
```
davidkyle, line 79: `sizeof` returns the size of the _thing_ in bytes, this conditional will always be true. Do we need this logic? I thought the specialisation of `writeInferenceResults()` was the fix here

maxhniebergall: yea, I was just looking into these errors and this is definitely wrong. I was trying to check for the number of dimensions, but I checked the wrong thing. I'll double check, but I believe that this reshape is required to be able to do that cat (which was what was throwing the error initially). The change to `writeInferenceResults()` was also required further down.

maxhniebergall: yea, I just double checked, without the reshape here we get the `zero-dimensional tensor (at position 0) cannot be concatenated` exception. But the if statement is wrong for the reason you mentioned.

maxhniebergall: I've just changed it to do what I was trying to do (I think), reshape the tensor to be 1-dimensional if it was zero dimensional.

davidkyle (approved):
LGTM

bin/pytorch_inference/Main.cc
davidkyle: I think you've got it now

bin/pytorch_inference/Main.cc
```diff
@@ -69,11 +69,17 @@ torch::Tensor infer(torch::jit::script::Module& module_,
...
         if (output.isTuple()) {
             // For transformers the result tensor is the first element in a tuple.
             all.push_back(output.toTuple()->elements()[0].toTensor());
         } else {
-            all.push_back(output.toTensor());
+            auto outputTensor = output.toTensor();
+            if (outputTensor.dim() == 0) { // If the output is a scaler, we need to reshape it into a 1D tensor
+                all.push_back(outputTensor.reshape({1, 1}));
+            } else {
+                all.push_back(outputTensor);
```
droberts195, line 81: Not sure how important it is for performance, but you might be able to use `std::move` here to avoid copying `outputTensor` into the vector. (The way the code was before, the temporary rvalue `output.toTensor()` was added directly to the vector, so would be moved by default. But in the new code it’s stored as an lvalue in a local variable, so cannot be moved unless you specifically ask for this. Maybe it’s not very important. It depends on the internal implementation of the object in PyTorch. If a copy involves a memory allocation then it’s will affect performance much more than if it’s just copying a few builtin types.)

maxhniebergall: Thanks Dave! I hope you're doing well!!

maxhniebergall: I ran the throughput/latency benchmark and the memory benchmark, and the numbers were extremely similar. There might be a slight performance improvement to using `std:move`, so I'll go with that.

Thanks Dave!

PR #2721, opened 2024-09-10
Bump 9.0.0

jan-elastic:


valeriy42 (approved):
LGTM. 

We should remember to remove the build of 7.17 when 9.0.0 is released.

PR #2726, opened 2024-09-12
Remove debug line that was printing out huge tensors causing timeouts in QA

maxhniebergall:
The recent change to support zero dimensional tensors was failing in QA tests, and the root cause was this debug line which printed out each of the results, which can be quite large for some models. 

I aslo removed std::move from the reshape because it was causing compile time warnings, but that wasn't the cause of this problem.

This issue only affected QA so it is a non-issue


davidkyle (approved):
LGTM

maxhniebergall:
We're seeing two different issues in CI:
On Macos x86_64 we have this error:
```
ASSERTION FAILURE:
- file   : CIoManagerTest.cc
- line   : 91
- message: critical check processedData.length() == threadReader.data().length() has failed [10000 != 4096]
```

@edsavage have you seen this?


the other two errors are about requiring java 21

PR #2727, opened 2024-09-15
[ML] Use java 21 for ES tests.

edsavage:
Elasticsearch main branch now requires Java 21 to build. Bump the version so that the ES integration tests can build and run.

### DIFF
--- a/.buildkite/pipelines/run_es_tests_aarch64.yml.sh
+++ b/.buildkite/pipelines/run_es_tests_aarch64.yml.sh
@@ -13,7 +13,7 @@ steps:
   - label: "Java :java: Integration Tests for aarch64 :hammer:"
     key: "java_integration_tests_aarch64"
     command:
-      - "sudo yum -y install java-17-amazon-corretto-devel"
+      - "sudo wget -O - https://corretto.aws/downloads/latest/amazon-corretto-21-aarch64-linux-jdk.tar.gz | tar xvzf - -C /usr/lib/jvm"
       - 'buildkite-agent artifact download "build/*" . --step build_test_linux-aarch64-RelWithDebInfo'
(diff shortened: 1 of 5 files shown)


valeriy42 (approved):
LGTM. Thank you for figuring this out 🚀

.buildkite/pipelines/run_es_tests_aarch64.yml.sh
```diff
@@ -13,7 +13,7 @@ steps:
   - label: "Java :java: Integration Tests for aarch64 :hammer:"
     key: "java_integration_tests_aarch64"
     command:
-      - "sudo yum -y install java-17-amazon-corretto-devel"
+      - 'sudo mkdir -p /usr/lib/jvm/amazon-corretto-21 && wget -O - https://corretto.aws/downloads/latest/amazon-corretto-21-aarch64-linux-jdk.tar.gz | sudo tar xzf - -C /usr/lib/jvm/amazon-corretto-21 --strip=1'
```
jan-elastic, line 16: Can't you just do:
```
sudo yum install java-21-openjdk
```
See: https://docs.redhat.com/en/documentation/red_hat_build_of_openjdk/21/html-single/installing_and_using_red_hat_build_of_openjdk_21_on_rhel/index

jan-elastic (approved):
LGTM, but I'm wondering if the existing simpler approach would also work

.buildkite/pipelines/run_es_tests_aarch64.yml.sh
edsavage: Thanks @jan-elastic ! I'll merge the current PR as it's working and take a look at your (much simpler!) suggestion.

jan-elastic: sounds good to me

PR #2729, opened 2024-09-17
[ML] Fudge for CKMostCorrelatedTest/testScale

edsavage:
Add a bit of tolerance for a CKMostCorrelatedTest/testScale comparison. This _should_ reduce the frequency of failures on `macOS x86_64` CI builds that are a bit slooooower than the rest.


valeriy42 (approved):
LGTM

PR #2737, opened 2024-09-19
[ML] Change time shift sign for intuitive configuration

valeriy42:
The way the application of the force_time_shift custom rule was configured before, is to specify the amount of time the model's frame of reference should shift by. 

This is counter-intuitive for the users, since in spring when they need to move their clock forward, they would be expected to specify force_time_shift by -3600, aka backward.

This PR switches the sign inside the C++ implementation and allows a more intuitive configuration.

I marked it as a non-issue since the functionality has not yet been released.


tveasey (approved):
LGTM

PR #2755, opened 2024-09-25
[ML] Correct filter regex for Staging builds

edsavage:
Triggering a Staging build via a Github webhook
requires a slight tweak to the branch filter regular expression.

Tested by briefly modifying the regex in Buildkite pipeline settings and pushing an empty commit to the `8.15` branch. Checking the Staging pipeline in Buildkite shows that a build was indeed triggered, see - https://buildkite.com/elastic/ml-cpp-staging-builds/builds/111

Relates #2717


valeriy42 (approved):
LGTM

PR #2759, opened 2024-10-07
[ML] Fix forecasting parameters initialization for large models

valeriy42:
We fix initialization of the `decayRate` parameter when models have to spill to disk

### DIFF
--- a/lib/maths/time_series/CTrendComponent.cc
+++ b/lib/maths/time_series/CTrendComponent.cc
@@ -601,9 +601,10 @@ void CTrendComponent::forecast(core_t::TTime startTime,
         double variance{a * common::CBasicStatistics::mean(variance_) +
                         b * common::CBasicStatistics::variance(m_ValueMoments)};
 
-        TVector2x1 trend{confidenceInterval(
-            this->value(modelWeights, models, scaleTime(time, m_RegressionOrigin)),
-            variance, confidence)};
+        auto prediction = a*this->value(modelWeights, models, scaleTime(time, m_RegressionOrigin))
+                                + b*common::CBasicStatistics::mean(m_ValueMoments);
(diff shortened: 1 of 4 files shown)


lib/maths/time_series/CTrendComponent.cc
```diff
@@ -601,9 +601,10 @@ void CTrendComponent::forecast(core_t::TTime startTime,
         double variance{a * common::CBasicStatistics::mean(variance_) +
                         b * common::CBasicStatistics::variance(m_ValueMoments)};
 
-        TVector2x1 trend{confidenceInterval(
-            this->value(modelWeights, models, scaleTime(time, m_RegressionOrigin)),
-            variance, confidence)};
+        auto prediction = a*this->value(modelWeights, models, scaleTime(time, m_RegressionOrigin))
+                                + b*common::CBasicStatistics::mean(m_ValueMoments);
+        TVector2x1 trend{confidenceInterval(prediction, variance, confidence)
+            };
```
tveasey, line 607: I have a vague feeling that we might of done something like this originally, but decided against it (although I can't find the evidence in ml-cpp history so it happened a long time ago).

There is an inconsistency with other prediction methods so it at least warrants a comment, but since those are only used for short range predictions for anomaly detection they don't suffer the same drawbacks.

Take one obvious issue with this: if you forecast a time series with a linear ramp at some point it will revert back to the historic average. This is going to look pretty weird and generally such series wouldn't behave like this. As it stands we make our best estimate, but blow up the uncertainty if you start to forecast longer than the observed data range. My inclination would be to keep this as is.

tveasey (commented):
Good detective work on the issues with overflow to disk. My inclination is not to alter the forecast behaviour itself (more detail in the comment).

tveasey (approved):
LGTM

PR #2769, opened 2024-10-21
[ML] Windows 2022 build VM

edsavage:
Update the build system to use the Windows 2022 build VM.

The Windows 2022 build VM is provisioned with the Windows 11 SDK, build 22621 - https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/


edsavage:
buildkite build this

valeriy42 (approved):
LGTM. Good work 🚀  

Only one thing: Would you mind bumping the CMake version to 3.30.5? I know I put 3.27.9 in the issue initially, but since then I discovered that at least on Linux there is no real reason not to use the latest Cmake version. Since there were several improvements regarding build speed and parallel efficiency,  it would be good to benefit from them.

PR #2774, opened 2024-11-08
[ML] Build solely on macos 13 (Ventura) aarch64

edsavage:
From version 9.0.0 we drop support for macos x86_64 builds.

Therefore all support for cross compiling for macos x86_64 has also been
dropped.

Macos builds have been migrated to be executed solely on macos 13 (Ventura)
aarch64. In CI, these builds are performed on updated Orka VMs.

In addition, 3rd party dependencies (boost, pytorch) are now baked into
the Orka VM, so downloading these dependencies at run time is no longer
required.

Depends on https://github.com/elastic/elasticsearch/pull/104125


valeriy42 (approved):
LGTM. Great work! 🚀

PR #2780, opened 2024-11-19
[ML] Upgrade to Boost 1.86.0 on Windows 2022 builds

edsavage:
Update build scripts and docs to refer to Boost 1.86


valeriy42 (approved):
LGTM. You need to be careful when backporting, since 8.18 relies on different base image.

valeriy42 (approved):
LGTM

PR #2782, opened 2024-11-20
Update build configurations for version 8.17 and SNAPSHOT messages

valeriy42:
Adjust build configurations to include version 8.17 and update daily SNAPSHOT build messages accordingly.


jan-elastic (approved):
LGTM

PR #2773, opened 2024-10-28
[ML] Update Linux builder Docker container for 9.0

valeriy42:
This PR updates all Linux-related Dockerfiles and tutorials with the latest dependencies, libraries, and C++ compiler versions. 

### Changes

- **Dockerfile Updates**:
  - The base image is changed from `centos:7` to `rockylinux:8` to maintain compatibility with RHEL in the future.
  - Updated all GCC versions to use GCC 13.3 and binutils 2.41 for enhanced C++17 support and security hardening.
  - Other updated libraries:
    - `binutils` 2.37 -> 2.41
    - `libxml2` 2.9.14 -> 2.10.4.
    - `patchelf` 0.13 -> 0.18.0
    - `python` 3.10.0 -> 3.12.7
    - `cmake` 3.19.2 -> 3.30.5

- **Tutorial Updates**:
  - Revised instructions for setting up the Linux build environment to match new dependencies and configurations.
  - Provided updated example commands for package installation compatible with Rocky Linux 8 and similar distributions.
- **Buildkite**:
  - Updated the CI scripts to use the newest version of the Docker containers.

PR #2785, opened 2024-11-28
[ML] Update Linux builder Docker container for 9.0

valeriy42:
This is a rework of #2773 . Additionally, I removed the dependency on `liblzma` when building the boost library since some of the production docker containers don't have `liblzma` installed.

I added arguments like `--without-lzma`, but for Boost, it didn't lead to the desired effect. Hence, I had to resolve to temporarily rename the system libraries:
```docker
RUN \
  for i in $(ls /lib64/libzstd.*); do mv $i "$i.bak" ; done && \
  for i in $(ls /lib64/liblzma.*); do mv $i "$i.bak" ; done
```


edsavage (approved):
LGTM :rocket:

3rd_party/licenses/gcc-runtime-INFO.csv
```diff
@@ -1,2 +1,2 @@
 name,version,revision,url,license,copyright,sourceURL
-GCC Runtime Library,10.3.0,f00b5710a30f22efc3171c393e56aeb335c3cd39,https://gcc.gnu.org,GPL-3.0 WITH GCC-exception-3.1,"Copyright (C) 2021 Free Software Foundation, Inc.",http://ftpmirror.gnu.org/gcc/gcc-10.3.0/gcc-10.3.0.tar.gz
```
edsavage, line 2: NIT: We should update the revision here to be `b71f1de6e9cf7181a288c0f39f9b1ef6580cf5c8`

PR #2806, opened 2025-01-07
[ML] Remove use of XML serialisers in unit tests.

### LINKED ISSUE #2804: Remove XML dependencies
The XML parser is used only for test data parsing, while the normal input and output rely exclusively on JSON. Removing XML readers from the code would allow us to remove dependencies on libXML and RapidXML and simplify several classes.

edsavage:
Replacing the use of XML serialisers in unit tests with equivalent JSON serialisers allows the removal of dependencies on the 3rd party RapidXML library and also the bespoke XML parser and related code.

Labelling the PR as `>non-issue` as despite it's size it doesn't affect any external code or results.

Closes #2804


valeriy42 (commented):
Good work! Does this mean that we can now remove libxml2 as our dependency? If so, we probably also need to adjust the CMake files and the 3rd party license info.

lib/model/unittest/CEventRateDataGathererTest.cc
```diff
@@ -286,6 +238,134 @@ void importCsvData(CDataGatherer& gatherer,
     ifs.reset();
 }
 
+void testGathererMultipleSeries(const core_t::TTime startTime,
+                                const core_t::TTime bucketLength,
+                                core_t::TTime data1[],
+                                core_t::TTime data2[],
+                                std::string expectedPersonCounts[],
+                                SModelParams params,
```
valeriy42, line 246: can this be a const reference?

lib/model/unittest/CEventRateDataGathererTest.cc
```diff
@@ -286,6 +238,134 @@ void importCsvData(CDataGatherer& gatherer,
...
+    BOOST_REQUIRE_EQUAL(1, featureData.size());
+    BOOST_REQUIRE_EQUAL(model_t::E_IndividualCountByBucketAndPerson,
+                        featureData[0].first);
+    BOOST_REQUIRE_EQUAL(std::string("[(0, 3)]"),
+                        core::CContainerPrinter::print(featureData[0].second));
+}
+
+void testGathererMultipleSeries(const core_t::TTime startTime,
+                                const core_t::TTime bucketLength,
+                                SModelParams params,
```
valeriy42, line 312: `params` is not used in the function.

edsavage:
> Good work! Does this mean that we can now remove libxml2 as our dependency? If so, we probably also need to adjust the CMake files and the 3rd party license info.

I _think_ so. My only hesitation is that it might be needed by one of the boost libraries. I'll check on all platforms before adjusting CMake, docs, build scripts etc.

edsavage:
> > Good work! Does this mean that we can now remove libxml2 as our dependency? If so, we probably also need to adjust the CMake files and the 3rd party license info.
> 
> I _think_ so. My only hesitation is that it might be needed by one of the boost libraries. I'll check on all platforms before adjusting CMake, docs, build scripts etc.

We still have a dependency on `libxml2`  because we're using it to generate XML formatted Boost Test results - see `lib/test/CBoostTestXmlOutput.cc`. I think we want to keep it as-is since it's much easier for a human to read than the JUNIT output we also generate (for the Buildkite Test Analytics)

valeriy42 (approved):
LGTM. Also, thank you for including the SonarQube linting suggestions.

PR #2810, opened 2025-01-13
[ML] Include value of VERSION_QUALIFIER in DRA artifact names

edsavage:
If the VERSION_QUALIFIER environment variable is set, it's value should be included in the DRA artifact naming scheme and the product version (as returned by `--version`)


jan-elastic (approved):
LGTM

PR #2815, opened 2025-01-15
[ML] Fix compilation issue in debug build

edsavage:
Fix a TRACE statement which causes the debug build compilation to fail


valeriy42 (approved):
LGTM

PR #2816, opened 2025-01-30
[ML] Bump version to 8.19.0

valeriy42:
Update the Elasticsearch version to 8.19.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 org.gradle.daemon=false
 
-elasticsearchVersion=8.18.0
+elasticsearchVersion=8.19.0
 
 artifactName=ml-cpp
 


gradle.properties
```diff
@@ -1,6 +1,6 @@
 org.gradle.daemon=false
 
-elasticsearchVersion=8.18.0
+elasticsearchVersion=8.19.0
```
jan-elastic, line 3: I think you also need to update the [catalog-info](https://github.com/elastic/ml-cpp/blob/main/catalog-info.yaml#L159), so that we keep building 8.18 artifacts.

valeriy42: Yes, this change goes into the `main` in https://github.com/elastic/ml-cpp/pull/2817

jan-elastic: Just noticed that's a [separate PR](https://github.com/elastic/ml-cpp/pull/2817/)

jan-elastic (approved):
LGTM

PR #2818, opened 2025-01-30
[ML] Update build configurations for Elasticsearch 9.1.0 and adjust cron schedules for daily builds

valeriy42:
Update the Elasticsearch version to 9.1.0 and modify cron schedules for daily builds to add support for the new version.


jan-elastic (approved):
LGTM

PR #2820, opened 2025-02-03
[ML] Refine execution of unit tests

edsavage:
Give the ability to specify which tests (or suites of tests) will be executed. Regular expressions are also allowed. In addition any other valid Boost.TEST flag can be passed to the process. This is especially useful when running tests from the command line. For example:

```
TEST_FLAGS="--show-progress" TESTS="*/testPersist" cmake --build cmake-build-relwithdebinfo -j 8 -t test_model
```

would run the `testPersist` test in all suites in the `model` unit tests, and show progress while doing so.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,13 +61,7 @@ Please adhere to the general guideline that you should never force push to a pub
 
 **Repository**: https://github.com/elastic/ml-cpp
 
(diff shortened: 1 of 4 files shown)


valeriy42 (commented):
Can you please document this option to run tests in our contribution guidelines so it won't get lost?

valeriy42 (approved):
LGTM

PR #2821, opened 2025-02-19
[ML] Better handling of config updates

### LINKED ISSUE #2741: [ML] Custom rules are not updated while the job is running
When we create a custom rule for an open anomaly detection job, e.g. using Kibana or the Anomaly Detection Job Update API, the rule is stored in the Java job configuration and passed to the native process, but the job configuration is not updated on the fly. This behavior can be reproduced in recent Elasticsearch versions, at least 8.13+.

The only workaround is to stop and restart the job.

edsavage:
* Enable updating of custom rules
* Correctly update filters (scoped rules)

These changes fix a bug whereby memory corruption was caused when updating filters for an open job. The bug had previously been masked by code that reset the job's analysis config to its original state after every config update - the downside of this was that custom rules could not be updated on the fly, only by stopping and restarting the job.

Closes #2741

Initially opening as a draft PR to see how it behaves against the ES integration tests.


include/api/CAnomalyJobConfig.h
```diff
@@ -249,19 +261,6 @@ class API_EXPORT CAnomalyJobConfig {
...
-        //! Reparse the detector configuration object from within a stored
-        //! string representing the analysis config object.
-        //! This is necessary to correctly reinitialise scoped rule objects
-        //! folowing an update of the fiter rules configuration.
-        bool reparseDetectorsFromStoredConfig(const std::string& analysisConfig);
-
-        void setConfig(const std::string& analysisConfigString) {
-            m_AnalysisConfigString = analysisConfigString;
-        }
-
```
valeriy42, line 264: Can you please leave a comment in this PR on why do you remove those?

valeriy42 (approved):
Good job on fixing the bug. I have just a few minor comments.

include/api/CAnomalyJobConfig.h
edsavage: The code to `set`, `get` and `reparse` analysis config was introduced as part of the work to upgrade from boost 1.77 to 1.83. That upgrade included a major rewrite of boost unordered containers. Unfortunately it broke behaviour that we had come to rely on related to updating analysis config during a jobs' active lifetime. As a workaround the analysis config was reset to its previous state after an update, while allowing filter rules config to be updated. This PR provides a correct fix for the filter rules update, which also allows the analysis config to be updated safely as well. Therefore any code related to resetting the analysis config to its previous state is no longer needed.

PR #2828, opened 2025-03-17
[ML] Fix 3rd party lib depends for Win 2022 builds

edsavage:
The CMake configuration was failing to locate some vital 3rd party libraries and include them for redistribution for Windows 2022 builds. This isn't an issue for our internal machines for building and testing as those files are already present there, but is very much an issue for non-build machines that are missing the required libraries.

Labelling as `>non-issue` as this relates to the build system only and impacts versions not yet released.


valeriy42 (approved):
LGTM

PR #2831, opened 2025-03-17
[ML] Track memory usage in  CHierarchicalResultsNormalizer

### LINKED ISSUE #2244: [ML] Memory usage of hierarchical results normalizer is not accounted for
We do not account for the memory size of the hierarchical results normalizer.

When it was first introduced it was fairly close to being a fixed size object, so could reasonably be considered part of the "process overhead" - fixed data structures for which we add 10MB to the required memory of each job.

However, now that we do per partition scoring it is no longer reasonable to consider the normalizer part of the fixed overhead.  For a job with a large number of partition field values it can be a non-negligible size, and this memory should be properly accounted for.

valeriy42:
This PR makes CHierarchicalResultsNormalizer to track its memory usage.

Fixes #2244


lib/model/unittest/CHierarchicalResultsNormalizerTest.cc
```diff
@@ -0,0 +1,36 @@
...
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_SUITE(CHierarchicalResultsNormalizerTest)
+using namespace ml;
+
+namespace {
+constexpr std::size_t BUCKET_SIZE{3600};
+}
+
+BOOST_AUTO_TEST_CASE(HierarchicalResultsNormalizerShouldUpdateMemoryUsage) {
```
edsavage, line 25: Rename to `testHierarchicalResultsNormalizerShouldUpdateMemoryUsage`? To be in keeping with elsewhere..

edsavage (approved):
LGTM. Just the one comment regarding test case naming.

include/model/CHierarchicalResultsNormalizer.h
```diff
@@ -44,6 +45,19 @@ struct MODEL_EXPORT SNormalizer {
...
+        core::memory_debug::dynamicSize("s_Description", s_Description, mem);
+    }
+
+    std::size_t memoryUsage() const {
+        std::size_t mem = 0;
+        mem += core::memory::dynamicSize(s_Description);
+        return mem;
+    }
+
+    std::size_t staticSize() const { return sizeof(*this); }
```
tveasey, line 59: I think we only need a `staticSize` method if we're computing memory usage via a base class pointer in which case this should be an override. So I don't think it is needed here.

tveasey (approved):
LGTM (aside the build issue)

include/model/CHierarchicalResultsNormalizer.h
```diff
@@ -44,6 +45,19 @@ struct MODEL_EXPORT SNormalizer {
     //! Compute a checksum for this object.
     uint64_t checksum() const;
 
+    void debugMemoryUsage(const core::CMemoryUsage::TMemoryUsagePtr& mem) const {
```
tveasey, line 48: I suspect declaring inline in a `MODEL_EXPORT` class is causing link issue on windows.

valeriy42:
@wwang500 , this PR should lead to an increase in reported memory usage on jobs with a large number of partitions/influencers. Can you please verify this with the nightly runs?

wwang500:
Checked the test results from last two weeks, there are no step changes. large jobs or small jobs.

i.e., model_bytes from nightly jobs: 
image

PR #2841, opened 2025-03-26
[ML] Better messaging regarding OOM process termination

edsavage:
This PR provides a more detailed message when a process is terminated with SIGKILL.

On Linux, the OOM (Out Of Memory) system handler will kill processes, according to heuristics, when the OS runs low on memory.

 Our native processes (apart from controller) are configured so that they would be chosen first to be terminated in such a situation.

 The OOM handler terminates processes with a SIGKILL (signal 9). SIGKILL is not able to be handled by processes and will result in immediate termination, not allowing for any logging of the situation. However, the parent process - controller - can detect and report on the death of its children.

 Relates https://github.com/elastic/ml-team/issues/1158


valeriy42 (approved):
LGTM

PR #2846, opened 2025-04-04
[ML] Report the "actual" memory usage of the autodetect process

edsavage:
Determine the actual memory usage of the autodetect process as reported by the OS, e.g. on Linux this would be the value of the maximum resident set size returned by a call to `getrusage`.

Add this value to the model size stats record returned to the ES Java process so it can be included in the `job counts` tab for anomaly detection jobs. 

Relates https://github.com/elastic/elasticsearch/pull/131981


bin/autodetect/Main.cc
```diff
@@ -83,7 +84,8 @@ int main(int argc, char** argv) {
         ml::counter_t::E_TSADNumberMemoryLimitModelCreationFailures,
         ml::counter_t::E_TSADNumberPrunedItems,
         ml::counter_t::E_TSADAssignmentMemoryBasis,
-        ml::counter_t::E_TSADOutputMemoryAllocatorUsage};
+        ml::counter_t::E_TSADOutputMemoryAllocatorUsage,
+        ml::counter_t::E_TSADMaxResidentSetSize};
```
valeriy42, line 88: Does reporting only the **maximum** resident set size make sense? Can we also collect the **current** resident set size?

valeriy42 (commented):
Hi Ed. I did the first pass. 

We should discuss the naming of the new field. While "actual" conveys the intention of the value, it is confusing to the user. 

Also, does maximum resident set size actually correspond to the actual current memory usage or is it the historical peak process memory usage?

edsavage:
> Also, does maximum resident set size actually correspond to the actual current memory usage or is it the historical peak process memory usage?

The resident set size (RSS) represents the process's current RAM usage (so not counting pages that have been swapped out etc.), and the max RSS is the high water mark of that value. I think that reporting both would be useful for our purposes.

include/model/ModelTypes.h
```diff
@@ -719,7 +719,9 @@ enum EAssignmentMemoryBasis {
     E_AssignmentBasisUnknown = 0,           //!< Decision made in Java code
     E_AssignmentBasisModelMemoryLimit = 1,  //!< Use model memory limit
     E_AssignmentBasisCurrentModelBytes = 2, //!< Use current actual model size
-    E_AssignmentBasisPeakModelBytes = 3 //!< Use highest ever actual model size
+    E_AssignmentBasisPeakModelBytes = 3, //!< Use highest ever actual model size
+    E_AssignmentBasisSystemMemoryBytes = 4, //!< Use the current system memory size
+    E_AssignmentBasisMaxSystemMemoryBytes = 5 //!< Use the highest ever system memory size
```
valeriy42, line 724: I don't think we need this assignment basis reasons (at least so far)

bin/autodetect/Main.cc
```diff
@@ -83,7 +84,9 @@ int main(int argc, char** argv) {
         ml::counter_t::E_TSADNumberMemoryLimitModelCreationFailures,
         ml::counter_t::E_TSADNumberPrunedItems,
         ml::counter_t::E_TSADAssignmentMemoryBasis,
-        ml::counter_t::E_TSADOutputMemoryAllocatorUsage};
+        ml::counter_t::E_TSADOutputMemoryAllocatorUsage,
+        ml::counter_t::E_TSADResidentSetSize,
+        ml::counter_t::E_TSADMaxResidentSetSize};
```
valeriy42, line 89: Should we call these counters `SystemMemoryUsage` for consistency?

lib/model/ModelTypes.cc
```diff
@@ -1733,6 +1733,10 @@ std::string print(EAssignmentMemoryBasis assignmentMemoryBasis) {
         return "current_model_bytes";
     case E_AssignmentBasisPeakModelBytes:
         return "peak_model_bytes";
+    case E_AssignmentBasisSystemMemoryBytes:
+        return "system_memory_bytes";
+    case E_AssignmentBasisMaxSystemMemoryBytes:
+        return "max_system_memory_bytes";
```
valeriy42, line 1739: Are those necessary?

prodsecmachine:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/machine-learning-48e/pr-checks/255bb8b6-432b-490b-90c1-e2705588540d?refs=fc8888c1a787fd6fc5dcc1184dab2f1a90d93d76&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/machine-learning-48e/pr-checks/255bb8b6-432b-490b-90c1-e2705588540d/license?refs=fc8888c1a787fd6fc5dcc1184dab2f1a90d93d76&source=prChecksComment)

lib/model/CResourceMonitor.cc
valeriy42, line 1: I think, this change is missing to update `CResrouceMonitor::totalMemory()`.  Please, correct me if I'm wrong, but the way I understand this code, `totalMemory()` on Linux should now simply return `systemMemoryUsage()`.

lib/model/CResourceMonitor.cc
```diff
@@ -378,11 +379,12 @@ void CResourceMonitor::sendMemoryUsageReport(core_t::TTime bucketStartTime,
 CResourceMonitor::SModelSizeStats
 CResourceMonitor::createMemoryUsageReport(core_t::TTime bucketStartTime) {
     SModelSizeStats res;
+    CSystemMemoryUsage systemMemoryUsage;
     res.s_Usage = this->totalMemory();
-    res.s_AdjustedUsage = this->adjustedUsage(res.s_Usage);
+    res.s_AdjustedUsage = systemMemoryUsage(this->adjustedUsage(res.s_Usage));
```
valeriy42: I don't think we need this complexity here. We can keep reporting adjusted usage the way we did before. But we need to update the value for  `s_Usage`.  If you change the logic in `totalMemory()`, you don't need to change anything here.

edsavage: Thanks for the comments Valeriy.

 What I was thinking was since `s_AdjustedUsage` and `s_AdjustedPeakUsage` are used as the values for the fields MODEL_BYTES and PEAK_MODEL_BYTES respectively - https://github.com/elastic/ml-cpp/blob/c56874a35d5f4ce4362ca37a8c72839be64e5abb/lib/api/CModelSizeStatsJsonWriter.cc#L57-L61 that on Linux these should simply both be set to `systemMemoryUsage()` and left as-is elsewhere. In doing so I was trying to avoid "adjusting" the system memory usage as it should not be necessary to do so.

That said, I'll adjust the code as you suggest and see how the results look.

valeriy42: This makes sense.

lib/api/CModelSizeStatsJsonWriter.cc
```diff
@@ -25,6 +25,8 @@ const std::string JOB_ID{"job_id"};
 const std::string MODEL_SIZE_STATS{"model_size_stats"};
 const std::string MODEL_BYTES{"model_bytes"};
 const std::string PEAK_MODEL_BYTES{"peak_model_bytes"};
+const std::string SYSTEM_MEMORY_BYTES{"system_memory_bytes"};
+const std::string MAX_SYSTEM_MEMORY_BYTES{"max_system_memory_bytes"};
```
valeriy42, line 29: AFAICS, this keys are not used anymore.

lib/model/CResourceMonitor.cc
```diff
@@ -382,6 +384,8 @@ CResourceMonitor::createMemoryUsageReport(core_t::TTime bucketStartTime) {
     res.s_PeakUsage = static_cast<std::size_t>(
         core::CProgramCounters::counter(counter_t::E_TSADPeakMemoryUsage));
     res.s_AdjustedPeakUsage = this->adjustedUsage(res.s_PeakUsage);
+    res.s_SystemMemoryUsage = core::CProcessStats::residentSetSize();
+    res.s_MaxSystemMemoryUsage = core::CProcessStats::maxResidentSetSize();
```
valeriy42: I have the feeling that the memory handling becomes unnecessary complicated and obfuscated. The way I see it, what we want to do is:
```
// First, make core::CProcessStats::residentSetSize() return TOptionalSize
res.s_AdjustedUsage = core::CProcessStats::residentSetSize().value_or(this->adjustedUsage(res.s_Usage));
res.s_AdjustedPeakUsage = core::CProcessStats::maxResidentSetSize().value_or(this->adjustedUsage(res.s_PeakUsage));

```

Then you don't need `res.s_SystemMemoryUsage` and `res.s_MaxSystemMemoryUsage` at all. 

Is this correct?

lib/model/CResourceMonitor.cc
```diff
@@ -486,10 +497,18 @@ std::size_t CResourceMonitor::lowLimit() const {
...
 
 std::size_t CResourceMonitor::totalMemory() const {
-    return m_MonitoredResourceCurrentMemory + m_ExtraMemory +
-           static_cast<size_t>(core::CProgramCounters::counter(
-               counter_t::E_TSADOutputMemoryAllocatorUsage));
+    CSystemMemoryUsage systemMemoryUsage;
+    return systemMemoryUsage(m_MonitoredResourceCurrentMemory + m_ExtraMemory +
+                             static_cast<size_t>(core::CProgramCounters::counter(
+                                 counter_t::E_TSADOutputMemoryAllocatorUsage)));
+}
```
valeriy42: It seems very complicated. Should this be something like
```
// First change the return value from core::CProcessStats::residentSetSize() to TOptionalSize
return core::CProcessStats::residentSetSize().value_or(m_MonitoredResourceCurrentMemory + m_ExtraMemory +
                             static_cast<size_t>(core::CProgramCounters::counter(
                                 counter_t::E_TSADOutputMemoryAllocatorUsage)));
```

lib/core/CProcessStats_Linux.cc
```diff
@@ -11,6 +11,7 @@
 #include <core/CProcessStats.h>
 
 #include <core/CLogger.h>
+#include <core/CProgramCounters.h>
```
valeriy42: Is this include used here?

lib/core/CProcessStats_MacOSX.cc
```diff
@@ -8,9 +8,11 @@
  * compliance with the Elastic License 2.0 and the foregoing additional
  * limitation.
  */
-#include <core/CLogger.h>
 #include <core/CProcessStats.h>
 
+#include <core/CLogger.h>
+#include <core/CProgramCounters.h>
```
valeriy42: Are these included needed here?

lib/core/CProcessStats_Windows.cc
```diff
@@ -8,8 +8,10 @@
  * compliance with the Elastic License 2.0 and the foregoing additional
  * limitation.
  */
-#include <core/CLogger.h>
 #include <core/CProcessStats.h>
+
+#include <core/CLogger.h>
+#include <core/CProgramCounters.h>
```
valeriy42: Are these included need here?

valeriy42 (commented):
I have a few suggestions to make the change more explicit.

lib/model/CResourceMonitor.cc
edsavage: Thanks Valeriy. Yes, correct that `res.s_SystemMemoryUsage` and `res.s_MaxSystemMemoryUsage` are not needed now. I'll remove them. 

Your suggestion of using `TOptionalSize` is an interesting one. I don't think it will work quite as you describe though since although `residentSetSize()` returns no value on Mac, it does on Windows, and `maxResidentSetSize()` returns a value on all platforms.

lib/core/unittest/CNamedPipeFactoryTest.cc
```diff
@@ -158,35 +158,36 @@ BOOST_AUTO_TEST_CASE(testServerIsCppWriter) {
...
+//     BOOST_TEST_REQUIRE(threadReader.start());
+//
+//     std::atomic_bool dummy{false};
+//     ml::core::CNamedPipeFactory::TFileP file{
+//         ml::core::CNamedPipeFactory::openPipeFileWrite(TEST_PIPE_NAME, dummy)};
+//     BOOST_TEST_REQUIRE(file);
+//
+//     sleep(1);
+//     std::size_t charsLeft{TEST_SIZE};
+//     std::size_t blockSize{7};
```
valeriy42: Don't forget to uncomment the code before merging.

lib/model/CResourceMonitor.cc
```diff
@@ -497,10 +504,9 @@ std::size_t CResourceMonitor::lowLimit() const {
 }
 
 std::size_t CResourceMonitor::totalMemory() const {
-    CSystemMemoryUsage systemMemoryUsage;
-    return systemMemoryUsage(m_MonitoredResourceCurrentMemory + m_ExtraMemory +
-                             static_cast<size_t>(core::CProgramCounters::counter(
-                                 counter_t::E_TSADOutputMemoryAllocatorUsage)));
+    return m_MonitoredResourceCurrentMemory + m_ExtraMemory +
+           static_cast<size_t>(core::CProgramCounters::counter(
+               counter_t::E_TSADOutputMemoryAllocatorUsage));
```
valeriy42: This doesn't seem right. You probably need to call `CResourceMonitor::applyMemoryStrategy(std::size_t usage)` or something along these lines here.

docs/CHANGELOG.asciidoc
```diff
@@ -33,6 +33,7 @@
 === Enhancements
 
 * Track memory used in the hierarchical results normalizer. (See {ml-pull}2831[#2831].)
+* Report the actual memory usage of the autodetect process. (See {ml-pull}2846[#2846])
```
valeriy42: You need to move it to 9.2.0 now

PR #2849, opened 2025-04-13
[ML] Fix compilation errors in debug build

edsavage:
Adjust TRACE statements - that are only active when assertions are enabled - to resolve compilation errors in debug builds.

Labelling as a `>non-issue` as this only affects `DEBUG` builds.


valeriy42 (approved):
LGTM

PR #2858, opened 2025-06-26
[ML] Bump version to 9.2

valeriy42:


catalog-info.yaml
```diff
@@ -166,24 +166,28 @@ spec:
       schedules:
```
jan-elastic, line 166: You need to add 9.1 to line 159 too.

valeriy42: done

PR #2864, opened 2025-09-30
[ML] Fix polynomial trend failure in forecasts

### LINKED ISSUE #2772: [ML] Forecasting trend is unexpected given data
We observed that forecasting can exhibit unexpected and unintuitive trends given data with interruptions:
Image
or temporary decrease:
Image

I was able to reproduce the behavior using synthetic data:
Image

May be related to https://github.com/elastic/ml-cpp/issues/513

valeriy42:
This commit updates the CTimeSeriesDecompositionDetail class to reset prediction error statistics and clear the trend model when significant changes in seasonal components are detected. This ensures that the model does not rely on outdated polynomial trend extrapolations when seasonality re-emerges.

Fixes #2772


tveasey:
I think this is superseded now right?

PR #2866, opened 2025-10-07
[ML] Handle noise overfitting issue

valeriy42:
- Implement testPiecewiseLinearNoiseOverfitting test in CTimeSeriesSegmentationTest
- Test validates that piecewise linear decomposition doesn't overfit on pure noise data
- Similar to CTrendComponentTest/testModelBoundsWithFlatDataAndNoise but for segmentation
- Tests multiple noise levels (0.5, 1.0, 2.0, 5.0) and p-values (1e-3, 1e-4, 1e-5)
- Ensures minimal segmentation for pure noise data (should be just start and end points)
- Validates that variance reduction is minimal to prevent overfitting to noise
- Test passes successfully, confirming robust segmentation behavior

### DIFF
--- a/include/maths/time_series/CTrendComponent.h
+++ b/include/maths/time_series/CTrendComponent.h
@@ -277,6 +277,10 @@ private:
     //! Uses variance ratio as an indicator of autocorrelation to prevent overconfidence.
     double calculateEffectiveSampleSize() const;
 
+    //! Calculate adaptive BIC penalty based on data characteristics.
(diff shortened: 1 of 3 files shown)


valeriy42:
complete_model_predictions

For the test `CTrendComponentTest/testForecastPiecewiseLinear`, the modeling after introducing BIC doesn't seem unreasonable to me, although the old test threshold was exceeded.

PR #2873, opened 2025-10-28
[ML] Add Sandbox2 security integration for PyTorch inference

valeriy42:
PyTorch inference runs untrusted TorchScript models supplied by Elasticsearch users, so the native process must be treated as hostile. Today `pytorch_inference` relies largely on in-process seccomp and graph validation, but that still leaves a large syscall and filesystem surface exposed inside the same address space as libtorch. This PR hardens the Linux production path by spawning `pytorch_inference` inside Google Sandbox2 from the ML controller, so isolation, syscall policy, and filesystem access are enforced before the model binary starts executing.

## Architecture

- **`MlSandbox`** (`lib/sandbox/`): owns `CSandboxedProcessSpawner`, `CPytorchInferenceSandboxPolicy`, and `CSandbox2Diagnostics`. `SANDBOX2_AVAILABLE` is scoped to this library, not `MlCore`.
- **`CProcessSpawnerRouter`** (`bin/controller/`): controller-owned routing layer. Nominates `./pytorch_inference` explicitly, strips `--disableSandbox`, and chooses sandboxed vs legacy spawn based on the ES kill switch (`xpack.ml.trained_models.sandbox_enabled`).
- **`CDetachedProcessSpawner`** (`lib/core/`): generic POSIX `posix_spawn` spawner again — no `pytorch_inference`-specific dispatch or Sandbox2 knowledge.
- **`pytorch_inference`**: runtime-only seccomp gate on Linux when `ML_SANDBOXED=1`; skips in-process BPF when running under Sandbox2.

Build support vendors Abseil and the Sandboxed API on Linux, wires them into CMake, and adds the required license files.

## Kill switch

When sandboxing cannot be enforced (e.g. user namespaces unavailable) or the operator disables it, the controller falls back to the legacy `posix_spawn` + in-process seccomp path via `--disableSandbox`.

## Testing

- **`lib/sandbox/unittest`**: spawn, allowlist drift guard, fail-closed behaviour, and policy-violation differential tests.
- **Coverage modes**: the sandbox suite probes what the runner's kernel permits and covers `enforced` (real sandboxed spawn/terminate plus the filesystem-policy differential) or `fail_closed` (spawn refusal plus kill-switch hint) accordingly. The allowlist drift guard is mode independent and always runs. `ML_SANDBOX2_REQUIRE=enforced` pins the aarch64 host runner so enforced coverage cannot silently degrade; the container run is left unpinned and supplies the fail-closed half.
- **`.buildkite/scripts/steps/diagnose_userns.sh`**: reports which user-namespace stage a runner denies, on the host and across container configurations, without needing a build. Kept as a regression canary for the pin above.
- **`test/test_sandbox2_attack_defense.py`**: end-to-end attack-defense harness.
- Extended evil-model tests; seccomp filter kept coherent with Sandbox2 policy via `CPytorchInferenceSyscallAllowlist.h`.

## Review sequence (post-restructure)

1. `lib/sandbox/CSandboxedProcessSpawner_Linux.cc` + `CPytorchInferenceSandboxPolicy_Linux.cc` — spawn and policy.
2. `bin/controller/CProcessSpawnerRouter.cc` — kill-switch routing.
3. `lib/core/CDetachedProcessSpawner.cc` — generic legacy spawn baseline.
4. `lib/seccomp/CSystemCallFilter_Linux.cc` + `include/seccomp/CPytorchInferenceSyscallAllowlist.h` — legacy vs Sandbox2 syscall coherence.
5. `bin/pytorch_inference/Main.cc` — runtime seccomp gate.
6. `3rd_party/CMakeLists.txt`, `lib/sandbox/CMakeLists.txt`, `lib/core/CMakeLists.txt` — build wiring.
7. Tests: `lib/sandbox/unittest/CSandboxedProcessSpawnerTest_Linux.cc` — the user-namespace probe and the enforced/fail-closed dispatch it drives; `test/test_sandbox2_attack_defense.py`.
8. CI: `.buildkite/scripts/steps/run_tests.sh` (where each mode is covered), `diagnose_userns.sh` + `userns_probe.sh`.
9. Docs: `docs/changelog/2873.yaml`, `docs/sandbox2_production_failure_modes.md`.

## ES companion PRs

- https://github.com/elastic/elasticsearch/pull/154825
- https://github.com/elastic/elasticsearch/pull/154797


valeriy42:
buildkite run_serverless_tests

valeriy42:
buildkite run_qa_tests

valeriy42:
buildkite run_qa_tests

valeriy42:
buildkite run_qa_tests  QAF_TESTS_TO_RUN=pytorch_tests

valeriy42:
buildkite run_qa_tests

valeriy42:
buildkite run_pytorch_tests

valeriy42:
buildkite run_pytorch_tests

edsavage:
Hi Valeriy, here's my high level first pass. As you suggested I'm starting the review with `CDetachedProcessSpawner_Linux.cc`. I'm happy overall with the security intent and the QA-driven policy fixes look solid. My main concerns are structure and layering rather than the Sandbox2 approach itself.

### Layering - `pytorch_inference` in `lib/core`
`CDetachedProcessSpawner` is a generic permitted-path spawner, but the Linux implementation is deeply coupled to just the one executable (`pytorch_inference` appears throughout: dispatch via `find("pytorch_inference")`, policy building, `ML_SANDBOXED`, log text, etc.). That just doesn’t feel appropriate to belong in `lib/core`.

If Sandbox2 is genuinely PyTorch specific for now, the policy/dispatch shouldn’t live as hard-coded knowledge inside this library. Maybe we could devise something where the controller (or a dedicated helper it owns) decides Sandbox2 vs legacy spawn and supplies the policy, without `core` naming a particular binary. The “allowlist before substring check” comment mitigates a bypass, but it doesn’t fix the layering.

### The size of  `spawn(... failureReason)`
This overload is very large (~300+ lines). Please consider splitting it, e.g. kill-switch/arg normalisation, sandboxed spawn path, and legacy `posix_spawn` path (policy builder is already partly extracted, which helps). That would also make a later relocation of the Sandbox2 path easier.

### Duplication
`CTrackerThread` and `setupFileActions` are largely duplicated across the platform spawner files (with Linux adding sandbox PID tracking). I think it'd be worth extracting the shared POSIX pieces rather than growing a third near-copy. I’d treat that as spawner-internal sharing, not a general-purpose thread utility.

The same for the local `joinStrings` helpers — `CStringUtils::join` already exists; the `"(none)"` empty case can wrap that.

### Smaller nits
- `TFilteredEnviron` -> `SFilteredEnviron` (or a small `C` class): `T` is for type aliases per the style guide.
- `extractArgDirs` / spawn-context mount logging: These belong with the sandbox policy, not as generic spawner free functions. I'd strongly prefer named constants (and reuse the fixed-mount helpers) so that the log text can’t drift from the policy.
- In a file this `#ifdef`-heavy, labelling matching `#endif`s (`// SANDBOX2_AVAILABLE`, etc.) would help navigation.
- Optional: compare `ML_SANDBOXED` to `"1"` in `pytorch_inference` rather than “any value set”.

edsavage (commented):
Continuing on from the `CDetachedProcessSpawner_Linux.cc` notes. I'm reviewing the rest of the changes in the  recommended review order (baseline -> seccomp -> pytorch -> controller/IO -> build > tests > docs). Tip reviewed: `6fcdecd30`.

### `CDetachedProcessSpawner.cc` (non-Linux)
The `--disableSandbox` stripping and the `failureReason` overload look good. There is one gap vs Linux: `failureReason` is set for allowlist/`access` failures, but **not** for `setupFileActions` / `posix_spawnattr_init` / `posix_spawn` failures (those only `LOG_ERROR`). So the controller START responses on macOS will be less informative for those cases.

### Seccomp + `CPytorchInferenceSyscallAllowlist.h`
The BPF jump updates for `__NR_dup` look consistent, and `sandbox2AllowsAllLegacySyscalls()` is a nice drift guard. One clarification: the check asserts **Sandbox2 is a superset of the legacy BPF**, not full equality (Sandbox2 also allows epoll/renameat/prlimit64/etc.). The “keep in sync” wording could say that explicitly so future editors don’t assume a 1:1 list.

### `pytorch_inference` Main + cmdline
- Seccomp is skipped only when sandboxed. The kill-switch / non-Linux paths still install the filter - this looks correct.
- Please compare `ML_SANDBOXED` to `"1"` rather than `getenv(...) != nullptr`, so that a stray `ML_SANDBOXED=0` can’t disable both Sandbox2 and seccomp. The controller already sets `=1`.
- The argv[0]-is-option workaround in `CCmdLineParser` is fragile. I'd prefer fixing the Executor argv construction (and documenting the Sandbox2 quirk) so this special case can go away. As a related note - inside the sandbox spawn branch, `fullArgs` is built from `args` rather than `effectiveArgs`. These are currently equivalent, but `effectiveArgs` would be clearer.
- `SANDBOX2_DISABLED` is referenced in `Main.cc` guards but I don’t see it defined anywhere in the repo.  Is this an intentional leftover escape hatch, or dead?

It's probably also worth flagging for awareness (related to #3098) the  gating of `--skipModelValidation` behind `ML_ALLOW_SKIP_MODEL_VALIDATION` (default OFF) means distributed builds won’t recognise the flag ES may still send when graph validation is disabled. It's worth confirming the intended prod behaviour with the ES companion PRs before merge.

### Controller + `CIoManager`
The propagation of spawn failure reasons into START responses is a clear ops win, and the per-pipe `CIoManager` errors will help with FIFO/mount-namespace failures. The `propertiesFile` on the controller looks orthogonal to sandboxing - I couldn't quite see why it's necessary - a one-line note on why it’s in this PR would help. The `${SANDBOX2_LIBRARIES}` on the controller stays empty on the non-Linux path looks right to me.

### Build wiring
The Linux FATAL_ERROR if Sandbox2 isn’t built / unity disabled around Abseil/SAPI / licenses/pins look good.

I'd suggest to **save/restore `BUILD_SHARED_LIBS`** in `3rd_party/CMakeLists.txt` the same way as `BUILD_TESTING` / `CMAKE_UNITY_BUILD` - it’s currently forced OFF and left that way for the rest of configure. Configure-time `string(REPLACE)` / regex patches to SAPI CMake are acceptable while the GIT_TAG is pinned; a short comment block (or `*.patch` files) listing “required patches + why” would make the next bump safer.

### Tests
As far as I can tell `CDetachedProcessSpawnerTest_Linux.cc` coverage looks solid (allowlist, substring bypass, kill-switch symlink, sandboxed start). The attack-defense harness is explicitly **manual / not in CI** - can you please confirm it was run on this tip, and consider a slim CI smoke test (benign start + one denied write) later if feasible. Leaving policy-violation coverage only in a manual harness is the main residual test gap.

### Docs / cross-cutting questions
- The changelog summary is fine. I'd consider mentioning Linux-only + the `sandbox_enabled` / `--disableSandbox` kill switch for support.
- **Networking:** policy allows `__NR_connect` without an explicit network-allow helper. Is the Sandbox2 netns empty by default so IP egress is still blocked? If networking isn’t needed, dropping `connect` (and documenting netns isolation) would shrink the surface.
- **FS mounts:** RO `/etc` + `/sys` plus RW parent dirs from absolute `key=/path` args: Can you please confirm why `/etc`/`/sys` are required, and that ES only passes dedicated pipe dirs into those args.

These are all minor request / nits. My main concerns are the layering / size work from the earlier comment.

valeriy42:
Thanks for the thorough review, Ed. Addressed below by theme.

## `--skipModelValidation` (regression)

You were right to flag this. The `ML_ALLOW_SKIP_MODEL_VALIDATION` gate in this PR was a **regression**: on `main` the flag is registered unconditionally, but the PR wrapped it in `#ifdef` with default OFF and no build ever enabling it — so production binaries would reject `--skipModelValidation` and break `graph_validation_enabled: false`. Reverted to `main` behaviour; design discussion continues in #3098.

## Layering / structure

Sandbox2 dispatch and policy are relocated out of `lib/core`:

- New **`MlSandbox`** static library (`lib/sandbox/`) owns `CSandboxedProcessSpawner`, `CPytorchInferenceSandboxPolicy`, and `CSandbox2Diagnostics`. `SANDBOX2_AVAILABLE` is scoped here, not on `MlCore`.
- **`CDetachedProcessSpawner`** is generic POSIX spawn again (no `pytorch_inference` substring dispatch).
- **`CProcessSpawnerRouter`** in the controller nominates `./pytorch_inference` explicitly, strips `--disableSandbox`, and routes to sandbox vs legacy spawner.

This also subsumes the ~300-line `spawn()` split you suggested.

## Accepted nits

- `ML_SANDBOXED` compared to `"1"` (runtime-only gate in `pytorch_inference`; removed dead `SANDBOX2_DISABLED` / `SANDBOX2_AVAILABLE` compile guards).
- Removed dead `argv[0]`-is-option workaround in `CCmdLineParser` (Executor already passes `processPath` at index 0).
- `SFilteredEnviron`, `CStringUtils::join`, labelled `#endif`s, `BUILD_SHARED_LIBS` save/restore, unconditional `BUILD_TESTING` restore, SAPI patch rationale comments.
- Superset wording on syscall allowlist drift guard.
- Changelog mentions Linux-only + kill switch.
- Dropped controller `propertiesFile` (only used by manual harness).
- macOS `failureReason` on `posix_spawn` setup failures.

## CI / test coverage (previously vacuous)

The replacement unit test never actually ran: wrong binary path relative to CTest `WORKING_DIRECTORY`, and silent `BOOST_TEST_MESSAGE` skips.

Fixed with:
- New **`lib/sandbox/unittest`** suite registered in parallel test runner.
- **`ML_SANDBOX2_EXPECT=enforced`** on aarch64 (re-run outside Docker, following seccomp precedent).
- **`ML_SANDBOX2_EXPECT=fail_closed`** in x86_64 Docker entrypoint.
- Hard-fail if expectation absent or contradicts `unshare(CLONE_NEWUSER)` probe.
- Differential policy-violation test (unsandboxed write succeeds; sandboxed write blocked).
- Fail-closed test asserts spawn failure names the kill switch.

Manual attack-defense harness was run on earlier tips; CI now has enforcing coverage on aarch64 and fail-closed assertions on x86_64.

## Push back (with evidence)

**`__NR_connect`:** Sandbox2 default namespace includes `CLONE_NEWNET` and we never pass `UnrestrictedNetworking`, so egress is blocked despite `connect` being allowed. `__NR_socket` is not on the allowlist. Dropping `connect` from Sandbox2 alone would break `sandbox2AllowsAllLegacySyscalls()`; changing both filters is a separate PR. Added netns-isolation comment in policy.

**`argv[0]` workaround:** Already fixed at Executor construction; deleted the workaround rather than documenting it.

**`/etc` + `/sys` mounts:** Required for glibc (`ld.so.cache`, `nsswitch.conf`, `localtime`) and libtorch/OpenMP CPU topology under `/sys/devices/system/cpu`. RW mounts come only from absolute `key=/path` pipe args via `extractArgDirs`. Narrowing mounts is a follow-up with its own QA cycle.

## Still open / follow-up

- Slim CI policy-violation smoke in Docker if aarch64-outside-Docker proves flaky on some agents.
- Follow-up to narrow `/etc`/`/sys` bind mounts after QA.

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+    for (const auto& value : values) {
+        if (joined.empty() == false) {
+            joined += ',';
+        }
+        joined += value;
+    }
+    return joined.empty() ? "(none)" : joined;
+}
+
+std::string joinStrings(const std::vector<std::string>& values) {
```
jan-elastic: let's do
```
  template <typename Container>
  std::string joinStrings(const Container& values)
```
instead of the duplication

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+    return joined.empty() ? "(none)" : joined;
+}
+
+struct SArgDirExtraction {
+    std::set<std::string> m_ArgDirs;
+    std::vector<std::string> m_RejectedPipeArgs;
+    std::vector<std::string> m_PipeDirAliasMappings;
+};
+
+SArgDirExtraction extractArgDirs(const std::vector<std::string>& args) {
```
jan-elastic: please document what this method does

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+    SArgDirExtraction extraction;
+    for (const auto& arg : args) {
```
jan-elastic: why auto?

I think
```
for (const string& ...)
```
is a lot easier to read

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+        }
+
+        if (arg[eqPos + 1] != '/') {
+            extraction.m_RejectedPipeArgs.push_back(arg + " (path not absolute)");
+            continue;
+        }
+
+        std::string path = arg.substr(eqPos + 1);
+        size_t lastSlash = path.rfind('/');
+        if (lastSlash == std::string::npos || lastSlash == 0) {
```
jan-elastic: `npos` isn't possible, because it starts with a slash

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+              << sandbox2::Result::StatusEnumToString(result.final_status())
+              << ", reason_code=" << result.reason_code() << ']';
+    if (result.final_status() == sandbox2::Result::VIOLATION) {
+        formatted << " [syscall=" << result.reason_code()
+                  << " arch=" << sandboxPlatformArch() << ']';
+    }
+    return formatted.str();
+}
+#endif
+
```
jan-elastic: why close `#ifdef SANDBOX2_AVAILABLE` and open a new one?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+    return true;
+}
+}
+
+namespace ml {
+namespace core {
+
+namespace detail {
+
+class CTrackerThread : public CThread {
```
jan-elastic: please document purpose of `CTrackerThread` and what it does

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+        std::string originalTmpdir;
+        const char* tmpdir = ::getenv("TMPDIR");
+        if (tmpdir != nullptr) {
+            originalTmpdir = tmpdir;
+        }
+
+        // Sandbox2 forkserver uses Unix sockets with 108 char path limit
+        bool tmpdirOverridden{false};
+        if (tmpdir != nullptr && ::strlen(tmpdir) > 80) {
+            LOG_WARN(<< "TMPDIR path too long for Sandbox2 forkserver Unix socket (108 char limit),"
```
jan-elastic: Why fail on > 80 if the limit is 108?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+                     << " temporarily overriding to /tmp (len=" << ::strlen(tmpdir)
+                     << ", value=" << tmpdir << ")");
```
jan-elastic: Instead of overriding tmpdir, shouldn't you just stop with an error?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+        std::vector<std::string> fullArgs;
+        fullArgs.reserve(args.size() + 1);
+        fullArgs.push_back(processPath);
+        for (const auto& arg : args) {
+            fullArgs.push_back(arg);
+        }
+
+        // Get binary and library directories
+        std::string binDir = absPath.substr(0, absPath.rfind('/'));
+        std::string libDir = binDir.substr(0, binDir.rfind('/')) + "/lib";
```
jan-elastic: this is just `binDir + "/lib"` if I'm not mistaken

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+            assignFailureReason(failureReason, reason);
+            return false;
+        }
+
+        // Create executor with a sandbox marker and restored TMPDIR when needed.
+        std::vector<std::string> customEnv;
+        bool sandboxMarkerSet{false};
+        const char* currentTmpdir = ::getenv("TMPDIR");
+        const bool restoreTmpdir = !originalTmpdir.empty() &&
+                                   (currentTmpdir == nullptr || originalTmpdir != currentTmpdir);
```
jan-elastic: can't you just use `bool tmpdirOverridden`?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+        m_TrackerThread->addSandboxPid(childPid);
+
+        // The sandboxee is a child of the Sandbox2 forkserver rather than of the
+        // controller, so the tracker's waitpid() never sees it. Own the sandbox
+        // instance on a dedicated thread that keeps it alive for the lifetime of
+        // pytorch_inference, waits for its result, logs termination the same way
+        // checkForDeadChildren() does for regular children, and removes the PID
+        // from the tracker so PID reuse cannot make terminateChild() signal an
+        // unrelated process.
+        {
```
jan-elastic: why is this inside {} ?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+#else
+        // On Linux, 3rd_party/CMakeLists.txt fails configuration if sandbox2::sandbox2
+        // cannot be built, and MlCore defines SANDBOX2_AVAILABLE whenever that target
+        // exists. This branch is therefore a compile-time safety net only.
+        LOG_ERROR(<< "pytorch_inference built without Sandbox2 support - cannot spawn securely"
+                  << SANDBOX2_DISABLE_HINT);
+        assignFailureReason(failureReason, "pytorch_inference built without Sandbox2 support - cannot spawn securely" +
+                                               SANDBOX2_DISABLE_HINT);
+        return false;
+#endif
```
jan-elastic: I would annotate these long/nested `#ifdef` blocks

so here:
```
#endif  // SANDBOX2_AVAILABLE
    }
#endif  // __linux__
```

no I have to scroll up to understand the flow.

Or refactor everything inside
```
 if (processPath.find("pytorch_inference") != std::string::npos && disableSandbox == false)
```
in a method `spawnSandboxedPytorch`. This method is very long.

lib/core/CDetachedProcessSpawner.cc
```diff
@@ -241,35 +243,62 @@ CDetachedProcessSpawner::~CDetachedProcessSpawner() {
...
     }
 
+    // Operator kill switch: Elasticsearch may append --disableSandbox when
+    // xpack.ml.trained_models.sandbox_enabled=false. Sandbox2 exists only on
+    // Linux; on macOS the controller simply consumes the flag so it never
+    // reaches pytorch_inference (which does not register it).
+    TStrVec effectiveArgs;
+    effectiveArgs.reserve(args.size());
+    for (const auto& arg : args) {
+        if (arg != "--disableSandbox") {
```
jan-elastic: can we define this constant nowhere? it's used across multiple files

bin/pytorch_inference/CCmdLineParser.cc
```diff
@@ -71,21 +71,39 @@ bool CCmdLineParser::parse(int argc,
                         "Optionaly set number of allocations to parallelize model forwarding - default is 1")
             ("cacheMemorylimitBytes", boost::program_options::value<std::size_t>(),
                         "Optional memory in bytes that the inference cache can use - default is 0 which disables caching")
-            ("validElasticLicenseKeyConfirmed", boost::program_options::value<bool>(),
+            ("validElasticLicenseKeyConfirmed", boost::program_options::value<bool>()->implicit_value(true),
```
jan-elastic: why is this?

bin/pytorch_inference/CCmdLineParser.cc
```diff
@@ -71,21 +71,39 @@ bool CCmdLineParser::parse(int argc,
...
 
         boost::program_options::variables_map vm;
-        boost::program_options::parsed_options parsed =
-            boost::program_options::command_line_parser(argc, argv)
-                .options(desc)
-                .run();
-        boost::program_options::store(parsed, vm);
+        // Workaround for Sandbox2: if argv[0] is an option (Sandbox2 sets it incorrectly),
+        // parse it as an option using a vector of strings
+        if (argc > 0 && std::string(argv[0]).substr(0, 2) == "--") {
```
jan-elastic: I think c++20 has `starts_with("--")`

bin/pytorch_inference/CCmdLineParser.cc
```diff
@@ -71,21 +71,39 @@ bool CCmdLineParser::parse(int argc,
...
         ;
         // clang-format on
 
         boost::program_options::variables_map vm;
-        boost::program_options::parsed_options parsed =
-            boost::program_options::command_line_parser(argc, argv)
-                .options(desc)
-                .run();
-        boost::program_options::store(parsed, vm);
+        // Workaround for Sandbox2: if argv[0] is an option (Sandbox2 sets it incorrectly),
```
jan-elastic: that's strange? it that a known bug? it so, please add a reference for tracking?

lib/core/CDetachedProcessSpawner_Linux.cc
```diff
@@ -0,0 +1,946 @@
...
+        formatted << " [syscall=" << result.reason_code()
+                  << " arch=" << sandboxPlatformArch() << ']';
+    }
+    return formatted.str();
+}
+#endif
+
+#ifdef SANDBOX2_AVAILABLE
+//! Builds the syscall and filesystem policy for a sandboxed pytorch_inference.
+//! Keep the syscall allowlist in sync with lib/seccomp/CSystemCallFilter_Linux.cc.
```
jan-elastic: The file is called: CPytorchInferenceSyscallAllowlist.h

jan-elastic (approved):
LGTM. This seems super solid and a huge improvement over the current security model.

Left a bunch of small comments, but nothing blocking.

For the record: I didn't completely parse the syscall lists (I don't know what half of them are, nor why they're needed). Should I invest some time in that?

valeriy42:
@edsavage — follow-up now that the restructuring is pushed (`3dc4a5617`).

The Jul 28 reply described the intended changes; this commit lands them on `enhancement/sandbox2`. Summary for re-review:

**Layering (your main concern)**
- Deleted `lib/core/CDetachedProcessSpawner_Linux.cc` (~950 lines of pytorch-specific Sandbox2 logic).
- New **`MlSandbox`** library: `CSandboxedProcessSpawner`, `CPytorchInferenceSandboxPolicy`, `CSandbox2Diagnostics`.
- **`CProcessSpawnerRouter`** in the controller owns sandbox vs legacy routing and explicitly nominates `./pytorch_inference` — `lib/core` no longer names any binary.

**`spawn()` size / split**
- Sandboxed spawn lives in `CSandboxedProcessSpawner_Linux.cc`; generic POSIX spawn stays in `CDetachedProcessSpawner.cc`.

**Duplication**
- `CTrackerThread` / `setupFileActions` sharing across platform spawners was **not** extracted in this pass — left as follow-up to keep the relocation diff reviewable. `joinStrings` → `CStringUtils::join` was done.

**Nits from your comment**
- `SFilteredEnviron`, labelled `#endif`s, spawn-context logging colocated with policy builder, `ML_SANDBOXED == "1"` runtime gate, changelog wording, dropped controller `propertiesFile`.

**Tests / CI**
- `lib/sandbox/unittest` with `ML_SANDBOX2_EXPECT=enforced|fail_closed` (hard-fail if missing or mismatched).
- aarch64: enforced re-run outside Docker; x86_64 Docker: `fail_closed`.
- Differential policy-violation test added.

**Regression fix**
- `--skipModelValidation` reverted to unconditional `main` registration (the `ML_ALLOW_SKIP_MODEL_VALIDATION` gate was a production break).

PR description updated to match the new layout. Happy to tackle `CTrackerThread` extraction as a follow-up if you still want it before merge.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Done — replaced the duplicated helpers with a single `joinForLog` template delegating to `CStringUtils::join` in `lib/sandbox/CPytorchInferenceSandboxPolicy_Linux.cc`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Done — `extractArgDirs` is documented in `include/sandbox/CPytorchInferenceSandboxPolicy.h` (moved with the MlSandbox extraction).

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Leaving as `const auto&` here — the container element type is already explicit in the function signature (`const std::vector<std::string>& args`), and `auto` deduces to `const std::string&`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Agreed — simplified to `if (lastSlash == 0)` in both `CPytorchInferenceSandboxPolicy.cc` and `_Linux.cc`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Resolved by the MlSandbox extraction — policy code now lives in a single `#ifdef SANDBOX2_AVAILABLE` block in `CPytorchInferenceSandboxPolicy_Linux.cc`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Sandbox PID tracking no longer uses `CTrackerThread`; it moved to `CSandboxedProcessSpawner` with a detached `AwaitResult` monitor thread, documented in `include/sandbox/CSandboxedProcessSpawner.h`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Removed the TMPDIR override entirely in 6f5766177. The pinned Sandbox2 forkserver comms use `socketpair(AF_LOCAL, ...)`, not filesystem `sun_path`, so the 80/108-char threshold was unsupported.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Same commit — no override anymore; the controller environment (including TMPDIR) passes through to the sandboxee unchanged, with only `ML_SANDBOXED=1` added.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: `bin/` and `lib/` are siblings under the platform root (`$ROOT/bin` and `$ROOT/lib`), so `parent(binDir) + "/lib"` is correct. `binDir + "/lib"` would resolve to `$ROOT/bin/lib`, which does not exist in our layout.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Addressed when the TMPDIR workaround was removed — the `tmpdirOverridden` flag and related logic are gone from `CSandboxedProcessSpawner_Linux.cc`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: The inner `{}` intentionally limits the scope of `sandboxPid` and the moved `sandboxPtr` capture before `return true`. No functional issue.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Resolved by the MlSandbox restructuring — sandbox spawn is now `CSandboxedProcessSpawner::spawn()` in `lib/sandbox/CSandboxedProcessSpawner_Linux.cc`, and the `#endif // SANDBOX2_AVAILABLE` is labelled.

lib/core/CDetachedProcessSpawner.cc
valeriy42: `--disableSandbox` stripping is centralized in `bin/controller/CProcessSpawnerRouter.cc`; core no longer parses it.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: Reverted to `value<bool>()` without `implicit_value(true)` to match `main` and the ES companion PR (`--validElasticLicenseKeyConfirmed=true`). The flag-only form was a sandbox-era workaround, no longer needed.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: The argv workaround that used `substr(0, 2) == "--"` was removed entirely in 3dc4a5617; the spawner now supplies a real program path as `argv[0]`.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: Same — workaround deleted rather than documented. `CSandboxedProcessSpawner_Linux.cc` builds `fullArgs[0] = processPath`, so the parser no longer needs a special case.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Updated the sync comment in `include/sandbox/CPytorchInferenceSandboxPolicy.h` to reference `include/seccomp/CPytorchInferenceSyscallAllowlist.h` as the canonical list.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Same commit — no override anymore; the controller environment (including TMPDIR) passes through to the sandboxee unchanged, with only `ML_SANDBOXED=1` added.

valeriy42: test

valeriy42: test

valeriy42: Same commit — no override anymore; the controller environment (including TMPDIR) passes through to the sandboxee unchanged, with only `ML_SANDBOXED=1` added.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: `bin/` and `lib/` are siblings under the platform root (`$ROOT/bin` and `$ROOT/lib`), so `parent(binDir) + "/lib"` is correct. `binDir + "/lib"` would resolve to `$ROOT/bin/lib`, which does not exist in our layout.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Addressed when the TMPDIR workaround was removed — the `tmpdirOverridden` flag and related logic are gone from `CSandboxedProcessSpawner_Linux.cc`.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: The inner `{}` intentionally limits the scope of `sandboxPid` and the moved `sandboxPtr` capture before `return true`. No functional issue.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Resolved by the MlSandbox restructuring — sandbox spawn is now `CSandboxedProcessSpawner::spawn()` in `lib/sandbox/CSandboxedProcessSpawner_Linux.cc`, and the `#endif // SANDBOX2_AVAILABLE` is labelled.

lib/core/CDetachedProcessSpawner.cc
valeriy42: `--disableSandbox` stripping is centralized in `bin/controller/CProcessSpawnerRouter.cc`; core no longer parses it.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: Reverted to `value<bool>()` without `implicit_value(true)` to match `main` and the ES companion PR (`--validElasticLicenseKeyConfirmed=true`). The flag-only form was a sandbox-era workaround, no longer needed.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: The argv workaround that used `substr(0, 2) == "--"` was removed entirely in 3dc4a5617; the spawner now supplies a real program path as `argv[0]`.

bin/pytorch_inference/CCmdLineParser.cc
valeriy42: Same — workaround deleted rather than documented. `CSandboxedProcessSpawner_Linux.cc` builds `fullArgs[0] = processPath`, so the parser no longer needs a special case.

lib/core/CDetachedProcessSpawner_Linux.cc
valeriy42: Updated the sync comment in `include/sandbox/CPytorchInferenceSandboxPolicy.h` to reference `include/seccomp/CPytorchInferenceSyscallAllowlist.h` as the canonical list.

edsavage (approved):
LGTM

A few non-blocking follow-ups, just so they don't get lost:
  1. `CTrackerThread / setupFileActions` dedup across the macOS/Windows spawners.
  2. Narrowing the /etc, /sys bind mounts and reconsidering `__NR_connect` once the drift guard can be updated in tandem. 
  3. The slim in-Docker policy-violation smoke, if the aarch64-outside-Docker enforcing test proves flaky on any agents.

One thing worth confirming before merge: has the aarch64 enforcing run been stable across your CI agents?

valeriy42:
buildkite run_qa_tests

valeriy42:
**Frozen at `db5b14ac8`** — this PR is being replaced rather than rebased.

Last green integration evidence: Buildkite #3081.

This branch will not be rebased, force-pushed, or merged, and its history is preserved. The Sandbox2 integration is being reconstructed from current `main` as a sequential, independently reviewable PR series:

| PR | Repo | Scope |
|---|---|---|
| A | ml-cpp | Dormant Sandbox2/Abseil dependency, licences, checked-in version-pinned patches, dormant `MlSandbox` target. No production routing. |
| B | ml-cpp | One machine-readable syscall declaration generating both the applied BPF allowlist and the Sandbox2 grants; typed seccomp install result and readiness marker. |
| C | ml-cpp | Typed internal launch specification, single validated `0700` per-child IPC root, minimal mounts, private tmpfs, allowlisted mechanism probe. |
| D | ml-cpp | Explicit child lifecycle state machine, kill-and-reap guard, single `AwaitResult` owner, injectable seams, classified `pidfd_open` outcomes. |
| E | ml-cpp | Typed internal routing on the retained `start` protocol, pinned enforced/fail-closed CI on both architectures, repaired attack-defense harness. |
| F | elasticsearch | Default-`false` operator setting, unique child IPC paths, unchanged wire serialization, `PyTorchSandboxIT`. |

Why replace rather than repair: the recurring defects in this branch share one cause — there was no explicit capability graph and no resource lifecycle state machine, while tests frequently asserted an outcome without proving the mechanism that produced it. Slicing the work lets each boundary be reviewed against a stated contract instead of inferred from a large diff. Security logic is rebuilt from that contract rather than bulk-cherry-picked; only audited vendoring and build material is transplanted mechanically.

The established tab-separated `start` command between Elasticsearch and the controller is unchanged — no new verb, wire format, or response JSON. Sandbox2 ships default-off behind the operator setting `xpack.ml.trained_models.sandbox_enabled` and is activated through a staged rollout.

This PR stays open and will be closed as superseded once every replacement PR is linked here.

PR #2884, opened 2025-11-25
[ML] Remove cross-compilation step

valeriy42:
We maintained a cross-compilation step for the arm64 architecture on the Linux x86_64 host machine to build binaries for the Intel macOS platform. Since we no longer support this platform, this build step and the associated Docker images have become burdensome to maintain. The only remaining benefit was that the build and test were run in debug mode, allowing us to catch build errors that would otherwise be optimized away by the compiler.

This PR removes support for the cross-compilation step and introduces a build step in debug mode for the x86_64 architecture.

### DIFF
--- a/.buildkite/pipelines/build_linux.json.py
+++ b/.buildkite/pipelines/build_linux.json.py
@@ -87,6 +87,41 @@ def main(args):
                 ],
             })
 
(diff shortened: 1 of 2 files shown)


edsavage (approved):
LGTM

PR #2896, opened 2026-01-27
[ML] Add pytorch_process rss memory stat to inference response

darius-vil:
The idea is to:
1. Capture the stat in PytorchResultProcessor: https://github.com/elastic/elasticsearch/blob/01ace3927df065f1caf090653404f29688e8103a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/process/PyTorchResultProcessor.java#L271
2. Make it available to `/_ml/trained_model/_stats` [doc](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-get-trained-models-stats), which gets the stats from `PytorchResultProcessor.getStats()`: https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/process/PyTorchResultProcessor.java#L231


valeriy42:
superseded by #3160

PR #2991, opened 2026-03-13
[ML] Add quantized model ops to pytorch_inference allowlist

edsavage:
## Summary

- Adds `aten::mul_` and `quantized::linear_dynamic` to the `ALLOWED_OPERATIONS` set in `CSupportedOperations.cc`, fixing model graph validation failures for dynamically quantized models (e.g. ELSER v2 imported via Eland with `torch.quantization.quantize_dynamic`).
- Updates the model extraction tooling (`extract_model_ops.py`, `validate_allowlist.py`, `torchscript_utils.py`) to support a `"quantize": true` flag in reference model configs, so quantized variants are traced with dynamic quantization applied before graph extraction — mirroring the Eland import pipeline.
- Adds quantized ELSER v2 entries to `reference_models.json`, `validation_models.json`, and the `reference_model_ops.json` test fixture.

## Context

The `kibana-elasticsearch-snapshot-verify` pipeline was failing because `pytorch_inference` rejected ELSER v2 models with: "Model graph does not match any supported architecture. Unrecognised operations: aten::mul_, quantized::linear_dynamic". These operations are legitimate — they appear when Eland applies `torch.quantization.quantize_dynamic` on `nn.Linear` layers during model import.

## Test plan

- [x] Local build succeeds
- [x] `test_pytorch_inference` passes (includes `CModelGraphValidatorTest` which validates ops against the reference golden file)
- [ ] CI builds pass on all platforms
- [ ] `kibana-elasticsearch-snapshot-verify` pipeline should no longer fail on ELSER v2 model loading

Labelling as `>non-issue` as this relates to an as yet unreleased enhancement.

Relates https://github.com/elastic/ml-cpp/pull/2936

### DIFF
--- a/bin/pytorch_inference/unittest/testfiles/reference_model_ops.json
+++ b/bin/pytorch_inference/unittest/testfiles/reference_model_ops.json
@@ -267,6 +267,7 @@
     },
     "elastic-eis-elser-v2": {
       "model_id": "elastic/eis-elser-v2",
+      "quantized": false,
       "ops": [
         "aten::Int",
         "aten::ScalarImplicit",
@@ -303,6 +304,7 @@
     },
     "elastic-elser-v2": {
       "model_id": "elastic/elser-v2",
+      "quantized": false,
       "ops": [
         "aten::Int",
         "aten::ScalarImplicit",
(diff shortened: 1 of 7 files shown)


valeriy42 (approved):
LGTM

bin/pytorch_inference/unittest/testfiles/reference_model_ops.json
```diff
@@ -267,6 +267,7 @@
     },
     "elastic-eis-elser-v2": {
       "model_id": "elastic/eis-elser-v2",
+      "quantized": false,
```
darius-vil, line 270: nit: maybe the name `"quantized": true/false` is oversimplified and hides an important nuance
I read this flag as "this model is quantized", while in reality it's "this model has dynamic quantization applied to specific layer types (nn.linear)"

Calling it `dynamic_quantization` would be a slight improvement - gives a better signal on what this flag controls.

Long term, if there is ever a need, we could replace a boolean with:
```
"quantize_layers": ["nn.linear"],
"target": "qint8"
```

dev-tools/extract_model_ops/torchscript_utils.py
```diff
@@ -11,13 +11,52 @@
...
+from pathlib import Path
 
 import torch
 from transformers import AutoConfig, AutoModel, AutoTokenizer
 
 
+def load_model_config(config_path: Path) -> dict[str, dict]:
+    """Load a model config JSON file and normalise entries.
+
+    Each entry is either a plain model-name string or a dict with
```
darius-vil, line 26: nit: alternatively we could make the json consistent and drop the normalization all together

PR #3073, opened 2026-07-08
Make minor-freeze version bump idempotent for already-completed bumps

edsavage:
## Summary

When the Centralized Version Bump pipeline re-triggered the `9.5.0` minor bump (to unblock another service) ml-cpp's `ml-cpp-version-bump` [build #59](https://buildkite.com/elastic/ml-cpp-version-bump/builds/59/canvas) failed because the bump had **already completed**. Release engineering asked that the version-bump logic be idempotent: it should not fail if the minor branch has already been cut and `main` has already been bumped to the next expected version.

**Root cause (Leg A):** `validate_minor_freeze_params` (used by `create_minor_branch.sh`) hard-required `main == NEW_VERSION`. After a completed freeze, `main` is at `MAIN_NEW_VERSION` (e.g. `9.6.0`) while `NEW_VERSION` is still `9.5.0`, so validation raised and the step failed. Leg B (`validate_main_minor_bump`) was already idempotent (no-ops when `current == MAIN_NEW_VERSION`).

## Changes

- **`dev-tools/version_bump_validation.py`**: `validate_minor_freeze_params` now accepts `main` being at either:
  - `NEW_VERSION` — freeze not yet applied (normal pre-freeze), or
  - the derived `MAIN_NEW_VERSION` — freeze already completed, **only when the release branch already exists** (idempotent re-run).

  `main` advancing to `MAIN_NEW_VERSION` *without* the release branch being cut is still rejected as a genuinely inconsistent state. The validator also logs a clear "minor freeze already completed (idempotent no-op)" line for the Buildkite log.
- **`dev-tools/unittest/test_version_bump_validation.py`**: added regression tests for the completed-freeze re-run, the branch-cut-but-main-not-yet-bumped intermediate state, and the two states that must still be rejected.

## Effect on the pipeline

With this fix, a re-run after a completed freeze no-ops cleanly:

- Leg A: `create_minor_branch.sh` validates OK, sees the branch already exists, sets `ml_cpp_minor_branch_needed=false`, and exits 0.
- Leg B: `bump_main_minor_freeze.sh` validates OK (no-op), finds no `gradle.properties`/`.backportrc.json` diff, prints "nothing to do", sets `ml_cpp_main_bump_needed=false`, and exits 0.

## Test plan

- [x] `python3 -m pytest dev-tools/unittest/test_version_bump_validation.py -q` passes (26 passed, 2 git-integration skipped)
[...]

PR #3064, opened 2026-07-02
[ML] Automate minor version bump in CI pipeline

edsavage:
Supersedes #3063 (same commits; reopened from personal fork per CONTRIBUTING.md).

## Summary

Extends the `ml-cpp-version-bump` Buildkite pipeline with `WORKFLOW=minor` for feature-freeze (minor version) releases, complementing the existing patch workflow (#3030).

**Phase 1 (validate):** checks that `main` is at `NEW_VERSION` (e.g. `9.5.0`), derives `MAIN_NEW_VERSION` (e.g. `9.6.0`), and sets meta-data for which legs still need work.

**Phase 2 (minor freeze):**
- **Leg A:** create release branch `BRANCH` from `main` (direct ref push)
- **Leg B:** bump `main` to `MAIN_NEW_VERSION`, update `.backportrc.json`, open PR
- Slack notification (step notify)
- DRA artifact wait (production branches only)

Also includes `testing-MAJOR.MINOR` sandbox branch support for **manual Buildkite testing only** — version rules strip the `testing-` prefix; git operations use the full ref name; main bump and DRA wait are skipped so sandbox runs do not touch production branches.

### Automated (local / PR CI)

- [x] `./dev-tools/run_dev_tools_tests.sh` — validation rules, pipeline JSON, phase-2 upload, DRA gating (48 tests at last run)
- [x] `ml-cpp-pr-builds` green on this PR

### Buildkite — production parameters (validate only)

Recommended before merge; does not create branches or open PRs when `DRY_RUN=true`:

```
WORKFLOW=minor
BRANCH=9.5
NEW_VERSION=9.5.0
DRY_RUN=true
```

Branch: `edsavage:feature/minor-version-bump`

### Buildkite — sandbox E2E (orchestration)

Proven on [build #54](https://buildkite.com/elastic/ml-cpp-version-bump/builds/54) (passed):

```
WORKFLOW=minor
BRANCH=testing-9.5
NEW_VERSION=9.5.0
VERSION_BUMP_NO_MERGE=true
```

Exercises: validate → create `testing-9.5` @ `9.5.0` → skip main bump → Slack notify → DRA skip.

`testing-*` is intentional for pipeline regression testing and inert when release-eng uses production `BRANCH=9.5`.

### Buildkite — idempotency (sandbox)

With `testing-9.5` already present after a successful run:

```
WORKFLOW=minor
BRANCH=testing-9.5
NEW_VERSION=9.5.0
DRY_RUN=true
```

[...]

PR #3074, opened 2026-07-13
Add newly cut minor branch to snapshot build pipeline during minor freeze

edsavage:
## Summary

The minor-freeze automation cuts the release branch and bumps `main`, but it did **not** register the new branch in the daily snapshot build pipeline defined in `catalog-info.yaml`. As a result the newly cut branch got no daily SNAPSHOT builds, and any scheduled/triggered build for it would be blocked by the snapshot pipeline's `filter_condition`. This was missed for `9.5`.

This PR both **automates** the registration for future freezes and **fixes the current gap** by adding `9.5`.

## Changes

- **`dev-tools/update_catalog_snapshot.py`** (new): a scoped, comment-preserving text edit (stdlib only — no YAML round-trip, so comments/formatting are preserved) that, for a `MAJOR.MINOR` branch:
  - adds `build.branch == "<branch>"` to the snapshot pipeline `filter_condition`, immediately after `main` (newest-first ordering);
  - inserts a `Daily <branch>` schedule (before `Daily main`) with the next free "half past the hour" cron slot (`30 HH * * *`), avoiding collisions with existing schedules.
  - Idempotent, and only the snapshot pipeline block is touched (the staging pipeline already matches any `MAJOR.MINOR` via regex and has no schedules).
- **`dev-tools/bump_main_minor_freeze.sh`** (Leg B): invokes the updater so the `catalog-info.yaml` change lands in the same PR into `main` as the `.backportrc.json` update. The PR summary/body now lists the snapshot change.
- **`catalog-info.yaml`**: registers the already-cut `9.5` branch (filter + `Daily 9.5` @ `30 05 * * *`).
- **`dev-tools/unittest/test_update_catalog_snapshot.py`** (new): unit tests for filter/schedule insertion, ordering, next-free-hour selection, idempotency, partial-state completion, input validation, section scoping, plus a check that editing the real `catalog-info.yaml` yields valid YAML with unique cronlines.

## Diff to the live file

```
filter_condition: ... "main" || "9.5" || "9.4" || "9.3" || "8.19" || "7.17"

+ Daily 9.5:
+   branch: '9.5'
+   cronline: 30 05 * * *
+   message: Daily SNAPSHOT build for 9.5
```

## Test plan

- [x] `python3 -m pytest dev-tools/unittest/ -q` passes (73 passed, 2 git-integration skipped)
- [x] `bash -n dev-tools/bump_main_minor_freeze.sh` OK
- [x] Editing the real `catalog-info.yaml` parses as valid YAML; all snapshot cronlines remain unique
- [x] Re-running the updater for `9.5` is a no-op (idempotent)
- [ ] Next minor freeze: confirm the Leg B PR into `main` includes the new `Daily <branch>` schedule + filter entry

## Notes / follow-ups

- Cron slots are assigned as `max(existing hour) + 1` at `30 HH`. With branches being EOL'd over time this won't realistically exhaust the 0–23 range; the script errors clearly if it ever would, prompting a prune of EOL schedules.
- This does not remove EOL branch schedules — that remains a manual/cleanup decision.

## Related

- Same minor-freeze automation as #3064, #3070, #3072, #3073.

### DIFF
--- a/dev-tools/unittest/test_update_catalog_snapshot.py
+++ b/dev-tools/unittest/test_update_catalog_snapshot.py
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0 and the following additional limitation. Functionality enabled by the
+# files subject to the Elastic License 2.0 may only be used in production when
+# invoked by an Elasticsearch process with a license key installed that permits
+# use of machine learning features. You may not use this file except in
+# compliance with the Elastic License 2.0 and the foregoing additional
+# limitation.
+
+"""Pytest tests for dev-tools/update_catalog_snapshot.py.
+
+Verifies a newly cut release branch is registered in the snapshot build pipeline
+(filter_condition + daily schedule) with a minimal, comment-preserving diff, that
+the operation is idempotent, and that the real catalog-info.yaml stays valid YAML.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+_DEV_TOOLS = Path(__file__).resolve().parents[1]
+if str(_DEV_TOOLS) not in sys.path:
+    sys.path.insert(0, str(_DEV_TOOLS))
+
+import update_catalog_snapshot as ucs  # noqa: E402
+
+_REPO_ROOT = _DEV_TOOLS.parent
+_CATALOG = _REPO_ROOT / "catalog-info.yaml"
+
+# Minimal fixture mirroring the snapshot pipeline block followed by another
+# section, so scoping (only the snapshot block is edited) is exercised.
+_FIXTURE = """\
+# Declare the snapshot build pipeline
+---
+apiVersion: "backstage.io/v1alpha1"
+kind: "Resource"
+metadata:
+  name: "ml-cpp-snapshot-builds"
+spec:
+  implementation:
+    spec:
+      provider_settings:
+        build_branches: true
+        filter_condition: build.branch == "main" || build.branch == "9.4" || build.branch == "7.17"
+        filter_enabled: true
+      schedules:
+        Daily 7_17:
+          branch: '7.17'
+          cronline: 30 04 * * *
+          message: Daily SNAPSHOT build for 7.17
+        Daily 9.4:
+          branch: '9.4'
+          cronline: 30 01 * * *
+          message: Daily SNAPSHOT build for 9.4
+        Daily main:
+          branch: main
+          cronline: 30 00 * * *
+          message: Daily SNAPSHOT build for main
+      skip_intermediate_builds: true
+
+# Declare the staging build pipeline
+---
+metadata:
+  name: "ml-cpp-staging-builds"
+spec:
+  implementation:
+    spec:
+      provider_settings:
+        filter_condition: 'build.branch == "main"'
+"""
+
+
+def test_adds_branch_to_filter_after_main() -> None:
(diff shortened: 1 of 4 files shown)


dev-tools/update_catalog_snapshot.py
```diff
@@ -0,0 +1,175 @@
...
+    # after the last schedule entry (before the next same-or-lower indent key).
+    main_entry = re.compile(rf"^{re.escape(key_indent)}Daily main:[ \t]*$", re.MULTILINE)
+    main_m = main_entry.search(section, m.end())
+    if main_m is not None:
+        insert_at = main_m.start()
+        return section[:insert_at] + block + section[insert_at:], True
+
+    # Fallback: insert right after the schedules: line.
+    insert_at = m.end() + 1  # past the trailing newline of the schedules: line
+    return section[:insert_at] + block + section[insert_at:], True
```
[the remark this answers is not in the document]
edsavage: Good catch — fixed in 09e9f48a0. The fallback (no `Daily main` entry) now appends the new schedule after the last existing entry, just before the first following line indented at or below the `schedules:` key (the next sibling mapping key), matching the comment. Added a regression test (`test_fallback_appends_after_last_entry_when_no_daily_main`) covering this path.

dev-tools/unittest/test_update_catalog_snapshot.py
```diff
@@ -0,0 +1,209 @@
...
+def test_missing_snapshot_anchor_raises() -> None:
+    with pytest.raises(ValueError, match="anchor not found"):
+        ucs.add_release_branch_to_snapshot("no snapshot pipeline here\n", "9.5")
+
+
+@pytest.mark.skipif(not _CATALOG.is_file(), reason="catalog-info.yaml not found")
+def test_real_catalog_edit_is_valid_yaml_and_unique_crons() -> None:
+    yaml = pytest.importorskip("yaml")
+    text = _CATALOG.read_text(encoding="utf-8")
+    new_text, changed = ucs.add_release_branch_to_snapshot(text, "9.9")
```
valeriy42: When 9.9 ships, this assertion flips to `False` and the test fails with no clear error. Use a sentinel like `99.0` that will never be a real branch, or assert `changed is True or "9.9" already in filter_condition`.

dev-tools/unittest/test_update_catalog_snapshot.py
```diff
@@ -0,0 +1,209 @@
...
+
+def test_idempotent() -> None:
+    once, changed1 = ucs.add_release_branch_to_snapshot(_FIXTURE, "9.5")
+    assert changed1 is True
+    twice, changed2 = ucs.add_release_branch_to_snapshot(once, "9.5")
+    assert changed2 is False
+    assert twice == once
+
+
+def test_partial_state_filter_only_is_completed() -> None:
```
valeriy42, line 117: `test_partial_state_filter_only_is_completed` covers filter-present/schedule-absent. The inverse (schedule-present/filter-absent) is equally reachable via manual editing and is handled by the same independent-function design, but goes untested.

dev-tools/unittest/test_update_catalog_snapshot.py
edsavage: Good point — fixed in 1948ba81c. Switched the real-catalog tests to a `_SENTINEL_BRANCH = "99.0"` constant that will never actually ship, so they can't flip to a no-op (`changed == False`) once a plausible version is released.

dev-tools/unittest/test_update_catalog_snapshot.py
edsavage: Added `test_partial_state_schedule_only_is_completed` in 1948ba81c covering the inverse (schedule present, filter absent): it asserts the filter entry is added and the existing schedule isn't duplicated.

valeriy42 (approved):

PR #3078, opened 2026-07-20
Harden model validation: reject __setstate__ hooks and forbid inductor::_reinterpret_tensor

edsavage:
## Summary

Two related TorchScript model-validation hardening fixes, landed together as one coherent change (supersedes #3079). Both address privately reported security findings; details are tracked internally.

**1. `__setstate__` load-time execution gap**
- `torch::jit::load` runs a module's `__setstate__` during deserialization, *before* the post-load graph validator (which only inspects the forward graph) could ever run. A forbidden op hidden in `__setstate__` therefore executes at load and is invisible to the validator.
- **Remediation (matches the recommended one):** statically scan the archive and refuse **any** model that embeds `__setstate__`/`__getstate__` *before* load, without executing model code.

**2. `inductor::_reinterpret_tensor` heap-OOB bypass**
- After `aten::as_strided` was forbidden, `torch.ops.inductor._reinterpret_tensor` gives the same class of unchecked storage-offset OOB heap read/write.
- Add it to `FORBIDDEN_OPERATIONS` so the post-load graph validator fail-fasts (authoritative, syntax-independent). An op hidden in `__setstate__` is already refused pre-load by the hook ban above.

Post-load allowlist / forbid validation is otherwise unchanged.

## Test plan
- [x] Unit tests: custom state hooks rejected; forward-only malicious models have no hooks (caught post-load); `inductor::_reinterpret_tensor` rejected post-load; benign `e5_with_norm.pt` accepted; garbage input does not crash
- [x] CI pytorch_inference / model validator suites green
- [x] Confirm production-style scripted transformers do not embed custom state hooks

### Model-zoo compatibility evidence (blanket hook ban)

Models exported exactly as ES ingests them (eland 9.2.0 `TransformerModel.save()` → `torch.jit.trace`/`save`), each archive scanned with the same logic as `scanArchiveForCustomStateHooks`. Scanner validated by a positive control (explicit `__setstate__` correctly rejected).

| Family | Model | Task(s) | quant=False | quant=True |
|---|---|---|---|---|
| BERT | bert-base-uncased (full) | text_embedding, fill_mask | accepted | accepted |
| DistilBERT | distilbert-base-uncased | text_embedding | accepted | accepted |
| RoBERTa | all-distilroberta-v1 | text_embedding | accepted | accepted |
| ELECTRA | electra-small-discriminator | text_embedding | accepted | accepted |
| DeBERTa-v2 | deberta-v3-xsmall | fill_mask | accepted | accepted |
| MPNet | all-mpnet-base-v2 | text_embedding | accepted | accepted |

Plus all local fixtures (real E5 embedding, norm model, the three ES integration-test models). The `--quantize` path (the main theoretical risk) is clean: TorchScript serialises quantized packed params via a bound C++ mechanism, so the hook literals never appear in the archive.

Related: #3080 (seccomp arch/ABI check), #3081 (controller non-dumpable).


test/test_setstate_load_timing.py
```diff
@@ -0,0 +1,215 @@
...
+        proc = subprocess.run(
+            [sys.executable, "-c", loader],
+            capture_output=True, text=True, timeout=120,
+        )
+        stdout = proc.stdout
+        marker_before_return = (
+            _LOAD_MARKER in stdout
+            and stdout.index(_LOAD_MARKER) < stdout.index("LOAD_RETURNED")
+        ) if "LOAD_RETURNED" in stdout else (_LOAD_MARKER in stdout)
+
```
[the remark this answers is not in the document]
edsavage: Good catch — tightened in 5957acdb5. The benign probe now only counts as proof when the subprocess exits 0 **and** `LOAD_RETURNED` was printed, with the marker strictly before it; on failure it now prints the returncode and a stderr tail. So a crash that happened to print the marker first is no longer a false positive.

edsavage:
Updated after comparing with the reporter's `0001-reject-torchscript-state-hooks.patch`: a forbidden-op-only pre-scan would miss the real #12621 setstate RCE (allowlisted ops + memory gadgets). We now reject `__setstate__`/`__getstate__` substrings in the archive before load, and retain the forbidden-op scan as defense in depth.

edsavage:
### Model-zoo compatibility check for the blanket `__setstate__`/`__getstate__` ban

Ran the third test-plan item. Models were exported exactly as ES ingests them (eland 9.2.0 `TransformerModel.save()` → `torch.jit.trace`/`save`) and each archive scanned with the same record-substring logic as `scanArchiveForCustomStateHooks`. Scanner validated by a positive control (a module with an explicit `__setstate__` was correctly **REJECTED**, literal found in `code/__torch__.py`).

**Every legitimate model was accepted (zero false positives):**

| Family | Model | Task(s) | quant=False | quant=True |
|---|---|---|---|---|
| BERT | bert-base-uncased (full) | text_embedding, fill_mask | accepted | accepted |
| DistilBERT | distilbert-base-uncased | text_embedding | accepted | accepted |
| RoBERTa | all-distilroberta-v1 | text_embedding | accepted | accepted |
| ELECTRA | electra-small-discriminator | text_embedding | accepted | accepted |
| DeBERTa-v2 | deberta-v3-xsmall | fill_mask | accepted | accepted |
| MPNet | all-mpnet-base-v2 | text_embedding | accepted | accepted |

Plus all local fixtures: real E5 embedding (`e5_with_norm.pt`), `norm_model.pt`, and the three ES integration-test models (`tiny_text_embedding`, `tiny_text_expansion`, `supersimple_pytorch_model_it`).

Notable: the `--quantize` path was the main theoretical risk (dynamic-quantized `LinearPackedParams` define `__getstate__`/`__setstate__` in Python), but TorchScript serialises packed params via a bound C++ mechanism, so the ASCII literals never appear in the archive — confirmed on a standalone quantized `nn.Linear` and on all quantized full-model exports above.

Conclusion: the blanket ban does not reject any legitimate model across the encoder families eland/ES supports for NLP.

edsavage:
Test plan item 2 (CI pytorch_inference / model validator suites green) confirmed by [build 2844](https://buildkite.com/elastic/ml-cpp-pr-builds/builds/2844) on head `ece0c5d1f` — passed on Linux x86_64/aarch64, macOS aarch64 and Windows x86_64.

`ml_test_pytorch_inference`: **60 tests, 0 failures, 0 errors**, including this PR's new cases:
- `testPreLoadScanRejectsCustomStateHooks` / `...InSubmodule`
- `testPreLoadScanAcceptsBenignModel` / `...ForwardOnlyMaliciousModel`
- `testPreLoadScanHandlesGarbageInput`
- `testMaliciousReinterpretTensorRejectedPostLoad`
- `testPrepackedE5ModelWithNorm`

All test-plan items are now checked off.

edsavage:
buildkite run_pytorch_tests

bin/pytorch_inference/CModelGraphValidator.cc
```diff
@@ -95,6 +127,48 @@ void CModelGraphValidator::collectBlockOps(const ::torch::jit::Block& block,
...
+            }
+        }
+    } catch (const std::exception& e) {
+        // If the archive cannot be parsed as a stream we do not treat this as a
+        // rejection here: torch::jit::load will attempt the same parse and
+        // surface a clear error.  The post-load graph validator still applies.
+        LOG_WARN(<< "Pre-load state-hook scan skipped (could not parse archive): "
+                 << e.what());
+        return {};
+    }
```
valeriy42: The `try` block wraps **both** the `PyTorchStreamReader` constructor and every `reader.getRecord(name)` call inside the loop (lines 138–157). If `getRecord()` throws for a non-code record (e.g., due to a bad CRC), the catch fires, logs one warning, and returns `{}` — abandoning any hooks that appear in later records without having scanned them.

Attack path: craft an archive whose `getAllRecords()` succeeds (ZIP central directory is intact) but whose first-iterated record has a bad checksum that triggers a `getRecord()` exception. The `__setstate__`-containing code record appears later in iteration order. The scan returns `{}` (no hooks found); `verifySafeModelBeforeLoad` passes; `torch::jit::load` then runs `__setstate__` (it reads only the records it needs by name, potentially skipping the corrupt one).

**Recommendation:** Narrow the guard to the reader construction only; handle per-record failures with warn-and-continue:

```cpp
std::unique_ptr<caffe2::serialize::PyTorchStreamReader> reader;
try {
    auto adapter = std::make_shared<CMemoryReadAdapter>(data, size);
    reader = std::make_unique<caffe2::serialize::PyTorchStreamReader>(std::move(adapter));
} catch (const std::exception& e) {
    LOG_WARN(<< "Pre-load state-hook scan skipped (could not parse archive): " << e.what());
    return {};
}
for (const auto& name : reader->getAllRecords()) {
    std::pair<std::shared_ptr<char[]>, size_t> entry;
    try {
        entry = reader->getRecord(name);
    } catch (const std::exception& e) {
        LOG_WARN(<< "Pre-load scan: skipping record '" << name << "': " << e.what());
        continue;           // scan remaining records
    }
    std::string_view bytes{static_cast<const char*>(entry.first.get()), entry.second};
    // ... hook detection unchanged ...
}
```

This preserves the "don't block on unparseable archive" intent while ensuring a single bad record can't abort the scan.

bin/pytorch_inference/unittest/CModelGraphValidatorTest.cc
```diff
@@ -338,6 +338,17 @@ bool hasUnrecognisedOp(const CModelGraphValidator::SResult& result, const std::s
     return std::find(result.s_UnrecognisedOps.begin(), result.s_UnrecognisedOps.end(),
                      op) != result.s_UnrecognisedOps.end();
 }
+
+std::string readFileBytes(const std::string& path) {
```
valeriy42, line 342: On a missing fixture, `file` enters a fail state silently; the call site's `BOOST_REQUIRE(bytes.empty() == false)` fires with `"bytes.empty() == false"` — no indication of which path was missing. Adding a `BOOST_REQUIRE_MESSAGE(file.is_open(), "Could not open: " + path)` (or equivalent early check) would give CI-friendly diagnostics if fixtures go missing.

bin/pytorch_inference/CModelGraphValidator.cc
edsavage: Good catch — fixed in 7b2b58eb6. The guard now wraps only the `PyTorchStreamReader` construction; the per-record loop catches `getRecord()` failures individually and `continue`s (warn-and-skip) so a single deliberately-corrupt record can no longer abort the scan and hide a `__setstate__` hook in a later record. An unparseable archive still returns `{}` (torch::jit::load will surface the same parse error).

bin/pytorch_inference/unittest/CModelGraphValidatorTest.cc
edsavage: Done in 7b2b58eb6 — `readFileBytes` now does `BOOST_REQUIRE_MESSAGE(file.is_open(), "Could not open fixture: " + path)`, so a missing fixture fails loudly with the path instead of the opaque `bytes.empty() == false`.

valeriy42 (approved):

edsavage:
buildkite build

PR #3081, opened 2026-07-21
Mark ML controller non-dumpable before accepting commands

edsavage:
## Summary

Follow-up to a privately reported security finding (details tracked internally): the ML controller shared UID `1000` with the JVM / `pytorch_inference` and remained dumpable, so a compromised peer could write `/proc/<controller-pid>/mem`, overwrite `access@GOT`, and turn `access(path, X_OK)` into `dlopen(path, RTLD_LAZY)` (`X_OK == RTLD_LAZY == 1`) when handling a spawn command — loading attacker code under the controller's laxer RuntimeDefault-only seccomp.

## Fix

After logging is reconfigured and **before** opening command/output pipes (i.e. before accepting commands):

1. `prctl(PR_SET_DUMPABLE, 0)`
2. Confirm with `prctl(PR_GET_DUMPABLE) == 0`
3. **Fail closed** (`LOG_FATAL` + `EXIT_FAILURE`) if either step fails

No-op on non-Linux platforms (`/proc/<pid>/mem` same-UID writes are the Linux concern).

## Test plan

- [x] Linux CI: controller still starts and processes commands normally.
- [x] On Linux: after start, `/proc/<controller-pid>/status` shows `Dumpable: 0` (or equivalent).
- [x] Confirm fail-closed path if `PR_SET_DUMPABLE` cannot be applied (manual / mocked).

Related: #3078 (`__setstate__` pre-load), #3080 (seccomp arch/ABI check).

### DIFF
--- a/bin/controller/Main.cc
+++ b/bin/controller/Main.cc
@@ -81,8 +81,14 @@ bool makeProcessNonDumpable() {
         return false;
     }
     // Confirm the attribute stuck — a silent no-op would leave the process exposed.
-    if (::prctl(PR_GET_DUMPABLE) != 0) {
-        LOG_ERROR(<< "process remains dumpable after PR_SET_DUMPABLE(0)");
+    const int dumpable{::prctl(PR_GET_DUMPABLE)};
+    if (dumpable == -1) {
+        LOG_ERROR(<< "prctl PR_GET_DUMPABLE failed: " << std::strerror(errno));
+        return false;
+    }
+    if (dumpable != 0) {
+        LOG_ERROR(<< "process remains dumpable (" << dumpable
+                  << ") after PR_SET_DUMPABLE(0)");
         return false;
     }
     return true;
--- a/bin/controller/Main.cc
+++ b/bin/controller/Main.cc
@@ -87,8 +87,7 @@ bool makeProcessNonDumpable() {
         return false;
     }
     if (dumpable != 0) {
-        LOG_ERROR(<< "process remains dumpable (" << dumpable
-                  << ") after PR_SET_DUMPABLE(0)");
+        LOG_ERROR(<< "process remains dumpable (" << dumpable << ") after PR_SET_DUMPABLE(0)");
         return false;
     }
     return true;
--- a/bin/controller/Main.cc
+++ b/bin/controller/Main.cc
@@ -63,6 +63,36 @@
 #include <iostream>
 #include <string>
 
+#ifdef Linux
+#include <sys/prctl.h>
+#endif
+
+namespace {
+
+//! Prevent same-UID peers (e.g. a compromised pytorch_inference / JVM Attach
+//! agent in the shared pod) from writing this process's memory via
+//! /proc/<pid>/mem.  That path was used to overwrite access@GOT and turn a
+//! controller spawn into dlopen (elastic/security#12621).  Fail closed: if we
+//! cannot establish the boundary, refuse to accept commands.
+bool makeProcessNonDumpable() {
+#ifdef Linux
+    if (::prctl(PR_SET_DUMPABLE, 0) != 0) {
+        LOG_ERROR(<< "prctl PR_SET_DUMPABLE(0) failed: " << std::strerror(errno));
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


edsavage:
### Test plan verification

All three items checked off. Item 1 is confirmed by CI; items 2–3 were verified against the **real controller binary** from this PR's CI build (`ml-cpp-9.6.0-SNAPSHOT-linux-aarch64`, build `88ade68d3`, PR CI [build 2806](https://buildkite.com/elastic/ml-cpp-pr-builds/builds/2806)), run in a **native arm64 Linux** container (no emulation).

#### 1 & CI — controller starts and processes commands normally
- `ml_test_controller` (9 cases incl. `CCommandProcessorTest_testStartPermitted`) passed on Linux x86_64 + aarch64.
- End-to-end with the real binary: it logs the version, becomes non-dumpable, opens the command/output pipes, and answers a command:
  ```
  INFO  controller (64 bit): Version 9.6.0-SNAPSHOT (Build 88ade68d3bd13c) ...
  output pipe response: [{"id":1,"success":false,"reason":"Did not understand verb 'nope'"}]
  ```

#### 2 — non-dumpable after start (Linux)
This kernel's `/proc/<pid>/status` has no `Dumpable:` line, so I verified the authoritative observable of `PR_SET_DUMPABLE(0)`: `/proc/<pid>` flips to `root:root` and same-UID peers are denied `mem`/`maps` — i.e. exactly the `/proc/<pid>/mem` write vector from security#12621. Running as a non-root uid (`1000`), with a normal `sleep` as a **dumpable control**:

```
                          /proc/<pid>/mem owner
control  sleep (dumpable):  1000:1000        maps: readable
CONTROLLER    (this PR)  :  root:root         maps: Permission denied
same-uid peer open(/proc/<controller>/mem, O_RDWR): DENIED  [Errno 13] Permission denied
```

The controller is caught *after* `makeProcessNonDumpable()` and *before* it opens the command pipe, so the boundary is established before any command is accepted — as intended.

#### 3 — fail-closed (mocked)
`prctl(PR_SET_DUMPABLE,0)` doesn't fail in practice, so I compiled `makeProcessNonDumpable()` verbatim (only `LOG_ERROR`→`fprintf`, `::prctl`→mock) with the same caller logic and forced each failure branch:

```
A: SET ok, GET==0 (normal)                 -> exit 0 (continue accepting commands)
B: PR_SET_DUMPABLE fails (EPERM)           -> exit 1 (EXIT_FAILURE, LOG_FATAL)
C: SET 'ok' but still dumpable (GET!=0)    -> exit 1 (EXIT_FAILURE, LOG_FATAL)
```

Both failure paths refuse to accept commands (fail closed); the normal path continues.

edsavage:
buildkite run_pytorch_tests

bin/controller/Main.cc
```diff
@@ -63,6 +63,36 @@
...
+#ifdef Linux
+    if (::prctl(PR_SET_DUMPABLE, 0) != 0) {
+        LOG_ERROR(<< "prctl PR_SET_DUMPABLE(0) failed: " << std::strerror(errno));
+        return false;
+    }
+    // Confirm the attribute stuck — a silent no-op would leave the process exposed.
+    if (::prctl(PR_GET_DUMPABLE) != 0) {
+        LOG_ERROR(<< "process remains dumpable after PR_SET_DUMPABLE(0)");
+        return false;
+    }
```
valeriy42:
```suggestion
    int dumpable = ::prctl(PR_GET_DUMPABLE);
    if (dumpable == -1) {
        LOG_ERROR(<< "prctl PR_GET_DUMPABLE failed: " << std::strerror(errno));
        return false;
    }
    if (dumpable != 0) {
        LOG_ERROR(<< "process remains dumpable (" << dumpable << ") after PR_SET_DUMPABLE(0)");
        return false;
    }
```

valeriy42 (approved):
LGTM. Minor suggestion to improve diagnostic

bin/controller/Main.cc
edsavage: Good call — applied in 3de96c959. Now distinguishes the `PR_GET_DUMPABLE` failure case (`-1`) from a still-dumpable process and logs the actual value. Thanks!

PR #3090, opened 2026-07-22
[ML] Fix concurrent LFU cache count invariant under lock timeouts

edsavage:
## Summary

`testConcurrentReadsAndWrites` (in `CCompressedLfuCacheTest`) intermittently fails on loaded debug CI agents with:

```
Count mismatch 1097 (less 7 lost) vs 1092.
critical check cache.checkInvariants() has failed
```

### Root cause

`CCompressedLfuCache::lookup` incremented `m_NumberLookups` **inside the read-guard lambda**, which does not run when the read lock acquisition times out (`m_MaximumWait` is only 50 ms). When the read lock times out the lookup falls through to the miss path, which can still:

- **insert** a value (+1 to the tracked counts), or
- find the key already cached and **increment** its count (+1), or
- **time out** on the write lock and bump `m_LostCount`.

In every case the write side changed the bookkeeping without the matching lookup being counted, so the surviving counts could **exceed** `m_NumberLookups - m_LostCount`, tripping the strict-equality check in `checkInvariants()`. This is a benign accounting asymmetry, not cache corruption, but it produces a real flaky test under contention.

There is also an inherent, opposite race: an item seen as a hit under the read lock can be evicted before the write lock increments its count, so counts can also legitimately be *lost*. The pre-existing `!=` (strict equality) check was fragile in both directions.

### Fix

- **Count every lookup exactly once**, before taking the read lock, so a read-lock timeout no longer desynchronises the lookup count from the write-side contributions.
- **Relax the count invariant to a one-sided bound for the timeout-capable (concurrent) cache**: counts may be lost but must never be fabricated (`totalCount <= accountedLookups`). The exact equality is retained for the single-threaded cache, which has no timeouts or races. The structural invariants (size / link / memory) remain strict.

### Test

Adds `testConcurrentCountInvariantWithTimeouts`, which drives read- and write-lock timeouts via a zero maximum wait under heavy contention. It fails **20/20** against the previous code and passes with the fix.

## Test plan

- [x] `testConcurrentReadsAndWrites` passes
- [x] New `testConcurrentCountInvariantWithTimeouts` passes with the fix and fails reliably (20/20) without it
- [x] Single-threaded cases (`testLookup`, `testMemoryUsage`, `testResize`) still pass (strict-equality invariant path unchanged)

### DIFF
--- a/docs/changelog/3090.yaml
+++ b/docs/changelog/3090.yaml
@@ -1,0 +1,5 @@
+pr: 3090
+summary: "Fix flaky concurrent LFU cache count invariant under lock timeouts"
+area: Machine Learning
+type: bug
+issues: []
--- a/include/core/CCompressedLfuCache.h
+++ b/include/core/CCompressedLfuCache.h
@@ -125,8 +125,14 @@ public:
 
         auto compressedKey = m_CompressKey(m_Dictionary, key);
 
+        // Count every lookup exactly once, even if we fail to acquire the read
+        // lock below. The fall-through miss path can still add a count (or record
+        // a lost update), so counting only inside the read guard would let the
+        // bookkeeping totals exceed the lookup count when the read lock times out
+        // under contention.
+        ++m_NumberLookups;
+
         if (this->guardRead(TIME_OUT, [&] {
-                ++m_NumberLookups;
                 auto hit = this->hit(compressedKey);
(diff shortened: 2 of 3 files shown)


valeriy42 (approved):
LGTM

PR #3122, opened 2026-07-27
[ML] Re-enable auto-merge for backport PRs (gated on required CI rollup)

edsavage:
## Summary

Re-enables auto-merge for auto-backport PRs, reverting the temporary disable in #3114.

#3114 disabled `gh pr merge --auto` because, at the time, the release branches did **not** require the ml-cpp build/test checks — so auto-merge would merge a backport as soon as the lightweight checks (CLA, snyk) passed, before the C++ matrix finished (this actually happened with #3106–#3111).

That gap is now closed:
- The pipeline publishes a single overall rollup status, `buildkite/ml-cpp-pr-builds` (`publish_commit_status=True`).
- `buildkite/ml-cpp-pr-builds` is now a **required** status check on `main` and on `9.5` / `9.4` / `8.19`.
- The rollup reflects the whole build/test matrix, including any triggered QA/PyTorch downstream — those `trigger` steps are `async: false`, so a downstream failure propagates into the parent build and hence the rollup.

So `gh pr merge --auto --squash` now waits for CI to be green (and the one-review requirement to be met) before merging. If CI never runs or fails, the PR just stays open — fail-safe.

Approval continues to be posted by `github-actions[bot]` (a distinct identity from the ephemeral-token PR author), satisfying the org "[org] Require a PR" ruleset on the release branches.

Related: #3099 (CI-trigger fix #3103), #3112 (set_commit_status), #3114 (temporary disable).

## Test plan
- [x] Merge a `>bug` PR carrying version + `auto-backport` labels.
- [x] Confirm each backport PR is approved by `github-actions[bot]` and has auto-merge armed.
- [x] Confirm a backport merges only **after** its `buildkite/ml-cpp-pr-builds` rollup goes green (not before).
- [ ] Confirm a backport whose CI fails does **not** auto-merge. _(Guaranteed by the required `buildkite/ml-cpp-pr-builds` check — GitHub auto-merge cannot complete unless it is `success`; gating proven on #3078/#3115. Not yet exercised by a naturally-failing backport build.)_

PR #3127, opened 2026-07-27
[ML] Sweep stale backport-pending labels (auto-merged backports)

edsavage:
## Summary

The `remove-backport-pending` job in `backport.yml` clears the `backport-pending` label from a source PR once all of its backports have merged. It's driven by the backport PR's `closed`(merged) event.

That event-driven path **never fires when a backport PR is auto-merged**: auto-merge is armed with `GITHUB_TOKEN`, so GitHub attributes the eventual merge to `github-actions[bot]` and **suppresses the workflow runs it would otherwise trigger** — the same recursion guard that stops `GITHUB_TOKEN`-authored PRs from starting CI (already documented in the backport job). The upshot is the source PR keeps `backport-pending` indefinitely and it has to be removed by hand (as happened on #3078 and #3115).

## Fix

Add an hourly (and `workflow_dispatch`) **`sweep-backport-pending`** backstop job that clears `backport-pending` from any source PR whose backport PRs have all merged/closed. It:

- is independent of *how* the backports merged (auto-merge, manual, admin bypass), and also covers **manually created** backports;
- reuses the same title-reference convention (`[9.5] Original title (#<source>)`) as the event-driven job;
- includes a **1-hour post-merge guard** so it never strips the label in the small window before the backport PRs have been opened.

The event-driven job is retained (immediate in the human-merge case); the sweep is purely additive.

### Notes
- New triggers `schedule` + `workflow_dispatch` are added. The existing `pull_request_target` jobs are guarded on `github.event.pull_request.*`, so they're skipped for schedule/dispatch events (and the sweep job is gated to only those events).
- Depends on backport PR titles carrying `(#<source>)` — same assumption as today's `remove-backport-pending`. Manually opened backports should follow that convention to be tracked.

## Test plan
- [ ] Merge to `main`.
- [ ] Trigger the workflow manually (`workflow_dispatch`) and confirm the `sweep-backport-pending` job runs, correctly *keeps* the label on a source PR that still has open backports (respecting the 1h guard), and *removes* it once all backports are merged/closed.
- [ ] After a real auto-backport round completes (all backports auto-merged), confirm `backport-pending` is cleared within the hour without manual intervention.

Follows up the auto-backport-and-merge work (#3094, #3103, #3122); addresses the known `GITHUB_TOKEN`-suppression cleanup gap noted there.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -3,6 +3,12 @@ name: Backport
 on:
   pull_request_target:
     types: ["labeled", "closed"]
+  # Hourly sweep + on-demand run for the sweep-backport-pending backstop job below.
+  # (The pull_request_target jobs are guarded on github.event.pull_request and are
+  # skipped for these events.)
+  schedule:
+    - cron: "17 * * * *"
+  workflow_dispatch: {}
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


edsavage:
Superseded by #3128, which fixes the root cause directly: arm auto-merge with the ephemeral App token so the deferred merge is attributed to a non-`github-actions[bot]` identity and therefore triggers `remove-backport-pending` normally. That's the GitHub-recommended remedy (and what kibana gets for free via `kibanamachine`), so the scheduled sweep here is no longer needed.

PR #3131, opened 2026-07-28
[ML] Run QA and PyTorch suites concurrently in a single triggered build

edsavage:
## Summary

When both the QA (`ml_cpp_pr`) and PyTorch suites are requested on a PR, trigger a **single** downstream appex-qa build with a comma-separated `QAF_TESTS_TO_RUN` instead of two separate trigger steps.

Previously the two suites were triggered as independent downstream builds. Because the `appex-qa-stateful-custom-ml-cpp-build-testing` pipeline has `skip_queued_branch_builds=true`, two builds on the same branch collided and one suite was silently skipped. Emitting a single trigger with both markers lets the appex-qa generator fan them out into concurrent parallel jobs.

## Changes

- `.buildkite/pipeline.json.py`: merge the `run_qa_tests` / `run_pytorch_tests` trigger blocks into one. `QAF_TESTS_TO_RUN` is built from the requested suites:
  - QA only → `ml_cpp_pr`
  - PyTorch only → `pytorch_tests`
  - both → `ml_cpp_pr,pytorch_tests`

The separate PyTorch allowlist-validation step is unaffected.

## ⚠️ Ordering dependency

This is **Part B**. It depends on **Part A** — the appex-qa generator change in `elastic/qaf-tests#309` — which teaches the generator to split a comma-separated `QAF_TESTS_TO_RUN` into parallel jobs. **qaf-tests#309 must merge first.** Until then, requesting *both* labels would pass `ml_cpp_pr,pytorch_tests` to the old generator, which treats it as a single invalid pytest `-m` expression. Single-suite requests are unaffected.

## Testing

- [x]  Local dry-runs of the generator confirm valid pipeline JSON and correct `QAF_TESTS_TO_RUN` for QA-only, PyTorch-only, and both-labels cases.
- [x]  End-to-end validation of concurrent jobs performed on a qaf-tests feature branch build.

### DIFF
--- a/.buildkite/pipeline.json.py
+++ b/.buildkite/pipeline.json.py
@@ -90,11 +90,20 @@ def main():
             if config.run_qa_tests or config.run_pytorch_tests:
                 qa_suites = []
                 if config.run_qa_tests:
-                    qa_suites.append(os.environ.get("QAF_TESTS_TO_RUN") or "ml_cpp_pr")
+                    # QAF_TESTS_TO_RUN may itself be a comma-separated override
+                    # (e.g. a subset of QA markers), so split it into individual
+                    # suites rather than appending the whole string verbatim.
+                    override = os.environ.get("QAF_TESTS_TO_RUN")
+                    if override:
+                        qa_suites.extend(s.strip() for s in override.split(",") if s.strip())
+                    else:
+                        qa_suites.append("ml_cpp_pr")
                 if config.run_pytorch_tests:
                     qa_suites.append("pytorch_tests")
-                env["QAF_TESTS_TO_RUN"] = ",".join(qa_suites)
-                pipeline_steps.append(pipeline_steps.generate_step("Upload QA tests runner pipeline",
+                # De-duplicate while preserving order so an override that already
+                # includes pytorch_tests does not produce a duplicate suite.
+                env["QAF_TESTS_TO_RUN"] = ",".join(dict.fromkeys(qa_suites))
+                pipeline_steps.append(pipeline_steps.generate_step("Upload QA/PyTorch tests runner pipeline",
                                                                    ".buildkite/pipelines/run_qa_tests.yml.sh"))
         if config.build_aarch64 and not config.skip_version_bump_pr_ci:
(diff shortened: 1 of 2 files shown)


edsavage:
addressed both of copilot's comments in `5f74468`:

1. **Comma-separated `QAF_TESTS_TO_RUN` override → duplicated suites** (line 96): the env override is now split on commas into individual suites, and the final list is de-duplicated while preserving order (`dict.fromkeys`). So both-labels + an override of `ml_cpp_pr,pytorch_tests` now yields `ml_cpp_pr,pytorch_tests` rather than `...,pytorch_tests,pytorch_tests`.

2. **Misleading step label** (line 97): renamed to **"Upload QA/PyTorch tests runner pipeline"** since the single trigger now covers QA and/or PyTorch. (The label is display-only — `generate_step` doesn't derive a key from it and nothing depends on it.)

Verified with local dry-runs: QA-only, PyTorch-only, both, both + a pytorch-inclusive override (de-dups), and a QA subset override — all produce the expected `QAF_TESTS_TO_RUN`.

edsavage:
Addressed the low-confidence note from the latest Copilot review (re: PyTorch-only runs being labelled/commit-statused as "QA Tests").

Rather than partially reverting to `run_pytorch_tests.yml.sh` for the PyTorch-only case (which would reintroduce two code paths and drop the `ES_BRANCH`/`STACK_VERSION` forwarding that `run_qa_tests.yml.sh` provides), `run_qa_tests.yml.sh` now derives a descriptor from `QAF_TESTS_TO_RUN` and uses it for the step label, echo, and `github_commit_status` context:

- `ml_cpp_pr` -> `Trigger Appex QA Tests`
- `pytorch_tests` -> `Trigger Appex PyTorch Tests`
- `ml_cpp_pr,pytorch_tests` -> `Trigger Appex QA + PyTorch Tests`

Substring matching keeps it correct for marker expressions (e.g. `ml_cpp_pr and not slow`). Note these contexts are not required status checks on any branch, so the rename has no branch-protection impact. Verified via dry-run across all cases; emitted YAML parses.

PR #3160, opened 2026-08-17
[ML] Periodically report pytorch_inference RSS

### LINKED ISSUE #2885: [ML] Report actual memory usage for pytorch_inference process
Add the capability to report the actual OS memory usage (RSS) for the `pytorch_inference` process, similar to what was implemented for `autodetect` in #2846.

**Background**

PR #2846 introduced reporting of actual memory usage (via `getrusage` RSS) for the `autodetect` process. This provides valuable insight into the real memory footprint of anomaly detection jobs as reported by the OS, rather than relying solely on internal memory tracking.

The `pytorch_inference` process currently:
- Has the infrastructure to report RSS values via `writeProcessStats()` (called on-demand via `E_ProcessStats` control message)
- Uses `CProcessStats::residentSetSize()` and `CProcessStats::maxResidentSetSize()` in `Main.cc`
- Does **not** periodically report this information back to the Java process

**Proposed Changes**

1. Add periodic reporting of system memory usage for `pytorch_inference`, similar to how `autodetect` updates `E_TSADSystemMemoryUsage` and `E_TSADMaxSystemMemoryUsage` program counters.

2. Include the RSS values in the output stream that can be consumed by the Java side. Options include:
   - Adding new fields to an existing result type
   - Creating a new periodic stats message
   - Extending the response from `E_ProcessStats` to be sent periodically

3. The values to report:
   - `system_memory_bytes` - current resident set size (`CProcessStats::residentSetSize()`)
   - `max_system_memory_bytes` - peak resident set size (`CProcessStats::maxResidentSetSize()`)

**Files likely to be modified**

- `bin/pytorch_inference/Main.cc` - Add periodic memory reporting
- `bin/pytorch_inference/CResultWriter.cc` / `CResultWriter.h` - Potentially extend output format
- `bin/pytorch_inference/CCommandParser.cc` / `CCommandParser.h` - If new message types are needed

**Relates to**

- elastic/ml-cpp#2846 (autodetect actual memory reporting)
- elastic/elasticsearch#139233 (Java-side changes for trained model stats)

edsavage:
## Summary

Adds a periodic memory reporter to `pytorch_inference` so Elasticsearch can track the process's real resident set size (RSS) for a trained model deployment, independent of inference traffic.

- A dedicated timer thread emits process stats every 10s (a single RSS read; negligible cost), so memory is observed even when a deployment is idle or between requests.
- Each report carries both the current RSS (`memory_rss`) and the OS peak / high-water mark (`memory_max_rss`). The peak is transmitted explicitly rather than derived on the ES side from the sample stream, so transient spikes between reports are not lost — this is the signal used to keep model assignment and adaptive scaling OOM-safe.
- The process-stats document key is renamed from `process_stats` to `stats` to match the Elasticsearch `PyTorchResult` / `InferenceProcessStats` parser.
- The reporter thread is shut down promptly (condition-variable signalled) before the rest of the process tears down, so it can never write to a closing stream.

Closes https://github.com/elastic/ml-cpp/issues/2885
Relates https://github.com/elastic/elasticsearch/pull/156891

## Release-ordering note (important)

ES's `InferenceProcessStats` parser is strict and treats an unknown field as a fatal parse error for the result stream. Because this PR (re)introduces `memory_max_rss`, **the Elasticsearch-side change that accepts `memory_max_rss` as an optional field must be merged before this PR is merged into ml-cpp.** See the companion Elasticsearch PR.

Merging this PR before the ES-side fix is not simply "safe with the constraint on the version bump" — once this PR lands, the nightly ml-cpp snapshot will start emitting `memory_max_rss`. The automatic ES version bump PR will pick up that snapshot and run ES integration tests and QAF tests against an ES parser that still rejects unknown fields, causing those tests to fail with a fatal parse error. The version bump PR will be stuck until the ES-side fix lands. To avoid that: merge the ES-side PR first, then merge this one.

## Test plan

- [x] `CResultWriterTest.testWriteProcessStats` updated for the `stats` schema + `memory_max_rss`; passes (all 7 `CResultWriterTest` cases pass).
- [x] Manual: run `pytorch_inference` with a model and confirm periodic `stats` docs are emitted ~every 10s with `memory_rss` and `memory_max_rss`. Observed `{"request_id":"ignore","stats":{"memory_rss":0,"memory_max_rss":93339648}}` after one 10s tick on macOS (`memory_rss` is 0 on macOS — platform-specific; both fields will be non-zero on Linux).
- [x] Confirm prompt shutdown (no stray writes / hangs on exit). Process exits with code 0 cleanly after stdin closes.

### DIFF
--- a/bin/pytorch_inference/CResultWriter.cc
+++ b/bin/pytorch_inference/CResultWriter.cc
@@ -30,7 +30,9 @@ const std::string CResultWriter::ACK{"ack"};
 const std::string CResultWriter::ACKNOWLEDGED{"acknowledged"};
 const std::string CResultWriter::NUM_ALLOCATIONS{"num_allocations"};
 const std::string CResultWriter::NUM_THREADS_PER_ALLOCATION{"num_threads_per_allocation"};
-const std::string CResultWriter::PROCESS_STATS{"process_stats"};
+// This key must match the field name the Elasticsearch PyTorchResult parser expects
+// for process stats (see InferenceProcessStats / PyTorchResult on the Java side).
+const std::string CResultWriter::PROCESS_STATS{"stats"};
 const std::string CResultWriter::MEMORY_RESIDENT_SET_SIZE{"memory_rss"};
 const std::string CResultWriter::MEMORY_MAX_RESIDENT_SET_SIZE{"memory_max_rss"};
 
--- a/bin/pytorch_inference/CResultWriter.h
+++ b/bin/pytorch_inference/CResultWriter.h
@@ -73,6 +73,12 @@ public:
     void writeSimpleAck(const std::string_view& requestId);
 
     //! Write memory usage information to the output stream.
+    //!
+    //! Emits both the current resident set size (\c memory_rss) and the peak
+    //! resident set size (\c memory_max_rss, the OS high-water mark). The peak is
+    //! transmitted explicitly rather than derived on the Elasticsearch side from
+    //! the stream of samples, so transient spikes between reports are not lost;
+    //! this is the memory signal used to keep model assignment OOM-safe.
     void writeProcessStats(const std::string_view& requestId,
                            const std::size_t residentSetSize,
                            const std::size_t maxResidentSetSize);
--- a/bin/pytorch_inference/Main.cc
+++ b/bin/pytorch_inference/Main.cc
@@ -36,13 +36,24 @@
 #include <torch/csrc/api/include/torch/types.h>
 #include <torch/script.h>
 
+#include <atomic>
+#include <chrono>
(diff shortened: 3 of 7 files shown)


edsavage:
buildkite run_serverless_tests ES_SERVERLESS_BRANCH=feature/ml-cpp-repo-override

edsavage:
buildkite run_serverless_tests ES_SERVERLESS_BRANCH=feature/ml-cpp-repo-override

edsavage:
buildkite run_serverless_tests

edsavage:
buildkite build this

valeriy42 (approved):

PR #3161, opened 2026-08-17
[ML] Migrate DRA GCS publishing from Release Manager to dra-prep plugin

vincentDeuschle:
## What this changes

Replaces the Release Manager Docker step with the `elastic/dra-prep-buildkite-plugin`
(+`unified-release-dra-processing` pipeline trigger).

**Before**: A `docker run` against `docker.elastic.co/infra/release-manager` was
responsible for classifying artifacts, generating the DRA manifest, and pushing
to GCS. Vault credentials were required (`secret/ci/elastic-ml-cpp/gcs/creds/prelertartifacts`).

**After**:
1. `elastic/dra-prep` plugin (this repo) — runs `dractl` to classify artifacts,
   generate the manifest, and upload to a temporary GCS path using Workload
   Identity Federation (no Vault credentials needed).
2. `unified-release-dra-processing` pipeline (always runs on `main` in
   elastic/unified-release) — picks up the manifest from the temp path, moves
   artifacts to the final GCS product path, and handles container push/sign.

**Net result**: DRA artifacts are available at the same GCS path as before.

## Files changed

- **`.buildkite/scripts/stage_artifacts.sh`** (new) — downloads `.zip` and `.csv` files from the `create_dra_artifacts` step and stages them into `artifacts/` for `dractl` to process.
- **`.buildkite/pipelines/upload_dra_to_gcs.yml.sh`** (rewritten) — generator now emits a DRA Prep step (plugin) + Trigger DRA processing step instead of the Release Manager Docker step. Version is read from `gradle.properties`; workflow is derived from `BUILD_SNAPSHOT` (already set per-pipeline in `catalog-info.yaml`).
- **`.buildkite/scripts/steps/upload_dra_to_gcs.sh`** (deleted) — the Release Manager Docker wrapper is no longer needed.

## Why

Part of the platform-wide Release Manager deprecation ([#project-stack-releases-improvements](https://elastic.slack.com/archives/C0B59CBMVSS)).

## Notes

- DRA prep is skipped on feature branches where `create_dra_artifacts` produces no artifacts — the `depends_on` chain handles this automatically.
- The S3 upload path (`upload_dra_to_s3`) is unchanged — it serves day-to-day Elasticsearch builds and is separate from DRA.
- **WIF registration** must be requested in [#project-stack-releases-improvements](https://elastic.slack.com/archives/C0B59CBMVSS) before this can be merged and run successfully.
- `catalog-info.yaml` manages pipelines via Terrazzo — no manual Buildkite UI change needed post-merge.

## What you own post-merge

- `stage_artifacts.sh` — update if artifact paths or naming changes.
- The `elastic/dra-prep` plugin version pin — bump when a new version is released.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.buildkite/scripts/stage_artifacts.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -euo pipefail
+
+WORKFLOW="${DRA_WORKFLOW:?DRA_WORKFLOW is required}"
+
+echo "--- :compression: Downloading ${WORKFLOW} artifacts from create_dra_artifacts step"
+rm -rf build/distributions artifacts
+mkdir -p build/distributions artifacts
+
+buildkite-agent artifact download 'build/distributions/*.zip' . --step create_dra_artifacts
```
edsavage, line 22: `create_dra.sh` still downloads Windows-style artifact paths (`build\\distributions\\*.zip`) when combining platform zips, and the S3 upload step uses the broader `build/distributions/*` glob.

This stager only downloads POSIX `*.zip` / `*.csv`. If any needed zip is stored under the backslash path in Buildkite's artifact store, DRA prep can miss it while S3 still gets it.

Could you confirm against a real snapshot build artifact listing that the combined zips + `dependencies-*.csv` from `create_dra_artifacts` are always under forward-slash paths? If not, using `build/distributions/*` (same as S3) then copying would be safer.

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
```diff
@@ -8,20 +8,48 @@
...
       provider: gcp
+    plugins:
+      - elastic/dra-prep#v0.1.5:
+          product_id: "ml-cpp"
+          stack_version: "${STACK_VERSION}"
+          workflow: "${DRA_WORKFLOW}"
+
+  - label: ":pipeline: Trigger DRA processing"
+    trigger: "unified-release-dra-processing"
+    async: true
```
edsavage, line 51: The old GCS step ran in-pipeline, so a collect/upload failure failed the snapshot/staging build. This trigger is `async: true`, so the ml-cpp build can go green while DRA never lands in the final GCS path.

If that's intentional (unified-release owns retries/visibility) it's worth stating in the PR. If we still want the parent build to fail when processing fails, `async: false` (or an equivalent status check) is closer to the old behaviour.

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
```diff
@@ -8,20 +8,48 @@
...
-    command:
-      - 'buildkite-agent artifact download "build/distributions/*" --step create_dra_artifacts .'
-      - '.buildkite/scripts/steps/upload_dra_to_gcs.sh'
+    command: ".buildkite/scripts/stage_artifacts.sh"
+    env:
+      DRA_WORKFLOW: "${DRA_WORKFLOW}"
     agents:
       provider: gcp
+    plugins:
+      - elastic/dra-prep#v0.1.5:
```
edsavage, line 44: Two related merge-gate questions:

1. **WIF** — the PR notes registration in #project-stack-releases-improvements must happen before this can run. Merging without that means the next snapshot/staging DRA publish fails with no Release Manager fallback. Is WIF registered for `ml-cpp` yet?

2. **Plugin pin** — `elastic/dra-prep#v0.1.5` looks fine if that's the org-recommended version. A pointer to the migration doc or another repo on the same pin would help us not drift.

Also: this step uses a vanilla GCP agent (no JDK image). Please confirm `dra-prep`/`dractl` is fully provided by the plugin on that agent — S3 still uses the JDK17 image, so it's easy to assume tools that aren't there.

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
```diff
@@ -8,20 +8,48 @@
...
 
 . .buildkite/scripts/common/base.sh
 
+STACK_VERSION=$(awk -F= '/^elasticsearchVersion/ {print $2}' gradle.properties | xargs echo)
+
+if [ -n "${VERSION_QUALIFIER:-}" ] ; then
+    STACK_VERSION="${STACK_VERSION}-${VERSION_QUALIFIER}"
+fi
+
+if [ "${BUILD_SNAPSHOT:-true}" = "false" ] ; then
```
edsavage: Nit / robustness: this treats anything other than the string `false` as snapshot (and appends `-SNAPSHOT`). `create_dra.sh` only appends `-SNAPSHOT` when `BUILD_SNAPSHOT = "true"`.

Catalog sets `"true"` / `"false"`, so today's pipelines are fine. An unset/empty/`True` value would still diverge (prep suffixes, create_dra does not). Safer to copy the `create_dra.sh` check exactly.

edsavage (commented):
Thanks for doing this @vincentDeuschle , @navyau09 . I'm generally pretty happy with it but I have a few concerns/questions before giving it the thumbs up. They're inline below.

Also (not in the diff, but an easy follow-up): `dev-tools/ingest_build_timings.py` still maps `upload_dra_artifacts_to_gcs`. After the step key rename to `dra-prep`, timing ingest for this step will go dark until that map is updated.

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
navyau09: Hey @edsavage This is intentional. In the `DRA-prep `model, `unified-release-dra-processing` owns visibility, retries, and the final GCS move. The ml-cpp build’s responsibility ends once it hands off the manifest.

.buildkite/scripts/stage_artifacts.sh
vincentDeuschle: I checked the code in `create_dra.sh` and while it does download buildkite artifacts using backslash paths, the script itself is running on linux agents. If I read the code [here](https://github.com/elastic/ml-cpp/blob/main/.buildkite/scripts/steps/create_dra.sh#L40-L55) correctly, it is combining all downloaded artifacts under a POSIX path.

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
navyau09: **WIF** is already configured and it should work : https://github.com/elastic/infra/blob/ca83670976b72a26ddfe887834a5f2aae8d9c56d/terraform/providers/gcp/env/release-artifacts/elastic-release.tfvars

**Plugin Pin**: `v0.1.5` is the current version and we will update the version if there is any change on our end to dractl. 
we discussed using a floating tag (v0 or @latest) but decided to keep pinning for now as it gives us control over when product pipelines pick up dractl changes. As we will start making a bigger change to support Two-Speed Stack Release project.

`dractl` is installed by the plugin itself as part of its post-command hook and it does not rely on any tooling from the JDK17 image

.buildkite/pipelines/upload_dra_to_gcs.yml.sh
vincentDeuschle: Thanks, I updated the behaviour accordingly.

vincentDeuschle:
> Thanks for doing this @vincentDeuschle , @navyau09 . I'm generally pretty happy with it but I have a few concerns/questions before giving it the thumbs up. They're inline below.
> 
> Also (not in the diff, but an easy follow-up): `dev-tools/ingest_build_timings.py` still maps `upload_dra_artifacts_to_gcs`. After the step key rename to `dra-prep`, timing ingest for this step will go dark until that map is updated.

Thanks @edsavage , I updated the PR based on your comments. Please let me know if you have any further concerns. For the rollout, I think it might be useful to test this out on a dev branch first, before we merge that into main. @edsavage @navyau09 would you agree to that?

navyau09:
> > Thanks for doing this @vincentDeuschle , @navyau09 . I'm generally pretty happy with it but I have a few concerns/questions before giving it the thumbs up. They're inline below.
> > Also (not in the diff, but an easy follow-up): `dev-tools/ingest_build_timings.py` still maps `upload_dra_artifacts_to_gcs`. After the step key rename to `dra-prep`, timing ingest for this step will go dark until that map is updated.
> 
> Thanks @edsavage , I updated the PR based on your comments. Please let me know if you have any further concerns. For the rollout, I think it might be useful to test this out on a dev branch first, before we merge that into main. @edsavage @navyau09 would you agree to that?

Hey @vincentDeuschle Yes, agreed. Please trigger a test build on the branch

vincentDeuschle:
> > > Thanks for doing this @vincentDeuschle , @navyau09 . I'm generally pretty happy with it but I have a few concerns/questions before giving it the thumbs up. They're inline below.
> > > Also (not in the diff, but an easy follow-up): `dev-tools/ingest_build_timings.py` still maps `upload_dra_artifacts_to_gcs`. After the step key rename to `dra-prep`, timing ingest for this step will go dark until that map is updated.
> > 
> > 
> > Thanks @edsavage , I updated the PR based on your comments. Please let me know if you have any further concerns. For the rollout, I think it might be useful to test this out on a dev branch first, before we merge that into main. @edsavage @navyau09 would you agree to that?
> 
> Hey @vincentDeuschle Yes, agreed. Please trigger a test build on the branch

I did a test run which succeeded, but I haven't had time yet to verify the uploaded artifacts. I'll follow up here tomorrow. https://buildkite.com/elastic/ml-cpp-snapshot-builds/builds/6950#job-01a01a44-440d-451b-b641-6c8af690e018

edsavage (approved):
LGTM

vincentDeuschle:
> > > > Thanks for doing this @vincentDeuschle , @navyau09 . I'm generally pretty happy with it but I have a few concerns/questions before giving it the thumbs up. They're inline below.
> > > > Also (not in the diff, but an easy follow-up): `dev-tools/ingest_build_timings.py` still maps `upload_dra_artifacts_to_gcs`. After the step key rename to `dra-prep`, timing ingest for this step will go dark until that map is updated.
> > > 
> > > 
> > > Thanks @edsavage , I updated the PR based on your comments. Please let me know if you have any further concerns. For the rollout, I think it might be useful to test this out on a dev branch first, before we merge that into main. @edsavage @navyau09 would you agree to that?
> > 
> > 
> > Hey @vincentDeuschle Yes, agreed. Please trigger a test build on the branch
> 
> I did a test run which succeeded, but I haven't had time yet to verify the uploaded artifacts. I'll follow up here tomorrow. https://buildkite.com/elastic/ml-cpp-snapshot-builds/builds/6950#job-01a01a44-440d-451b-b641-6c8af690e018

I compared the artifacts uploaded to GCS of this test run (build id `9.6.0-190ae695`) with the latest scheduled build on main (build id `9.6.0-0bf10cac`). The list of file names under both `gs://elastic-artifacts-snapshot/ml-cpp/<build_id>/*` prefixes is identical.

I will merge this PR and continue to monitor through our scheduled snapshot builds.

PR #3167, opened 2026-08-30
Fail CMake configure if 3rd-party git clone fails (#3164) (#8.19)

edsavage:
Backport of #3164 to 8.19.

Adds `RESULT_VARIABLE` checking to both `pull-eigen.cmake` and `pull-valijson.cmake` so a transient network failure during CMake configure produces a clear error message rather than a cryptic compiler error later.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edsavage:
Closing in favour of auto-backport from #3164 after merge.

PR #3181, opened 2026-09-09
[ML] Land dormant Sandbox2/Abseil dependency foundation

valeriy42:
## Summary

Reconstructs the dormant dependency/build foundation for Sandbox2 from
current `main`, as a clean first slice ahead of the sandbox policy,
spawner, and controller-routing changes that land in follow-up PRs. No
controller or `pytorch_inference` routing changes in this PR — Sandbox2
is not selectable from any production code path yet.

Frozen PR #2873 attempted this feature in one large branch; this PR
takes just its dependency/build layer and replaces its inline
`file(WRITE)`/`string(REGEX REPLACE)` source rewrites with checked-in,
version-pinned patches that fail the configure step loudly instead of
silently no-op'ing on upstream drift.

- `3rd_party/CMakeLists.txt`: FetchContent `sandboxed-api` `v20241008` on
  Linux, applying 4 checked-in patches via `git apply` (fails configure
  loudly on upstream drift, idempotent across reconfigure).
- `3rd_party/patches/sandboxed-api/`: the 4 patches (disable vendored
  gtest, stop `-fno-exceptions` propagating into ml-cpp targets, make
  Python3 optional, link zlib + static libstdc++/libgcc into the
  forkserver binary).
- `3rd_party/licenses/{abseil,sandbox2}-*`: license/attribution files.
- `lib/sandbox/`: dormant `MlSandbox` target (`CMlSandboxAvailability`
  query only — no policy/spawner/diagnostics) plus a Linux-only
  forkserver runtime smoke test that forks/execs/reaps a
  dynamically-linked payload via `PolicyBuilder::AddLibrariesForBinary()`.

Verified locally (Linux x86_64, both a plain configure and
`-DCMAKE_UNITY_BUILD=ON`) before pushing, and green on this repo's own
Linux/macOS/Windows CI, license/security scans, and Java integration
suites.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/3rd_party/CMakeLists.txt
+++ b/3rd_party/CMakeLists.txt
@@ -39,3 +39,121 @@ execute_process(
     COMMAND ${CMAKE_COMMAND} -P ./pull-valijson.cmake
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
 )
+
+# Build Abseil and Sandbox2 on Linux only. MlSandbox (lib/sandbox) is a
+# dormant target: it is built everywhere Sandbox2 is available, but nothing
+# in the controller/pytorch_inference wiring routes to it yet. See
+# docs/projects/mlcpp-sandbox2-pr2873 in the elastic-workspace harness for
+# the staged-rollout contract this dependency foundation exists for.
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  include(FetchContent)
+
+  # Save and restore CMake cache state this block flips so it cannot change
+  # the caller's build configuration for anything outside Sandbox2/Abseil.
+  set(_saved_BUILD_TESTING ${BUILD_TESTING})
+  set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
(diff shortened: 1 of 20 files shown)


3rd_party/CMakeLists.txt
```diff
@@ -39,3 +39,120 @@ execute_process(
...
+    0001-abseil-cpp-disable-gtest.patch
+    0002-no-fno-exceptions-propagation.patch
+    0003-python3-optional.patch
+    0004-forkserver-zlib-static-libstdcxx.patch
+  )
+
+  FetchContent_Declare(
+    sandboxed-api
+    GIT_REPOSITORY https://github.com/google/sandboxed-api.git
+    GIT_TAG ${ML_SANDBOXED_API_TAG}
```
edsavage, line 86: I think we should pin the tag to its actual commit SHA here, rather than use a (mutable) tag string, to guarantee we get exactly what we want.

edsavage (approved):
LGTM, just the one minor comment.

PR #3182, opened 2026-09-09
[ML] Generate syscall policies from one declaration; fail-closed degraded seccomp

valeriy42:
Replace the hand-maintained BPF jump-offset table in `CSystemCallFilter_Linux.cc` with a program builder that derives every jump from the allowlist vector's own size/index. The applied program is generated from `CPytorchInferenceSyscallAllowlist.h`, a single machine-readable declaration, instead of a parallel hardcoded list.

`CSystemCallFilter::installSystemCallFilter()` returns a typed `ESystemCallFilterInstallOutcome` across all three platform implementations instead of `void`, and logs an `ml.seccomp.installed` readiness marker on success. `pytorch_inference/Main.cc` gains a `decideDegradedModeAction()` decision that would terminate before `CIoManager::initIo()` on any degraded-mode seccomp failure; that termination stays behind an internal switch defaulting to false until the controller has a typed way to know a degraded-mode launch was a deliberate operator choice rather than the only option available. Flipping it on today would terminate every launch on a host lacking seccomp BPF, with no operator fallback to select instead. The four non-PyTorch callers now make their unchanged log-and-continue policy explicit instead of silently discarding the result.

Also adds an explicit structured `degradedModeAttestationMarker()` so a controller/Elasticsearch observer can assert seccomp installation directly instead of inferring it from the absence of a fatal log line, and carries forward two pytorch_inference/libtorch compatibility fixes from #2873 into the shared declaration (`clone3` allowed by literal syscall number, and `prlimit64`) with a named regression test, so a from-scratch rewrite doesn't silently drop them.

`CSeccompFilterBuilderTest.cc` decodes the actually-built BPF program to prove it matches the declaration and that jump offsets are derived, not hand-maintained, plus fault-injected coverage of `decideDegradedModeAction()` for every install-failure class.

A future Sandbox2 filesystem/network policy doesn't exist yet in this branch's history; this change establishes the single declaration for that policy to consume once it's written.

Verified on a real x86_64 host: `ml_test_seccomp` (7 test cases) and the real `pytorch_inference`/`autodetect`/`categorize`/`normalize`/`data_frame_analyzer` binaries all build, link, and run correctly against the new typed API; the generated BPF program installs via a real `prctl(PR_SET_SECCOMP)` call.

### DIFF
--- a/include/seccomp/CPytorchInferenceSyscallAllowlist.h
+++ b/include/seccomp/CPytorchInferenceSyscallAllowlist.h
@@ -1,0 +1,128 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the following additional limitation. Functionality enabled by the
+ * files subject to the Elastic License 2.0 may only be used in production when
+ * invoked by an Elasticsearch process with a license key installed that permits
+ * use of machine learning features. You may not use this file except in
+ * compliance with the Elastic License 2.0 and the foregoing additional
+ * limitation.
+ */
+#ifndef INCLUDED_ml_seccomp_CPytorchInferenceSyscallAllowlist_h
+#define INCLUDED_ml_seccomp_CPytorchInferenceSyscallAllowlist_h
+
+#include <vector>
+
+#ifdef __linux__
+#include <sys/syscall.h>
+#endif
+
+namespace ml {
+namespace seccomp {
+namespace pytorch_inference {
+
+#ifdef __linux__
+
+// statx, rseq and clone3 won't be defined on a RHEL/CentOS 7 build machine,
+// but might exist on the kernel we run on, so fall back to the raw numbers.
+#if defined(__x86_64__)
+#ifndef __NR_statx
+#define ML_NR_statx 332
+#else
+#define ML_NR_statx __NR_statx
+#endif
+#ifndef __NR_rseq
+#define ML_NR_rseq 334
+#else
+#define ML_NR_rseq __NR_rseq
+#endif
+#elif defined(__aarch64__)
+#ifndef __NR_statx
+#define ML_NR_statx 291
+#else
+#define ML_NR_statx __NR_statx
+#endif
+#ifndef __NR_rseq
+#define ML_NR_rseq 293
+#else
+#define ML_NR_rseq __NR_rseq
+#endif
+#endif
+#ifndef __NR_clone3
+#define ML_NR_clone3 435
+#else
+#define ML_NR_clone3 __NR_clone3
+#endif
+
+//! Syscalls permitted by the legacy in-process BPF filter
+//! (CSystemCallFilter_Linux.cc) for every process that installs it, currently
+//! shared by pytorch_inference, autodetect, categorize, normalize and
+//! data_frame_analyzer. This is the single machine-readable declaration that
+//! the applied BPF program is generated from: CSystemCallFilter_Linux.cc
+//! contains no independent syscall list and no manually maintained jump
+//! offsets. A future Sandbox2 policy is expected to consume the same
+//! declaration for its explicit grants, so both mechanisms stay in sync.
+//!
+//! Carry-forward note: PR #2873 fixed several pytorch_inference/libtorch
+//! compatibility gaps the hard way, and this declaration is a rewrite from
+//! scratch rather than a copy of that work, so it deliberately keeps two of
+//! them. ML_NR_clone3 (see 57f00ed1b) and __NR_prlimit64 (see 03b1ee4a) are
+//! carried into this shared declaration so a future Sandbox2 policy
+//! inherits them automatically instead of rediscovering them the same way;
+//! CSeccompFilterBuilderTest.cc asserts both stay present. The x86_64
+//! legacy filesystem syscalls below (see ec7d3ed85) were already part of
+//! this filter's syscall set prior to this declaration and remain
+//! unchanged. PR #2873's futex-op broadening (see d9a856d5f) and CI
+//! link-order/test-bundle packaging fixes (see 730933db, f8b0a534) apply to
+//! the Sandbox2 policy and its Buildkite pipeline respectively, not to this
+//! file — carry those forward when that code is written instead of
+//! rediscovering them.
+inline std::vector<int> legacyBpfAllowedSyscalls() {
+    std::vector<int> syscalls {
+#if defined(__x86_64__)
+        __NR_access, __NR_open, __NR_dup2, __NR_unlink, __NR_stat, __NR_lstat,
+            __NR_time, __NR_readlink, __NR_getdents, // for forecast temp storage
+            __NR_rmdir, // for forecast temp storage
+            __NR_mkdir, // for forecast temp storage
+            __NR_mknod,
+#elif defined(__aarch64__)
+        __NR_faccessat,
+#endif
+            __NR_fcntl, // for fdopendir
+            __NR_getrusage,
+            __NR_getpid,    // for pthread_kill
+            ML_NR_statx,    // for create_directories
+            __NR_getrandom, // for unique_path
+            __NR_mknodat, __NR_newfstatat, __NR_readlinkat, __NR_dup3,
+            __NR_getpriority, // for nice
+            __NR_setpriority, // for nice
+            __NR_read, __NR_write, __NR_writev, __NR_lseek, __NR_clock_gettime,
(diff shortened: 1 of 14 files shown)


include/seccomp/CPytorchInferenceSyscallAllowlist.h
```diff
@@ -0,0 +1,128 @@
...
+#include <vector>
+
+#ifdef __linux__
+#include <sys/syscall.h>
+#endif
+
+namespace ml {
+namespace seccomp {
+namespace pytorch_inference {
+
```
edsavage, line 23: I think that both the name and the `pytorch_inference` namespace are misleading here. The header comment notes this allowlist is shared by all five ML binaries (pytorch_inference, autodetect, categorize, normalize, data_frame_analyzer), but the name implies it is exclusive to pytorch_inference. A reader of `autodetect/Main.cc` has no way to know that `installSystemCallFilter()` internally calls `pytorch_inference::legacyBpfAllowedSyscalls()`.

I get the forward-looking intention (that the Sandbox2 policy will eventually consume this declaration specifically for pytorch_inference), but that makes the naming more load-bearing, not less. When the other four binaries eventually get their own policies the name will actively mislead. Something like `CMlLegacyBpfSyscallAllowlist.h` with the function in `ml::seccomp` (not a `pytorch_inference` sub-namespace) would still be accurate today while keeping the door open for the future split.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -30,125 +42,64 @@ namespace {
...
-#endif
-    BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_faccessat, 45, 0),
-#else
-#error Unsupported hardware architecture
+    // Jump to the deny row (immediately after the last syscall row below,
+    // i.e. numSyscalls rows ahead) for calls using the x32 ABI, without
+    // checking any allowlisted syscall.
+    program.push_back(BPF_JUMP(BPF_JMP | BPF_JGT | BPF_K, UPPER_NR_LIMIT,
+                               static_cast<std::uint8_t>(numSyscalls), 0));
 #endif
```
edsavage, line 80: Both here and in the per-syscall loop below (`jumpToAllow`), the cast to `std::uint8_t` will silently overflow if the allowlist ever grows past 255 entries. The current list is around 45 so this is fine in practice, but worth guarding explicitly with something like:
```cpp
assert(numSyscalls <= std::numeric_limits<std::uint8_t>::max() &&
       "Syscall allowlist too large for BPF jump offset");
```

lib/seccomp/unittest/CSeccompFilterBuilderTest.cc
```diff
@@ -0,0 +1,240 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(testCarryForwardSyscallsPresent) {
+    // PR #2873 fixed these pytorch_inference/libtorch compatibility gaps
+    // the hard way; this declaration is a rewrite from scratch, and must
+    // not silently drop them. Each assertion below is a named regression
+    // test for one carried-forward fix within this file's scope.
+    const std::set<int> declared{
+        ml::seccomp::pytorch_inference::legacyBpfAllowedSyscalls().begin(),
+        ml::seccomp::pytorch_inference::legacyBpfAllowedSyscalls().end()};
```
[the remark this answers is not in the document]
valeriy42: Fixed in 0df628b77: `testCarryForwardSyscallsPresent` binds one `std::vector<int>` then builds the `std::set` from that vector's iterators — same pattern as `testAppliedProgramMatchesDeclaration`.

lib/seccomp/CSystemCallFilter_Linux.cc
```diff
@@ -30,125 +42,64 @@ namespace {
...
-    BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_getuid, 4, 0),
-    BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_exit_group, 3, 0),
-    BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_brk, 2, 0),
-    BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, __NR_exit, 1, 0),
+    // Every syscall row jumps to the terminal SECCOMP_RET_ALLOW row on match.
+    // The jump distance is derived from the row's own index and the total
+    // count, so adding, removing or reordering an entry in allowedSyscalls
+    // never requires touching any other row.
+    for (std::uint32_t i = 0; i < numSyscalls; ++i) {
+        const auto jumpToAllow = static_cast<std::uint8_t>(numSyscalls - i);
```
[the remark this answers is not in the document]
valeriy42: Addressed in 0df628b77: production inventory is `constexpr` with `static_assert(std::size(kLegacyBpfAllowedSyscalls) <= 255)` in `CMlLegacyBpfSyscallAllowlist.h`, so a 256th entry fails compile in every TU that includes it (not a RelWithDebInfo-stripped `assert`).

include/seccomp/CSystemCallFilter.h
```diff
@@ -41,9 +43,85 @@ namespace seccomp {
...
+//! that a legacy/degraded-mode pytorch_inference launch actually installed
+//! its in-process seccomp filter before processing untrusted model input.
+//! Replaces attesting readiness by inference — "no fatal log line appeared
+//! before initIo() ran" — with an explicit signal a test or observer can
+//! assert on directly. Returns empty when
+//! installation did not succeed: a failed degraded launch already exits
+//! before initIo() (see decideDegradedModeAction()) and must never emit
+//! this marker, since doing so would falsely attest a filter that isn't
+//! there. Logged over the existing per-process log pipe; this is not a new
+//! startup channel.
```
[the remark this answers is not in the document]
valeriy42: Comment updated in 0df628b77: `degradedModeAttestationMarker` returns empty on non-`Installed`; terminate-before-`initIo()` only when `decideDegradedModeAction(..., terminateOnFailure=true)` — production still passes `false`.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
```diff
@@ -0,0 +1,128 @@
...
+#include <vector>
+
+#ifdef __linux__
+#include <sys/syscall.h>
+#endif
+
+namespace ml {
+namespace seccomp {
+namespace pytorch_inference {
+
```
valeriy42: Renamed in 0df628b77 to `include/seccomp/CMlLegacyBpfSyscallAllowlist.h`; dropped `pytorch_inference` namespace; `legacyBpfAllowedSyscalls()` stays in `ml::seccomp`.

lib/seccomp/CSystemCallFilter_Linux.cc
valeriy42: Same as the `jt` cap fix in 0df628b77: compile-time `static_assert` on the declaration array size (`<= 255`), not a runtime `assert`.

edsavage (approved):
LGTM

Just a couple of minor observations, that don't necessarily have to be addressed in this PR.

There's a macOS outcome misclassification in `CSystemCallFilter_MacOSX.cc`. When `writeTempRulesFile()` returns empty we return `E_MechanismUnavailable`, but that path also covers the `mkstemps/temp-file` I/O failures,  which are setup failures, not due to an unavailable sandbox mechanism. I _think_ we should be mapping to `E_FilterInstallFailed` (or another distinct profile-setup outcome).

Also, I'm thinking we could guard the builder itself. The `static_assert` protects the production allowlist, but `buildSyscallAllowlistProgram(const std::vector<int>&)` still casts `numSyscalls/jumpToAllow` to `std::uint8_t` unconditionally, so a future/test caller passing >255 entries would still silently wrap the jump offsets. An early-out (return empty) or bounds check in the builder would make the contract self-enforcing.

valeriy42:
Addressed the two leftover observations from the LGTM in efb9acdce:

**macOS empty `writeTempRulesFile()`.** This path already returned `E_FilterInstallFailed` in 0df628b, not `E_MechanismUnavailable`. Added an intent comment so that mapping is not treated as a taxonomy bug and remapped back. `sandbox_init` failure stays `E_FilterInstallFailed`.

**Builder `uint8_t` wrap.** `static_assert` still caps the production allowlist at 255. `buildSyscallAllowlistProgram(const std::vector<int>&)` now returns empty when the caller passes more than 255 entries, and `installSystemCallFilter()` treats an empty program as `E_FilterInstallFailed` instead of installing wrapped jump offsets. Tests pin the 255/256 boundary.

PR #3184, opened 2026-09-09
[ML] Typed filesystem/network launch policy for Sandbox2 pytorch_inference

valeriy42:
Stacks on #3182 (PR B). PR C of the Sandbox2 clean rebuild plan (see the elastic-workspace harness, docs/projects/mlcpp-sandbox2-pr2873).

Replaces raw argument-directory inference in `CPytorchInferenceSandboxPolicy` with a typed launch spec: every `input`/`output`/`restore`/`logPipe` path is validated against the pinned child-root contract (`$TMPDIR/ml-child-ipc/<child-id>`) before any policy is built. Rejects relative, root-level, dot-dot, out-of-root, wrong-depth, duplicate, mutable-symlink/alias, and cross-option child-id-mismatch paths - never widens a mount to recover a rejected argument.

Also minimizes the filesystem policy: enumerates and justifies all seven historically bulk-mounted fixed directories, replaces whole `/etc` with five individually justified files, never binds host `/proc`/`/sys` (relies on Sandbox2's own namespaced procfs/sysfs), uses a private bounded tmpfs at `/tmp` instead of the host's, and consumes the syscall allowlist already shared with the legacy BPF filter instead of hand-duplicating it.

Adds a purpose-built allowlisted mechanism probe (`ml_sandbox_probe`) proving allowed IPC access, denied host reads, denied external egress, loopback reachability, and mount enumeration, plus a portable validator unit-test suite and a Linux-only mechanism integration test.

Verified this session (macOS host, no Sandbox2/Linux toolchain available): the validator compiles clean with `-Wall -Wextra` outside the ml-cpp build system, and a standalone driver exercises every rejection/acceptance path against real `mkdtemp`/`mkdir`/`symlink` fixtures - all pass. Not yet verified: the `SANDBOX2_AVAILABLE` build, the CMake wiring, and the mechanism-probe integration test itself - needs a Linux CI/devbox pass. Opening as draft for that reason.

### DIFF
--- a/include/sandbox/CPytorchInferenceSandboxPolicy.h
+++ b/include/sandbox/CPytorchInferenceSandboxPolicy.h
@@ -1,0 +1,144 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the following additional limitation. Functionality enabled by the
+ * files subject to the Elastic License 2.0 may only be used in production when
+ * invoked by an Elasticsearch process with a license key installed that permits
+ * use of machine learning features. You may not use this file except in
+ * compliance with the Elastic License 2.0 and the foregoing additional
+ * limitation.
+ */
+#ifndef INCLUDED_ml_sandbox_CPytorchInferenceSandboxPolicy_h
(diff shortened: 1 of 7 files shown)


valeriy42:
Superseded: reopening from an elastic/ml-cpp branch (not the fork) so this stacks correctly with gh-stack alongside #3181/#3182 - see the replacement PR linked here shortly.

PR #3185, opened 2026-09-09
[ML] Typed filesystem/network launch policy for Sandbox2 pytorch_inference

valeriy42:
Stacks on #3182.

Replaces raw argument-directory inference in `CPytorchInferenceSandboxPolicy` with a typed launch spec: every `input`/`output`/`restore`/`logPipe` path is validated against a pinned child-root contract (`$TMPDIR/ml-child-ipc/<child-id>`) before any policy is built. Rejects relative, root-level, dot-dot, out-of-root, wrong-depth, duplicate, mutable-symlink/alias, and cross-option child-id-mismatch paths - never widens a mount to recover a rejected argument.

Also minimizes the filesystem policy: enumerates and justifies all seven historically bulk-mounted fixed directories, replaces whole `/etc` with five individually justified files, never binds host `/proc`/`/sys` (relies on Sandbox2's own namespaced procfs/sysfs), uses a private bounded tmpfs at `/tmp` instead of the host's, and consumes the syscall allowlist already shared with the legacy BPF filter instead of hand-duplicating it.

Adds a purpose-built allowlisted mechanism probe (`ml_sandbox_probe`) proving allowed IPC access, denied host reads, denied external egress, loopback reachability, and mount enumeration, plus a portable validator unit-test suite and a Linux-only mechanism integration test.

Verified this session: the validator's core logic compiles clean with `-Wall -Wextra -Werror` and passes a standalone driver covering every rejection/acceptance path against real `mkdtemp`/`mkdir`/`symlink` fixtures. The `SANDBOX2_AVAILABLE`/Linux path compiles clean against stub sandbox2/seccomp headers (no vendored Sandbox2 headers available on this host). Not yet verified: an actual Sandbox2 run of the mechanism probe and the real Linux CMake/build integration - needs a Linux CI or devbox pass, in progress. Also fixes a Windows build break this change would otherwise have introduced (POSIX-only realpath/PATH_MAX used unconditionally in a file ml-cpp builds on every platform).

### DIFF
--- a/include/sandbox/CPytorchInferenceSandboxPolicy.h
+++ b/include/sandbox/CPytorchInferenceSandboxPolicy.h
@@ -1,0 +1,148 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the following additional limitation. Functionality enabled by the
+ * files subject to the Elastic License 2.0 may only be used in production when
+ * invoked by an Elasticsearch process with a license key installed that permits
+ * use of machine learning features. You may not use this file except in
+ * compliance with the Elastic License 2.0 and the foregoing additional
+ * limitation.
+ */
+#ifndef INCLUDED_ml_sandbox_CPytorchInferenceSandboxPolicy_h
+#define INCLUDED_ml_sandbox_CPytorchInferenceSandboxPolicy_h
+
+#include <string>
+#include <vector>
+
+#ifdef SANDBOX2_AVAILABLE
+#include <sandboxed_api/sandbox2/policybuilder.h>
+#endif
+
+namespace ml {
+namespace sandbox {
+
+//! Reasons a path-bearing launch argument fails typed validation against the
+//! pinned child-root contract (see validateChildIpcLaunchSpec below). Every
+//! value here must fail *before* a policy is constructed; none of them widen
+//! a mount to recover.
+enum class EChildIpcPathRejection {
+    E_NotAbsolute,    //!< value does not start with '/'.
+    E_RootLevelPath,  //!< value has no mountable parent directory below '/'.
+    E_ContainsDotDot, //!< value has a ".." path component.
+    E_CanonicalizationFailed, //!< the trusted base or the value's parent directory could not be
+    //!< resolved (realpath() on POSIX, _fullpath() on Windows).
+    E_OutsideTrustedBase, //!< canonical parent is not beneath the trusted $TMPDIR.
+    E_WrongDepth, //!< canonical parent is not exactly $TMPDIR/ml-child-ipc/<child-id>.
+    E_ChildIdMismatch, //!< two path options resolved to a different <child-id>.
+    E_MutableSymlinkOrAlias, //!< the literal and canonical parent directories diverge.
+    E_Duplicate //!< the same literal argument was supplied more than once.
+};
+
+//! One rejected path-bearing argument and why.
+struct SRejectedChildIpcPath {
+    std::string s_Arg;
+    EChildIpcPathRejection s_Reason;
+};
+
+//! A typed, validated launch specification for a single sandboxed
+//! pytorch_inference child, derived from its path-bearing launch options
+//! (input, output, restore, logPipe). Replaces raw argument-directory
+//! inference: every accepted path is provably beneath the one pinned
+//! per-child IPC root, never inferred from arbitrary argv content.
+struct SChildIpcLaunchSpec {
+    //! <child-id> path component shared by every accepted path option.
+    //! Empty whenever the overall result is not s_Ok - either no
+    //! recognized path option was present, or at least one was rejected
+    //! (SChildIpcValidationResult clears the whole spec on any rejection).
+    std::string s_ChildId;
+    //! Canonical $TMPDIR/ml-child-ipc/<child-id> - the directory the native
+    //! controller creates (mode 0700) before policy construction, and the
+    //! only host directory CSandboxedProcessSpawner maps to
+    //! /run/elastic/ml-ipc. Empty iff s_ChildId is empty.
+    std::string s_ChildIpcRoot;
+    //! Canonical paths of every accepted path-bearing argument, always
+    //! s_ChildIpcRoot plus exactly one leaf component.
+    std::vector<std::string> s_PipePaths;
+};
+
+//! Result of validating a pytorch_inference launch command line against the
+//! pinned child-root contract.
+struct SChildIpcValidationResult {
+    //! True only when at least one path option was present and every
+    //! path option that was present was accepted. False means the caller
+    //! must fail the spawn - never fall back to a partially-built policy.
+    bool s_Ok = false;
+    SChildIpcLaunchSpec s_Spec;
+    std::vector<SRejectedChildIpcPath> s_Rejected;
+};
+
+//! Validate every input/output/restore/logPipe argument in args against the
+//! pinned child-root contract: each must canonicalize to a parent directory
+//! of exactly trustedTmpDir/ml-child-ipc/<child-id>, for one consistent
+//! <child-id>, with no ".."; no relative, root, or out-of-root path; no
+//! divergent literal/canonical parent; and no duplicate literal argument.
+//! Scalar (non path-bearing) options are never inspected as candidate paths.
+//! trustedTmpDir must already be the canonical form of the operator's
+//! Environment.tmpDir(); this function does not itself decide what counts
+//! as trusted.
+SChildIpcValidationResult validateChildIpcLaunchSpec(const std::string& trustedTmpDir,
+                                                     const std::vector<std::string>& args);
+
+#ifdef SANDBOX2_AVAILABLE
+
+//! What buildPytorchInferenceFilesystemPolicy does with one of the seven
+//! historically bulk-mounted fixed directories
+//! (/lib /lib64 /usr/lib /usr/lib64 /etc /proc /sys). Mounting whole /etc or
+//! binding the host's /proc or /sys directly is non-conformant.
+enum class EFixedMountAction {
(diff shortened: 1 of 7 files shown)


include/sandbox/CPytorchInferenceSandboxPolicy.h
```diff
@@ -0,0 +1,148 @@
...
+//! The minimization decision applied to each of the seven historically
+//! bulk-mounted fixed directories, with its justification. /etc is Skip
+//! (see allowlistedEtcFiles() for the narrower replacement); /proc and /sys
+//! are the Sandbox2-namespaced procfs/sysfs, never a host bind (the
+//! ml_sandbox_probe mechanism test asserts this held for a real launch, via
+//! its pid_namespace check). /lib, /lib64, /usr/lib, /usr/lib64 remain whole
+//! read-only directories: the dynamic loader resolves libtorch/glibc shared
+//! objects from them at runtime from an unbounded, platform-dependent set,
+//! so per-file allowlisting would duplicate the loader's own search logic.
+const std::vector<SFixedMountDecision>& fixedMountDecisions();
```
edsavage: This function would probably be better suited to be defined in the .cc file alongside the only caller of it (buildPytorchInferenceFilesystemPolicy())

include/sandbox/CPytorchInferenceSandboxPolicy.h
```diff
@@ -0,0 +1,148 @@
...
+
+//! Individual /etc files pytorch_inference/libtorch are demonstrated to
+//! need, replacing a whole-/etc bind. Extend only with a named consumer.
+const std::vector<std::string>& allowlistedEtcFiles();
```
edsavage: Same as fixedMountDecisions, this would be better suited to live in the .cc..

Also, if that were the case EFixedMountAction and SFixedMountDecision could also be moved there too.

lib/sandbox/unittest/CPytorchInferenceSandboxPolicyMechanismTest_Linux.cc
```diff
@@ -0,0 +1,182 @@
...
+    // plus "." and "..", never a full directory bind. A regression back to
+    // AddDirectory("/etc", true) would spike this into the dozens/hundreds,
+    // so an upper bound catches it without hard-coding the exact count.
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "etc_enumeration"), "counted");
+    BOOST_TEST_REQUIRE(std::stoi(detailFor(resultsContent, "etc_enumeration")) <= 10);
+
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "pid_namespace"), "namespaced");
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "loopback_reachable"), "ok");
+
+    ::unlink((childRoot + "/probe.txt").c_str());
```
edsavage: I'm a bit concerned about these clean up steps getting missed if we hit assertion above. Maybe a RAII mechanism would be safer?

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
```diff
@@ -0,0 +1,378 @@
...
+    };
+    return DECISIONS;
+}
+
+const std::vector<std::string>& allowlistedEtcFiles() {
+    // NOTE: /etc/ssl/certs/ca-certificates.crt is the Debian/Ubuntu trust
+    // bundle path; the ml-cpp CI build image is CentOS7/RHEL-based, whose
+    // equivalent is /etc/pki/tls/certs/ca-bundle.crt. This list has not yet
+    // been verified against the actual supported-distro trust bundle path -
+    // an open item, not resolved here.
```
edsavage: Should we track this in a separate issue?

edsavage (approved):
LGTM

Just a few minor suggestions.

include/sandbox/CPytorchInferenceSandboxPolicy.h
valeriy42: Done in ab6823534: `EFixedMountAction`, `SFixedMountDecision`, `fixedMountDecisions()`, and `allowlistedEtcFiles()` moved into the `.cc` anonymous namespace; the header now exposes only the validator types and `buildPytorchInferenceFilesystemPolicy`.

include/sandbox/CPytorchInferenceSandboxPolicy.h
valeriy42: Same as the sibling thread — mount decision tables are implementation details in `CPytorchInferenceSandboxPolicy.cc` now (ab6823534).

lib/sandbox/unittest/CPytorchInferenceSandboxPolicyMechanismTest_Linux.cc
valeriy42: Done in ab6823534: `CMechanismProbeFixture` RAII guard in the mechanism IT destructor cleans up `probe.txt`, `results.txt`, child root, `ml-child-ipc`, and the mkdtemp dir even when `BOOST_REQUIRE_*` aborts.

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
valeriy42: Opened elastic/ml-cpp#3200 to track verifying the distro-specific TLS trust-bundle path; the in-code NOTE in `allowlistedEtcFiles()` links it (ab6823534). Not resolved in this PR.

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    // Private, bounded tmpfs - never the host's shared /tmp.
+    policyBuilder.AddTmpfs("/tmp", tmpfsSizeBytes);
+
+    // The one per-child IPC root, mapped read-write to a fixed in-sandbox
+    // path. spec must already be s_Ok (validateChildIpcLaunchSpec), so
+    // s_ChildIpcRoot is exactly $TMPDIR/ml-child-ipc/<child-id> - never
+    // ml-child-ipc itself, never a sibling child's directory.
+    policyBuilder.AddDirectoryAt(spec.s_ChildIpcRoot, "/run/elastic/ml-ipc", /*is_ro=*/false);
```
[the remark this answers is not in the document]
valeriy42: Implemented in ab6823534: `buildPytorchInferenceFilesystemPolicy` now takes `const SChildIpcValidationResult&`; on `s_Ok == false` or a non-`ml-child-ipc` `s_ChildIpcRoot` it `SetError`s and skips `AddDirectoryAt`.

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
```diff
@@ -0,0 +1,378 @@
...
+            // like "--inp=/path". None of those are a mount-widening bypass:
+            // an unrecognized option is never added to s_PipePaths, so its
+            // directory is simply never mounted and the spawn either fails
+            // closed (pipe unreachable) or gets rejected elsewhere. The sole
+            // production caller, ProcessPipes.addArgs() in
+            // elasticsearch/x-pack/plugin/ml, always emits the exact
+            // concatenated "--input=" + value form, so this is a defensive
+            // fail-closed gap rather than an active exploit path. Left
+            // unfixed rather than special-cased.
+            continue;
```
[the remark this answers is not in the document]
valeriy42: Declining this PR: the fail-closed NOTE at `CPytorchInferenceSandboxPolicy.cc` documents that only `--name=value` tokens are parsed; production `ProcessPipes` emits that form and unrecognized spellings are never mounted.

lib/sandbox/unittest/CPytorchInferenceSandboxPolicyMechanismTest_Linux.cc
```diff
@@ -0,0 +1,182 @@
...
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "host_read_etc_shadow"), "denied");
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "private_tmpfs_write"), "allowed");
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "external_egress"), "denied");
+
+    // Mount conformance: /etc must list only allowlistedEtcFiles() (5 entries)
+    // plus "." and "..", never a full directory bind. A regression back to
+    // AddDirectory("/etc", true) would spike this into the dozens/hundreds,
+    // so an upper bound catches it without hard-coding the exact count.
+    BOOST_REQUIRE_EQUAL(outcomeFor(resultsContent, "etc_enumeration"), "counted");
+    BOOST_TEST_REQUIRE(std::stoi(detailFor(resultsContent, "etc_enumeration")) <= 10);
```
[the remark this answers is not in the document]
valeriy42: Declining this PR: the `<= 10` upper bound still catches a regression to whole-`/etc` bind (dozens/hundreds of entries); tightening to an exact allowlist count can follow once the probe contract is frozen.

lib/sandbox/unittest/payloads/ml_sandbox_probe.cc
```diff
@@ -0,0 +1,198 @@
...
+        report("host_read_etc_shadow", "allowed");
+    }
+
+    // Private tmpfs: writability alone doesn't prove /tmp is a private
+    // tmpfs rather than a host bind - a regressed policy that AddDirectory's
+    // the real host /tmp would still pass a plain write check on any
+    // world-writable host. statfs()'s f_type is the actual mechanism
+    // distinguishing tmpfs from a bind-mounted host directory.
+    struct statfs tmpStatfs {};
+    const bool isTmpfs = ::statfs("/tmp", &tmpStatfs) == 0 && tmpStatfs.f_type == TMPFS_MAGIC;
```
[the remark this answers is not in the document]
valeriy42: Declining this PR: `TMPFS_MAGIC` assertion is not added here; private tmpfs behavior is already covered by the `private_tmpfs_write` mechanism outcome in the probe IT.

PR #3188, opened 2026-09-10
[ML] Typed routing behind the established controller protocol

valeriy42:
## Summary

Stacks on #3187. Adds typed controller-side routing behind two symmetric controller tokens.

- `CCommandProcessor` parses at most one of `--disableSandbox` (operator kill switch, forces legacy) or `--requireSandbox` (operator opt-in, forces Sandbox2) for the exact configured `pytorch_inference` path. Duplicate occurrences of either token, a mismatched process path, or both tokens present together are all rejected before spawn. The selected token is stripped before the child ever sees it.
- New `CProcessSpawnerRouter` dispatches the decided route to `CSandboxedProcessSpawner` or `CDetachedProcessSpawner` with no automatic fallback: a required Sandbox2 launch that fails returns a failed start response, it never retries through the legacy spawner.
- A `start` command with neither token always takes the legacy route - permanent behaviour for any caller that sends no routing token (support/debug scripts, direct controller invocation, the test harness), not a rollout seam. Elasticsearch is expected to always send exactly one of the two tokens per launch, chosen from its own operator setting's live value.
- Structured once-per-launch observability signal (`sandbox2_launch`: `deployment_id`, `model_id`, `route`, `sandbox2_established`, `mode`, `legacy_reason`, `sandbox2_compiled_in`) so a consumer can distinguish an operator kill switch from the no-token default, and "Sandbox2 supported but no routing token sent" from "built without Sandbox2 support."
- `ML_SANDBOXED` is stripped from every legacy-route child's environment (POSIX and Windows) so an inherited or injected value can never fail-open the mandatory in-process seccomp filter.
- Staged user-namespace capability probe plus CI wiring: aarch64 runs enforced-mode coverage; x86_64 runs fail-closed-only coverage, since x86_64 Buildkite k8s pods get `EPERM` on `mount("proc", ...)` and there is currently no userns-capable x86_64 CI runner.
- Repaired `test_sandbox2_attack_defense.py`: reached markers, unsandboxed positive controls, per-case cleanup/reap assertions, the real `ml-child-ipc/<child-id>` IPC layout, an assertion that the controller's own `sandbox2_launch` signal reports the expected route before any security-boundary check runs, and explicit `--requireSandbox`/`--disableSandbox` tokens per case instead of a global env-var default.
- Producer-side controller-protocol capability token (`3rd_party/controller-protocol.version`, now version 2) for a future cross-repo compatibility check.

## Filesystem-policy fixes found during enforced-mode qualification

End-to-end qualification on the qaf harness with `xpack.ml.trained_models.sandbox_enabled=true` (real `mode:enforced` route, not the legacy fallback) surfaced several launch/filesystem-policy gaps that previously only manifested once a real `pytorch_inference` ran under the enforced sandbox. Fixed here:

- **Per-child IPC directory is created before validation.** The native controller now creates `$TMPDIR/ml-child-ipc/<child-id>` (mode 0700) before `validateChildIpcLaunchSpec()`'s live `realpath()` calls, on both spawn paths. Elasticsearch only ever constructs the IPC path strings; nothing created the directory they name.
- **Per-child IPC root is mounted at the same path inside and outside the sandbox** (no `/run/elastic/ml-ipc` remap), because `pytorch_inference` receives its `--input=`/`--output=`/`--restore=`/`--logPipe=` argv as host paths under that root - a remap left them unresolvable in the sandbox mount namespace.
- **Sandbox2-specific syscall allowlist** ported from the frozen reference (#2873): Sandbox2's namespace/threading setup exercises syscalls the legacy in-process BPF filter never needed, so `legacyBpfAllowedSyscalls()` alone is insufficient.
- **`/proc` is mounted (PID-namespaced) inside the sandbox rootfs.** Sandbox2 mounts a fresh PID-namespaced procfs on the outer root, then `pivot_root`s into the chroot and detaches the old root, so the rootfs had no `/proc`. Without it `readlink(/proc/self/exe)` fails with `ENOENT`, breaking Intel oneMKL's runtime dispatcher (it reads `/proc/self/exe` to self-locate and `dlopen` its CPU-specific `libmkl_*.so.3` kernels) - every enforced `pytorch_inference` aborted with `Intel oneMKL FATAL ERROR: Cannot load <mkl-loader>`. The mount binds the already-namespaced procfs (never the host's): inside the sandbox `/proc` shows only the sandboxee's own PIDs. `/sys` is left unmounted.
- **Regression test:** `ml_sandbox_probe` now checks `/proc/self/exe` is readable inside the real sandbox, asserted by `CPytorchInferenceSandboxPolicyMechanismTest_Linux` - turning the MKL crash into a build-time failure.

Verified: ml-cpp sandbox unit tests 33/33; qaf boot check and `test_scenario_buildly` (8/8) pass under `mode:enforced` with zero MKL crashes and no legacy fallbacks.


lib/sandbox/CPytorchInferenceSandboxPolicy.cc
```diff
@@ -168,6 +184,29 @@ bool childIpcRootHasExpectedShape(const std::string& childIpcRoot) {
...
+    // that is inert until a Windows caller exists.
+    if (::_mkdir(dir.c_str()) == 0) {
+        return true;
+    }
+    return errno == EEXIST;
+#else
+    if (::mkdir(dir.c_str(), 0700) == 0) {
+        return true;
+    }
+    return errno == EEXIST;
```
edsavage, line 207: There's a mismatch between docs and implementation: the comment above (and the `E_CreationFailed` enum doc in the header, "a non-directory in the way") says a non-directory occupying `dir` is reported as a failure, but `mkdir` on an existing regular file or symlink returns `EEXIST`, so this returns `true` (success). Only an `ENOTDIR` from a non-directory *parent component* would be caught. It's not exploitable (validateChildIpcLaunchSpec canonicalizes and rejects afterward via `E_MutableSymlinkOrAlias`/shape checks), but either fix the comment or tighten with an `lstat`/`S_ISDIR` check on the `EEXIST` branch so the code matches its stated contract.

bin/controller/CProcessSpawnerRouter.cc
```diff
@@ -0,0 +1,313 @@
...
+    // inside CSandboxedProcessSpawner_Linux.cc::spawn() is the other), and
+    // it runs strictly before spawn() dispatches to either backend - on the
+    // legacy route just as much as the Sandbox2 route, since the signal
+    // below always wants a real deployment_id. Ensure the directory exists
+    // here too, rather than relying on the Sandbox2 spawner (which may not
+    // even run on this route) to have already done it. A creation failure
+    // is not logged again here: an empty deployment_id in the signal is
+    // itself the observable symptom, and the Sandbox2 spawner (when that
+    // route is actually taken) logs the failure with detail.
+    ml::sandbox::ensureChildIpcDirectory(trustedTmpDir, args);
```
edsavage, line 122: I can't spot anything in the router or spawner that ever removes `$TMPDIR/ml-child-ipc/<child-id>`, and this call now creates it on *every* sandbox-eligible spawn, including the legacy/degraded route, purely so the `sandbox2_launch` signal has a real deployment_id. So a plain no-token launch of `pytorch_inference` now leaves a 0700 directory behind even when Sandbox2 is never involved. I'm just wondering what reaps these across the child's lifetime and on controller restart? If Elasticsearch owns cleanup then a brief comment would help.

bin/controller/CProcessSpawnerRouter.cc
```diff
@@ -0,0 +1,313 @@
...
+        // through to the legacy spawner as the frozen router's #ifdef
+        // Linux masked this exact case by doing.
+        LOG_ERROR(<< "Refusing to launch '" << processPath << "': configured as a sandboxed process path, but this "
+                  << "build was not compiled with Sandbox2 support");
+        spawned = false;
+#endif
+    } else {
+        // Not a sandboxed process path: unrelated processes always go via
+        // the legacy spawner, unchanged from today's behaviour.
+        spawned = m_LegacySpawner.spawn(processPath, args, childPid);
```
edsavage, line 273: If a caller passes `route == E_Sandbox2` but `processPath` is not a configured sandboxed path, this falls through to the legacy spawner and emits *no* `sandbox2_launch` signal (`sandboxEligible == false`). CCommandProcessor makes this unreachable in production (it rejects `--requireSandbox` on non-sandboxed paths), but this router is a standalone public class whose contract permits the combination. A `LOG_WARN`/assert on "E_Sandbox2 requested for a non-sandboxed path" would harden it against future callers rather than silently downgrading to legacy.

bin/controller/Main.cc
```diff
@@ -206,8 +206,17 @@ int main(int argc, char** argv) {
...
-
-    ml::controller::CCommandProcessor processor{permittedProcessPaths, *outputStream};
+    // Unconditional on every platform, deliberately: this list only
+    // nominates which process path the --disableSandbox/--requireSandbox
+    // controller tokens are meaningful for, it does not by itself require
+    // Sandbox2 for that path. A plain (no-token) launch of
+    // ./pytorch_inference always takes the legacy route (see
+    // CCommandProcessor), so listing it here fails nothing on macOS,
+    // Windows, or a Linux build without Sandbox2 support.
+    ml::controller::CCommandProcessor::TStrVec sandboxedProcessPaths{"./pytorch_inference"};
```
edsavage, line 216: This comment ("fails nothing on macOS, Windows, or a Linux build without Sandbox2 support") is only true for the *no-token* path. Now that `./pytorch_inference` is listed as being sandboxed on every platform, an explicit `--requireSandbox` for it on a non-Sandbox2 build routes to `E_Sandbox2`, then hits the router's `#else`, and then **fails the launch**. Which is all arguably correct (since the operator asked for a sandbox the build can't provide), but it's probably worth (a) confirming that ES never sends `--requireSandbox` on non-Linux, and (b) tightening this comment so it doesn't read as "no token ever fails here."

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
```diff
@@ -124,6 +125,187 @@ inline std::vector<int> legacyBpfAllowedSyscalls() {
...
+    };
+    return syscalls;
+}
+
+//! Syscalls covered by Sandbox2 PolicyBuilder helpers (AllowRead/AllowWrite/
+//! AllowOpen/etc. in buildPytorchInferencePolicy()) that are also present in
+//! legacyBpfAllowedSyscalls() - tracked so sandbox2AllowsAllLegacySyscalls() can
+//! assert Sandbox2 never grants strictly less than the legacy filter without
+//! requiring every one of these to be repeated in sandbox2ExplicitSyscalls().
+inline std::vector<int> sandbox2HelperCoveredSyscalls() {
```
edsavage: `sandbox2AllowsAllLegacySyscalls()` trusts this hand-maintained list to reflect what the Sandbox2 PolicyBuilder helpers (`AllowRead`/`AllowWrite`/...) actually grant. A future sandboxed-api upgrade could change helper coverage, and the assertion would still pass while the real applied policy diverges (granting less than the legacy filter). I think its worth a one-line note here that this list must be re-verified whenever the pinned Sandbox2 version is bumped.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
```diff
@@ -124,6 +125,187 @@ inline std::vector<int> legacyBpfAllowedSyscalls() {
...
+        __NR_getrandom,
+        __NR_exit,
+        __NR_exit_group,
+        __NR_newfstatat,
+        __NR_fstat,
+        __NR_getuid,
+        __NR_getgid,
+        __NR_geteuid,
+        __NR_getegid,
+        ML_NR_statx,
```
edsavage: This newly-added list (and also `sandbox2ExplicitSyscalls()` above) now depends on `ML_NR_statx`/`ML_NR_rseq`, which are only `#define`d inside the `__x86_64__`/`__aarch64__` branches near the top of this header, yet are used _unconditionally_ under `__linux__`. `ML_NR_clone3` has a global fallback; these two don't, so a Linux build on any other arch fails with a confusing "undeclared identifier" from right here. Given that ml-cpp only supports x86_64/aarch64 today this is a minor nit, but a `#else #error "unsupported architecture"` closing the arch block would make the unsupported-arch failure explicit.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
```diff
@@ -124,6 +125,187 @@ inline std::vector<int> legacyBpfAllowedSyscalls() {
...
+        __NR_exit,
+        __NR_getuid,
+        __NR_getgid,
+        __NR_geteuid,
+        __NR_getegid,
+        __NR_setpriority,
+        __NR_getpriority,
+        __NR_tgkill,
+        __NR_statfs,
+        __NR_connect,
```
edsavage, line 226: `__NR_connect` is granted to the Sandbox2 policy here (and also carried in the legacy list), but this policy adds no network namespace, so `connect` is actually reachable inside the sandbox. Is this because it's needed for the log-pipe/unix socket?  Maybe add a justifying trailing comment? Just a one-liner on why `connect` (and `mknod`/`mknodat`) are required would keep the allowlist self-documenting.

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
```diff
@@ -413,10 +545,13 @@ buildPytorchInferenceFilesystemPolicy(const std::string& binDir,
...
-    // path. validated.s_Ok and s_ChildIpcRoot shape were checked above.
-    policyBuilder.AddDirectoryAt(validated.s_Spec.s_ChildIpcRoot, "/run/elastic/ml-ipc",
-                                 /*is_ro=*/false);
+    // The one per-child IPC root, mapped read-write at the same path inside
+    // and outside the sandbox. validated.s_Ok and s_ChildIpcRoot shape were
+    // checked above. Same-path (not a remapped in-sandbox path) because
+    // pytorch_inference receives its --input=/--output=/--restore=/--logPipe=
+    // argv from Elasticsearch as host paths under this root; a remap would
+    // leave those paths unresolvable inside the sandbox's own mount namespace.
+    policyBuilder.AddDirectory(validated.s_Spec.s_ChildIpcRoot, /*is_ro=*/false);
```
edsavage, line 554: Nit on the `/dev/null`/`/dev/urandom`/`/dev/random` loop earlier in this function (inherited from #3187, and unchanged here, so anchored on this line): `is_ro = std::strcmp(devFile, "/dev/null") != 0` is correct (only `/dev/null` is writable, the RNG devices read-only) but took me a while to understand. Again a one-line comment ("only /dev/null needs to be writable") would help readability.

edsavage (commented):
Just a few changes and some nits below, plus one non-defect I want to state explicitly (the deliberately-off hard-termination switch). No real blockers.

bin/pytorch_inference/Main.cc
```diff
@@ -296,31 +296,59 @@ int main(int argc, char** argv) {
...
+    // security boundary and ML_SANDBOXED is exactly "1", so the whole step -
+    // install, degraded-mode decision, attestation marker - is skipped.
+    // Attempting it from inside an already-sandboxed environment would
+    // either fail (which would terminate every enforced-route launch once
+    // hard termination above is activated) or succeed and emit the
+    // legacy-route attestation marker on a launch the controller's
+    // sandbox2_launch signal reports as "route":"sandbox2".
+    const bool sandbox2Launched{ml::seccomp::sandbox2LaunchedChild()};
+    const ml::seccomp::SInProcessFilterResult seccompResult{ml::seccomp::applyInProcessSeccompFilter(
+        sandbox2Launched, TERMINATE_ON_DEGRADED_SECCOMP_FAILURE,
```
edsavage, line 344: With `TERMINATE_ON_DEGRADED_SECCOMP_FAILURE` deliberately `false`, a legacy-route launch whose in-process seccomp install *fails* (mechanism unavailable) still proceeds to process untrusted model input unsandboxed. This is intentional and very well documented here and in `CSystemCallFilter.h`, gated on the future guarantee that every launch carries an explicit token. We should make sure there's a tracked follow-up issue for the flip, since this is the one place the design is knowingly open.

edsavage (approved):
LGTM

Just a few non-blocking changes/nits

edsavage:
Drive-by (from investigating the failing `build_test_linux-x86_64-RelWithDebInfo-debug` step on [build 3194](https://buildkite.com/elastic/ml-cpp-pr-builds/builds/3194)) — not something this PR introduced, but it surfaces here so flagging it.

**Root cause:** the errors are in vendored Abseil, not our code:

```
_deps/absl-src/absl/profiling/internal/exponential_biased.cc:56:69:
  error: conversion from 'int64_t' to 'double' may change value [-Werror=conversion]
_deps/absl-src/absl/profiling/internal/exponential_biased.cc:69:10:
  error: conversion from 'double' to 'int64_t' may change value [-Werror=float-conversion]
cc1plus: all warnings being treated as errors
```

Two recently-landed changes combine to make this fatal:

1. `add_compile_options(${ML_CXX_FLAGS})` (`CMakeLists.txt:122`) runs before `add_subdirectory(3rd_party)`, so our `-Wconversion` (`cmake/compiler/gcc.cmake`) leaks into every FetchContent'd target, including Abseil.
2. #3198 enabled `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON` for the debug Linux build, which promotes those third-party warnings to errors.

Since the dormant Sandbox2/Abseil foundation (#3181) actually compiles Abseil on Linux, this now affects **every** PR's debug step — #3188 just happens to be where I saw it.

**Suggested fix/work-around** — stop leaking our lints into vendored code rather than relaxing `-Werror`. Options, in order of preference:

1. Declare the third-party deps as `SYSTEM` so their warnings are suppressed:
   ```cmake
   FetchContent_Declare(sandboxed-api ... SYSTEM)   # CMake >= 3.25
   # or: add_subdirectory(${sandboxed-api_SOURCE_DIR} ${sandboxed-api_BINARY_DIR} EXCLUDE_FROM_ALL SYSTEM)
   ```
2. In the Linux block of `3rd_party/CMakeLists.txt` (which already saves/restores `BUILD_TESTING`, `CMAKE_UNITY_BUILD`, etc.), also turn off warnings-as-errors for that subtree:
   ```cmake
   set(CMAKE_COMPILE_WARNING_AS_ERROR OFF)
   ```
3. Narrowest: `-Wno-error=conversion -Wno-error=float-conversion` on the Abseil targets only.

Happy to open a small PR for option 1/2 if that's useful.

edsavage:
Opened #3208 with the preferred fix (disable warnings-as-errors for the third-party subtree only). Once it merges and this branch picks it up, the debug Linux build should go green. Feel free to pull it in sooner if you want to unblock this PR immediately.

valeriy42:
Re-review requested after force-push to `9eadbddd6`.

- Landlock / confinement-ladder work from the tip SPIKE commits is split into stacked #3215 (squashed to two commits on `feature/sandbox2-pr-f-landlock-fallback`).
- This PR (#3188) now ends at `9eadbddd6` with review-thread fixes in `11dc1407f` / `9eadbddd6` (see inline replies).

Follow-ups: #3213 (flip degraded seccomp termination), #3214 (reap `ml-child-ipc/<id>`).

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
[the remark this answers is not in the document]
valeriy42: Fixed in `11dc1407f`: `makeChildIpcDirectory()` now accepts `EEXIST` only when `lstat` shows an owner-only mode-0700 directory; a regular file or looser directory yields `E_CreationFailed`. Tests in `CPytorchInferenceSandboxPolicyTest.cc`.

valeriy42: Same fix as the Copilot thread (`11dc1407f`).

bin/controller/CProcessSpawnerRouter.cc
valeriy42: Documented on `ensureChildIpcDirectory()` in `11dc1407f`: per-deployment directory, reused across restarts, empty after FIFO unlink. Removal tracked in #3214.

bin/controller/CProcessSpawnerRouter.cc
valeriy42: Addressed in `11dc1407f` by documenting on `ERoute::E_Sandbox2` and the router `else` branch that the enum default is not a routing decision for non-sandboxed paths (so we do not `LOG_WARN` on every `./autodetect` launch).

bin/controller/Main.cc
valeriy42: Comment rewritten in `11dc1407f`: no-token never fails; `--requireSandbox` without Sandbox2 fails closed on this PR; ES sends tokens only on Linux.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
valeriy42: Removed `sandbox2AllowsAllLegacySyscalls()` / `sandbox2HelperCoveredSyscalls()` in `11dc1407f` — the policy builder’s `legacyBpfAllowedSyscalls()` loop is what actually mirrors the legacy filter.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
valeriy42: Added `#else #error` for unsupported Linux arch in `11dc1407f`.

include/seccomp/CMlLegacyBpfSyscallAllowlist.h
valeriy42: Trailing comments on `__NR_mknod*` (mkfifo) and `__NR_connect` (AF_UNIX pipe connect) in `sandbox2ExplicitSyscalls()` in `11dc1407f`.

lib/sandbox/CPytorchInferenceSandboxPolicy.cc
valeriy42: One-line `/dev/null` writable-only comment in `11dc1407f`.

bin/pytorch_inference/Main.cc
valeriy42: Follow-up filed as #3213; `Main.cc` links it in `9eadbddd6`. Landlock/step-down behaviour for hosts that cannot run Sandbox2 is in stacked #3215 (not this PR).

lib/core/CDetachedProcessSpawner_Windows.cc
```diff
@@ -175,22 +230,63 @@ bool CDetachedProcessSpawner::spawn(const std::string& processPath,
...
+    // because lpEnvironment below must be a native UTF-16 block passed with
+    // CREATE_UNICODE_ENVIRONMENT - CreateProcess() does not support mixing
+    // an ANSI command line/application name with a Unicode environment
+    // block. processPath/cmdLine are converted to wide strings with
+    // CStringUtils::narrowToWide() (the established conversion helper in
+    // this codebase) purely for this call; they are not the source of the
+    // regression this switch fixes (see below).
+    const std::wstring wideProcessPath{CStringUtils::narrowToWide(
+        processPathHasExeExt ? processPath : processPath + ".exe")};
+    std::wstring wideCmdLine{CStringUtils::narrowToWide(cmdLine)};
```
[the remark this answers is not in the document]
valeriy42: No code change: argv paths are ASCII in practice; the pre-change `CreateProcess` (ANSI) path was equally lossy for non-ASCII environment values. The `CreateProcessW` switch was required for correct Unicode `lpEnvironment` handling when stripping `ML_SANDBOXED`.

lib/sandbox/unittest/CSandboxUserNamespaceProbeTest_Linux.cc
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    int status = 0;
+    BOOST_TEST_REQUIRE(::waitpid(child, &status, 0) == child);
+
+    if (WIFEXITED(status) == 0) {
+        // Killed by a signal: not a meaningful staged result, but also not
+        // the specific exec-failure signature (126/127) - treat as a
+        // staged failure rather than a hard harness-broken failure.
+        return EProbeOutcome::E_StagedFailure;
```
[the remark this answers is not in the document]
valeriy42: Fixed in `11dc1407f`: signaled probe payload is `E_Crashed` and fails the harness in every `ML_SANDBOX2_REQUIRE` mode.

test/test_sandbox2_attack_defense.py
```diff
@@ -0,0 +1,1202 @@
...
+                    try:
+                        data = os.read(self.fd, 4096)
+                    except BlockingIOError:
+                        continue
+                    except OSError as e:
+                        if self.running:
+                            self.error = str(e)
+                        break
+                    if not data:
+                        break
```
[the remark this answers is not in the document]
valeriy42: No code change: on Linux, FIFO hang-up is reported only after a writer has connected and left (`pipe_poll` writer count), so the reader does not exit on an empty `read()` before the child connects.

PR #3189, opened 2026-09-10
[ML] Stabilize uncertain trend forecasts

### LINKED ISSUE #2772: [ML] Forecasting trend is unexpected given data

valeriy42:
## Summary

Fixes #2772, where an uncertain fitted trend could dominate a long forecast and produce implausible curvature, including values outside a percentage metric's natural range.

## Mathematical rationale

The trend component returns a polynomial regression prediction `f(x)` and its covariance-derived forecast variance.  Previously, a forecast evaluated the fitted polynomial directly at every future time.  This means a small but poorly determined quadratic coefficient is multiplied by `x^2`; over a long horizon that term can become much larger than the local signal even when the model is highly uncertain about it.  The existing prediction-horizon guard does not cover this case: the requested duration can be operationally reasonable while the selected trend is still statistically unreliable to extrapolate.

The fix keeps the model's near-term prediction but attenuates its *change from the first forecast point* according to the fraction of variance explained by the fitted trend:

```text
w(t) = V_trend / (V_trend + V_extrapolation(t))
y_fixed(t) = f(t0) + w(t) * (f(t) - f(t0))
```

Here `t0` is the first forecast time, `V_trend` is the component's long-term variance, and `V_extrapolation(t)` is the additional uncertainty incurred by projecting the fitted polynomial from the regression origin to `t`.  Thus `w(t)` is close to one when the trend is reliable, preserving the original prediction; it falls toward zero when extrapolation uncertainty dominates, smoothly anchoring the forecast at the boundary instead of allowing an uncertain high-order term to diverge.  This uses no new persisted state, settings, metric bounds, or customer-wide runtime cost.

The covariance calculation must use the regression basis.  The trend model is parameterized at its regression origin, so the variance of the change `f(t) - f(t0)` uses the basis delta:

```text
[0, t - t0, t^2 - t0^2]
```

Using `[1, t, t^2]` instead incorrectly includes the intercept and computes the variance of an absolute prediction rather than the extrapolated change.  This PR corrects that basis.

## Behaviour and coverage

- Exact zero extrapolation variance retains the original trend prediction (`w = 1`); this prevents a regression for deterministic affine trends.
- Invalid or negative numerical variances take the conservative anchored fallback.
- The forecast remains continuous at the first prediction point.
- Added tests cover the restored #2772 state, prefix continuity, affine trends, and exact zero-uncertainty trends.

This targets the intended observability workload: ordinary trend forecasts remain unchanged when their extrapolation is well supported, while temporary changes and weakly identified curvature cannot dominate a long-horizon forecast.


valeriy42:
[old-vs-new-synthetic-forecast-comparison-10x-horizon.pdf](https://github.com/user-attachments/files/32064294/old-vs-new-synthetic-forecast-comparison-10x-horizon.pdf)

Synthetic examples before and after the change

lib/maths/time_series/CTrendComponent.cc
```diff
@@ -585,7 +594,25 @@ void CTrendComponent::forecast(core_t::TTime startTime,
...
+                extrapolationWeight = 1.0;
+            } else if (longTermVariance > 0.0 && std::isfinite(longTermVariance)) {
+                // Preserve extrapolation while its uncertainty is comparable to
+                // the variation observed in the series.
+                extrapolationWeight = std::sqrt(
+                    std::min(2.0 * longTermVariance / extrapolationVariance, 1.0));
+                extrapolationWeight = std::isfinite(extrapolationWeight) ? extrapolationWeight
+                                                                         : 0.0;
+            }
+        }
```
tveasey, line 614: I find this kind of hard to follow. How about the equivalent:
```
double extrapolationWeight{0.0};
if (extrapolationVariance == 0.0) {
    extrapolationWeight = 1.0;
} else if (extrapolationVariance > 0.0 && std::isfinite(extrapolationVariance) &&
           longTermVariance > 0.0 && std::isfinite(longTermVariance)) {
    // Preserve extrapolation while its uncertainty is comparable to (within 2x)
    // the variation observed in the series.
    extrapolationWeight = std::sqrt(
        std::min(2.0 * longTermVariance / extrapolationVariance, 1.0));
}
```

tveasey (approved):
I like the strategy. I have a proposal for simplifying the edge case handling, but otherwise LGTM.

lib/maths/time_series/CTrendComponent.cc
valeriy42: Applied the flattened exclusive-case form. Zero still maps to weight 1; invalid/negative still fall through to 0. Dropped the post-sqrt isfinite check because the argument is in [0, 1] given the guards.

valeriy42:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/ml-cpp/pull/3199)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

